PR #15, opened 2018-06-19
Roadmap in separate markdown file, linked to from Readme

ernopp:
Previously discussed in https://github.com/spatialos/UnityGDK/pull/12


README.md
```diff
@@ -51,6 +51,9 @@ For prerequisites, installation and setup, see the [Installation and setup](docs
 ## Migration from the SpatialOS Unity SDK
 Currently the Unity GDK is in its early development, so there is no migration path from the [Unity SDK](https://github.com/spatialos/UnitySDK) to the Unity GDK. There will be a migration path in later releases.
 
+## Roadmap
+See the latest Roadmap [here](docs/Roadmap.md) and don't hesitate to feedback!
```
jessicafalk: `to give feedback` instead of `to feedback`

docs/Roadmap.md
```diff
@@ -0,0 +1,7 @@
+Our latest roadmap is below - please don't hesitate to feedback!
```
jessicafalk: `to give feedback` instead of `to feedback`

docs/Roadmap.md
```diff
@@ -0,0 +1,7 @@
...
+
+| | Faster development iteration | Native workflows| High performance|
+|------|---|----|--|
+|First look|**Build less often**<br/>Run client and server running simultaneously in your Unity Editor.<br/><br/>**Transform Feature Module**<br/>Buffer and interpolate position and rotation updates.|**Automatic synchronisation**<br/>Generated code synchronises changes to components for you with a full override control option.|**Harness the power of ECS**<br/>Tale full advantage of the Unity Entity Component System (ECS).
```
jessicafalk: remove the `running`: `Run client and server simultaneously in your Unity Editor`

`Take` instead of `Tale`

README.md
ernopp: this is fine :)

docs/Roadmap.md
ernopp: this is fine :)

ElleEss: "...give us feedback."   <- even better.

ElleEss:
(1) IMHO, it's better in its own doc.
(2) It needs a title at this level `#`.
(3) It would be better in the docs directory rather than a top-level directory.
(4) 'First look' doesn't tie in with the words 'pre-alpha' so we need to say when 'first look' is available from and what date this roadmap applies from (because it will change).

PR #24, opened 2018-06-20
Implement world cleanup to allow reconnecting.

martinzlocha:

-------

#### Description
If you ever try to switch scenes, you will find that you get a whole bunch of null refs due to worlds not being cleaned up properly between runs.

There is an issue that its not possible to reconnect because the ID of the client does not change. This issue is tracked in UTY-483 and won't be fixed here because its out of scope.

#### Tests
Create two scenes and switch between them.

#### Documentation
N/A

#### Primary reviewers

### DIFF
--- a/workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
+++ b/workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
@@ -27,10 +27,6 @@ namespace Improbable.Gdk.Core
             }
 
             WorkerId = workerId;
-            World = new World(WorkerId);
-            WorkerRegistry.SetWorkerForWorld(this);
-
-            View = new MutableView(World);
             Origin = origin;
         }
 
@@ -45,20 +41,45 @@ namespace Improbable.Gdk.Core
                 Connection = ConnectionUtility.LocatorConnectToSpatial((LocatorConfig) config, GetWorkerType);
             }
 
-            if (Connection == null)
+            if (Connection == null || !Connection.IsConnected)
             {
                 return false;
             }
 
+            World = new World(WorkerId);
+            WorkerRegistry.SetWorkerForWorld(this);
+
+            View = new MutableView(World);
             View.Connection = Connection;
 
-            Application.quitting += () =>
-            {
-                ConnectionUtility.Disconnect(Connection);
-                Connection = null;
-            };
+            RegisterSystems();
+
             View.Connect();
             return true;
+        }
+
+        public void Clear()
+        {
+            if (Connection == null || World == null)
+            {
+                return;
+            }
+
+            // This is required because the systems will still tick once when the world is disposed causing exceptions
+            foreach (var manager in World.BehaviourManagers)
+            {
+                var system = manager as ComponentSystem;
+                if (system != null)
+                {
+                    system.Enabled = false;
+                }
+            }
+
+            World.Dispose();
+            ConnectionUtility.Disconnect(Connection);
+
+            View = null;
+            World = null;
         }
 
         public virtual void RegisterSystems()
--- a/workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
+++ b/workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
@@ -80,6 +80,7 @@ namespace Improbable.Gdk.Core
 
             View = null;
             World = null;
+            Connection = null;
         }
 
         public virtual void RegisterSystems()
--- a/workers/unity/Assets/Gdk/Core/Worker/WorkerRegistry.cs
+++ b/workers/unity/Assets/Gdk/Core/Worker/WorkerRegistry.cs
@@ -32,6 +32,13 @@ namespace Improbable.Gdk.Core
             return worker;
         }
 
+        public static void Clear()
+        {
+            WorkerTypeToInitializationFunction.Clear();
+            WorkerTypeToAttributeSet.Clear();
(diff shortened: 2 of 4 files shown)


workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -45,22 +41,47 @@ public bool Connect(ConnectionConfig config)
                 Connection = ConnectionUtility.LocatorConnectToSpatial((LocatorConfig) config, GetWorkerType);
             }
 
-            if (Connection == null)
+            if (Connection == null || !Connection.IsConnected)
             {
                 return false;
             }
 
+            World = new World(WorkerId);
```
jessicafalk: do we want to do that? what if the user has some systems running independent from any spatial os component?
that was one of the main reasons to initialise everything in the constructor

the other one is to ensure that the worker is in a valid state from the very beginning.

martinzlocha: I am not sure creating the world in the constructor is the best approach because it would make the logic to dispose and recreate it much harder. Also currently the process of initializing the worker and connecting it is all done synchronously in one tick so even if its not created in the constructor no ticks will be lost. We could move this to the start of the Worker.Connect call but I dont think this will make a big difference.

jamiebrynes7: Well imagine you didn't want to connect immediately on play, this change makes it so that everything being initialised during the connection attempt. Ideally we've have the basic initialisation pre-connect and then Spatial-y things on connect then clean up Spatial-y things on disconnect.

I don't think the current structure really supports that particularly well though, unless the user wants to do all pre-connection stuff in GameObjects (which is likely) but there's no guarantee of that.

martinzlocha: That makes sense, currently we don't have good support for non Spatial-y things because when I clear the systems I disable all of them. Currently there are many unanswered questions on dealing with the world when the scene changes etc so a lot of this is out of the scope for this ticket.

But I think separating it into a PreConnect and Connect function might work.

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -45,22 +41,47 @@ public bool Connect(ConnectionConfig config)
...
+            WorkerRegistry.SetWorkerForWorld(this);
+
+            View = new MutableView(World);
             View.Connection = Connection;
 
-            Application.quitting += () =>
```
jessicafalk: This should stay here. Having `OnDisable` called in the bootstrap and the application calling `OnQuit` is not the same.

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
jessicafalk: Why not leave it in the constructor? :)

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -45,22 +41,47 @@ public bool Connect(ConnectionConfig config)
...
+                {
+                    system.Enabled = false;
+                }
+            }
+
+            World.Dispose();
+            ConnectionUtility.Disconnect(Connection);
+
+            View = null;
+            World = null;
```
jessicafalk: Also need to set Connection to null

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
firtina-improbable: cc @JonasImprobable had worked on a bug about this I think

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
martinzlocha: My thought was that the user should be allowed to call Worker.Clear() and then after some time be able to do Worker.Connect() again. If I initialize it in the constructor, then clean everything out then doing Worker.Connect() will throw exceptions. I think that there should be a way for the user to recreate the world and all dependencies for the connection without having to reinitialize the worker.

martinzlocha:
@jessicafalk The ticket is slightly unclear, the issue is that when you switch to a different scene where you don't have gameobjects from the other scene you get a bunch of null exceptions because the rigidbody for that gameobject is null. Another issue is that currently we don't provide a way to clean up the world in general so this PR aimed to address that.

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
martinzlocha: We need to clean up the world when the game object is disabled as well because if we don't there is a bunch of null refs caused by rigidbodies being null because they are no longer in the scene. We could also have Application.quitting call clean in addition to this though. Is there ever a situation where OnDisable isn't called but Application.quitting is?

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
martinzlocha: Thank you, missed that :)

martinzlocha:
After an offline discussion it was agreed that there needs a design doc to agree on how we handle disconnection and cleanup in general because this does not remove the ECS entities. Closing this PR for now.

PR #30, opened 2018-06-22
UTY-488 and UTY-494: Docs Linter Improvements

jamiebrynes7:

-------

#### Description
Couple small changes:
* Enforce local links for links in the repository. This makes the linking more reliable and easier to test (for example a remote link to a invalid heading in a markdown file will not fail, whereas local links will) as well as running through the linter faster.
* Allowed local links to directories
* Added more logging for when a remote link check throws an exception.
* Made changes to all the `md` files that were in violation of the first point. 
* Added dotenv file to track the remote url. This is used for the enforcement of links and in the docs verification branch.

#### Tests
Ran `ci/docs-lint.sh` and added tests to the project for the new conditions.
#### Documentation
This is documentation? We might want to add a style guide for markdown to enforce local links.
#### Primary reviewers
@firtina-improbable @ElleEss


tools/DocsLinter/Program.cs
```diff
@@ -9,6 +9,9 @@ namespace DocsLinter
 {
   internal class Program
   {
+    private const string GithubRepoBlobPath = "github.com/spatialos/unitygdk/blob";
+    private const string GithubRepoTreePath = "github.com/spatialos/unitygdk/tree";
```
firtina-improbable: Why do we need the `/blob` / `/tree` at the end?

tools/DocsLinter/Tests/LinkCheckingTests.cs
```diff
@@ -35,6 +35,18 @@ public void CheckRemoteLink_returns_true_for_arbitrary_status_code()
       var result = Program.CheckRemoteLink(string.Empty, GetUrlForStatusCode(301)); // Permanent redirect
       Assert.IsTrue(result);
     }
+
+    [Test]
+    public void CheckRemoteLink_returns_false_for_remote_links_to_repo_files()
+    {
+      var repoLinkBlob = "https://www.github.com/spatialos/UnityGDK/blob/README,md";
```
firtina-improbable: Do you mean `README.md` ? (`,` vs `.`)

tools/DocsLinter/Program.cs
jamiebrynes7: Those are two different ways to reference files in the repo. If i just have `github.com/spatialos/unitygdk` it will pick up links to things like the issues board and stuff like that

tools/DocsLinter/Tests/LinkCheckingTests.cs
jamiebrynes7: Yes 😬

tools/DocsLinter/Program.cs
firtina-improbable: Got it :)

jared-improbable: Won't this fail for end users that have forked the GDK? Seems like an annoying thing for forkers to have to change.

jamiebrynes7: I guess why are forkers changing our docs and running the docs linter?

jared-improbable: It seems like a valid use-case, if they're modifying/adding things they want to document, and have a build pipeline.

jamiebrynes7: The problem is that you can't generically know the repo URL without parsing `git remotes` or similar. We can't guarantee that everyone has the `spatialos/UnityGDK` as `origin` or similar.

We could do a regex? `.*github.com/.*/unitygdk/blob/.*` and similar? But then we enforce the naming of the repo which might be okay.

firtina-improbable: We can generalize further :D `/github\.com\/[^/]+\/[^/]+\/(blob|tree)/`

jamiebrynes7: Discussed offline with @jared-improbable. Will add a `.env` file containing a reference to the repo link. This is a simple change for forkers to make if they also want this workflow!

firtina-improbable: Nice :)

tools/DocsLinter/DotenvLoader.cs
```diff
@@ -0,0 +1,43 @@
...
+      foreach (var dotenvRow in dotenvFileContents)
+      {
+        var index = dotenvRow.IndexOf("=");
+
+        if (index < 0)
+        {
+          continue;
+        }
+
+        var key = dotenvRow.Substring(0, index).Trim();
```
jared-improbable: `dotEnvRow.Split('=').Select(string.Trim)` ? :)

tools/DocsLinter/DotenvLoader.cs
```diff
@@ -0,0 +1,43 @@
...
+
+      var dotenvFileContents = File.ReadAllLines(".env");
+
+      foreach (var dotenvRow in dotenvFileContents)
+      {
+        var index = dotenvRow.IndexOf("=");
+
+        if (index < 0)
+        {
+          continue;
```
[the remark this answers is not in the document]
jamiebrynes7: Yeah why not :)

tools/DocsLinter/DotenvLoader.cs
jamiebrynes7: You could potentially have a `=` inside a bash variable I think

PR #32, opened 2018-06-25
UTY-328: Add Caller Worker Id to Command Requests

jamiebrynes7:

-------

#### Description
Previously we only provided the attribute set, now we provide the id too!
#### Tests
Tested locally.
#### Documentation
N/A. 
#### Primary reviewers
@jessicafalk

### DIFF
--- a/code_generator/Templates/UnityCommandPayloadGenerator.tt
+++ b/code_generator/Templates/UnityCommandPayloadGenerator.tt
@@ -23,7 +23,7 @@ namespace <#= qualifiedNamespace #>
 
                 internal Translation Translation { get; }
 
-                public string CallerWorkerId;
+                public string CallerWorkerId {get; }
 
                 public List<string> CallerAttributeSet { get; }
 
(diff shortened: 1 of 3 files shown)


code_generator/Templates/UnityCommandPayloadGenerator.tt
```diff
@@ -23,17 +23,21 @@ namespace <#= qualifiedNamespace #>
 
                 internal Translation Translation { get; }
 
+                public string CallerWorkerId;
```
jessicafalk: should we also make it readonly or do we want to change the `CallerWorkerId` after creating the request.

jamiebrynes7: Possibly, might be worth making everything else `readonly` then

jamiebrynes7: Nvm just noticed the `{ get; }` on everything else

PR #41, opened 2018-06-26
Added an attribute to remove non-codegen components

samiwh:
#### Description
Currently we hard code in temporary components in the clean system. This makes this generic and opens up the possibility of having auto cleaned components in other features.
#### Tests
There is a clean reactive component test already, which this passes
#### Documentation
temporary-component.md
#### Primary reviewers
@zeroZshadow @jamiebrynes7 @JonasImprobable

### DIFF
--- a/docs/content/temporary-components.md
+++ b/docs/content/temporary-components.md
@@ -1,0 +1,50 @@
+**Warning:** The [pre-alpha](https://docs.improbable.io/reference/latest/shared/release-policy#maturity-stages) release is for evaluation purposes only, with limited documentation - see the guidance on [Recommended use](../../README.md#recommended-use).
+
+-----
+
+
+## Temporary components
+
+The attribute `Improbable.Gdk.Core.RemoveAtEndOfTick`, can be applied to any ECS component, extending either `IComponentData` or `ISharedComponentData`.
+All components with this attribute will be removed from all entities during the `InternalSpatialOSCleanGroup` at the end of `SpatialOSSendGroup`.
+
+Note: The `InternalSpatialOSCleanGroup` runs after both the `SpatialOSUpdateGroup` and the `CustomSpatialOSSendGroup`.
+
+### Example
+```csharp
+[RemoveAtEndOfTick]
+struct SomeTemporaryComponent : IComponentData
+{
+}
+```
+
+If a system, `AddComponentSystem`, adds `SomeTemporaryComponent` to some entities, then `ReadComponentSystem.OnUpdate()` (see the sample below) will run during the same update loop.
+
+```csharp
+[UpdateAfter(typeof(AddComponentSystem))]
+[UpdateInGroup(typeof(SpatialOSUpdateGroup))]
+class ReadComponentSystem : ComponentSystem
+{
+    private struct Data
+    {
+        int Length;
+        [Readonly] ComponentDataArray<SomeTemporaryComponent> TemporaryComponent;
+    }
+
+    [Inject] Data data;
+
+    protected override void OnUpdate()
+    {
+        // perform work related to the temporary component
+    }
+}
+``` 
+
+At the end of the update loop, `SomeTemporaryComponent` will automatically be removed from all entities.
+
+----
+**Give us feedback:** We want your feedback on the Unity GDK and its documentation  - see [How to give us feedback](../../README.md#give-us-feedback).
+
+[//]: # (Editorial review status: Engineer review only)
+[//]: # (Questions to deal with: need to describe what a tick is. Needs to link to a doc on system execution order. Needs to link to docs describing the existing components which use this)
+[//]: # (1. Editorial review  - JIRA TICKET: UTY-628)
--- a/workers/unity/Assets/Gdk/Core/Attributes/RemoveAtEndOfTick.cs
+++ b/workers/unity/Assets/Gdk/Core/Attributes/RemoveAtEndOfTick.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using Unity.Entities;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    ///     Any component with this attribute will be removed from all entities by the CleanReactiveComponentSystem
+    ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false, Inherited = false)]
+    public class RemoveAtEndOfTick : Attribute
+    {
+    }
+}
+
--- a/workers/unity/Assets/Gdk/Core/Components/WorkerEntityComponents.cs
+++ b/workers/unity/Assets/Gdk/Core/Components/WorkerEntityComponents.cs
@@ -1,21 +1,42 @@
+using System.Collections.Generic;
 using Unity.Entities;
 
 namespace Improbable.Gdk.Core
 {
+    /// <summary>
+    ///     Component denoting a worker entity
+    /// </summary>
     public struct WorkerEntityTag : IComponentData
     {
     }
 
+    /// <summary>
+    ///     Component added to the worker entity immediately after establishing a connection to a SpatialOS deployment
+    ///     Removed immediately after disconnecting
+    /// </summary>
     public struct IsConnected : IComponentData
     {
     }
 
+    /// <summary>
+    ///     Component added to the worker entity immediately after establishing a connection to a SpatialOS deployment
+    ///     Removed at the end of the tick it was added
+    /// </summary>
+    [RemoveAtEndOfTick]
     public struct OnConnected : IComponentData
     {
     }
 
+    /// <summary>
+    ///     Component added to the worker entity immediately after disconnecting from SpatialOS
+    ///     Removed at the end of the tick it was added
+    /// </summary>
+    [RemoveAtEndOfTick]
     public struct OnDisconnected : ISharedComponentData
     {
+        /// <summary>
+        ///     The reported reason for disconnecting
+        /// </summary>
         public string ReasonForDisconnect;
     }
 }
--- a/workers/unity/Assets/Gdk/Core/Systems/CleanReactiveComponentsSystem.cs
+++ b/workers/unity/Assets/Gdk/Core/Systems/CleanReactiveComponentsSystem.cs
@@ -1,15 +1,22 @@
-using Improbable.Collections;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 using Unity.Entities;
 
 namespace Improbable.Gdk.Core
 {
+    /// <summary>
+    /// Removes GDK reactive components and components with attribute RemoveAtEndOfTick from all entities
+    /// </summary>
     [UpdateInGroup(typeof(SpatialOSSendGroup.InternalSpatialOSCleanGroup))]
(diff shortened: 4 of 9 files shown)


workers/unity/Assets/Gdk/Core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -19,53 +26,76 @@ protected override void OnCreateManager(int capacity)
...
-            onConnectedComponentGroup = GetComponentGroup(typeof(OnConnected));
-            onDisconnectedComponentGroup = GetComponentGroup(typeof(OnDisconnected));
         }
 
         private void GenerateComponentGroups()
         {
             foreach (var translationUnit in view.TranslationUnits.Values)
             {
-                translationUnit.CleanUpComponentGroups = new List<ComponentGroup>();
+                translationUnit.CleanUpComponentGroups = new Collections.List<ComponentGroup>();
```
jamiebrynes7: Unnecessary change?

workers/unity/Assets/Gdk/Core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -19,53 +26,76 @@ protected override void OnCreateManager(int capacity)
...
             }
+
+            MethodInfo addRemoveComponentActionMethod =
+                typeof(CleanReactiveComponentsSystem).GetMethod("AddRemoveComponentAction",
+                    BindingFlags.NonPublic | BindingFlags.Instance);
+
+            // Find all components with the RemoveAtEndOfTick attribute
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+            {
+                foreach (var type in assembly.GetTypes())
```
jamiebrynes7: LINQ ?

`assembly.GetTypes().Where((type) => ...).ForEach((type) => ...)`

workers/unity/Assets/Gdk/Core/Systems/CleanReactiveComponentsSystem.cs
samiwh: Accident. Will fix.

workers/unity/Assets/Gdk/Core/Systems/CleanReactiveComponentsSystem.cs
samiwh: I personally think that makes this harder to read. But I will see if there is something I can do with this.

workers/unity/Assets/Gdk/Core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -1,15 +1,22 @@
-using Improbable.Collections;
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using JetBrains.Annotations;
```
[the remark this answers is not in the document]
samiwh: It's for the [UsedImplicitly] annotation. That stops resharper/rider from complaining. I didn't realise it had added an include. Will take a look at whether this will compile on vanilla vs and see if I can get rid of the warning another way.

docs/content/temporary-components.md
```diff
@@ -0,0 +1,43 @@
+**Warning:** The [pre-alpha](https://docs.improbable.io/reference/latest/shared/release-policy#maturity-stages) release is for evaluation purposes only, with limited documentation - see the guidance on [Recommended use](../../README.md#recommended-use).
+
+-----
+
+
+## Temporary components
+
+The attribute `Improbable.Gdk.Core.RemoveAtEndOfTick`, can be applied to any ECS component, extending either `IComponentData` or `ISharedComponentData`.
```
firtina-improbable: `Tick` > `Update` perhaps?

workers/unity/Assets/Gdk/Core/Components/WorkerEntityComponents.cs
```diff
@@ -2,20 +2,40 @@
...
+    /// Component added to the worker entity immediately after establishing a connection to a SpatialOS deployment
+    /// Removed immediately after disconnecting
+    /// </summary>
     public struct IsConnected : IComponentData
     {
     }
 
+    /// <summary>
+    /// Component added to the worker entity immediately after establishing a connection to a SpatialOS deployment
+    /// Removed at the end of the tick is was added
```
firtina-improbable: "Removed at the end of the tick is was added" sounds odd

workers/unity/Assets/Gdk/Core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -35,37 +38,64 @@ private void GenerateComponentGroups()
...
+                typeof(CleanReactiveComponentsSystem).GetMethod("AddRemoveComponentAction",
+                    BindingFlags.NonPublic | BindingFlags.Instance);
+
+            // Find all components with the RemoveAtEndOfTick attribute
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+            {
+                foreach (var type in assembly.GetTypes())
+                {
+                    if (type.GetCustomAttribute<RemoveAtEndOfTick>(false) == null
+                        || typesToRemove.Contains(type)
```
firtina-improbable: I recommend moving `typesToRemove.Contains` check at the end of this conditional block, or even after it as a separate `continue`. This is because it's currently between two type-based conditions

docs/content/temporary-components.md
```diff
@@ -0,0 +1,43 @@
+**Warning:** The [pre-alpha](https://docs.improbable.io/reference/latest/shared/release-policy#maturity-stages) release is for evaluation purposes only, with limited documentation - see the guidance on [Recommended use](../../README.md#recommended-use).
```
jessicafalk: @ElleEss should go through the doc before merging it in.

docs/content/temporary-components.md
```diff
@@ -0,0 +1,43 @@
...
+
+-----
+
+
+## Temporary components
```
jessicafalk: Still think that some kind of motivation as to why we need temporary components is missing here. The doc just jumps straight to the implementation

docs/content/temporary-components.md
```diff
@@ -0,0 +1,43 @@
...
+
+The attribute `Improbable.Gdk.Core.RemoveAtEndOfTick`, can be applied to any ECS component, extending either `IComponentData` or `ISharedComponentData`.
+All components with this attribute will be removed from all entities during the `InternalSpatialOSCleanGroup` at the end of `SpatialOSSendGroup`.
```
jessicafalk: do we have a ticket to update the doc after we added the documentation about the groups?

docs/content/temporary-components.md
```diff
@@ -0,0 +1,43 @@
...
+    [Inject] Data data;
+
+    protected override void OnUpdate()
+    {
+        // perform work related to the temporary component
+    }
+}
+``` 
+
+At the end of the update loop, `SomeTemporaryComponent` will automatically be removed from all entities.
```
jessicafalk: should we mention here during which system update that happens

docs/content/temporary-components.md
jamiebrynes7: I think this is consistent with the current implementation of update groups. We just haven't written the doc explaining exactly what these are yet 😬

jessicafalk: it's more about adding a reference to that doc as soon as it exists ;)

docs/content/temporary-components.md
samiwh: It's removed at the end of the update look but it's not at the end of update. Also that is an implementation detail (although I'd be surprised if we made it so that SpatialOS ticks are not tied to update)
The intention was to show that it is based around the SpatialOS tick.

workers/unity/Assets/Gdk/Core/Systems/CleanReactiveComponentsSystem.cs
samiwh: Will do

docs/content/temporary-components.md
samiwh: Will add as a reviewer.

docs/content/temporary-components.md
samiwh: That's mostly because we are missing all the things that go with it and the three or so docs that also need to be included. This is factored out from other stuff.
The reason for it to be there is mostly the temporary components on the worker entity and also a prerequisite for the timer systems and various other things.

docs/content/temporary-components.md
samiwh: There is a ticket for that doc to exist.
There is another one to reorganise the docs completely too.

docs/content/temporary-components.md
samiwh: I'll mention the group. The system itself is an implementation detail. It's also easier to miss that when renaming and moving things and leave the docs out of date.

workers/unity/Assets/Gdk/Core/Components/WorkerEntityComponents.cs
samiwh: s/is/it

docs/content/temporary-components.md
firtina-improbable: We may need to link to what a `tick` is in that case, and what "being at the end of it" means specifically

jamiebrynes7:
@samiwh Please add a unit test for this feature. :)

workers/unity/Assets/Gdk/Core/Attributes/RemoveAtEndOfTick.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Unity.Entities;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    /// Any component with this attribute will be removed from all entities by the CleanReactiveComponentSystem
+    /// Can be added to components extending either <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Struct | AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
```
zeroZshadow: Just wondering, what kind of classes will this be on?

samiwh: I think that is a mistake. Will remove it. (Could conceivably add it to components but that will break if done right now)

docs/content/temporary-components.md
samiwh: We should, but there isn't a doc for that yet. It will likely be in the system execution order doc.

workers/unity/Assets/Gdk/Core/Components/WorkerEntityComponents.cs
```diff
@@ -1,21 +1,58 @@
...
+    ///     Component added to the worker entity immediately after establishing a connection to a SpatialOS deployment
+    ///     Removed at the end of the tick it was added
+    /// </summary>
+    [RemoveAtEndOfTick]
     public struct OnConnected : IComponentData
     {
+        private var x = MyCoponent.SomeList;
     }
 
+    public struct MyComponent : IComponentData
```
firtina-improbable: Is this block supposed to contain these lines?

jamiebrynes7: What is this lol

PR #45, opened 2018-06-28
UTY-360: Add Support for Nested Type and Enum Generation

jamiebrynes7:
#### Description
A number of changes had to be made to enable recursively generating types:
1. The `enumSet` of all enums in a schema corpus now includes enums defined in types.
2. Refactored `UnityTypeGenerator` into two parts:
 a. `UnityTypeContent` which contains the template for the actual struct content. It recursively calls generate on itself and `UnityEnumContent`
 b. `UnityTypeGenerator` which contains the imports and namespace declaration and then generates the content using the above tt file.
3. `UnityEnumGenerator` had the same refactor as above.
4. Added an `IndentString` function to enable proper alignment of nested types.
5. Committed generated code to allow for easy diffing when we change the code generator.

Also uncommented the nested schema to show that it works! 
#### Tests
Ran `End2End` tests on the CodeGenerator. It compiled and ran properly!
#### Documentation
N/A - this is a change that gets us closer to full schema compatibility.
#### Primary reviewers
@jessicafalk @samiwh

### DIFF
--- a/code_generator/.gitignore
+++ b/code_generator/.gitignore
@@ -328,7 +328,6 @@ ASALocalRun/
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
 
-Generated/
+/Generated/
 End2End/schema_json_ast/
-End2End/GeneratedDependencies/Generated/
 End2End/GeneratedDependencies/GeneratedCSharpWorker/
--- a/code_generator/End2End/schema/nested_typenames.schema
+++ b/code_generator/End2End/schema/nested_typenames.schema
@@ -1,26 +1,24 @@
 package improbable.test_schema;
 
-// TODO: UTY-360 Uncomment this file when Nested Type definitions are supported.
-//
-// type NestedTypeName {
-//     type Other {
-//         type NestedTypeSameName {
-//             type Other0 {
-//                 int32 foo = 1;
-//             }
-//             enum NestedEnum {
-//                 enum_value = 1;
-//             }
-//             Other0 other_zero = 1;
-//             NestedEnum enum_field = 2;
-//         }
-//         NestedTypeSameName same_name = 1;
-//     }
-//     Other other = 1;
-// }
+type TypeName {
+    type Other {
+        type NestedTypeName {
+            type Other0 {
+                int32 foo = 1;
+            }
+            enum NestedEnum {
+                enum_value = 1;
+            }
+            Other0 other_zero = 1;
+            NestedEnum enum_field = 2;
+        }
+        NestedTypeName same_name = 1;
+    }
+    Other other_type = 1;
+}
 
-// component NestedComponent {
-//     id = 20152;
-//     NestedTypeName nested_type = 1;
-// }
-//
+component NestedComponent {
+    id = 20152;
+    TypeName nested_type = 1;
+}
+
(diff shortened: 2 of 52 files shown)


code_generator/End2End/schema/nested_typenames.schema
```diff
@@ -1,26 +1,24 @@
...
-//             Other0 other_zero = 1;
-//             NestedEnum enum_field = 2;
-//         }
-//         NestedTypeSameName same_name = 1;
-//     }
-//     Other other = 1;
-// }
+type NestedTypeName {
+    type Other {
+        type NestedTypeSameName {
```
samiwh: I'm very confused by this file. What is this the same name as?

samiwh: Could we name the components based on what tier they are in, or something that makes it clearer what is meant to be going on here.

jamiebrynes7: No idea, I think this was copied from the WorkerSDK?

jamiebrynes7: Yeah why not

code_generator/src/Generation/Generators/CommonGeneratorUtils.cs
```diff
@@ -6,5 +6,10 @@ public static string GetGeneratedHeader()
         {
             return "// ===========\r\n// DO NOT EDIT - this file is automatically regenerated.\r\n// ===========";
         }
+
+        public static string IndentString(string input)
```
zeroZshadow: This doesn't just indent a string, could this have a better name?

code_generator/Templates/UnityTypeContent.tt
```diff
@@ -0,0 +1,47 @@
...
+<# } #>
+<# 
+    var enumGenerator = new UnityEnumContent();
+    foreach (var nestedEnum in nestedEnums) 
+    {
+#>
+
+    <#= CommonGeneratorUtils.IndentString(enumGenerator.Generate(nestedEnum)) #>
+<# } #>
+}
```
zeroZshadow: this needs a newline

code_generator/.gitignore
```diff
@@ -328,7 +328,6 @@ ASALocalRun/
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
 
-Generated/
+/Generated/
 End2End/schema_json_ast/
-End2End/GeneratedDependencies/Generated/
```
jessicafalk: do we want to remove this line?

jamiebrynes7: Yes, at the request of @samiwh we check in the example generated code. That way, any changes that someone makes to the templates is easier to parse by looking at the example generated code.

code_generator/src/Generation/Generators/CommonGeneratorUtils.cs
jamiebrynes7: Yes probably, `IndentEveryNewline` ? 😬

code_generator/.gitignore
samiwh: I'd argue that was the entire point of adding the tests. We could also do this by having a unit test that specifies what the generated code should look like (the original suggestion) but these seems good enough.

samiwh:
@jamiebrynes7 Ideally the generated code would be ordered before the other commits so we could see the diff that this pr creates.

jamiebrynes7:
@samiwh My git ninja skills aren't that great

PR #48, opened 2018-07-02
Make Linting Consistent Again (Or Something)

jamiebrynes7:
#### Description
This PR aims to make linting more consistent by specifying line-lengths in the `.editorconfig` files. Also moved the `.editorconfig` to the root directory which will be inherited by all sub-directories. 

#### Tests
N/A
#### Documentation
N/A
#### Primary reviewers


code_generator/.editorconfig
```diff
@@ -10,6 +10,10 @@ charset = utf-8
 insert_final_newline = true
```
jared-improbable: Any reason this and the config in `tools` are separate? Can we unify everything?

jamiebrynes7: I don't think so unless we add soft linking to the setup scripts, which we could do but feels a little brittle. 

Its not ideal to have it replicated though

jared-improbable: Why can't the single file live in the root of the GDK folder?

jamiebrynes7: Oh I misread the docs 🤦‍♂️ We can just have a parent one. Will change

tools/DocsLinter/SimplifiedMarkdownDoc.cs
```diff
@@ -6,176 +6,176 @@
...
-  ///   A data class that represents a local link. I.e. - "../README.md#heading"
-  /// </summary>
-  public struct LocalLink : ILink
-  {
-    public Heading? Heading;
-    public string FilePath;
 
     /// <summary>
-    ///   Constructor for that parses the Markdig link object.
+    ///     Abstract base class for a link.
```
firtina-improbable: s/base class/interface/

workers/unity/Assets/Playground/Scripts/Player/ProcessLaunchCommandSystem.cs
```diff
@@ -18,7 +18,8 @@ private struct LaunchCommandData
             [ReadOnly] public EntityArray Entity;
             public ComponentDataArray<SpatialOSLauncher> Launcher;
 
-            [ReadOnly] public ComponentArray<CommandRequests<Generated.Playground.Launcher.LaunchEntity.Request>> CommandRequests;
```
firtina-improbable: Is this PR going to fix this file flip flopping

jamiebrynes7: Yes it should by explicitly setting the line-length.

tools/DocsLinter/SimplifiedMarkdownDoc.cs
jamiebrynes7: Driveby time 🔫

workers/unity/Assets/Gdk/Core/Components/ISpatialComponentData.cs
```diff
@@ -1,3 +1,5 @@
+using Unity.Mathematics;
```
jessicafalk: remove this line. We don't need it anymore :)

PR #49, opened 2018-07-02
UTY-383: Implement component update structs

JonasImprobable:

-------

#### Description
- Added Option struct type
- Added codegenerated update structs
- ComponentsUpdated reactive component now contains updates instead of component snapshots
#### Tests
- Unit tests for Option type
- Manual for everything else
#### Documentation
- Documentation on reactive components was updated
#### Primary reviewers
@samiwh @firtina-improbable


workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
```diff
@@ -0,0 +1,188 @@
...
+        public void Option_payload_can_be_cleared()
+        {
+            var option = new Option<bool>(true);
+            Assert.AreEqual((bool1)true, option.HasValue);
+            option.Clear();
+            Assert.AreEqual((bool1)false, option.HasValue);
+        }
+
+        [Test]
+        public void TryGetValue_on_empty_option()
```
firtina-improbable: Please rename this test method to give a hint on what it asserts, e.g. `TryGetValue_returns_false_when_empty`

firtina-improbable: Also for the methods below too

workers/unity/Assets/Gdk/Physics/Systems/ApplyTransformUpdatesSystem.cs
```diff
@@ -21,8 +23,24 @@ protected override void OnUpdate()
...
             for (var i = 0; i < transformUpdateData.Length; i++)
             {
-                var newUpdates = transformUpdateData.TransformUpdate[i].Buffer;
-                transformUpdateData.BufferedTransform[i].TransformUpdates.AddRange(newUpdates);
+                var transformUpdates = transformUpdateData.TransformUpdate[i].Buffer;
+                var lastTransformSnapshot = transformUpdateData.BufferedTransform[i].LastTransformSnapshot;
+                var transformSnapshots = new List<SpatialOSTransform>();
+                foreach (var update in transformUpdates)
+                {
+                    lastTransformSnapshot.Location = update.Location.HasValue
```
firtina-improbable: It may look better to do an if block here

code_generator/GdkCodeGenerator.csproj
```diff
@@ -81,7 +81,11 @@
     <Compile Include="src\Generation\Generators\Parts\UnityComponentConversionGeneratorPart.cs" />
     <Compile Include="src\Generation\Generators\Parts\UnityComponentDataGeneratorPart.cs" />
     <Compile Include="src\Generation\Generators\Parts\UnityComponentGeneratorPart.cs" />
+<<<<<<< HEAD
```
firtina-improbable: Merge conflict

JonasImprobable: Fixed.

workers/unity/Assets/Gdk/Physics/Systems/ApplyTransformUpdatesSystem.cs
JonasImprobable: Fixed.

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
JonasImprobable: Fixed.

code_generator/Templates/UnityComponentUpdateGenerator.tt
```diff
@@ -0,0 +1,20 @@
...
+    var componentDetails = GetComponentDetails();
+    var generatedHeader = CommonGeneratorUtils.GetGeneratedHeader();
+#>
+<#= generatedHeader #>
+
+using Improbable.Gdk.Core;
+
+namespace <#= qualifiedNamespace #>
+{ 
+    public struct <#= componentDetails.TypeName #>Update : ISpatialComponentUpdate
```
zeroZshadow: Please make this a sub class of the existing `<#= componentDetails.TypeName #>` namespacing class

code_generator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -172,10 +177,33 @@ namespace <#= qualifiedNamespace #>
                 }
 <# } #>
                 componentData.DirtyBit = false;
+
 <# if (componentDetails.IsBlittable) { #>
-                view.UpdateComponent(entity, componentData, UpdatesPool);
+                view.SetComponentData(entity, componentData);
```
jessicafalk: why did we change this line?

workers/unity/Assets/Gdk/Core/Components/Option.cs
```diff
@@ -0,0 +1,120 @@
+using System;
+
+namespace Improbable.Gdk.Core
+{
+    public struct Option<T> : IEquatable<Option<T>>
+    {
+        private bool hasValue;
```
zeroZshadow: Shouldn't this be a BlittableBool ?

workers/unity/Assets/Gdk/Physics/Systems/ApplyTransformUpdatesSystem.cs
```diff
@@ -10,7 +11,8 @@ public class ApplyTransformUpdatesSystem : ComponentSystem
         public struct TransformUpdateData
         {
             public int Length;
-            public ComponentArray<ComponentsUpdated<SpatialOSTransform>> TransformUpdate;
+            public ComponentDataArray<SpatialOSTransform> Transform;
```
jessicafalk: we are not using this one in the system. Is it needed?

workers/unity/Assets/Gdk/Physics/Components/BufferedTransform.cs
```diff
@@ -6,6 +6,7 @@ namespace Improbable.Gdk.TransformSynchronization
     public class BufferedTransform : Component
     {
         public System.Collections.Generic.List<SpatialOSTransform> TransformUpdates;
+        public SpatialOSTransform LastTransformSnapshot;
```
jessicafalk: We are not using this variable anywhere. Only updating its content whenever we have a new update

workers/unity/Assets/Gdk/Physics/Systems/ApplyTransformUpdatesSystem.cs
```diff
@@ -21,8 +23,28 @@ protected override void OnUpdate()
...
+
+                    if (update.Rotation.HasValue)
+                    {
+                        lastTransformSnapshot.Rotation = update.Rotation.Value;
+                    }
+
+                    if (update.Tick.HasValue)
+                    {
+                        lastTransformSnapshot.Tick = update.Tick.Value;
+                    }
```
firtina-improbable: In the latest commit it seems you had removed the call to `transformSnapshots.Add()`

Do we need to add to `transformSnapshots` for every `update`, or only `update` which has one or more of these three values modified?

workers/unity/Assets/Gdk/Core/Components/Option.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        public T GetValueOrDefault(T defaultValue)
+        {
+            return hasValue ? value : defaultValue;
+        }
+
+        public override bool Equals(object other)
+        {
+            if (other is Option<T>)
```
jamiebrynes7: The body of this function can be replaced with: 

```return other is Option<T> && this.Equals((Option<T>)other);```

File: workers/unity/Assets/Gdk/Core/Components/Option.cs [hunk 11]
```diff
@@ -0,0 +1,120 @@
...
+                if (!hasValue)
+                {
+                    throw new InvalidOperationException("Called Value on empty Option.");
+                }
+
+                return value;
+            }
+            set
+            {
+                if (value == null)
```
jamiebrynes7: This boxes I think? (value: `T` -> `object`) since `T` can be a value type

workers/unity/Assets/Gdk/Core/Components/Option.cs
jessicafalk: is there a reason why we have a private and a public version of the variable? The private one doesn't look like it should be readonly or final, so we can just make it public in general

workers/unity/Assets/Gdk/Core/Components/Option.cs
```diff
@@ -0,0 +1,120 @@
...
+        public bool Equals(Option<T> other)
+        {
+            if (!hasValue && !other.HasValue)
+            {
+                return true;
+            }
+
+            if (hasValue && other.HasValue)
+            {
+                return value.Equals(other.Value);
```
jamiebrynes7: This line boxes as well I think (Value: `T` -> `object`) since `T` can be a value type

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
```diff
@@ -0,0 +1,179 @@
...
+namespace Improbable.Gdk.Core.EditmodeTests
+{
+    [TestFixture]
+    public class OptionTests
+    {
+        [Test]
+        public void Parameterless_constructor_creates_empty_option()
+        {
+            var option = new Option<bool>();
+            Assert.AreEqual(false, (bool)option.HasValue);
```
jessicafalk: this looks like it needs linting

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        [Test]
+        public void Constructor_with_parameters_creates_option_with_payload()
+        {
+            var payload = true;
+            var option = new Option<bool>(payload);
+            Assert.AreEqual(true, (bool)option.HasValue);
+            Assert.AreEqual(payload, option.Value);
```
jamiebrynes7: You repeat this same assertion below

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            option.Value = payload;
+            Assert.AreEqual(true, (bool)option.HasValue);
+            Assert.AreEqual(payload, option.Value);
+        }
+
+        [Test]
+        public void Accessing_value_of_empty_option_throws()
+        {
+            var option = new Option<bool>();
+            Assert.AreEqual(false, (bool)option.HasValue);
```
jamiebrynes7: No need to test this, you test it above

code_generator/Templates/UnityComponentUpdateGenerator.tt
zeroZshadow: So this makes it `SpatialOSPosition.Update`, instead of `SpatialOSPositionUpdate`. Which will be consistent with the other components, and avoid name conflicts

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                var value = option.Value;
+            });
+        }
+
+        [Test]
+        public void Setting_null_value_throws()
+        {
+            var option = new Option<string>();
+            Assert.AreEqual(false, (bool)option.HasValue);
```
jamiebrynes7: Same comment, repeated tests

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
```diff
@@ -0,0 +1,179 @@
+using System;
+using NUnit.Framework;
+
+namespace Improbable.Gdk.Core.EditmodeTests
+{
+    [TestFixture]
+    public class OptionTests
```
jamiebrynes7: Add some tests with reference types too!

code_generator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -172,10 +177,33 @@ namespace <#= qualifiedNamespace #>
...
 <# } else { #>
-                view.UpdateComponentObject(entity, componentData, UpdatesPool);
+                view.SetComponentObject(entity, componentData);
+<# } #>
+
+<# if (fieldDetailsList.Count > 0) { #>
```
jessicafalk: I believe it is possible to get completely empty component update ops: updates from components that don't contain any fields. The way it's set up right now, we would simply swallow them. Do we want to add a `ComponentsUpdated` tag for these cases as well so that can be processed by any custom system? (not sure what the exact use case would be, maybe pings according to @zeroZshadow? )

code_generator/Templates/UnityComponentConversionGenerator.tt
JonasImprobable: The view previously provided the following API:
- UpdateComponent()
- UpdateComponentObject()
which both update the component snapshot and add a message to their respective reactive components.

In the new solution we update the component snapshot separately from adding the reactive message. This is also more in line with how we are doing events, commands, auth changes.

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: @zeroZshadow Yes, I derped here. Should be a blittablebool.
@jessicafalk Having public fields would allow users to tamper with the integrity of an option. You could e.g. set the payload while hasValue is false rather than using the designated interface for setting getting, clearing.

workers/unity/Assets/Gdk/Physics/Systems/ApplyTransformUpdatesSystem.cs
JonasImprobable: We are using this in line 47. :)

workers/unity/Assets/Gdk/Physics/Components/BufferedTransform.cs
JonasImprobable: This is needed in `ApplyTransformUpdatesSystem` because we now need to reconstruct component snapshots from the updates since we don't get the component snapshots directly anymore.

workers/unity/Assets/Gdk/Physics/Systems/ApplyTransformUpdatesSystem.cs
JonasImprobable: Holy shit, I derped here. Yes, that line wasn't supposed to be removed.

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
JonasImprobable: Yep, haven't run the linter yet. Will do.

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: Nice. I love the simplicity.

code_generator/Templates/UnityComponentConversionGenerator.tt
JonasImprobable: If you receive a component update, it has to be that either your component fields were updated or you received events (or both). In case we only receive events as part of a component update, the current idea is to emit the EventsReceived tags but not emit ComponentsUpdated tags.

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
JonasImprobable: Good point. Removed.

workers/unity/Assets/Gdk/Core/Components/Option.cs
```diff
@@ -0,0 +1,120 @@
...
+                    throw new InvalidOperationException("Called Value on empty Option.");
+                }
+
+                return value;
+            }
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentException("newValue is null.");
```
samiwh: Please make a new exception type for this.

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
samiwh: Put this in a one time setup block and assert on it there. Then they other tests can assume this for setup

workers/unity/Assets/Gdk/Core/Components/Option.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        public BlittableBool HasValue => hasValue;
+
+        public T Value
+        {
+            get
+            {
+                if (!hasValue)
+                {
+                    throw new InvalidOperationException("Called Value on empty Option.");
```
samiwh: Please make a new exception type for this.

workers/unity/Assets/Gdk/Core/Components/Option.cs
samiwh: You might want to make this something other than a type T. Possibly an internal type than can be implicitly cast to T but can also be set with null (basically a nullable T wrapper). It would create the oddity that you can have an option ref with a value but that value being null. But that might be fine.

We could also try branching off `typeof(T).isValueType` and having things deal with the two different cases there. Either way it might be better if `Value` is garunteed to be a type that cab be assigned null.

samiwh: I'd mildly prefer not having a setter here at all though. Users shouldn't ever be setting value directly. It's just an accessor that can throw if there is no value present.

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: How granular do you want this? as in do you want
- an OptionArgumentException
- an ArgumentIsNullException
- an OptionArgumentIsNullException?

jared-improbable: There is already a `System.ArgumentNullException` that should be used. https://msdn.microsoft.com/en-us/library/system.argumentnullexception(v=vs.110).aspx

code_generator/Templates/UnityComponentUpdateGenerator.tt
JonasImprobable: Great idea. Will do.

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: Discussed offline: Made Options immutable and made empty options distinct from options with payload null.

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: As far as I understand, this boxes depending on whether the type of `value` overrides the default `Object.Equals` method. This sounds like something that is outside of our control.

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
JonasImprobable: Removed.

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
JonasImprobable: Removed redundant assertions.

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: Also, is this really necessary? What's wrong with the standard exceptions?

workers/unity/Assets/Gdk/Core/Tests/Editmode/Components/OptionTests.cs
JonasImprobable: Added tests for reference types.

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: Is exception has been removed as Options are immutable now.

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: Is this really necessary? What's wrong with the standard exceptions?

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/ExhaustiveBlittableSingularTranslation.cs
```diff
@@ -27,20 +27,21 @@ public class Translation : ComponentTranslation, IDispatcherCallbacks<global::Im
...
-                new ComponentPool<ComponentsUpdated<SpatialOSExhaustiveBlittableSingular>>(
-                    () => new ComponentsUpdated<SpatialOSExhaustiveBlittableSingular>(),
-                    (component) => component.Buffer.Clear());
-
             private static readonly ComponentPool<AuthoritiesChanged<SpatialOSExhaustiveBlittableSingular>> AuthsPool =
                 new ComponentPool<AuthoritiesChanged<SpatialOSExhaustiveBlittableSingular>>(
                     () => new AuthoritiesChanged<SpatialOSExhaustiveBlittableSingular>(),
                     (component) => component.Buffer.Clear());
 
+            private static readonly ComponentPool<ComponentsUpdated<SpatialOSExhaustiveBlittableSingular.Update>> UpdatesPool =
```
jamiebrynes7: Why has this moved around?

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/ExhaustiveBlittableSingularTranslation.cs
```diff
@@ -168,7 +168,136 @@ public void OnComponentUpdate(ComponentUpdateOp<global::Improbable.TestSchema.Ex
                 }
 
                 componentData.DirtyBit = false;
-                view.UpdateComponent(entity, componentData, UpdatesPool);
+
+                view.SetComponentData(entity, componentData);
+
+                var componentFieldsUpdated = false;
+                if (update.field1.HasValue)
```
jamiebrynes7: Please combine this with the `if`s above

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/ExhaustiveBlittableSingularTranslation.cs
```diff
@@ -168,7 +168,136 @@ public void OnComponentUpdate(ComponentUpdateOp<global::Improbable.TestSchema.Ex
...
+                    componentFieldsUpdated = true;
+                }
+                if (update.field17.HasValue)
+                {
+                    componentFieldsUpdated = true;
+                }
+
+                if (componentFieldsUpdated)
+                {
+                    var gdkUpdate = new SpatialOSExhaustiveBlittableSingular.Update();
```
jamiebrynes7: This is probably fine for now, but we should look at optimising this in the future given that we have already looped over this. Bitmask maybe?

workers/unity/Assets/Gdk/Core/Components/ReactiveComponents.cs
```diff
@@ -9,7 +9,7 @@ public abstract class MessagesReceived<T> : Component
         public List<T> Buffer = new List<T>();
     }
 
-    public class ComponentsUpdated<T> : MessagesReceived<T>
+    public class ComponentsUpdated<T> : MessagesReceived<T> where T : ISpatialComponentUpdate
```
jamiebrynes7: 👏

workers/unity/Assets/Gdk/Core/Components/Option.cs
samiwh: Standard exceptions don't tell you what has gone wrong. You can find the message but the main thing you get is the exception type.

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/ExhaustiveBlittableSingularTranslation.cs
JonasImprobable: In out current implementation, each component is guaranteed to have authority changed messages but will only have componentsUpdated messages if defined component fields. The change in the ordering is to reflect the change in importance.

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/ExhaustiveBlittableSingularTranslation.cs
JonasImprobable: Combined the two code blocks. We now preemptively make an update struct without knowing whether there will be field updates but that's probably fine.

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/ExhaustiveBlittableSingularTranslation.cs
JonasImprobable: Combined the two code blocks. We now preemptively make an update struct without knowing whether there will be field updates but that's probably fine.

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/TypeName.cs
```diff
@@ -94,6 +95,9 @@ public enum NestedEnum : uint
                 }
                 
             }
+
```
jessicafalk: where did all these new lines come from?

code_generator/Templates/UnityComponentConversionGenerator.tt
samiwh: You can get a completely empty update. I'm not sure where you would do this, but anything that sends only changes could easily do this. Especially if rather than sending a timestamp, something only sends the time delta since the last update.

JonasImprobable: To make sure I understand. Is it possible to send a completely empty component update by e.g. calling `connection.SendComponentUpdate<Foo>(new Foo.Update());`? This would bot be a trivial thing to support and I also wonder whether something like this should be conceptually permitted by the worker SDK (why not have some sort of ping event instead?) so I suggest we consciously don't support empty updates for now and start supporting them later if necessary.

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/TypeName.cs
JonasImprobable: This doesn't seem to be caused by my PR. cc @jamiebrynes7

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: Added bespoke exception type.

workers/unity/Assets/Gdk/Core/Components/Option.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    public class CalledValueOnEmptyOptionException : Exception
+    {
+        public CalledValueOnEmptyOptionException()
+        {
+        }
+
+        public CalledValueOnEmptyOptionException(string message)
```
samiwh: Do we need anything other than this constructor

JonasImprobable: We probably don't. I just thought it would be "proper style" to add the other two constructors too. See https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions

workers/unity/Assets/Gdk/Core/Components/Option.cs
```diff
@@ -0,0 +1,116 @@
...
+        public bool Equals(Option<T> other)
+        {
+            if (!HasValue && !other.HasValue)
+            {
+                return true;
+            }
+
+            if (HasValue && other.HasValue)
+            {
+                return value.Equals(other.Value);
```
samiwh: Does this mean we can only make options of equatable structs?

samiwh: Actually looked it up. The answer is no, but this is a boxing allocation. Do we use equality in the updates structs anywhere?

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/ExhaustiveBlittableSingularTranslation.cs
samiwh: That seems good to me. An empty update is still an update.

workers/unity/Assets/Gdk/Core/Components/Option.cs
JonasImprobable: Just tested this. This should work for everything except for options with null payloads where you will attempt to call `null.Equals(something else)`. That I can fix. Everything else, value types, reference types, structs, should provide a default Object.Equals method.

JonasImprobable: Oh shit. It's the boxing problem all over again.

code_generator/End2End/GeneratedDependencies/Generated/improbable/testschema/TypeName.cs
jamiebrynes7: 🤷‍♂️

workers/unity/Assets/Gdk/Physics/Systems/ApplyTransformUpdatesSystem.cs
```diff
@@ -21,8 +23,31 @@ protected override void OnUpdate()
         {
             for (var i = 0; i < transformUpdateData.Length; i++)
             {
-                var newUpdates = transformUpdateData.TransformUpdate[i].Buffer;
-                transformUpdateData.BufferedTransform[i].TransformUpdates.AddRange(newUpdates);
+                var transformUpdates = transformUpdateData.TransformUpdate[i].Buffer;
+                var lastTransformSnapshot = transformUpdateData.BufferedTransform[i].LastTransformSnapshot;
+                var transformSnapshots = new List<SpatialOSTransform>();
```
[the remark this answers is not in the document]
JonasImprobable: Removed temporary list allocation.

PR #53, opened 2018-07-02
UTY-309: Remove LogError and replace by a dispatcher and error objects.

martinzlocha:

-------

#### Description
We are using `Debug.LogError` in a few places in the GDK. This has been replaced by an ErrorDispatcher and in some cases by returning ConnectionError object.

Design doc: https://brevi.link/unity-gdk-log-error-replacement 

#### Tests
Ran the game sucessfully.

#### Documentation
N/A

#### Primary reviewers
@jessicafalk @samiwh

### DIFF
--- a/workers/unity/Assets/Gdk/Core/Config/ConnectionConfig.cs
+++ b/workers/unity/Assets/Gdk/Core/Config/ConnectionConfig.cs
@@ -14,7 +14,9 @@ namespace Improbable.Gdk.Core
         {
             if (string.IsNullOrEmpty(configValue))
             {
-                throw new System.ArgumentException($"No valid {configName} has been provided");
+                throw new ConnectionFailedException(
+                    $"Config validation failed with: No valid {configName} has been provided",
+                    ConnectionErrorReason.InvalidConfig);
             }
         }
     }
--- a/workers/unity/Assets/Gdk/Core/Error/ILogDispatcher.cs
+++ b/workers/unity/Assets/Gdk/Core/Error/ILogDispatcher.cs
@@ -1,0 +1,9 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public interface ILogDispatcher
+    {
+        void HandleLog(LogType type, LogEvent logEvent);
+    }
+}
--- a/workers/unity/Assets/Gdk/Core/Error/LogEvent.cs
+++ b/workers/unity/Assets/Gdk/Core/Error/LogEvent.cs
@@ -1,0 +1,42 @@
+using System.Collections.Generic;
+using System.Text;
+
+namespace Improbable.Gdk.Core
+{
+    public struct LogEvent
+    {
+        public string Message;
+        public Dictionary<string, object> Data;
+
+        public LogEvent(string message)
+        {
+            Message = message;
+            Data = new Dictionary<string, object>();
+        }
+
+        public LogEvent WithField(string key, object value)
+        {
+            Data.Add(key, value);
+            return this;
+        }
+
+        public override string ToString()
+        {
+            var builder = new StringBuilder();
+
+            builder.AppendLine(Message);
+
+            if (Data.Count > 0)
+            {
+                builder.AppendLine("Log event context:");
+
+                foreach (var entry in Data)
+                {
+                    builder.AppendLine($"'{entry.Key}': '{entry.Value}'");
+                }
+            }
+
+            return builder.ToString();
+        }
+    }
+}
--- a/workers/unity/Assets/Gdk/Core/Error/LoggingDispatcher.cs
+++ b/workers/unity/Assets/Gdk/Core/Error/LoggingDispatcher.cs
@@ -1,0 +1,29 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public class LoggingDispatcher : ILogDispatcher
+    {
+        public void HandleLog(LogType type, LogEvent logEvent)
+        {
+            switch (type)
+            {
+                case LogType.Error:
+                    Debug.LogError(logEvent);
+                    break;
+                case LogType.Warning:
+                    Debug.LogWarning(logEvent);
+                    break;
+                case LogType.Log:
+                    Debug.Log(logEvent);
+                    break;
+                case LogType.Assert:
+                    Debug.LogAssertion(logEvent);
+                    break;
+                case LogType.Exception:
+                    Debug.LogError(logEvent);
+                    break;
+            }
+        }
+    }
+}
--- a/workers/unity/Assets/Gdk/Core/Exceptions/ConnectionFailedException.cs
+++ b/workers/unity/Assets/Gdk/Core/Exceptions/ConnectionFailedException.cs
@@ -1,0 +1,21 @@
+using System;
+
+namespace Improbable.Gdk.Core
+{
+    public enum ConnectionErrorReason
+    {
+        CannotEstablishConnection,
+        DeploymentNotFound,
+        InvalidConfig
+    }
+
+    public class ConnectionFailedException : Exception
+    {
+        public ConnectionErrorReason Reason;
+
+        public ConnectionFailedException(string message, ConnectionErrorReason reason) : base(message)
+        {
+            Reason = reason;
+        }
+    }
+}
--- a/workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
+++ b/workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
@@ -14,6 +14,7 @@ namespace Improbable.Gdk.Core
     {
         public const long WorkerEntityId = -1337;
         public Entity WorkerEntity { get; }
+        public ILogDispatcher LogDispatcher { get; }
 
         public readonly Dictionary<int, ComponentTranslation> TranslationUnits =
             new Dictionary<int, ComponentTranslation>();
@@ -33,11 +34,14 @@ namespace Improbable.Gdk.Core
 
         private readonly Dictionary<long, Entity> entityMapping;
 
-        public MutableView(World world)
+        public const string UnexpectedAuthorityChangeError = "Unexpected state transition during authority change.";
+
+        public MutableView(World world, ILogDispatcher logDispatcher)
         {
             entityManager = world.GetOrCreateManager<EntityManager>();
             entityMapping = new Dictionary<long, Entity>();
             gameObjectManager = new GameObjectManager();
+            LogDispatcher = logDispatcher;
 
             setComponentObjectAction = (Action<Entity, ComponentType, object>) Delegate.CreateDelegate(
                 typeof(Action<Entity, ComponentType, object>), entityManager, setComponentObjectMethodInfo);
@@ -69,7 +73,10 @@ namespace Improbable.Gdk.Core
             Entity entity;
             if (!TryGetEntity(entityId, out entity))
             {
-                Debug.LogErrorFormat(Errors.NoCorrespondingEntityForEntityId, typeof(T).Name, entityId);
+                LogDispatcher.HandleLog(LogType.Error, new LogEvent(
+                        "Entity not found when attempting to add a component to an entity.")
+                    .WithField(IdType.EntityId.ToString(), entityId)
+                    .WithField("OpName", typeof(T).Name));
                 return;
             }
 
@@ -91,7 +98,10 @@ namespace Improbable.Gdk.Core
             Entity entity;
             if (!TryGetEntity(entityId, out entity))
             {
-                Debug.LogErrorFormat(Errors.NoCorrespondingEntityForEntityId, typeof(T).Name, entityId);
+                LogDispatcher.HandleLog(LogType.Error, new LogEvent(
+                        "Entity not found when attemting to remove a component from an entity.")
+                    .WithField(IdType.EntityId.ToString(), entityId)
+                    .WithField("OpName", typeof(T).Name));
             }
 
             RemoveComponent<T>(entity);
@@ -144,7 +154,10 @@ namespace Improbable.Gdk.Core
             Entity entity;
             if (!TryGetEntity(entityId, out entity))
             {
-                Debug.LogErrorFormat(Errors.NoCorrespondingEntityForEntityId, typeof(T).Name, entityId);
+                LogDispatcher.HandleLog(LogType.Error, new LogEvent(
+                        "Entity not found when attempting to set component object.")
+                    .WithField(IdType.EntityId.ToString(), entityId)
+                    .WithField("OpName", typeof(T).Name));
                 return;
             }
 
@@ -206,7 +219,9 @@ namespace Improbable.Gdk.Core
             Entity entity;
             if (!TryGetEntity(entityId, out entity))
             {
-                Debug.LogErrorFormat(Errors.NoCorrespondingEntityForEntityId, typeof(T).Name, entityId);
+                LogDispatcher.HandleLog(LogType.Error, new LogEvent("Entity not found during authority change.")
+                    .WithField(IdType.EntityId.ToString(), entityId)
+                    .WithField("OpName", typeof(T).Name));
                 return;
             }
 
@@ -215,8 +230,10 @@ namespace Improbable.Gdk.Core
                 case Authority.Authoritative:
                     if (!HasComponent<NotAuthoritative<T>>(entity))
                     {
-                        Debug.LogErrorFormat(Errors.IncorrectAuthorityTransition, Authority.Authoritative,
-                            Authority.NotAuthoritative);
+                        LogDispatcher.HandleLog(LogType.Error, new LogEvent(UnexpectedAuthorityChangeError)
+                            .WithField("Type", typeof(Authority))
+                            .WithField("NewState", Authority.Authoritative)
+                            .WithField("OldState", Authority.NotAuthoritative));
                         return;
                     }
 
@@ -226,8 +243,10 @@ namespace Improbable.Gdk.Core
                 case Authority.AuthorityLossImminent:
                     if (!HasComponent<Authoritative<T>>(entity))
                     {
-                        Debug.LogErrorFormat(Errors.IncorrectAuthorityTransition, Authority.AuthorityLossImminent,
-                            Authority.Authoritative);
+                        LogDispatcher.HandleLog(LogType.Error, new LogEvent(UnexpectedAuthorityChangeError)
+                            .WithField("Type", typeof(Authority))
+                            .WithField("NewState", Authority.AuthorityLossImminent)
+                            .WithField("OldState", Authority.Authoritative));
                         return;
                     }
 
@@ -236,8 +255,10 @@ namespace Improbable.Gdk.Core
                 case Authority.NotAuthoritative:
                     if (!HasComponent<Authoritative<T>>(entity))
                     {
-                        Debug.LogErrorFormat(Errors.IncorrectAuthorityTransition, Authority.NotAuthoritative,
-                            Authority.Authoritative);
+                        LogDispatcher.HandleLog(LogType.Error, new LogEvent(UnexpectedAuthorityChangeError)
+                            .WithField("Type", typeof(Authority))
+                            .WithField("NewState", Authority.NotAuthoritative)
+                            .WithField("OldState", Authority.Authoritative));
                         return;
                     }
 
@@ -272,7 +293,9 @@ namespace Improbable.Gdk.Core
         {
             if (entityMapping.ContainsKey(entityId))
             {
-                Debug.LogErrorFormat(Errors.AddEntityButEntityIdAlreadyExists, entityId);
+                LogDispatcher.HandleLog(LogType.Error, new LogEvent(
+                        "Tried to add an entity but there is already an entity associated with that EntityId.")
+                    .WithField(IdType.EntityId.ToString(), entityId));
                 return;
             }
 
@@ -292,7 +315,9 @@ namespace Improbable.Gdk.Core
             Entity entity;
             if (!TryGetEntity(entityId, out entity))
             {
-                Debug.LogErrorFormat(Errors.DeleteNonExistentEntity, entityId);
+                LogDispatcher.HandleLog(LogType.Error, new LogEvent(
+                        "Tried to delete an entity but there is already an entity associated with that EntityId.")
+                    .WithField(IdType.EntityId.ToString(), entityId));
                 return;
             }
 
@@ -327,21 +352,6 @@ namespace Improbable.Gdk.Core
             var messageBuffer = GetOrCreateComponent(entity, pool);
             messageBuffer.Buffer.Add(message);
             SetComponentObject(entity, messageBuffer);
-        }
-
-        private static class Errors
-        {
-            public static string NoCorrespondingEntityForEntityId =
-                "Received {0} for EntityId {1}, but there is no entity associated with that EntityId.";
-
-            public const string IncorrectAuthorityTransition =
-                "Unexpected authority transition to state: {0}. Expected previous state: {1}.";
-
-            public const string AddEntityButEntityIdAlreadyExists =
-                "Tried to add an entity with EntityId {0}, but there is already an entity associated with that EntityId.";
-
-            public const string DeleteNonExistentEntity =
-                "Tried to delete an entity with EntityId {0}, but there is no entity associated with that EntityId";
         }
     }
 }
--- a/workers/unity/Assets/Gdk/Core/Systems/CustomSpatialOSSendSystem.cs
+++ b/workers/unity/Assets/Gdk/Core/Systems/CustomSpatialOSSendSystem.cs
@@ -1,4 +1,5 @@
 using Unity.Entities;
+using UnityEngine;
 
 namespace Improbable.Gdk.Core
 {
@@ -16,7 +17,12 @@ namespace Improbable.Gdk.Core
             worker = WorkerRegistry.GetWorkerForWorld(World);
 
             spatialOSSendSystem = World.GetOrCreateManager<SpatialOSSendSystem>();
-            spatialOSSendSystem.RegisterCustomReplicationSystem(typeof(T));
+            if (!spatialOSSendSystem.TryRegisterCustomReplicationSystem(typeof(T)))
+            {
+                worker.View.LogDispatcher.HandleLog(LogType.Error, new LogEvent(
+                        "Custom Replication System for this component already exists.")
+                    .WithField("ComponentType", typeof(T)));
+            }
         }
     }
 }
--- a/workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
+++ b/workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
@@ -1,18 +1,11 @@
 using System.Collections.Generic;
 using Unity.Entities;
-using UnityEngine;
 
 namespace Improbable.Gdk.Core
 {
     [UpdateInGroup(typeof(SpatialOSSendGroup.InternalSpatialOSSendGroup))]
     public class SpatialOSSendSystem : ComponentSystem
     {
-        internal static class Warnings
-        {
-            public const string CustomReplicationSystemAlreadyExists =
-                "Custom Replication System for type {0} already exists!";
-        }
-
         private WorkerBase worker;
         private MutableView view;
 
@@ -44,15 +37,15 @@ namespace Improbable.Gdk.Core
             }
         }
 
-        public void RegisterCustomReplicationSystem(ComponentType type)
+        public bool TryRegisterCustomReplicationSystem(ComponentType type)
         {
             if (!registeredReplicators.Contains(type.TypeIndex))
             {
-                Debug.LogWarningFormat(Warnings.CustomReplicationSystemAlreadyExists, type);
-                return;
+                return false;
             }
 
-            registeredReplicators.Remove(type.TypeIndex);
+            // The default replication system is removed, instead the custom one is responsible for replication.
+            return registeredReplicators.Remove(type.TypeIndex);
         }
 
         protected override void OnUpdate()
--- a/workers/unity/Assets/Gdk/Core/Utility/ConnectionUtility.cs
+++ b/workers/unity/Assets/Gdk/Core/Utility/ConnectionUtility.cs
@@ -8,15 +8,7 @@ namespace Improbable.Gdk.Core
     {
         public static Connection ConnectToSpatial(ReceptionistConfig config, string workerType, string workerId)
         {
-            try
-            {
-                config.Validate();
-            }
-            catch (System.ArgumentException e)
-            {
-                Debug.LogError($"Config validation failed with: {e.Message}");
-                return null;
-            }
+            config.Validate();
 
             Debug.Log("Attempting connection to SpatialOS...");
 
@@ -31,24 +23,12 @@ namespace Improbable.Gdk.Core
 
         public static Connection LocatorConnectToSpatial(LocatorConfig config, string workerType)
         {
-            try
-            {
-                config.Validate();
-            }
-            catch (System.ArgumentException e)
-            {
-                Debug.LogError($"Config validation failed with: {e.Message}");
-                return null;
-            }
+            config.Validate();
 
             using (var locator = new Locator(config.LocatorHost, config.LocatorParameters))
             {
                 Debug.Log("Attempting to retrieve deployment name...");
                 var deploymentName = GetDeploymentName(locator);
-                if (deploymentName == null)
-                {
-                    return null;
-                }
 
                 Debug.Log("Successfully obtained deployment name!");
 
@@ -95,8 +75,8 @@ namespace Improbable.Gdk.Core
 
             if (!connection.HasValue || !connection.Value.IsConnected)
             {
-                Debug.LogError("Failed to connect to SpatialOS.");
-                return null;
+                throw new ConnectionFailedException("Failed to connect to SpatialOS.",
+                    ConnectionErrorReason.CannotEstablishConnection);
             }
 
             Debug.Log("Successfully connected to SpatialOS!");
@@ -112,20 +92,21 @@ namespace Improbable.Gdk.Core
 
                 if (!deployments.HasValue)
                 {
-                    Debug.LogError("Failed to retrieve deployment.");
-                    return null;
+                    throw new ConnectionFailedException("Failed to retrieve deployment.",
+                        ConnectionErrorReason.DeploymentNotFound);
                 }
 
                 if (deployments.Value.Error != null)
                 {
-                    Debug.LogError($"Failed to obtain deployment name with error: {deployments.Value.Error}.");
-                    return null;
+                    throw new ConnectionFailedException(
+                        $"Failed to obtain deployment name with error: {deployments.Value.Error}.",
+                        ConnectionErrorReason.DeploymentNotFound);
                 }
 
                 if (deployments.Value.Deployments.Count == 0)
                 {
-                    Debug.LogError("Received an empty list of deployments.");
-                    return null;
+                    throw new ConnectionFailedException("Received an empty list of deployments.",
+                        ConnectionErrorReason.DeploymentNotFound);
                 }
 
                 return deployments.Value.Deployments[0].DeploymentName;
--- a/workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
+++ b/workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
@@ -19,18 +19,22 @@ namespace Improbable.Gdk.Core
 
         public abstract string GetWorkerType { get; }
 
-        protected WorkerBase(string workerId, Vector3 origin)
+        protected WorkerBase(string workerId, Vector3 origin) : this(workerId, origin, new LoggingDispatcher())
+        {
+        }
+
+        protected WorkerBase(string workerId, Vector3 origin, ILogDispatcher loggingDispatcher)
         {
             if (string.IsNullOrEmpty(workerId))
             {
-                throw new ArgumentException("WorkerId is null or empty.");
+                throw new ArgumentException("WorkerId is null or empty.", nameof(workerId));
             }
 
             WorkerId = workerId;
             World = new World(WorkerId);
             WorkerRegistry.SetWorkerForWorld(this);
 
-            View = new MutableView(World);
+            View = new MutableView(World, loggingDispatcher);
             Origin = origin;
         }
 
@@ -41,7 +45,7 @@ namespace Improbable.Gdk.Core
             World.Dispose();
         }
 
-        public bool Connect(ConnectionConfig config)
+        public void Connect(ConnectionConfig config)
         {
             if (config is ReceptionistConfig)
             {
@@ -51,10 +55,10 @@ namespace Improbable.Gdk.Core
             {
                 Connection = ConnectionUtility.LocatorConnectToSpatial((LocatorConfig) config, GetWorkerType);
             }
-
-            if (Connection == null)
+            else
             {
-                return false;
+                throw new InvalidConfigurationException($"Invalid connection config was provided: '{config}' Only" +
+                    "ReceptionistConfig and LocatorConfig are supported.");
             }
 
             Application.quitting += () =>
@@ -62,8 +66,8 @@ namespace Improbable.Gdk.Core
                 ConnectionUtility.Disconnect(Connection);
                 Connection = null;
             };
+
             View.Connect();
-            return true;
         }
 
         public virtual void RegisterSystems()
--- a/workers/unity/Assets/Gdk/Core/Worker/WorkerRegistry.cs
+++ b/workers/unity/Assets/Gdk/Core/Worker/WorkerRegistry.cs
@@ -19,7 +19,7 @@ namespace Improbable.Gdk.Core
         {
             if (WorldToWorker.ContainsKey(worker.World))
             {
-                throw new Exception(string.Format("A worker is already stored for world {0}", worker.World.Name));
+                throw new ArgumentException($"A worker is already stored for world '{worker.World.Name}'");
             }
 
             WorldToWorker[worker.World] = worker;
@@ -65,8 +65,7 @@ namespace Improbable.Gdk.Core
             Func<string, Vector3, WorkerBase> createWorker;
             if (!WorkerTypeToInitializationFunction.TryGetValue(workerType, out createWorker))
             {
-                Debug.LogErrorFormat("No worker found for worker type {0}", workerType);
-                return null;
+                throw new ArgumentException("No worker found for worker type '{0}'", workerType);
             }
 
             return createWorker(workerId, origin);
--- a/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
+++ b/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
@@ -68,11 +68,6 @@ namespace Playground
                     ? WorkerRegistry.CreateWorker<UnityClient>($"{workerType}-{Guid.NewGuid()}", new Vector3(0, 0, 0))
                     : WorkerRegistry.CreateWorker(workerType, workerId, new Vector3(0, 0, 0));
 
-                if (worker == null)
-                {
-                    return;
-                }
-
                 Workers.Add(worker);
 
                 connectionConfig = ConnectionUtility.CreateConnectionConfigFromCommandLine(commandLineArgs);
@@ -80,8 +75,9 @@ namespace Playground
 
             if (World.AllWorlds.Count <= 0)
             {
-                Debug.LogError(
-                    "No worlds have been created, due to invalid worker types being specified.");
+                throw new InvalidConfigurationException(
+                    "No worlds have been created, due to invalid worker types being specified. Check the config in" +
+                    "Improbable -> Configure editor workers.");
             }
 
             var worlds = World.AllWorlds.ToArray();
@@ -95,7 +91,16 @@ namespace Playground
             foreach (var worker in Workers)
             {
                 LoadLevel(worker);
-                worker.Connect(connectionConfig);
+
+                try
+                {
+                    worker.Connect(connectionConfig);
+                }
+                catch (ConnectionFailedException exception)
+                {
+                    worker.View.LogDispatcher.HandleLog(LogType.Error, new LogEvent(exception.Message)
+                        .WithField("Reason", exception.Reason));
+                }
             }
         }
 
--- a/workers/unity/Assets/Playground/Scripts/Spawning/ArchetypeInitializationSystem.cs
+++ b/workers/unity/Assets/Playground/Scripts/Spawning/ArchetypeInitializationSystem.cs
@@ -48,7 +48,10 @@ namespace Playground
 
             if (!(worker is UnityClient) && !(worker is UnityGameLogic))
             {
-                Debug.LogErrorFormat(Errors.UnknownWorkerType, World.Name);
+                view.LogDispatcher.HandleLog(LogType.Error, new LogEvent(
+                        "Worker type isn't supported by the ArchetypeInitializationSystem.")
+                    .WithField("WorldName", World.Name)
+                    .WithField("WorkerType", worker));
             }
 
             workerType = worker.GetWorkerType;
@@ -65,7 +68,10 @@ namespace Playground
                 if (!ArchetypeConfig.WorkerTypeToArchetypeNameToComponentTypes[workerType]
                     .TryGetValue(archetypeName, out componentTypesToAdd))
                 {
-                    Debug.LogErrorFormat(Errors.MappingNotFound, workerType, archetypeName);
+                    view.LogDispatcher.HandleLog(LogType.Error, new LogEvent(
+                            "No corresponding archetype mapping found.")
+                        .WithField("ArchetypeName", archetypeName)
+                        .WithField("WorkerType", workerType));
                     continue;
                 }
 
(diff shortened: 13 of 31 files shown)


martinzlocha:
@jamiebrynes7 The doc is: https://brevi.link/unity-gdk-log-error-replacement

workers/unity/Assets/Gdk/Core/Config/ScriptableWorkerConfiguration.asset
```diff
@@ -3,7 +3,7 @@
 --- !u!114 &11400000
 MonoBehaviour:
   m_ObjectHideFlags: 0
-  m_PrefabParentObject: {fileID: 0}
```
jessicafalk: you'll probably need to rebase on master. I just merged that fix in

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
```diff
@@ -48,7 +48,7 @@ public void RegisterCustomReplicationSystem(ComponentType type)
         {
             if (!registeredReplicators.Contains(type.TypeIndex))
             {
-                Debug.LogWarningFormat(Warnings.CustomReplicationSystemAlreadyExists, type);
+                view.ErrorDispatcher.LogErrorFormat(Warnings.CustomReplicationSystemAlreadyExists, type);
```
jessicafalk: do we want to change that to an error?

martinzlocha: I think it should be considered as an error, you shouldn't ever be registering the same system multiple times.

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -68,19 +68,14 @@ public void Awake()
...
-
                 Workers.Add(worker);
 
                 connectionConfig = ConnectionUtility.CreateConnectionConfigFromCommandLine(commandLineArgs);
             }
 
             if (World.AllWorlds.Count <= 0)
             {
-                Debug.LogError(
+                throw new ArgumentException(
```
samiwh: Please make new exception types for specific errors rather than use the base ones.
Also this isn't an argument exception.

samiwh:
Generally it's a lot easier to read if you split different units of work into different commits.

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
```diff
@@ -0,0 +1,12 @@
+namespace Improbable.Gdk.Core
+{
+    public struct ConnectionError
+    {
+        public readonly string Message;
```
samiwh: The point of this object is to help a user write retry logic. As a rule of thumb a string is human readable, enums etc are machine readable. So a string is needed for you to be able to log things in the case of failure. But you need some other type of error data for this to be useful.

workers/unity/Assets/Gdk/Core/Error/IErrorDispatcher.cs
```diff
@@ -0,0 +1,9 @@
+namespace Improbable.Gdk.Core
+{
+    public interface IErrorDispatcher
+    {
+        void LogError(string message);
+
+        void LogErrorFormat(string format, params object[] args);
```
samiwh: I don't think there has been any discussion as to what kind of dispatcher we want, or how it should be exposed.

I also don't think providing string formatting is the kind of thing an error dispatcher really cares about. String formatting is a native things to C# anyway.
You need something which specifies an error well enough that the callback can decide what to do with it (log locally, log to spatial, quit, try to fix it in some rare case, set some flag somewhere else etc...)
A string message is useful for logging, but you likely want to give more information than that.

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
samiwh: This could just be made into try register and return a bool.

samiwh: Then log as a warning.

jessicafalk: Yeah, this doesn't need to be an error. You shouldn't register a system multiple times, but it won't break anything

workers/unity/Assets/Gdk/Core/Error/IErrorDispatcher.cs
jessicafalk: Martin mentioned this to me yesterday. While I agree that an dispatcher shouldn't care about it, it mirros Unity's `Debug` class which makes it simple for the user.
In the end it depends on what we want to do with the dispatcher. Should it be a "logging dispatcher" or an "error handling dispatcher".  I would argue that we will actually need both to facilitate sending logging messages to spatial and to handle errors in a nice way,

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
martinzlocha: Added an enum field.

workers/unity/Assets/Gdk/Core/Error/IErrorDispatcher.cs
martinzlocha: The reasoning for having LogErrorFormat was that we use the method a lot in our code with Debug and I felt that it would be ugly to wrap everything in string.Format and only have LogError.

I would say that we primarily want this to be a logging error dispatcher. If we can handle an error an error object should be returned instead.

jessicafalk: This is how we did it in the SDK. This does not work completely as it won't allow us to add additional parameters (e.g. entity id) to it. We basically lose all context.

I don't think there is a reason for having a logging error dispatcher. Either it's a generic logging dispatcher or it's actually handling the error in some way

martinzlocha: Ok I think it will be best if I add a few possible implementations into the doc and we can decide based on that.

samiwh: The point of this isn't to be like Unity's debug class (unless we decide that is needed). It's to generally register that there should be an error. A string doesn't even need to be the main part of that api, it just needs the ability to send some sort of object that describes an error. Having multiple overloaded for string formatting seems odd.

jessicafalk: ok, we want a dispatcher that will allow the user to handle an error in some way. This could be printing an error message, but this could also be some kind of retry-logic. in that case I agree that this shouldn't mirror Unity's debug class

jessicafalk: Though we will need to extend the current logging at some point to be able to send debugging statements to spatial (not as part of this pr) while still logging inside the editor

samiwh: You also don't need to use string.format (although most IDEs seem to prefer it to LogFormat). You can use `$"some formatted string with {someVariable} in it"`.

samiwh: We do want the default dispatcher to just log though.

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    public struct ConnectionError
+    {
+        public readonly string Message;
+        public readonly ConnectionErrorReseaon Reseaon;
+
+        public ConnectionError(string message, ConnectionErrorReseaon reseaon)
+        {
+            Message = message;
+            Reseaon = reseaon;
```
samiwh: Typo :P

workers/unity/Assets/Gdk/Core/Error/InvalidConfigurationException.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Improbable.Gdk.Core
+{
+    public class InvalidConfigurationException : Exception
+    {
+        public InvalidConfigurationException()
+        {
+        }
+
+        public InvalidConfigurationException(string message) : base(message)
```
samiwh: I think this is the only constructor we need.

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
```diff
@@ -44,15 +44,15 @@ private void GenerateComponentGroups()
             }
         }
 
-        public void RegisterCustomReplicationSystem(ComponentType type)
+        public bool RegisterCustomReplicationSystem(ComponentType type)
```
samiwh: If it returns bool, it's best to call it Try...

workers/unity/Assets/Gdk/Core/Error/IErrorDispatcher.cs
martinzlocha: @samiwh @jessicafalk I have added two proposed solutions into the original design doc. One of them is the current one and the second uses Error Objects similar to the ones which are returned by worker.Connect()

martinzlocha: Ok I have implemented the error handler to use the IErrorContext interface. Some errors were very specific so I didnt not create the struct in a separate file because they were only used in one spot. For the others I created them separately.

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
```diff
@@ -0,0 +1,20 @@
+namespace Improbable.Gdk.Core
+{
+    public enum ConnectionErrorReseaon
+    {
+        CannotEstablishConnection,
+        DeploymentNotFound,
+        InvalidConfig
+    }
+
+    public struct ConnectionError : IErrorContext
```
jessicafalk: just checked out your branch and it can't find `IErrorContext`. did you forget to add a file?

workers/unity/Assets/Gdk/Core/Utility/ConnectionUtility.cs
```diff
@@ -89,45 +75,67 @@ private static ConnectionParameters CreateConnectionParameters(ConnectionConfig
...
+                    error = new ConnectionError
+                    {
+                        Message = "Received an empty list of deployments.",
+                        Reason = ConnectionErrorReseaon.DeploymentNotFound
+                    };
+
                     return null;
                 }
 
+                error = default(ConnectionError);
```
jessicafalk: I don't think we should assign error any value, if there is no error

workers/unity/Assets/Gdk/Core/Utility/ConnectionUtility.cs
```diff
@@ -89,45 +75,67 @@ private static ConnectionParameters CreateConnectionParameters(ConnectionConfig
...
                     return null;
                 }
 
                 if (deployments.Value.Deployments.Count == 0)
                 {
-                    Debug.LogError("Received an empty list of deployments.");
+                    error = new ConnectionError
+                    {
+                        Message = "Received an empty list of deployments.",
+                        Reason = ConnectionErrorReseaon.DeploymentNotFound
```
jessicafalk: misspelled `ConnectionErrorReseaon` :)

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -19,18 +19,22 @@ public abstract class WorkerBase : IDisposable
 
         public abstract string GetWorkerType { get; }
 
-        protected WorkerBase(string workerId, Vector3 origin)
+        protected WorkerBase(string workerId, Vector3 origin) : this(workerId, origin, new LoggingDispatcher())
```
jessicafalk: remove this constructor and `IErrorDispatcher errorDispatcher` optional / assign it to `new LoggingDispatcher()` as the default value

workers/unity/Assets/Gdk/Core/Error/IErrorDispatcher.cs
```diff
@@ -0,0 +1,9 @@
+namespace Improbable.Gdk.Core
+{
+    public interface IErrorDispatcher
+    {
+        void HandleError(IErrorContext context);
+
+        void HandleWarning(IErrorContext context);
```
jessicafalk: on second thought: do we need a `HandleWarning` in an `IErrorDispatcher`? 
We always only talked about error handling, warnings don't really need to be treated differently as they shouldn't drastically change the overall execution flow (if they are, they should be errors imo)

workers/unity/Assets/Gdk/Core/Config/ConnectionConfig.cs
```diff
@@ -14,7 +14,8 @@ protected void ValidateConfig(string configValue, string configName)
         {
             if (string.IsNullOrEmpty(configValue))
             {
-                throw new System.ArgumentException($"No valid {configName} has been provided");
+                throw new System.ArgumentException(
```
jessicafalk: let's use `InvalidConfigurationException ` here too :)

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -211,8 +233,12 @@ public void AddGameObjectEntity(Entity entity, GameObject gameObject)
                 case Authority.Authoritative:
                     if (!HasComponent<NotAuthoritative<T>>(entity))
                     {
-                        Debug.LogErrorFormat(Errors.IncorrectAuthorityTransition, Authority.Authoritative,
-                            Authority.NotAuthoritative);
+                        ErrorDispatcher.HandleError(new InvalidStateTransition
```
jessicafalk: that error struct is also missing :)

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -211,8 +233,12 @@ public void AddGameObjectEntity(Entity entity, GameObject gameObject)
...
+                        {
+                            NewAuthority = Authority.Authoritative,
+                            OldAuthority = Authority.NotAuthoritative,
+                            Attribute = "authority"
```
jessicafalk: as I don't see the actual implementation, it's a bit hard to tell whether `Attribute` is really needed. If yes, don't use a string, but create an enum

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
martinzlocha: Yup :/ Forgot to add all new files. Added.

workers/unity/Assets/Gdk/Core/Utility/ConnectionUtility.cs
martinzlocha: If you pass an `out` parameter to a function it always has to be assigned. Even if this is not intuitive I think we have to keep it. Its similar to the implementation of TryGetValue.

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
martinzlocha: Changed to be more generic.

workers/unity/Assets/Gdk/Core/Error/IErrorDispatcher.cs
martinzlocha: One reason to keep them separate could be that for warnings you only want to log them and for errors you want to stop the game and show an error window for example.

jessicafalk: but then you can just use `Debug.LogWarning`: is there ever a reason why we would want to do something different from just logging them

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
```diff
@@ -0,0 +1,20 @@
+namespace Improbable.Gdk.Core
+{
+    public enum ConnectionErrorReason
+    {
+        CannotEstablishConnection,
+        DeploymentNotFound,
+        InvalidConfig
+    }
+
+    public struct ConnectionError : IErrorContext
```
jessicafalk: using structs results in boxing for every single `HandleError` call. Would it be better to use classes here? @samiwh as he is more familiar with boxing rules then I am

workers/unity/Assets/Gdk/Core/Error/IErrorDispatcher.cs
martinzlocha: I'm not sure if there is anything else. I guess its a trade off between having and extra option to handle stuff differently if its a warning/error or making it simpler for someone to implement an IErrorDispatcher. But I don't think doing Debug.LogWarning in some cases instead of passing stuff to the ErrorHandler is the right approach because you might decide that you want to handle warnings differently and then you will have to change it everywhere you used Debug.LogWarning.

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
jamiebrynes7: It will box if it needs to convert to an object (which seems like it will given the use of an interface)

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public string Message;
+        public ConnectionErrorReason Reason;
+
+        public override string ToString()
+        {
+            return Message;
```
jamiebrynes7: Do we want to include the reason enum in the string or do we expect the message to encapsulate that information?

workers/unity/Assets/Gdk/Core/Error/IErrorDispatcher.cs
jamiebrynes7: Equally, this interface needs another name then 😉

workers/unity/Assets/Gdk/Core/Error/InvalidEntityOperation.cs
```diff
@@ -0,0 +1,13 @@
+namespace Improbable.Gdk.Core
+{
+    public struct InvalidEntityOperation : IErrorContext
+    {
+        public long EntityId;
+        public string Message;
+
+        public override string ToString()
+        {
+            return Message;
```
jamiebrynes7: Similar comment, do we want to also print the `EntityId` here?

workers/unity/Assets/Gdk/Core/Error/InvalidStateTransition.cs
```diff
@@ -0,0 +1,15 @@
+using Improbable.Worker;
+
+namespace Improbable.Gdk.Core
+{
+    public struct InvalidStateTransition<T> : IErrorContext
```
jamiebrynes7: Do we want a constraint on `T` ?

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -69,7 +71,12 @@ public void Dispose()
             Entity entity;
             if (!TryGetEntity(entityId, out entity))
             {
-                Debug.LogErrorFormat(Errors.NoCorrespondingEntityForEntityId, typeof(T).Name, entityId);
+                ErrorDispatcher.HandleError(new EntityNotFound
+                {
+                    OpName = typeof(T).Name,
+                    Id = entityId,
```
jamiebrynes7: Please make the order of the initialisation consistent here :)

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -222,8 +247,11 @@ public void AddGameObjectEntity(Entity entity, GameObject gameObject)
                 case Authority.AuthorityLossImminent:
                     if (!HasComponent<Authoritative<T>>(entity))
                     {
-                        Debug.LogErrorFormat(Errors.IncorrectAuthorityTransition, Authority.AuthorityLossImminent,
-                            Authority.Authoritative);
+                        ErrorDispatcher.HandleError(new InvalidStateTransition<Authority>
+                        {
+                            NewState = Authority.AuthorityLossImminent,
+                            OldState = Authority.NotAuthoritative
```
jamiebrynes7: The original arguments were `Authority.AuthorityLossImminent` and `Authority.Authoritative`. Was this an intended change?

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
```diff
@@ -1,16 +1,19 @@
...
 using Unity.Entities;
-using UnityEngine;
 
 namespace Improbable.Gdk.Core
 {
     [UpdateInGroup(typeof(SpatialOSSendGroup.InternalSpatialOSSendGroup))]
     public class SpatialOSSendSystem : ComponentSystem
     {
-        internal static class Warnings
+        public struct CustomReplicationSystemAlreadyExists : IErrorContext
```
jamiebrynes7: Does this struct need to be `public` ?

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -19,18 +20,23 @@ public abstract class WorkerBase : IDisposable
 
         public abstract string GetWorkerType { get; }
 
-        protected WorkerBase(string workerId, Vector3 origin)
+        protected WorkerBase(string workerId, Vector3 origin, Option<IErrorDispatcher> errorDispatcher)
```
jamiebrynes7: This `Option<>` basically serves as a default argument? Is there any reason you can't do: 

`protected WorkerBase(string workerId, Vector3 origin, IErrorDispatcher errorDispatcher = new LoggingDispatcher())`

workers/unity/Assets/Playground/Scripts/Player/ProcessLaunchCommandSystem.cs
```diff
@@ -18,8 +18,7 @@ private struct LaunchCommandData
             [ReadOnly] public EntityArray Entity;
             public ComponentDataArray<SpatialOSLauncher> Launcher;
 
-            [ReadOnly]
-            public ComponentArray<CommandRequests<Generated.Playground.Launcher.LaunchEntity.Request>> CommandRequests;
+            [ReadOnly] public ComponentArray<CommandRequests<Generated.Playground.Launcher.LaunchEntity.Request>> CommandRequests;
```
jamiebrynes7: Please revert this (if you merge with master you'll have the linting fixes that stop this flip flopping)

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
martinzlocha: I think keeping it public allows users to verify whether the error they got in the handler is of type CustomReplicationSystemAlreadyExists.

martinzlocha:
After offline discussion decision has been reached to make the ErrorHandler more general and call it LogHandler.

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
samiwh: It will however this is an infrequently called api (also boxing almost always has to happen on actually logging objects). 
We could try allocating a class from a pool in future if we find this is creating an appreciable amount of garbage. I would also expect an IL2CPP compiler to optimise this out.

samiwh: TL;DR I don't think this is a performance bottleneck, but it's easy to revisit if we need to. Could also make this just a struct and have other meta data in a user defined context obejct, which will be null by default. That way nothing would be unboxed (and possibly boxed), but it would be a worse API.

workers/unity/Assets/Gdk/Core/Error/ConnectionError.cs
samiwh: I'd just have a string field and not bother with the method, but I don't have a strong opinion here. The logger may well add it's own stuff to whatever it outputs.

workers/unity/Assets/Gdk/Core/Error/LoggingDispatcher.cs
```diff
@@ -0,0 +1,30 @@
+using Debug = UnityEngine.Debug;
+
+namespace Improbable.Gdk.Core
+{
+    public enum LogLevel
```
jared-improbable: Why do we need yet another Log level enum? Unity already has https://docs.unity3d.com/ScriptReference/LogType.html

workers/unity/Assets/Gdk/Core/Error/UnknownWorkerType.cs
```diff
@@ -0,0 +1,12 @@
+namespace Improbable.Gdk.Core
+{
+    public struct UnknownWorkerType : ILogContext
```
jared-improbable: Wouldn't it be handy to include the name of the unknown worker type?

workers/unity/Assets/Gdk/Core/Systems/CustomSpatialOSSendSystem.cs
```diff
@@ -16,7 +16,7 @@ protected override void OnCreateManager(int capacity)
             worker = WorkerRegistry.GetWorkerForWorld(World);
 
             spatialOSSendSystem = World.GetOrCreateManager<SpatialOSSendSystem>();
-            spatialOSSendSystem.RegisterCustomReplicationSystem(typeof(T));
+            spatialOSSendSystem.TryRegisterCustomReplicationSystem(typeof(T));
```
jared-improbable: Not checking the return value? Shouldn't the error handling for this be outside of `TryRegister...`?
If something starts with `Try`, I wouldn't expect it to do any error reporting, and to leave it up to the caller.

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
```diff
@@ -44,15 +47,19 @@ private void GenerateComponentGroups()
...
         }
 
-        public void RegisterCustomReplicationSystem(ComponentType type)
+        public bool TryRegisterCustomReplicationSystem(ComponentType type)
         {
             if (!registeredReplicators.Contains(type.TypeIndex))
             {
-                Debug.LogWarningFormat(Warnings.CustomReplicationSystemAlreadyExists, type);
-                return;
+                view.LogDispatcher.HandleLog(new CustomReplicationSystemAlreadyExists
```
jared-improbable: This doesn't belong here - it should be handled by the caller.

workers/unity/Assets/Gdk/Core/Utility/ConnectionUtility.cs
```diff
@@ -89,45 +75,67 @@ private static ConnectionParameters CreateConnectionParameters(ConnectionConfig
             return connectionParameters;
         }
 
-        private static Connection TryToConnect(Future<Connection> connectionFuture)
+        private static Connection TryToConnect(Future<Connection> connectionFuture, out ConnectionError error)
```
jared-improbable: This feels odd to me.
Why not return a struct something like
```
struct ConnectionResult
{
	private Connection connection;
	public Error? {get;}
    public 	Connection {get {Error.HasValue ? throw : connection }  }
}
```

workers/unity/Assets/Gdk/TestUtils/UnityTestWorker.cs
```diff
@@ -1,3 +1,4 @@
+using Improbable.Collections;
```
jessicafalk: import not needed?

workers/unity/Assets/Gdk/TestUtils/UnityTestWorker.cs
```diff
@@ -9,7 +10,7 @@ public class UnityTestWorker : WorkerBase
 
         public override string GetWorkerType => WorkerType;
 
-        public UnityTestWorker(string workerId, Vector3 origin) : base(workerId, origin)
+        public UnityTestWorker(string workerId, Vector3 origin) : base(workerId, origin, new LoggingDispatcher())
```
jessicafalk: no need to pass in the loggingdispatcher here

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -68,19 +68,14 @@ public void Awake()
                     ? WorkerRegistry.CreateWorker<UnityClient>($"{workerType}-{Guid.NewGuid()}", new Vector3(0, 0, 0))
                     : WorkerRegistry.CreateWorker(workerType, workerId, new Vector3(0, 0, 0));
 
-                if (worker == null)
```
jessicafalk: why did this get removed? The worker could still be null

workers/unity/Assets/Playground/Scripts/Workers/UnityClient.cs
```diff
@@ -1,3 +1,4 @@
+using Improbable.Collections;
```
jessicafalk: remove that import

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -19,18 +19,22 @@ public abstract class WorkerBase : IDisposable
...
-                throw new ArgumentException("WorkerId is null or empty.");
+                throw new ArgumentException("WorkerId is null or empty.", workerId);
             }
 
             WorkerId = workerId;
             World = new World(WorkerId);
             WorkerRegistry.SetWorkerForWorld(this);
 
-            View = new MutableView(World);
+            View = new MutableView(World, errorDispatcher);
```
jessicafalk: rename to loggingDispatcher instead of errorDispatcher :)

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
martinzlocha: I dont think it can be null anymore because now an exception is thrown in the case where it was null.

workers/unity/Assets/Gdk/Core/Utility/ConnectionUtility.cs
martinzlocha: One issue with this approach which I have is that I also call `string GetDeploymentName(Locator locator, out ConnectionError error)` within `TryToConnect`. So I would also have to change this to return a struct with a string and an error to be consistent. I think thats uglier.

jared-improbable: I strongly disagree. Using an `out` parameter to return an error isn't idiomatic C#. Throwing an exception is, but apparently we don't want to do that for some reason.

martinzlocha: @samiwh disagreed with throwing an exception when it is expected that it can happen and should be part of the normal workflow.

jessicafalk: I agree with that in general. The exact use case for the error context object was, if I remember correctly, when an error happens in one of the systems. In that case throwing an exception might not work as expected, but you'll want to handle that error in some way.

Here: I agree, would prefer to throw an exception

samiwh: If we think errors codes will be harder to write or more confusing to users because they are less idiomatic then I'm happy to use exceptions instead. Exceptions also play very slightly nicer with tasks (and we have played around with making connect an async task).

There are two main reasons I was against using exceptions.

1. These failures aren't exceptional, they are just normal control flow; connections can sometimes fail when everything is configured correctly. With a file that usually means it needs user intervention to correct. I'd argue this is more akin to `Dictionary.TryGetValue`.
If we disagree on this then falling back to exceptions seems like a reasonable default.

2. I personally find try catch blocks hard to read, and retry logic for things that throw exceptions can look odd.
(However I will conceded that the code for generating the error often looks better)

samiwh: The other reason was unity catching all exceptions and making them easy to miss.

martinzlocha: If you want to see the possible implementation of using exceptions its in #c68689a. If we decide to go with error objects we can reverse it.

martinzlocha:
@jessicafalk @samiwh @jared-improbable @jamiebrynes7 I have reworked the LogDispatcher based on Jareds comments, please have a look.

code_generator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -108,7 +108,10 @@ namespace <#= qualifiedNamespace #>
                 Unity.Entities.Entity entity;
                 if (!view.TryGetEntity(op.EntityId.Id, out entity))
                 {
-                    Debug.LogErrorFormat(TranslationErrors.OpReceivedButNoEntity, op.GetType().Name, op.EntityId.Id);
+                    view.LogDispatcher.HandleLog(new LogEvent("Entity not found.", LogType.Error)
```
samiwh: Entity not found doesn't really give enough information. It should say in what context it was looking.

code_generator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -108,7 +108,10 @@ namespace <#= qualifiedNamespace #>
...
+                        .WithField("Id", op.EntityId.Id)
+                        .WithField("IdType", IdType.EntityId)
```
samiwh: This doesn't let you put both an ecs and a spatail ID in one message. Might be better to key the id on the id type.

workers/unity/Assets/Playground/Scripts/Spawning/ArchetypeInitializationSystem.cs
```diff
@@ -48,7 +48,9 @@ protected override void OnCreateManager(int capacity)
 
             if (!(worker is UnityClient) && !(worker is UnityGameLogic))
             {
-                Debug.LogErrorFormat(Errors.UnknownWorkerType, World.Name);
+                view.LogDispatcher.HandleLog(new LogEvent("Unknown worker type.", LogType.Error)
```
samiwh: Same comment as before. The message should tell the user everything they need for how to fix this. So ideally there should be some context in this. In this case you can actually say exactly what has gone wrong and how to fix it.

workers/unity/Assets/Gdk/Core/Error/LoggingDispatcher.cs
samiwh: Possible reason for another type - the number of SpatialOS log levels is larger than the Unity one. I don't think it's a massive problem if we only expose log, error, and warning though.

Other possible reason is that the logger is slightly more general than what Unity has. A user could want to handle fatal errors for example.

samiwh:
@jared-improbable (can't reply to the original comment) Is the objection to multiple types that information was missing or that, creating types is awkward?

Background - The original idea for types is that a handler could do things for specific messages. E.g. reduce noise when receiving the same message multiple times, log counts of errors to a dev console etc. Also means that specific messages can be handled in a specific way. With a completely generic approach it becomes hard to identify a particular error.

workers/unity/Assets/Gdk/Core/Exceptions/ConnectionFailedException.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Improbable.Gdk.Core
+{
+    public enum ConnectionErrorReason
+    {
+        CannotEstablishConnection,
+        DeploymentNotFound,
+        InvalidConfig
+    }
+
+    public class ConnectionFailedException : Exception
```
jared-improbable: Wouldn't it be useful to ensure the `ToString` of this exception includes the error reason?

workers/unity/Assets/Gdk/Core/Systems/CustomSpatialOSSendSystem.cs
```diff
@@ -16,7 +17,12 @@ protected override void OnCreateManager(int capacity)
             worker = WorkerRegistry.GetWorkerForWorld(World);
 
             spatialOSSendSystem = World.GetOrCreateManager<SpatialOSSendSystem>();
-            spatialOSSendSystem.RegisterCustomReplicationSystem(typeof(T));
+            if (!spatialOSSendSystem.TryRegisterCustomReplicationSystem(typeof(T)))
+            {
+                worker.View.LogDispatcher.HandleLog(new LogEvent("Custom Replication System for this component " +
```
jared-improbable: Avoid using "!" in messages. Keep it low-key and un-excited :)

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
```diff
@@ -44,15 +37,14 @@ private void GenerateComponentGroups()
...
         {
             if (!registeredReplicators.Contains(type.TypeIndex))
             {
-                Debug.LogWarningFormat(Warnings.CustomReplicationSystemAlreadyExists, type);
-                return;
+                return false;
             }
 
-            registeredReplicators.Remove(type.TypeIndex);
+            return registeredReplicators.Remove(type.TypeIndex);
```
jared-improbable: Why is the `Register` method *removing* a replication system?

workers/unity/Assets/Gdk/Core/Worker/WorkerRegistry.cs
```diff
@@ -19,7 +19,7 @@ public static void SetWorkerForWorld(WorkerBase worker)
         {
             if (WorldToWorker.ContainsKey(worker.World))
             {
-                throw new Exception(string.Format("A worker is already stored for world {0}", worker.World.Name));
+                throw new ArgumentException($"A worker is already stored for world {worker.World.Name}");
```
jared-improbable: Put some single `'` around the name to help distinguish it from the message.

workers/unity/Assets/Gdk/Core/Worker/WorkerRegistry.cs
```diff
@@ -65,8 +65,7 @@ public static WorkerBase CreateWorker(string workerType, string workerId, Vector
             Func<string, Vector3, WorkerBase> createWorker;
             if (!WorkerTypeToInitializationFunction.TryGetValue(workerType, out createWorker))
             {
-                Debug.LogErrorFormat("No worker found for worker type {0}", workerType);
-                return null;
+                throw new ArgumentException("No worker found for worker type {0}", workerType);
```
jared-improbable: Put some single ' around the name to help distinguish it from the message.

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -68,19 +68,14 @@ public void Awake()
...
-
                 Workers.Add(worker);
 
                 connectionConfig = ConnectionUtility.CreateConnectionConfigFromCommandLine(commandLineArgs);
             }
 
             if (World.AllWorlds.Count <= 0)
             {
-                Debug.LogError(
+                throw new InvalidConfigurationException(
```
jared-improbable: Where can users find out how to fix this?

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -51,19 +55,18 @@ public bool Connect(ConnectionConfig config)
             {
                 Connection = ConnectionUtility.LocatorConnectToSpatial((LocatorConfig) config, GetWorkerType);
             }
-
-            if (Connection == null)
+            else
             {
-                return false;
+                throw new InvalidConfigurationException("Invalid connection config was provided.");
```
jared-improbable: How can they fix this?

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -19,18 +19,22 @@ public abstract class WorkerBase : IDisposable
...
+        protected WorkerBase(string workerId, Vector3 origin) : this(workerId, origin, new LoggingDispatcher())
+        {
+        }
+
+        protected WorkerBase(string workerId, Vector3 origin, ILogDispatcher loggingDispatcher)
         {
             if (string.IsNullOrEmpty(workerId))
             {
-                throw new ArgumentException("WorkerId is null or empty.");
+                throw new ArgumentException("WorkerId is null or empty.", workerId);
```
jared-improbable: The second parameter is meant to be the name of the parameter that was empty, so you want `nameof(workerid)`

workers/unity/Assets/Gdk/Core/Error/LogEvent.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            builder.AppendLine(Message);
+
+            if (Data.Count > 0)
+            {
+                builder.AppendLine("Log event context:");
+
+                foreach (var entry in Data)
+                {
+                    builder.AppendLine($"{entry.Key}: {entry.Value}");
```
jared-improbable: Put `'` around key/values to help pick the value out, and make empty string obvious.

code_generator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -108,7 +108,9 @@ namespace <#= qualifiedNamespace #>
                 Unity.Entities.Entity entity;
                 if (!view.TryGetEntity(op.EntityId.Id, out entity))
                 {
-                    Debug.LogErrorFormat(TranslationErrors.OpReceivedButNoEntity, op.GetType().Name, op.EntityId.Id);
+                    view.LogDispatcher.HandleLog(new LogEvent("Entity not found during OnAddComponent in <#= componentDetails.ComponentName #>Translation.", LogType.Error)
```
jared-improbable: Simplify the message, since the information can be included via the attached data. (and again below)

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -91,7 +96,10 @@ public void RemoveComponent<T>(long entityId)
             Entity entity;
             if (!TryGetEntity(entityId, out entity))
             {
-                Debug.LogErrorFormat(Errors.NoCorrespondingEntityForEntityId, typeof(T).Name, entityId);
+                LogDispatcher.HandleLog(new LogEvent("Entity not found when attemting to remove a compoenent from" +
```
jared-improbable: Typo: `compoenent`

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -91,7 +96,10 @@ public void RemoveComponent<T>(long entityId)
...
+                        "an entity..", LogType.Error)
```
jared-improbable: Double`..` at the end.

workers/unity/Assets/Gdk/Core/Error/ILogDispatcher.cs
```diff
@@ -0,0 +1,9 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public interface ILogDispatcher
+    {
+        void HandleLog(LogEvent logEvent);
```
jared-improbable: Does it make more sense to detach the log level from the LogEvent, e.g. changing the signature to `HandleLog(LogType, LogEvent)`?
That also opens up a more Unity-like API of `LogError`, `LogWarning`, etc.

In fact, Unity already has `ILogHandler` - https://docs.unity3d.com/ScriptReference/ILogHandler.html
Was this taken into account during the design phase?

```
interface ILogHandler
    {
        public void LogFormat(LogType logType, Object context, string format, params object[] args);
        public void LogException(Exception exception, Object context);
    }
```

workers/unity/Assets/Gdk/Core/Exceptions/ConnectionFailedException.cs
martinzlocha: I don't think so, the messages are already sufficiently clear and different based on the reason. The aim of adding the Reason field was to make it easier to programmatically distinguish the errors.

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
martinzlocha: From my understanding of the code if you want to use a custom replicator you need to remove the default one.

code_generator/Templates/UnityComponentConversionGenerator.tt
zeroZshadow: This should be `.Translation` everywhere.

zeroZshadow: Could the LogDispatcher be cached on the translator during construction?

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
zeroZshadow: When you register a custom replicator, it removes the default replicator from the list that it iterates over every tick.

code_generator/Templates/UnityComponentConversionGenerator.tt
martinzlocha: I don't think so given that its possible for users to create their own LogDispatcher for each worker for example.

martinzlocha:
@zeroZshadow We had an offline discussion about long comment messages but currently we are enforcing the 120 char limit and even if the solution of using + is ugly it is preferable.

code_generator/Templates/UnityComponentConversionGenerator.tt
zeroZshadow: Each worker gets their own Translation instance, so that should be fine

workers/unity/Assets/Gdk/Core/Error/ILogDispatcher.cs
martinzlocha: Changed the signature as suggested above :)

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
martinzlocha: Added more information to the exception to help users fix this.

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
martinzlocha: As above,

code_generator/Templates/UnityComponentConversionGenerator.tt
martinzlocha: Ahh yes I can get how to do it now, thanks. Changed :)

workers/unity/Assets/Gdk/Core/Error/LoggingDispatcher.cs
```diff
@@ -0,0 +1,23 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public class LoggingDispatcher : ILogDispatcher
+    {
+        public void HandleLog(LogType type, LogEvent logEvent)
+        {
+            switch (type)
```
jessicafalk: should we also handle `LogType.Assert` and `LogType.Exception`

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -19,18 +19,22 @@ public abstract class WorkerBase : IDisposable
...
+        protected WorkerBase(string workerId, Vector3 origin) : this(workerId, origin, new LoggingDispatcher())
+        {
+        }
+
+        protected WorkerBase(string workerId, Vector3 origin, ILogDispatcher loggingDispatcher)
         {
             if (string.IsNullOrEmpty(workerId))
             {
-                throw new ArgumentException("WorkerId is null or empty.");
+                throw new ArgumentException("WorkerId is null or empty.", nameof(workerId));
```
jessicafalk: is `nameof` needed here?

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
```diff
@@ -51,19 +55,19 @@ public bool Connect(ConnectionConfig config)
             {
                 Connection = ConnectionUtility.LocatorConnectToSpatial((LocatorConfig) config, GetWorkerType);
             }
-
-            if (Connection == null)
+            else
             {
-                return false;
+                throw new InvalidConfigurationException($"Invalid connection config was provided: '{config}' Only" +
+                    "ReceptionistConfig and LocatorConfig is supported.");
```
jessicafalk: `are` instead of `is`

workers/unity/Assets/Gdk/Core/Worker/WorkerBase.cs
martinzlocha: Yes, this was pointed out by Jared, we want the name of the variable to be shown not the value of the variable.

workers/unity/Assets/Gdk/Core/Error/LoggingDispatcher.cs
martinzlocha: Added both, the only issue is that we can't use Debug.LogException because that needs an exception not a string. Will use Debug.LogError in that case.

workers/unity/Assets/Gdk/Core/Error/EntityNotFound.cs
```diff
@@ -0,0 +1,8 @@
+namespace Improbable.Gdk.Core
```
jamiebrynes7: The name of this file doesn't match the content ;)

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
jared-improbable: This probably warrants a comment, because it looks very wrong.

PR #54, opened 2018-07-03
Initial script to setup a Windows Buildkite agent

jared-improbable:
This will be adapted to Ansible for the final version.

Testing on a blank instance provided by @petemounce 

## Primary Reviewers
@petemounce @DoomGerbil

### DIFF
--- a/ci/create-windows-image.ps1
+++ b/ci/create-windows-image.ps1
@@ -1,0 +1,274 @@
+### Run from an Administrator command prompt with:
+### powershell -executionpolicy Unrestricted ci/create-image.sh
+### The machine will reboot after the script runs.
+
+### Required env vars:
+### $env:UNITY_SERIAL
+### $env:UNITY_USERNAME
+### $env:UNITY_PASSWORD
+
+# https://chocolatey.org/docs/how-to-setup-offline-installation
+$CHOCOLATEY_VERSION="0.10.11"
+$GIT_VERSION="2.18.0"
+$RESHARPER_CLI_VERSION="2018.1.2"
+$SPATIAL_VERSION="1.1.9"
+
+# From https://public-cdn.cloud.unity3d.com/hub/prod/releases-win32.json
+$UNITY_VERSION="2018.1.6f1"
+$UNITY_RELEASE_HASH="57cc34175ccf"
+
+# Direct link to installer at https://my.visualstudio.com/Downloads?q=Visual%20Studio%202017
+# VS Professional 15.7
+$VISUAL_STUDIO_HASH="09e0efec-0cfd-4015-b732-e8543589a6e0/1c89b3fa4ca0d287138916dca1625a0e"
+
+### To get a logfile in "c:\programdata\ssh\logs", add the following in the embedded script segment named $sshdConfigContent, below
+###     SyslogFacility LOCAL0
+###     LogLevel DEBUG
+
+### TODO: Install Improbable robots SSH keys for github
+### TODO: Install spatial service tokens
+
+Import-Module BitsTransfer
+$ErrorActionPreference = "Stop"
+
+mkdir c:\build
+
+Write-Host "Installing Chocolatey..."
+$env:chocolateyVersion = "$CHOCOLATEY_VERSION"
+Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
+
+Write-Host "Installing git..."
+Start-Process -Wait -NoNewWindow -FilePath "choco" -ArgumentList "install", "git", "--yes", "--version=$GIT_VERSION"
+
+Write-Host "Installing Resharper CLI tools..."
+Start-Process -Wait -NoNewWindow -FilePath "choco" -ArgumentList "install", "resharper-clt.portable", "--yes", "--version=$RESHARPER_CLI_VERSION"
+
+Write-Host "Installing Spatial CLI..."
+Start-Process -Wait -NoNewWindow -FilePath "choco" -ArgumentList "install", "spatial", "--yes", "--version=$SPATIAL_VERSION"
+
+Start-BitsTransfer -Source "https://download.unity3d.com/download_unity/$UNITY_RELEASE_HASH/Windows64EditorInstaller/UnitySetup64-$UNITY_VERSION.exe" -Destination "c:\build\unity-installer.exe"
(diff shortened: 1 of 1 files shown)


ci/create-windows-image.ps1
```diff
@@ -0,0 +1,274 @@
...
+### $env:UNITY_PASSWORD
+
+# https://chocolatey.org/docs/how-to-setup-offline-installation
+$CHOCOLATEY_VERSION="0.10.11"
+$GIT_VERSION="2.18.0"
+$RESHARPER_CLI_VERSION="2018.1.2"
+$SPATIAL_VERSION="1.1.9"
+
+# From https://public-cdn.cloud.unity3d.com/hub/prod/releases-win32.json
+$UNITY_VERSION="2018.1.6f1"
```
jessicafalk: that should be 2018.2.0b10

ci/create-windows-image.ps1
```diff
@@ -0,0 +1,274 @@
...
+Write-Host "Installing Spatial CLI..."
+Start-Process -Wait -NoNewWindow -FilePath "choco" -ArgumentList "install", "spatial", "--yes", "--version=$SPATIAL_VERSION"
+
+Start-BitsTransfer -Source "https://download.unity3d.com/download_unity/$UNITY_RELEASE_HASH/Windows64EditorInstaller/UnitySetup64-$UNITY_VERSION.exe" -Destination "c:\build\unity-installer.exe"
+Start-BitsTransfer -Source "https://download.unity3d.com/download_unity/$UNITY_RELEASE_HASH/TargetSupportInstaller/UnitySetup-Linux-Support-for-Editor-$UNITY_VERSION.exe" -Destination "c:\build\unity-linux-installer.exe"
+Start-BitsTransfer -Source "https://download.unity3d.com/download_unity/$UNITY_RELEASE_HASH/TargetSupportInstaller/UnitySetup-Mac-Mono-Support-for-Editor-$UNITY_VERSION.exe" -Destination "c:\build\unity-mac-installer.exe"
+Start-BitsTransfer -Source "https://download.visualstudio.microsoft.com/download/pr/$VISUAL_STUDIO_HASH/vs_professional.exe" -Destination "c:\build\vs-installer.exe"
+
+Write-Host "Installing Unity..."
+Start-Process -Wait -NoNewWindow -FilePath /build/unity-installer.exe -ArgumentList "/S"
```
zeroZshadow: Since we're not specifying the target directory `/D`, this might break in the future if Unity every changes this.

ci/create-windows-image.ps1
```diff
@@ -0,0 +1,274 @@
...
+
+# Direct link to installer at https://my.visualstudio.com/Downloads?q=Visual%20Studio%202017
+# VS Professional 15.7
+$VISUAL_STUDIO_HASH="09e0efec-0cfd-4015-b732-e8543589a6e0/1c89b3fa4ca0d287138916dca1625a0e"
+
+### To get a logfile in "c:\programdata\ssh\logs", add the following in the embedded script segment named $sshdConfigContent, below
+###     SyslogFacility LOCAL0
+###     LogLevel DEBUG
+
+### TODO: Install Improbable robots SSH keys for github
```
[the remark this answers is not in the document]
DoomGerbil: @petemounce Correct me if I'm wrong, but the plan has been to host the actual build machine configuration files in a specific location in the platform repo, which would sidestep this issue.

Of course, for a temporary experiment, it's likely fine to host this config here without credentials, but yeah, we can't have our private SSH keys hosted in (or accessible to a VM built purely from) a public repo.

ci/create-windows-image.ps1
```diff
@@ -0,0 +1,274 @@
...
+### TODO: Install spatial service tokens
```
DoomGerbil: Probably ditto here about the private credential access.

ci/create-windows-image.ps1
```diff
@@ -0,0 +1,274 @@
...
+$openSSHAdminUser = 'c:\ProgramData\ssh'
+$openSSHAuthorizedKeys = Join-Path $openSSHAdminUser 'authorized_keys'
+
+If (-Not (Test-Path $openSSHAdminUser)) {
+    New-Item -Path $openSSHAdminUser -Type Directory
+}
+
+# BEGIN IMPROBABLE - use Google services
+# See https://cloud.google.com/compute/docs/storing-retrieving-metadata#default
+$keyUrl = "http://metadata.google.internal/computeMetadata/v1/project/attributes/ssh-keys"
```
DoomGerbil: This will work only if run in a GCE project.  Is that a problem for this config, since it's hosted in the public repo?

I guess, to rephrase, since this will be visible to people who don't have access to our infrastructure, is it OK that this will not work for them?

ci/create-windows-image.ps1
jared-improbable: The odds of that are pretty low. I want us to do the default thing that Unity does to keep in line with what many users will do. And if Unity do, against all best practice and convention, change from `%PROGRAMFILES%\Unity`, then when we upgrade to that version, we update our scripts to reflect the new reality, rather than forcing the installation into where we want it to be.

ci/create-windows-image.ps1
jared-improbable: This is meant as a guide for setting up Ansible in the platform repo later.  I'll add a comment to the effect that "This is our stuff, you can use it as a guide, but it's not useful to you for more than that."

ci/create-windows-image.ps1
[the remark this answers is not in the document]
jared-improbable: Correct, this is more of an experiment to fast-forward Windows support for BuildKite - the configs will live in the platform repo, and we'll just pin to those from this repo.

ci/create-windows-image.ps1
```diff
@@ -14,8 +18,8 @@ $RESHARPER_CLI_VERSION="2018.1.2"
 $SPATIAL_VERSION="1.1.9"
 
 # From https://public-cdn.cloud.unity3d.com/hub/prod/releases-win32.json
-$UNITY_VERSION="2018.1.6f1"
-$UNITY_RELEASE_HASH="57cc34175ccf"
+$UNITY_VERSION="2018.2.0b10"
```
jared-improbable: @jessicafalk @zeroZshadow Once this is all rolled out, this change (plus a repin of the GDK to point to the new Windows image) will be all we need to do to support a new Unity release. :)

jessicafalk: I am SO looking forward to this :D

PR #59, opened 2018-07-03
UTY-420: Unity GDK Tooling in .Net Core - Part 1: Code Generation

jamiebrynes7:
#### Description
Well this was an adventure. A list of changes:
1. Committed generated code (that someone forgot to do 😠)
2. Pulled `Mono.TextTemplating` in as source because their latest Nuget package is like 1.5 years old and doesn't support .Net Core 2.0
3. Rewrote `Improbable.TextTemplating` using the `Mono.TextTemplating` source!
4. Moved GdkCodeGenerator into its own folder to separate from the other projects in the sln
5. Twiddled the `csproj` files to work with `dotnet`. 
6. Ensured that the End2End tests work with `dotnet`
7. Updated the `spatialos*build.json` files to use `dotnet`
8. Update the CI build scripts.
9. Update the documentation

A caveat, currently any changes to `Improbable.TextTemplating` or `Mono.TextTemplating` will require the user to manually copy over the `dll` files. If we try to do this automatically, MSBuild locks itself due to the `UsingTask` declaration in `GdkCodeGeneration.csproj`. 

#### Tests
Ran `dotnet test` on both `GdkCodeGenerator.csproj` and `Tests.csproj` in the End2End repository.
#### Documentation
Re-pointed links and updated setup guide.
#### Primary reviewers
@jared-improbable @samiwh


jamiebrynes7:
I added the steps to install .NET Core with the Visual Studio install, but will also include .NET Core 2.x explicitly as a dependency

ci/includes/pinned-tools.sh
```diff
@@ -30,4 +30,4 @@ export LINTER="cleanupcode.exe"
 
 export MSBUILD="$(powershell  –ExecutionPolicy Bypass ./ci/find-msbuild.ps1)"
```
samiwh: Do we use this anymore?

jared-improbable: This and the associated script can be nuked, though it could happen in a follow-up PR once this is all stabilized.

samiwh: 👍

jessicafalk:
Closing both .NET Core PRs in favor of: https://github.com/spatialos/UnityGDK/pull/73
Sorry @jamiebrynes7: just couldn't get CI to pass without merging both PRs together and adding some additional fixes

PR #62, opened 2018-07-04
Fixed incorrect error messages reported for world commands

MPeti01:
#### Description
Fixed a bug where incorrect errors were reported when a response arrived for a world command and the entity that sent the command is gone. Incorrect command names were being reported, and requests IDs were passed in place of entity IDs, this is now corrected.

Also differentiated between a request ID not existing in the dictionary, and the entity for a request not existing on the worker anymore. The latter is expected to happen occasionally under normal circumstances, so lowered the error to a warning.

#### Tests
Minor bugfix
#### Documentation
Minor bugfix
#### Primary reviewers
@samiwh @jamiebrynes7

### DIFF
--- a/workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
+++ b/workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
@@ -300,7 +300,7 @@ namespace Improbable.Gdk.Core
         public void OnDeleteEntityResponse(DeleteEntityResponseOp op)
         {
             Entity entity;
-            if (!TryGetEntityFromRequestId(op.RequestId.Id, "CreateEntityResponse", out entity))
+            if (!TryGetEntityFromRequestId(op.RequestId.Id, "DeleteEntityResponse", out entity))
             {
                 return;
             }
@@ -314,7 +314,7 @@ namespace Improbable.Gdk.Core
         public void OnReserveEntityIdResponse(ReserveEntityIdsResponseOp op)
         {
             Entity entity;
-            if (!TryGetEntityFromRequestId(op.RequestId.Id, "CreateEntityResponse", out entity))
+            if (!TryGetEntityFromRequestId(op.RequestId.Id, "ReserveEntityIdsResponse", out entity))
             {
                 return;
             }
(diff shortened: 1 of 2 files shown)


workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
```diff
@@ -300,7 +300,7 @@ public void OnCreateEntityResponse(CreateEntityResponseOp op)
         public void OnDeleteEntityResponse(DeleteEntityResponseOp op)
         {
             Entity entity;
-            if (!TryGetEntityFromRequestId(op.RequestId.Id, "CreateEntityResponse", out entity))
+            if (!TryGetEntityFromRequestId(op.RequestId.Id, "DeleteEntityResponse", out entity))
```
jamiebrynes7: Can use `op.GetType().FullName` I think? Less brittle to changes in the future

jessicafalk: That will allocate on the heap, but agree that it would be nice to not have it hardcoded

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
```diff
@@ -352,7 +352,7 @@ private bool TryGetEntityFromRequestId(uint requestId, string responseName, out
             long entityId;
             if (!RequestIdToEntityId.TryGetValue(requestId, out entityId))
             {
-                Debug.LogErrorFormat(TranslationErrors.CannotFindEntityForCommandResponse, requestId, responseName);
+                Debug.LogErrorFormat(TranslationErrors.RequestDoesNotExist, requestId, responseName);
```
jessicafalk: we need to make sure that this goes in before Martin's PR OR rebase on top of it, if his goes in first to ensure that we don't start using `LogError` again

samiwh: I suspect this one will be ready first. In which case the only change to the other ones will be a string change.

PR #63, opened 2018-07-04
UTY-594 automatic unity installation / Unity beta 11 update

zeroZshadow:

-------

#### Description
When running on CI, this will now install the Unity editor required for the GDK
This version is automatically read from the ProjectVersion.txt provided by the unity project.

This also upgrades the project to beta 11 as beta 10 is no longer available!

An additional CI nightly step should be added that ensures the latest released Unity is always installed on all agents.

#### Tests
I've upgraded to 2018.1.0b11 and ran CI, which downloaded the correct version and build properly

#### Documentation
-

#### Primary reviewers
@jamiebrynes7 @jessicafalk


jared-improbable:
What happens if an agent hasn't run a build yet, AND Unity push a new version and remove the current from the json endpoint, AND then a build is scheduled to run on that agent? That will just fall over.

Perhaps rather than discovering the hash+version live, we should pin to the hash+version, and add a subcommand to make it easy to list available versions to upgrade and pin to.

ci/prepare-unity.sh
```diff
@@ -0,0 +1,57 @@
...
+  ! ( isLinux || isMacOS );
+}
+
+# Only run inside TeamCity
+if [ -z "${TEAMCITY_CAPTURE_ENV+x}" ]; then
+    exit 0
+fi
+
+if isWindows; then
+    MODULES="linux mac-mono"
```
jessicafalk: need to include the mobile modules

ci/prepare-unity.sh
```diff
@@ -0,0 +1,57 @@
...
+  [[ "$(uname -s)" == "Darwin" ]];
+}
+
+function isWindows() {
+  ! ( isLinux || isMacOS );
+}
+
+# Only run inside TeamCity
+if [ -z "${TEAMCITY_CAPTURE_ENV+x}" ]; then
+    exit 0
```
jessicafalk: add some kind of message here to signal to the user that this can only be run in TC /  why should it only be run on TC? wouldn't that also make it easier for us, if we can just run this script to get the new version?

ci/prepare-unity.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Setup unity_downloader
+markStartOfBlock "Installing unity_downloader"
+mkdir -p "${GOPATH}"
+go get -v -d -u "${UNITY_PACKAGE}"
+markEndOfBlock "Installing unity_downloader"
+
+# Run unity_downloader with platform specific modules
+markStartOfBlock "Download Unity"
+go run "${GOPATH}/src/${UNITY_PACKAGE}/main.go" "${UNITY_VERSION}" "${UNITY_ROOT}" "--modules" ${MODULES}
```
jessicafalk: what is the current behavior of this downloader? does it just download or also do all the other steps that we discussed yesterday. Basically: how easy is it for us to install a new Unity version on our machines?

ci/prepare-unity.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Check if version is installed
+if isUnityImprobablePathPresent; then
+    exit 0
+fi
+
+# Setup unity_downloader
+markStartOfBlock "Installing unity_downloader"
+mkdir -p "${GOPATH}"
+go get -v -d -u "${UNITY_PACKAGE}"
```
jessicafalk: not too happy of introducing go into this repo as well...
also: this will fail for anyone who has not access to the unity downloader repo

ci/prepare-unity.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+GOPATH="$(pwd)/go"
+export GOPATH
+UNITY_PACKAGE="github.com/improbable/unity_downloader"
+
+markStartOfBlock "$0"
+
+# Check if version is installed
+if isUnityImprobablePathPresent; then
```
jessicafalk: this is kind of confusing as it checks for `UNITY_DIR` which could be our internal path that we use for TC agents: `C:/Unity/..` or it could be that `UNITY_HOME` is set to some arbitrary value. maybe rename it or move this check `[ -d "$UNITY_DIR" ]` directly into the if statement

also: add a log message here as well. We need to tell people what's going on, if we ever need to debug the logs ;)

ci/includes/pinned-tools.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+UNITY_ROOT="C:/Unity"
```
jessicafalk: `IMPROBABLE_UNITY_ROOT`

ci/prepare-unity.sh
jared-improbable: I think we need bigger signposts that, like `ci/create-windows-image.ps1`, this script only applies to internal Improbable infrastructure and shouldn't be run by the general public.

ci/prepare-unity.sh
zeroZshadow: I really don't want this to install unity for our clients, hence this only runs for CI to avoid some manual work until the new CI is here.

ci/prepare-unity.sh
zeroZshadow: I'll add a log that unity is already installed if this passes.
I do not understand the issue with the path check?

ci/includes/pinned-tools.sh
zeroZshadow: 👍

zeroZshadow:
Based on offline feedback, the unity_downloader tool will be refactored to use pinning instead of getting versions on the fly

zeroZshadow:
Needs a fix to work with the new 2018.2.0f1 BETA
Should not rely on version naming to see if it is beta or release. And instead simply check which of the editor url's is valid.

ci/prepare-unity.sh
jessicafalk: what I meant by that is that you changed the implementation of `isUnityImprobablePathPresent`. Previously it checked whether Unity exists on the path for our TC agents. Now it checks whether it exists in `UNITY_DIR`, which could be either the "Improbable path" or the "User path". I guess renaming the method would be sufficient. 

Though with Jared's PR that needs to change anyway. so maybe directly do the `[ -d "$UNITY_DIR" ]` check here instead of calling `isUnityImprobablePathPresent`

ci/prepare-unity.sh
```diff
@@ -0,0 +1,67 @@
...
+set -e -u -x -o pipefail
+
+cd "$(dirname "$0")/../"
+
+# Only run inside TeamCity
+if [ -z "${TEAMCITY_CAPTURE_ENV+x}" ]; then
+    exit 0
+fi
+
+source ci/includes/pinned-tools.sh
```
jessicafalk: try to keep it consistent with the other scripts. We first source all `*.sh` and then add logic. So move the TeamCity check after the `source ...` commands

for the platform-check functions: let's move them to pinned-tools in case we need them somewhere else as well

ci/prepare-unity.sh
```diff
@@ -0,0 +1,67 @@
...
+elif isLinux; then
+    echo "Building is not supported on linux"
+    exit 1
+fi
+
+GOPATH="$(pwd)/go"
+export GOPATH
+
+UNITY_PACKAGE="github.com/improbable/unity_downloader"
+UNITY_PINNED_HASH=$(cat "workers/unity/ProjectSettings/ProjectVersionHash.txt")
```
jessicafalk: group all variable initializations together and then start exporting

ci/prepare-unity.sh
```diff
@@ -0,0 +1,67 @@
...
+# Setup unity_downloader
+markStartOfBlock "Installing unity_downloader"
+mkdir -p "${GOPATH}"
+go get -v -d -u "${UNITY_PACKAGE}"
+markEndOfBlock "Installing unity_downloader"
+
+# Run unity_downloader with platform specific modules
+markStartOfBlock "Download Unity"
+pushd "${GOPATH}/src/${UNITY_PACKAGE}"
+go run main.go install "${UNITY_VERSION}" "${UNITY_PINNED_HASH}" "${IMPROBABLE_UNITY_ROOT}" "--modules" ${MODULES}
```
jessicafalk: two spaces at the beginning

zeroZshadow: 2 or 4 spaces? I'm seeing this mixed throughout our tools

ci/build.sh
```diff
@@ -10,6 +10,10 @@ PROJECT_DIR="$(pwd)"
 
 markStartOfBlock "$0"
 
+markStartOfBlock "Prepare Unity"
+ci/prepare-unity.sh
+markEndOfBlock "Prepare Unity"
+
 markStartOfBlock "Building Tools"
 "${MSBUILD}" tools/ImpNuget/ImpNuget.csproj //property:Configuration=Release //clp:ErrorsOnly //nologo //m 1>&2
```
jessicafalk: you haven't rebased in a while, haven't you? ;)

zeroZshadow:
Please also have a look at `https://github.com/improbable/unity_downloader` and let me know if it is oke.

zeroZshadow:
Will redo on new master

PR #66, opened 2018-07-06
Add asmdefs as LF line endings.

jared-improbable:
-------
#### Description
Add asmdefs as LF line endings.
#### How is this documented (for example: release note, upgrade guide, feature page, in-code documentation)?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@
 *.prefab text eol=lf
 *.asset text eol=lf
 *.asmdef text eol=lf
+*.sh text eol=lf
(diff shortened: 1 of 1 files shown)


firtina-improbable:
Does this mean the files should be modified for this PR?

jamiebrynes7:
@firtina-improbable I don't think so, the `.gitattributes` file only describes how files should be checked out. I.e. - `auto` means whatever the system you are on.

On that, should we also set `*.sh` to `LF` to stop people getting trolled by that (possibly the problem in #70). Driveby @jared-improbable ?

PR #67, opened 2018-07-06
UTY-555: Add emulator support for Android

DanailPenev:
#### Description
- Added logic that detects whether the device that the game is currently running on is an Android emulator (widely accepted way of checking, should work for all popular emulators).
- Added an Improbable menu item to run spatial locally with all required parameters to improve workflow.

#### Tests
- Code gets executed on Android emulator device only.

#### Documentation
-
#### Primary reviewers

### DIFF
--- a/workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
+++ b/workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
@@ -9,15 +9,16 @@ using System.Net.Sockets;
 using UnityEditor;
 using UnityEngine;
 using UnityEngine.Experimental.UIElements;
+using Debug = UnityEngine.Debug;
 
 namespace Improbable.Gdk.Android.Editor
 {
     internal class AndroidSpatialRunner
     {
         private static readonly string ProjectDirectory = Path.Combine(Application.dataPath, "..", "..", "..");
-        private static readonly string ProcessName = "spatial";
-        private static readonly string ProcessParams = "local launch";
-        private static readonly string Config = "android_launch.json";
+        private const string ProcessName = "spatial";
+        private const string ProcessParams = "local launch";
+        private const string Config = "android_launch.json";
 
         [MenuItem("Improbable/Run Spatial for Android")]
         public static void RunAndroidSpatial()
@@ -34,6 +35,7 @@ namespace Improbable.Gdk.Android.Editor
                 throw new NullReferenceException("Could not find local IP Address. Make sure you are connected to the Internet.");
             }
 
+            Debug.Log(ipAddress.ToString());
             return ipAddress.ToString();
         }
 
--- a/workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
+++ b/workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
@@ -1,0 +1,51 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Sockets;
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.Experimental.UIElements;
+
+namespace Improbable.Gdk.Android.Editor
+{
+    internal class AndroidSpatialRunner
+    {
+        private static readonly string ProjectDirectory = Path.Combine(Application.dataPath, "..", "..", "..");
+        private static readonly string ProcessName = "spatial";
+        private static readonly string ProcessParams = "local launch";
+        private static readonly string Config = "android_launch.json";
+
+        [MenuItem("Improbable/Run Spatial for Android")]
+        public static void RunAndroidSpatial()
+        {
+            Process.Start(GetProcessStartInfo());
+        }
+
+        private static string GetLocalIpAddress()
+        {
+            var host = Dns.GetHostEntry(Dns.GetHostName());
+            var ipAddress = host.AddressList.FirstOrDefault(ip => ip.AddressFamily == AddressFamily.InterNetwork);
+            if (ipAddress == null)
+            {
+                throw new NullReferenceException("Could not find local IP Address. Make sure you are connected to the Internet.");
+            }
+
+            return ipAddress.ToString();
+        }
+
+        private static ProcessStartInfo GetProcessStartInfo()
+        {
+            return new ProcessStartInfo
+            {
+                FileName = ProcessName,
+                Arguments = $"{ProcessParams} {Config} --runtime_ip={GetLocalIpAddress()}",
+                WorkingDirectory = ProjectDirectory,
+                UseShellExecute = true
+            };
+        }
+    }
+}
--- a/workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
+++ b/workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
@@ -1,6 +1,4 @@
 using System;
-using System.Collections;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
@@ -8,7 +6,6 @@ using System.Net;
 using System.Net.Sockets;
 using UnityEditor;
 using UnityEngine;
-using UnityEngine.Experimental.UIElements;
 using Debug = UnityEngine.Debug;
 
 namespace Improbable.Gdk.Android.Editor
@@ -32,7 +29,8 @@ namespace Improbable.Gdk.Android.Editor
             var ipAddress = host.AddressList.FirstOrDefault(ip => ip.AddressFamily == AddressFamily.InterNetwork);
             if (ipAddress == null)
             {
-                throw new NullReferenceException("Could not find local IP Address. Make sure you are connected to the Internet.");
+                throw new NullReferenceException(
+                    "Could not find local IP Address. Make sure you are connected to the Internet.");
             }
 
             Debug.Log(ipAddress.ToString());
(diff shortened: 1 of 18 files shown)


workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            var host = Dns.GetHostEntry(Dns.GetHostName());
+            var ipAddress = host.AddressList.FirstOrDefault(ip => ip.AddressFamily == AddressFamily.InterNetwork);
+            if (ipAddress == null)
+            {
+                throw new NullReferenceException(
+                    "Could not find local IP Address. Make sure you are connected to the Internet.");
+            }
+
+            Debug.Log(ipAddress.ToString());
```
martinzlocha: Did you mean to leave this here? If it should stay maybe comment what the IP represents.

workers/unity/Assets/Gdk/Android/Utility/DeviceInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+        private static readonly string Manufacturer = GetValue("MANUFACTURER");
+        private static readonly string Brand = GetValue("BRAND");
+        private static readonly string Device = GetValue("DEVICE");
+        private static readonly string Product = GetValue("PRODUCT");
+
+        public static bool IsEmulator()
+        {
+            return Fingerprint.StartsWith("generic") || Fingerprint.StartsWith("unknown") ||
+                Model.Contains("google_sdk") || Model.Contains("Emulator") ||
+                Model.Contains("Android SDK built for x86") || Manufacturer.Contains("Genymotion") ||
```
martinzlocha: Will this always work on other operating systems? Might be good to make it more general if possible.

DanailPenev: Resolved offline. The code checks if the device is an emulator of any type (not just Android Studio) and is an implementation that is widely used in different Mobile frameworks. As it is a reflective java call it works on Windows/Mac/Linux.

workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
```diff
@@ -0,0 +1,51 @@
...
+        private const string ProcessParams = "local launch";
+        private const string Config = "android_launch.json";
+
+        [MenuItem("Improbable/Run Spatial for Android")]
+        public static void RunAndroidSpatial()
+        {
+            Process.Start(GetProcessStartInfo());
+        }
+
+        private static string GetLocalIpAddress()
```
jessicafalk: return `IPAddress` instead of `string`. This will make it simpler to use, if we need it anywhere else and for the current use case the conversion to string will happen automatically

workers/unity/Assets/Gdk/Android/Utility/DeviceInfo.cs
jessicafalk: `Android SDK built for x86`: you can create emulators for arm. does that also check for them?

workers/unity/Assets/Playground/Improbable.Playground.asmdef
```diff
@@ -9,7 +9,8 @@
         "Unity.Entities",
         "Unity.Entities.Hybrid",
         "Improbable.CSharp",
-        "Unity.Mathematics"
+        "Unity.Mathematics",
+        "Improbable.Gdk.Android"
```
jessicafalk: move it up to the other `Improbable.Gdk....` references to keep some kind of ordering

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -51,6 +54,17 @@ public void Awake()
...
 #endif
             }
+#if UNITY_ANDROID
+            else if (Application.isMobilePlatform)
+            {
+                var worker = WorkerRegistry.CreateWorker<UnityClient>($"Client-{Guid.NewGuid()}", new Vector3(0, 0, 0));
+                Workers.Add(worker);
+                if (DeviceInfo.IsEmulator())
+                {
+                    connectionConfig = ReceptionistConfig.CreateConnectionConfigForAndroidEmulator();
```
jessicafalk: are we not setting any connection config in case the device is a phone?

DanailPenev: This will be covered in a separate PR which deals with local connection for physical devices. :)

workers/unity/Assets/Gdk/Android/Utility/DeviceInfo.cs
DanailPenev: Resolved offline: This should theoretically cover ARM devices as well but it is really difficult to test (as computers are on x86 architecture and ARM emulators are extremely slow). If we get feedback in future that this is a required feature and it does not work we can look into it. Should be fine though.

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -4,6 +4,9 @@
 using Improbable.Gdk.Core;
 using Unity.Entities;
 using UnityEngine;
+#if UNITY_ANDROID
+using Improbable.Gdk.Android;
```
jessicafalk: right now it doesn't look like we need this namespace anywhere in the bootstrap / or I'm just not seeing it

DanailPenev: DeviceInfo.IsEmulator() is in that namespace

workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            var host = Dns.GetHostEntry(Dns.GetHostName());
+            var ipAddress = host.AddressList.FirstOrDefault(ip => ip.AddressFamily == AddressFamily.InterNetwork);
+            if (ipAddress == null)
+            {
+                throw new NullReferenceException(
+                    "Could not find local IP Address. Make sure you are connected to the Internet.");
+            }
+
+            Debug.Log($"Runtime IP: {ipAddress}");
```
jared-improbable: Is this debug print needed?

workers/unity/Assets/Gdk/Core/Config/ReceptionistConfig.cs
```diff
@@ -23,5 +23,14 @@ public static ReceptionistConfig CreateConnectionConfigFromCommandLine(Dictionar
                 parsedArgs, RuntimeConfigNames.LinkProtocol, RuntimeConfigDefaults.LinkProtocol);
             return config;
         }
+
+        public static ReceptionistConfig CreateConnectionConfigForAndroidEmulator()
+        {
+            return new ReceptionistConfig
+            {
+                ReceptionistHost = "10.0.2.2",
```
jared-improbable: Make this configurable?

workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
DanailPenev: It's for convenience because you have to type the IP manually on the device and this lets you quickly see it in the console (i.e. not have to run ```ipconfig``` in terminal).

workers/unity/Assets/Gdk/Core/Config/ReceptionistConfig.cs
DanailPenev: It is a default alias for the IP of the host device the emulator is running on. By configurable do you mean moving it to a global variable at the top of the file or something else?

workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
mattyoung-improbable: If this is something the user has to type in on the device maybe state that in the log message?

workers/unity/Assets/Gdk/Android/Editor/AndroidSpatialRunner.cs
```diff
@@ -0,0 +1,51 @@
...
+        private const string ProcessParams = "local launch";
+        private const string Config = "android_launch.json";
+
+        [MenuItem("Improbable/Run Spatial for Android")]
+        public static void RunAndroidSpatial()
+        {
+            Process.Start(GetProcessStartInfo());
+        }
+
+        private static IPAddress GetLocalIpAddress()
```
mattyoung-improbable: Would be better if this was a pure getter, without the side effect of printing the Log message (What happens if this needs to be called twice in the future?)

workers/unity/Assets/Gdk/Core/Config/ReceptionistConfig.cs
mattyoung-improbable: Can make it a global for now with an informative name, like HostDeviceIP or something?

workers/unity/Assets/Gdk/Android/Utility/DeviceInfo.cs
gkassabli: Would be nice to have a comment enumerating which emulators does this code cover. Later on it would be easier to check if particular emulator is covered or not

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
gkassabli: Shall we add a comment denoting that?

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
```diff
@@ -364,7 +364,8 @@ private bool TryGetEntityFromRequestId(uint requestId, string responseName, out
             }
             else if (!view.TryGetEntity(entityId, out entity))
             {
-                Debug.LogWarningFormat(TranslationErrors.CannotFindEntityForWorldCommandResponse, entityId, responseName);
+                Debug.LogWarningFormat(TranslationErrors.CannotFindEntityForWorldCommandResponse, entityId,
```
jessicafalk: use our new logging dispatcher :)

PR #69, opened 2018-07-06
Stop use of `spatial worker build`

jared-improbable:

-------

#### Description
* Remove use of `spatial worker build`, et al
* A follow-up PR will be made to move all of the files into the right places once this is merged
* `ci/build.sh` no longer builds out the client/gamelogic workers. That's now only done in `build-test.sh`, which is called by TeamCity.
* Unified .gitignore files

#### Tests
CI, local runs
#### Documentation
Docs updated to remove CLI use
#### Primary reviewers
@jessicafalk @jamiebrynes7


ci/build-client.sh
```diff
@@ -0,0 +1,28 @@
...
+
+"${UNITY_EXE}" -projectPath "${UNITY_PROJECT_DIR}" \
+    -batchmode \
+    -quit \
+    -nographics \
+    -logfile "$(pwd)/build/build_logs/UnityClientBuild.log" \
+    -executeMethod "Improbable.Gdk.Legacy.BuildSystem.WorkerBuilder.Build" \
+    +buildWorkerTypes "UnityClient" \
+    +buildTarget "local"
+markEndOfBlock "Building UnityClient"
```
jamiebrynes7: Nit: Add a line between `markEndOfBlock` and the last argument of the previous command for readability

jamiebrynes7: Similarly for `build-gamelogic.sh`

ci/codegen.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
jamiebrynes7: This file will become redundant soon I think (with the .NET Core changes)

ci/includes/pinned-tools.sh
```diff
@@ -31,3 +38,13 @@ export LINTER="cleanupcode.exe"
 export MSBUILD="$(powershell  –ExecutionPolicy Bypass ./ci/find-msbuild.ps1)"
 
 export NUNIT3_CONSOLE="code_generator/packages/NUnit.ConsoleRunner.3.8.0/tools/nunit3-console.exe"
+
+function unpackTo() {
+  local SOURCE=$1
+  local TARGET=$2
+
+  mkdir -p "${TARGET}"
+  unzip -o -q "${SOURCE}" -d "${TARGET}"
```
jamiebrynes7: Including a dependency on unzip - may need to include in setup guide

code_generator/End2End/GeneratedDependencies/.spatialos/schema_codegen_cache/cache.json
```diff
@@ -0,0 +1 @@
+{"InputFilesLastModified":[{"InputPath":"C:\\Users\\jared\\AppData\\Local\\Temp\\improbable_extracted_packages\\29e5a4460683a6afacf50e42a559bb35\\proto\\google\\protobuf\\descriptor.proto","LastModified":1530866039},{"InputPath":"C:\\Users\\jared\\AppData\\Local\\Temp\\improbable_extracted_packages\\29e5a4460683a6afacf50e42a559bb35\\proto\\improbable\\entity_state.proto","LastModified":1530866039},{"InputPath":"E:\\git\\UnityGDK\\code_generator\\End2End\\schema\\blittable_types.schema","LastModified":1530880098},{"InputPath":"E:\\git\\UnityGDK\\code_generator\\End2End\\schema\\exhaustive_test.schema","LastModified":1530880098},{"InputPath":"E:\\git\\UnityGDK\\code_generator\\End2End\\schema\\nested_typenames.schema","LastModified":1530880098},{"InputPath":"E:\\git\\UnityGDK\\code_generator\\End2End\\schema\\nonblittable_types.schema","LastModified":1530880098}],"MiscData":[".\\GeneratedCSharpWorker",".spatialos\\schema_codegen_cache\\generated_proto_ir",".spatialos\\schema_codegen_cache\\generated_proto_dependency_ir",".spatialos\\schema_codegen_cache\\generated_descriptor","csharp",".\\GeneratedCSharpWorker","13.0.1"]}
```
jamiebrynes7: Accidental inclusion?

code_generator/src/CodeGenerator.cs
```diff
@@ -120,21 +129,21 @@ private void ShowHelpMessage()
 
         private bool ValidateOptions()
         {
-            if (options.JsonDirectory == null)
+            if (options.NativeOutputDirectory == null)
             {
-                Console.WriteLine("Input directory not specified.");
+                Console.WriteLine("Output directory not specified");
```
jamiebrynes7: `"Native output directory...

ci/build-client.sh
```diff
@@ -0,0 +1,28 @@
...
+
+markStartOfBlock "Building UnityClient"
+
+# Back-compat: we need to run this to generate the bridge configurations.
+spatial build build-config UnityClient
+
+"${UNITY_EXE}" -projectPath "${UNITY_PROJECT_DIR}" \
+    -batchmode \
+    -quit \
+    -nographics \
```
jessicafalk: I am pretty sure that this line has caused multiple issues in the past for the SDK. Should we remove it?

ci/build-test.sh
```diff
@@ -6,3 +6,5 @@ cd "$(dirname "$0")/../"
 
 ci/build.sh
```
jessicafalk: should we remove this file and directly call ci/codegen.sh? It looks like the changes in both of our .NET PRs make `ci/build.sh` completely obsolete.

code_generator/End2End/GeneratedDependencies/Generated.csproj
```diff
@@ -40,8 +40,8 @@
     </StartupObject>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Improbable.CSharp">
```
jessicafalk: did we want to remove this reference?

ci/build-client.sh
jared-improbable: Done

code_generator/End2End/GeneratedDependencies/.spatialos/schema_codegen_cache/cache.json
jared-improbable: Removed

ci/includes/pinned-tools.sh
jared-improbable: It's included in git bash, so since we already depend on that, it should be fine for a bit longer?

code_generator/src/CodeGenerator.cs
jared-improbable: Done

ci/build-client.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
jessicafalk: should we make `build-client` and `build-gamelogic` parameterizable so that users could build the workers for whichever `buildTarget` they want

ci/download-core-sdk.sh
```diff
@@ -0,0 +1,38 @@
...
+mkdir -p "${CORE_SDK_DIR}/schema"
+mkdir -p "${CORE_SDK_DIR}/tools"
+mkdir -p "${CORE_SDK_DIR}/worker_sdk"
+
+spatial package retrieve "worker_sdk" "core-dynamic-x86-win32"       "${PINNED_CORE_SDK_VERSION}" "${CORE_SDK_DIR}/worker_sdk/core-dynamic-x86-win32"
+spatial package retrieve "worker_sdk" "core-dynamic-x86_64-win32"    "${PINNED_CORE_SDK_VERSION}" "${CORE_SDK_DIR}/worker_sdk/core-dynamic-x86_64-win32"
+spatial package retrieve "worker_sdk" "core-dynamic-x86_64-linux"    "${PINNED_CORE_SDK_VERSION}" "${CORE_SDK_DIR}/worker_sdk/core-dynamic-x86_64-linux"
+spatial package retrieve "worker_sdk" "core-bundle-x86_64-macos"     "${PINNED_CORE_SDK_VERSION}" "${CORE_SDK_DIR}/worker_sdk/core-bundle-x86_64-macos"
+spatial package retrieve "worker_sdk" "csharp"                       "${PINNED_CORE_SDK_VERSION}" "${CORE_SDK_DIR}/worker_sdk/csharp"
+spatial package retrieve "tools"      "schema_compiler-x86_64-win32" "${PINNED_CORE_SDK_VERSION}" "${CORE_SDK_DIR}/tools/schema_compiler-x86_64-win32"
```
jessicafalk: do we need to download a mac version as well?

docs/content/code-generator.md
```diff
@@ -66,25 +66,12 @@ After post processing, a code generation job is ran on each C# object that repre
...
-  "command": "../../bin/code_generator/GdkCodeGenerator.exe",
-  "arguments": [
-      "--json-dir=./improbable/json/usr",
-      "--output-dir=./Assets/Generated/"
-  ]
-}
-```
-
-If any changes you make change the input parameters or add new ones, you should add them to those two json files to ensure that `spatial codegen` runs correctly!
+If any changes you make change the input parameters or add new ones, you should add them `ci/codegen.sh` so it runs correctly!
```
jessicafalk: you should add them ***to*** `ci/codegen.sh`**

.gitignore
```diff
@@ -1,19 +1,42 @@
...
 
-*.csproj.user
+# Autogenerated VS/MD/Consulo solution and project files
 .vs/
-.spatialos/
-build/
-bin/
-obj/
-packages/
+.consulo/
```
jessicafalk: what is that?

ci/includes/pinned-tools.sh
```diff
@@ -31,3 +38,13 @@ export LINTER="cleanupcode.exe"
...
+}
+
+PINNED_CORE_SDK_VERSION="13.0.1"
```
jessicafalk: I think it would be better to move this into its own file to avoid any kind of merge conflicts later

ci/build-client.sh
jared-improbable: Done, they now accept a parameter

ci/download-core-sdk.sh
```diff
@@ -0,0 +1,52 @@
...
+  mkdir -p "${TARGET}"
+  unzip -o -q "${SOURCE}" -d "${TARGET}"
+}
+
+PINNED_CORE_SDK_VERSION="$(cat core-sdk.version)"
+
+CORE_SDK_DIR="$(pwd)/build/core_sdk"
+UNITY_PROJECT_DIR="$(pwd)/workers/unity"
+NATIVE_DEPENDENCIES_PATH="${UNITY_PROJECT_DIR}/Assets/Plugins/Improbable/Core"
+MANAGED_DEPENDENCIES_PATH="${UNITY_PROJECT_DIR}/Assets/Plugins/Improbable/Sdk"
```
jamiebrynes7: Do  we want to clean these directories before unpacking the `dll` and `so` files.

ci/build-client.sh
firtina-improbable: would it make sense to have another parameter to be "UnityClient" or "UnityGameLogic" to prevent code duplication?

jessicafalk: get `ci/build.sh` back and remove `ci/build-client.sh` and `ci/build-gamelogic.sh` xD

would prefer that actually as it also makes it easier to add more worker types later

docs/content/code-generator.md
jared-improbable: Done

ci/download-core-sdk.sh
jared-improbable: Seems appropriate to add that as part of the Mac work?

ci/build-client.sh
jared-improbable: I think we should do this work once we have more than 2 workers - we may have a better design once we have a bigger set of example use cases.

ci/download-core-sdk.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+# Modify the PINNED_CORE_SDK_VERSION variable in `ci/includes/pinned-tools.sh` and run this script to download
```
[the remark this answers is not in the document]
jared-improbable: Fixed to reflect the new reality

ci/download-core-sdk.sh
jared-improbable: Done

ci/build-client.sh
jessicafalk: I would really prefer to do this now as this seems far more error-prone. e.g. just accidentally changing one script and not the other

ci/download-core-sdk.sh
jessicafalk: if we add the worker_sdk mac dll here, I would argue that we can also add the rest of the mac support here. But I'm fine with creating an additional PR, if that is preferred

PR #71, opened 2018-07-06
Added original request context to world command responses

MPeti01:
#### Description
A feature was added to commands to include a copy of the original request in the response, to give more information about the situation where it was sent. However, the world commands were not updated, so this is currently only available for user commands. This PR adds that missing functionality.

Details of the implementation are slightly different, due to 1) the fact that world commands are different in that they don't have target entity IDs, 2) the fact that it didn't seem like a good idea to add a significant amount of code and 3 similar-but-different Request classes to easily breakable copy-pastish code, 3) the fact that the APIs are currently inconsistent between the two kinds and changing that should probably be discussed.

Also got rid of an unreachable error message while changing that part of the code.
#### Primary reviewers
@samiwh @mattyoung-improbable @zeroZshadow


workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
```diff
@@ -184,7 +196,18 @@ public class WorldCommandsTranslation : ComponentTranslation
...
+
+        private Dictionary<uint, DeleteEntityRequest> requestIdToDeleteEntityRequest
+            = new Dictionary<uint, DeleteEntityRequest>();
+
+        private Dictionary<uint, ReserveEntityIdsRequest> requestIdToReserveEntityIdsRequest
+            = new Dictionary<uint, ReserveEntityIdsRequest>();
+
+        private Dictionary<uint, EntityQueryRequest> requestIdToEntityQueryRequest
+            = new Dictionary<uint, EntityQueryRequest>();
+        // Also for other 3!
```
jamiebrynes7: Can you further qualify or remove this comment?

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
```diff
@@ -1,36 +1,38 @@
 using System.Collections.Generic;
 using Improbable.Gdk.Core.Components;
 using Improbable.Worker;
+using Improbable.Worker.Query;
```
jamiebrynes7: I think this import isn't required

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
```diff
@@ -1,36 +1,38 @@
...
+using Tests;
```
jamiebrynes7: Also what's up with this import?

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
```diff
@@ -285,53 +308,56 @@ public override void RegisterWithDispatcher(Dispatcher dispatcher)
 
         public void OnCreateEntityResponse(CreateEntityResponseOp op)
         {
+            CreateEntityRequest request = requestIdToCreateEntityRequest[op.RequestId.Id];
```
jamiebrynes7: A slightly better pattern may be:

```
CreateEntityRequest request;
if (!requestIdToCreateEntityRequest.TryGetValue(op.RequestId.Id, out request) 
{
	// Error handling
}
requestIdToCreateEntityRequest.Remove(op.RequestId.Id
```

Similarly for the other types of commands below.

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
MPeti01: Sorry, used to IntelliJ auto cleaning imports and committed in a hurry. Will remove.

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
MPeti01: Also result of rushing, I put that there before adding all 4 dictionaries. Note to self: read through the entire diff next time.

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
MPeti01: @samiwh suggested removing that because it should never happen, and if we wanted to check for errors that should never happen consistently then a lot would have to be changed in the GDK.

jamiebrynes7: I do agree that it should never happen, but I still feel that we should do error handling here (maybe I'm just being overly defensive)

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
```diff
@@ -285,53 +308,56 @@ public override void RegisterWithDispatcher(Dispatcher dispatcher)
...
+            requestIdToCreateEntityRequest.Remove(op.RequestId.Id);
+
+            var response =
+                new CreateEntityResponse((CommandStatusCode) op.StatusCode, op.Message, op.EntityId.Value.Id, request);
```
Ivaskovic: Any particular reason for the move up? It seems unnecessary to compute `response` in case the check fails and we return.

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
MPeti01: If there's no more comments on this I'll leave it as-is I think. Sami's point sounds reasonable to me (verifying things here and not in other similar places is weird)

workers/unity/Assets/Gdk/Core/Components/WorldCommandsTranslation.cs
```diff
@@ -8,29 +8,29 @@
 
 namespace Improbable.Gdk.Core
 {
-    internal struct CreateEntityRequest
+    public struct CreateEntityRequest
     {
         public Worker.Entity Entity;
         public Collections.Option<EntityId> EntityId;
```
samiwh: If this is public then we are generally trying to only expose Nullable rather than improbable collections. In our case I don't think that we actually have nullable in the public api either though.

MPeti01: So what should be done here? What does codegen do with collections defined in schema (I assume we should mirror that here)?

samiwh: Codegen will either use a gdk option type or give you a nullable; I can't remember if we have switched the former yet. However in this api we can change things up a bit. Here it is acceptable to just use nullable or to use a long (probably nullable). The world translations will need removing or changing soon anyway, so I wouldn't worry too much about keeping this consistent with codegen.

PR #73, opened 2018-07-09
upgrading to .NET Core 2.1

jessicafalk:

-------

#### Description
After trying to get https://github.com/spatialos/UnityGDK/pull/59 and https://github.com/spatialos/UnityGDK/pull/61 to pass CI for quite a while, I finally gave up and created a new branch merging in all the changes that happened in both PRs + some fixes to get CI to pass. Basically still the same: Trying to move to .NET Core (but .NET Core 2.1 instead of .NET Core 2.0) and fixing everything around it to be able to run all bash scripts successfully.

Commits are kind of ordered by folders / sections that needed to be updated

#### Tests
Ran TC CI like a thousand times until it finally passed

#### Documentation
updated docs


ci/build.sh
```diff
@@ -11,12 +11,8 @@ PROJECT_DIR="$(pwd)"
 markStartOfBlock "$0"
 
 markStartOfBlock "Building Tools"
-"${MSBUILD}" tools/ImpNuget/ImpNuget.csproj //property:Configuration=Release //clp:ErrorsOnly //nologo //m 1>&2
-pushd "tools/DocsLinter"
-    "../../bin/ImpNuget/ImpNuget.exe"
-popd
 
-"${MSBUILD}" tools/DocsLinter/DocsLinter.csproj //property:Configuration=Release //clp:ErrorsOnly //nologo //m 1>&2
+dotnet build tools/DocsLinter/DocsLinter.csproj
```
zeroZshadow: This is missing the `-c Release` flag

ci/build.sh
```diff
@@ -29,11 +25,7 @@ spatial worker build -t=local
 markEndOfBlock "Building Unity Project"
 
 markStartOfBlock "Code Generation End2End"
-pushd "code_generator/End2End/Tests"
-    "../../../bin/ImpNuget/ImpNuget.exe"
-popd
-
-"${MSBUILD}" code_generator/End2End/End2End.sln //property:Configuration=Release //clp:ErrorsOnly //nologo //m 1>&2
+dotnet build code_generator/End2End/End2End.sln
```
zeroZshadow: This is missing the -c Release flag as well

ci/build.sh
zeroZshadow: Since this is technically testing if the projects can be build, maybe `--no-incremental` should be passed to do a full rebuild?

ci/includes/pinned-tools.sh
```diff
@@ -24,10 +24,15 @@ function getUnityDir() {
 }
 
 UNITY_DIR="$(getUnityDir)"
+export UNITY_HOME="${UNITY_DIR}"
```
zeroZshadow: Why is this overwriting the user's UNITY_HOME?

tools/DocsLinter/Program.cs
```diff
@@ -195,8 +195,15 @@ private static void Main(string[] args)
 
             try
             {
-                var request = WebRequest.CreateHttp(remoteLink.Url);
+                var strippedUrl = remoteLink.Url;
```
zeroZshadow: What is this drive by change fixing?

ci/includes/pinned-tools.sh
jessicafalk: because if it's not set, the code generator tests fail

tools/DocsLinter/Program.cs
jessicafalk: well, if you would have waited until it's out of WIP, I would have added the reason :P
what happened with .NET Core 2.1: it will fail to retrieve URLs that have an anchor. Previously it simply ignored the anchor

jessicafalk:
@zeroZshadow changed all references from  netcoreapp2.0 to  netcoreapp2.1, but there doesn't seem to be a netstandard2.1 yet. So left that for now

ci/includes/pinned-tools.sh
jessicafalk: This will go away with Jared's PR anyway, I believe: https://github.com/spatialos/UnityGDK/pull/57

ci/includes/pinned-tools.sh
```diff
@@ -24,10 +24,15 @@ function getUnityDir() {
...
 export LINTER="cleanupcode.exe"
 
 export MSBUILD="$(powershell  –ExecutionPolicy Bypass ./ci/find-msbuild.ps1)"
 
-export NUNIT3_CONSOLE="code_generator/packages/NUnit.ConsoleRunner.3.8.0/tools/nunit3-console.exe"
+export NUNIT3_CONSOLE="tools/DocsLinter/packages/NUnit.ConsoleRunner.3.8.0/tools/nunit3-console.exe"
+
+DOTNET_VERSION="$(dotnet --version)"
+
+export MSBuildSDKsPath="/c/Program Files/dotnet/sdk/${DOTNET_VERSION}/Sdks"
```
jamiebrynes7: Why is this required?

jessicafalk: because otherwise the TC agents can't find `Microsoft.Net.Sdk`. Will need to check with IT today as it seems to be related to not installing .NET Core via Visual Studio, but for now I left it in here until it's resolved

docs/content/code-generator.md
```diff
@@ -13,7 +13,7 @@ The Unity GDK comes packaged with a custom code generator. This doc will detail
 
 The code generator operates on the [JSON AST](https://docs.improbable.io/reference/latest/shared/schema/custom-codegen#abstract-syntax-tree-format) that is generated by the schema compiler. A pre-built `dll` parses this JSON AST and converts it to C# objects.
 
-This parsing stage can be found in [CodeGenerator.cs](../../code_generator/src/CodeGenerator.cs):
+This parsing stage can be found in [CodeGenerator.cs](../../code_generator/GdkCodeGenerator/src/CodeGenerator.cs):
```
ElleEss: "..can be found..." is passive voice (see [Docs space in Confluence](https://improbableio.atlassian.net/wiki/spaces/GBU/pages/220725249?atlOrigin=eyJpIjoiNWZjYjgzNjIyZjdjNDQzM2IxMGRkMGNlNDJhMmY5OWUiLCJwIjoiYyJ9). Better to make it active voice:
"The parsing stage is in [CodeGenerator.cs](../../code_generator/src/CodeGenerator.cs):"

ElleEss: Also, I think the code snippet is the parsing stage, in which case, we should say so.
""The parsing stage, shown in the code snippet below, is in [CodeGenerator.cs](../../code_generator/src/CodeGenerator.cs),:"

docs/content/code-generator.md
```diff
@@ -22,28 +22,31 @@ This parsing stage can be found in [CodeGenerator.cs](../../code_generator/src/C
 
 The code generator then does post-processing on the collection of C# objects that represent the schema. 
 
-1. It will flag each type and component as [blittable or not blittable](https://docs.microsoft.com/en-us/dotnet/framework/interop/blittable-and-non-blittable-types). This information is required as blittable and non-blittable components need to be generated in a different manner. You can find the source for this in [BlittableFlagger.cs](../../code_generator/src/Generation/SchemaProcessing/BlittableFlagger.cs).
-2. It will mark data types that are used as event, command request, and/or command response payloads. You can find the source for this in [PayloadMarker.cs](../../code_generator/src/Generation/SchemaProcessing/PayloadMarker.cs).
+1. It will flag each type and component as [blittable or not blittable](https://docs.microsoft.com/en-us/dotnet/framework/interop/blittable-and-non-blittable-types). This information is required as blittable and non-blittable components need to be generated in a different manner. You can find the source for this in [BlittableFlagger.cs](../../code_generator/GdkCodeGenerator/src/Generation/SchemaProcessing/BlittableFlagger.cs).
```
ElleEss: I know this isn't new text but it doesn't need "will" as it's a universal constant - it always happens in these conditions (because our product is reliable).  So:
"It flags each type..."
Not a show-stopper but helps show confidence in our product.

docs/content/code-generator.md
```diff
@@ -22,28 +22,31 @@ This parsing stage can be found in [CodeGenerator.cs](../../code_generator/src/C
...
+2. It will mark data types that are used as event, command request, and/or command response payloads. You can find the source for this in [PayloadMarker.cs](../../code_generator/GdkCodeGenerator/src/Generation/SchemaProcessing/PayloadMarker.cs).
```
ElleEss: "It marks..." 

(Better for the reasons listed.)

ci/includes/pinned-tools.sh
samiwh: This might be overwriting it with a slightly different version. Can we change the name of our own UNITY_HOME to something else that can be exported for this specific session.

code_generator/Improbable.TextTemplating/Improbable.TextTemplating.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
samiwh: Can we import the nuget package? https://www.nuget.org/packages/Mono.TextTemplating

jessicafalk: We unfortunately can't because the last version there is 1.5 years old and doesn't support .NET Core 2

ci/includes/pinned-tools.sh
jessicafalk: as mentioned above, I would prefer to leave it in for now just so that CI passes. @jared-improbable this line can be removed when your PR goes in, right?

jared-improbable: Yep, this line isn't long for this world.

ci/includes/pinned-tools.sh
jared-improbable: Use `${PROGRAM_FILES}` as a hedge for now.

docs/content/code-generator.md
```diff
@@ -53,12 +56,12 @@ Note: the transformation of T4 to C# is in a pre-build step which is defined in
...
   ```
 
-Each T4 template is paired with a hand-written partial class that defines the data used in the template. These partial classes can be found [here](../../code_generator/src/Generation/Generators/Parts/).
+Each T4 template is paired with a hand-written partial class that defines the data used in the template. These partial classes can be found [here](../../code_generator/GDKCodeGenerator/src/Generation/Generators/Parts/).
 
   
 #### Code generation jobs
 
-After post processing, a code generation job is ran on each C# object that represents a single schema file. This job is defined in [SingleGenerationJob.cs](../../code_generator/src/Generation/SingleGenerationJob.cs). This job enumerates all the enums, types, and components in a single schema file and runs the correct generator for each.
+After post processing, a code generation job is ran on each C# object that represents a single schema file. This job is defined in [SingleGenerationJob.cs](../../code_generator/GDKCodeGenerator/src/Generation/SingleGenerationJob.cs). This job enumerates all the enums, types, and components in a single schema file and runs the correct generator for each.
```
jared-improbable: "is ran" -> "is run".
Could also be rewritten to be have a more active voice.

PR #77, opened 2018-07-10
UTY-464: Forward logs to Spatial.

martinzlocha:

-------

#### Description
Debug logs are currently contained inside a unity worker. Errors remain unnoticed as a result.

#### Tests
Deployed and generated errors, visible in console afterwards.

#### Documentation
Will add in a separate PR for this and LogError change in general.

#### Primary reviewers


workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            if (connection == null)
+            {
+                return;
+            }
+
+            LogLevel logLevel;
+            var message = logEvent.ToString();
+
+            switch (type)
```
jared-improbable: Can I suggest using a `private static readonly Dictionary<LogType, LogLevel>` that's initialized inline? May look much more readable.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+                    break;
+                default:
+                    logLevel = LogLevel.Debug;
+                    break;
+            }
+
+            Collections.Option<EntityId> entityId;
+            if (logEvent.Data.ContainsKey(IdType.EntityId.ToString()))
+            {
+                entityId = new EntityId((long) logEvent.Data[IdType.EntityId.ToString()]);
```
jared-improbable: What happens here if someone accidentally stores a non-`long` type, but uses the IdType.EntityID? This will throw and blow up!

workers/unity/Assets/Playground/Scripts/Workers/UnityGameLogic.cs
```diff
@@ -11,11 +11,18 @@ public class UnityGameLogic : WorkerBase
 
         public override string GetWorkerType => WorkerType;
 
-        public UnityGameLogic(string workerId, Vector3 origin) : base(workerId, origin)
+        public UnityGameLogic(string workerId, Vector3 origin) : base(workerId, origin, new ForwardingDispatcher())
```
jamiebrynes7: Why only `UnityGameLogic` and not `UnityClient` ?

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,67 @@
+using Improbable.Worker;
```
JonasImprobable: Nit: How about renaming the `Error` folder to something like `Logging`?

workers/unity/Assets/Playground/Scripts/Workers/UnityGameLogic.cs
```diff
@@ -11,11 +11,18 @@ public class UnityGameLogic : WorkerBase
...
         {
             PlayerLifecycleConfig.CreatePlayerEntityTemplate = PlayerTemplate.CreatePlayerEntityTemplate;
         }
 
+        public override void Connect(ConnectionConfig config)
+        {
+            base.Connect(config);
+
+            ((ForwardingDispatcher) View.LogDispatcher).SetConnection(Connection);
```
jared-improbable: Why not initialize a field of the type `ForwardingDispatcher`, pass that to `base`, and use that here to avoid a pretty dodgy-looking cast?

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public class ForwardingDispatcher : ILogDispatcher
+    {
+        private Connection connection;
+
+#if UNITY_EDITOR
```
JonasImprobable: There is no harm with logging this for headless workers. You can still see the logs in the unity log files. An additional benefit would be that a worker's behaviour is consistent between running in the editor and as standalone.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            this.connection = connection;
+        }
+
+        public void HandleLog(LogType type, LogEvent logEvent)
+        {
+#if UNITY_EDITOR
+            if (Application.isEditor)
+            {
+                loggingDispatcher.HandleLog(type, logEvent);
```
jessicafalk: I think we should always log to Unity as well. If they want to disable it, they can change the setting as described here: https://answers.unity.com/questions/1301347/how-to-disable-all-logging-on-release-build.html

JonasImprobable:
As a meta comment: Currently it looks like we are calling `View.LogDispatcher.HandleLog()` in pre-defined places. In case a worker throws any error (could be a null ref exception due to a bug in user code), those errors would still be silent with the current setup, no?

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
samiwh: Or they just change the dispatcher. That is what it is there for.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+                    break;
+            }
+
+            Collections.Option<EntityId> entityId;
+            if (logEvent.Data.ContainsKey(IdType.EntityId.ToString()))
+            {
+                entityId = new EntityId((long) logEvent.Data[IdType.EntityId.ToString()]);
+            }
+
+            connection.SendLogMessage(logLevel, "ForwardingDispatcher", message, entityId);
```
samiwh: ForwardingDispatcher is not a good logger name. That field is there to give users context on where the message came from. The type of the gdk dispatcher gives no information.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+            switch (type)
+            {
+                case LogType.Error:
+                    logLevel = LogLevel.Error;
+                    break;
+                case LogType.Warning:
+                    logLevel = LogLevel.Warn;
+                    break;
+                case LogType.Log:
+                    logLevel = LogLevel.Info;
```
samiwh: Do we want to log everything by default? I'd argue that warnings and above only by default. Sending all your debug information to the runtime is mostly just noise. Especially as right now we are mostly running things in the editor.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    public class ForwardingDispatcher : ILogDispatcher
+    {
+        private Connection connection;
+
+#if UNITY_EDITOR
+        private readonly LoggingDispatcher loggingDispatcher = new LoggingDispatcher();
+#endif
+
+        public void SetConnection(Connection connection)
```
samiwh: Set this in a constructor. This object shouldn't exist without one. The only case it could be null is if the worker disconnects, in which case we don't have a good flow for this.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+#if UNITY_EDITOR
+            if (Application.isEditor)
+            {
+                loggingDispatcher.HandleLog(type, logEvent);
+            }
+#endif
+
+            if (connection == null)
+            {
+                return;
```
samiwh: If you are disconnected but still running you still want to log locally. Most likely use case is running in the editor.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
samiwh: Is it worth making a helper function for getting well known bits of metadata?

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
martinzlocha: Would using connection.GetWorkerId() be better?

workers/unity/Assets/Playground/Scripts/Workers/UnityGameLogic.cs
martinzlocha: I don't think that is possible. The ForwardingDispatcher would have to be a static field for me to be able to pass it to `base` on construction. I guess I could it would be possible to create a method in base called SetDispatcher() but this would leave the constructor in an uninitialized state without setting the dispatcher.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
jessicafalk: I disagree, we might create workers that don't connect immediately. Unless you want to tightly couple the creation of workers with the connection to Spatial.

Right now we create the dispatcher when we create the worker. This dispatcher could still log to Unity, even if no connection is set.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
martinzlocha: Yeah thats a good point, I will change it once this gets approved. I don't want to move everything now, it will just confuse everyone when reviewing the PR.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,85 @@
...
+        public ForwardingDispatcher(LogLevel minimumLogLevel = LogLevel.Warn)
+        {
+            this.minimumLogLevel = minimumLogLevel;
+
+            Application.logMessageReceived += LogCallback;
+        }
+
+        private void LogCallback(string message, string stackTrace, LogType type)
+        {
+            if (type == LogType.Exception)
```
JonasImprobable: Does this mean that warning and errors are silent? Wouldn't it be better to expose the min-loglevel for forwarding arbitrary unity logs as a config parameter?

martinzlocha:
@JonasImprobable @jared-improbable @jamiebrynes7 @samiwh @jessicafalk I should have covered all comments in the PR, please have a look. Also now the dispatcher forwards exceptions as well.

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        public void SetConnection(Connection connection)
+        {
+            this.connection = connection;
+        }
+
+        public void HandleLog(LogType type, LogEvent logEvent)
+        {
+            loggingDispatcher.HandleLog(type, logEvent);
```
JonasImprobable: The structure of this logging setup feels a little chaotic.
In our current setup, we have a top level `ForwardingDispatcher` which takes logging messages from pre-set places and uses a child `LoggingDispatcher` to log messages locally. It also takes messages logged locally from `Application.logMessageReceived` and forwards those to fabric too.

In the UnitySDK, the setup looked like this:
Every UnitySDK error message is logged locally using e.g. `Debug.LogError()`, a single log forwarder forwards all application logs, from the UnitySDK or elsewhere, to fabric as long as a certain min-log level is exceeded. I find the latter approach more elegant and less error prone (see. potential problem with logging the same message to fabric twice)

jessicafalk: This approach was discussed, but is not feasible in our case. Using this, we would need to add a callback to `logMessageReceived`, whenever a `Debug.Log` is fired, we can automatically forward it to fabric. However, we lose all context and are not able to send e.g. the entity id with it. 

Therefore, we need to define our own logging dispatcher to be able to send additional information to fabric.

JonasImprobable: Can we not log messages in some well defined format and parse the message string to retrieve the context again? Hacky, I know.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,100 @@
...
+            { LogType.Assert, LogLevel.Error },
+            { LogType.Warning, LogLevel.Warn },
+            { LogType.Log, LogLevel.Info }
+        };
+
+        public ForwardingDispatcher(LogLevel minimumLogLevel = LogLevel.Warn)
+        {
+            this.minimumLogLevel = minimumLogLevel;
+
+            Application.logMessageReceived += LogCallback;
```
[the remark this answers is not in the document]
martinzlocha: I guess it is possible if someone does Debug.Log with LogType.Exception but I don't think anyone should ever do that but throw an exception as well.

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
jessicafalk: this doesn't provide the user with a nice API on how to log messages.
I agree, that it was very nice to directly use Debug.LogError, but it doesn't seem to be the right fit for our use case.

workers/unity/Assets/Gdk/Core/Logging/FieldName.cs
```diff
@@ -0,0 +1,9 @@
+namespace Improbable.Gdk.Core
+{
+    public static class FieldName
```
samiwh: Please get rid of this. You can't tell at a glance what this is or what it relates to. However, I'm happy for there to be a set of static logging helper functions that extracts the relevant information if it is the right stored as the right type in the dictionary.

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
samiwh: I don't see the logging something twice , which only happens when running two workers in the same editor, as a problem. Especially as you can get around it by just not having two workers with log dispatchers that capture all editor output. 

However when running two things in the same editor, capturing Untiy logs won't tell you which worker sent it or let you easily configure on a per worker basis what it does.

Parsing strings to regain context is something I want to avoid as it is incredibly error prone, and almost never a good idea.

workers/unity/Assets/Gdk/Core/Error/ForwardingDispatcher.cs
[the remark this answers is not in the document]
JonasImprobable: What about errors? It would be handy to see arbitrary errors (originated from the UnityGDK or elsewhere) in the fabric logs.

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
jessicafalk: regarding Unity logs: that can be easily mitigated by parsing in the worker id.
 
What I do agree with and forgot to mention earlier: We shouldn't use a LoggingDispatcher inside another dispatcher. This seems confusing and relies on a certain internal implementation of that dispatcher. Can we just use `Debug.Log` directly?

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -231,6 +235,7 @@ public void AddGameObjectEntity(Entity entity, GameObject gameObject)
                     if (!HasComponent<NotAuthoritative<T>>(entity))
                     {
                         LogDispatcher.HandleLog(LogType.Error, new LogEvent(UnexpectedAuthorityChangeError)
+                            .WithField(FieldName.LoggerName, "MutableView")
```
jessicafalk: factor that string out in a private variable

workers/unity/Assets/Generated/improbable/gdk/tests/ExhaustiveBlittableSingularTranslation.cs
```diff
@@ -67,7 +67,8 @@ public void OnAddComponent(AddComponentOp<global::Improbable.Gdk.Tests.Exhaustiv
                 if (!view.TryGetEntity(op.EntityId.Id, out entity))
                 {
                     LogDispatcher.HandleLog(LogType.Error, new LogEvent("Entity not found during OnAddComponent.")
-                        .WithField(IdType.EntityId.ToString(), op.EntityId.Id)
+                        .WithField(FieldName.LoggerName, "ExhaustiveBlittableSingular")
```
jessicafalk: same here: use a private field instead of a hardcoded string

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
JonasImprobable: @samiwh The logging something twice refers to my comment above. If we log an exception, the exception is also printed locally and then logged to fabric a second time.

The problem I have with the current approach is that errors that we don't explicitly look for happen silently. When looking at the logs of a deployment, it becomes very hard to find bugs that don't happen locally. Users also need to remember to use our custom logging API instead of  `Debug.Log()` which they are used to using. Whenever I want to log something, I find myself always going for `Debug.Log()` first out of pure habit. Having contextual information is nice but often times, I am just interested in knowing whether a bug exists. Except for cases where e.g. entityids are well known, I wonder how much added benefit there is If the log comes with some entityid or request id, basically just some number.

jessicafalk: We can additionally use `logMessageReceived` to ensure that all warnings and errors gets forwarded to fabric, but other than that we need to provide the functionality of providing more context to fabric when possible/necesasry

I think it's fine to print it in Unity and log it to fabric. This way we ensure that we send the important messages to fabric, but still keep a complete log file on the user's machine in case we need to debug something

martinzlocha: One issue I can see is that if we want to forward Exception, Error and Warning logs then the ForwardingDispatcher can't log any of those. So it would have to just use a normal Debug.Log in the console when it receives an error not to cause duplication. An alternative to this would be not to use ForwardingDispatcher when in unity editor and then not log when using the ForwardingDispatcher.

martinzlocha: I had a look at ways to use Debug.Log and it doesn't seem that there is a way to use it with multiple connections. Also after discussing this with Peter (AAA) he mentioned that he didn't like the behavior that messages are sent to console without explicitly specifying it so he is happy with the proposed implementation.

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                var loggerName = data[LogDataFieldName.LoggerName];
+                var name = loggerName as string;
+                if (name != null)
+                {
+                    return name;
+                }
+            }
+
+            return "UnnamedLogger";
```
jamiebrynes7: `UnknownLogger` ? Mild preference

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -75,7 +76,8 @@ public void Dispose()
             {
                 LogDispatcher.HandleLog(LogType.Error, new LogEvent(
                         "Entity not found when attempting to add a component to an entity.")
```
jamiebrynes7: Magic strings - make these consts

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -99,6 +99,7 @@ public void Start()
                 catch (ConnectionFailedException exception)
                 {
                     worker.View.LogDispatcher.HandleLog(LogType.Error, new LogEvent(exception.Message)
+                        .WithField(LogDataFieldName.LoggerName, "Bootstrap")
```
jamiebrynes7: Magic strings

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
martinzlocha: Moved all to constants :)

jessicafalk:
there are still merge conflicts according to github

martinzlocha:
@jessicafalk Yup, but this is just generated code so once it gets approved I will rebase and codegen again.

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -98,9 +120,9 @@ public void RemoveComponent<T>(long entityId)
             Entity entity;
             if (!TryGetEntity(entityId, out entity))
             {
-                LogDispatcher.HandleLog(LogType.Error, new LogEvent(
-                        "Entity not found when attemting to remove a component from an entity.")
-                    .WithField(IdType.EntityId.ToString(), entityId)
+                LogDispatcher.HandleLog(LogType.Error, new LogEvent(Errors.EntityNotFoundOnRemove)
+                    .WithField(LogDataFieldName.LoggerName, LoggerName)
+                    .WithField(LogDataFieldName.EntityId, entityId)
                     .WithField("OpName", typeof(T).Name));
```
jessicafalk: move `OpName` also to `LogDatafieldName`

jessicafalk: at least it seems like it's used often enough

martinzlocha: Changed :) Also found a few other strings which I replaced.

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,104 @@
...
+{
+    public static class LogDataFieldName
+    {
+        public const string EntityId = "EntityId";
+        public const string RequestId = "RequestId";
+        public const string LoggerName = "LoggerName";
+        public const string OpName = "OpName";
+        public const string Type = "Type";
+        public const string NewState = "NewState";
+        public const string OldState = "OldState";
```
jessicafalk: nit: sort alphabetically

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,104 @@
...
+            { LogType.Error, LogLevel.Error },
+            { LogType.Assert, LogLevel.Error },
+            { LogType.Warning, LogLevel.Warn },
+            { LogType.Log, LogLevel.Info }
+        };
+
+        public ForwardingDispatcher(LogLevel minimumLogLevel = LogLevel.Warn)
+        {
+            this.minimumLogLevel = minimumLogLevel;
+
```
jessicafalk: don't need that newline

workers/unity/Assets/Gdk/Core/Logging/LoggingUtils.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class LoggingUtils
+    {
+        public const string EntityId = "EntityId";
+        public const string LoggerName = "LoggerName";
+
+        public static Collections.Option<EntityId> ExtractEntityId(Dictionary<string, object> data)
+        {
+            if (data.ContainsKey(LoggingUtils.EntityId))
+            {
+                var dataEntityId = data[LoggingUtils.EntityId];
```
samiwh: This needs to be TryGet

workers/unity/Assets/Gdk/Core/Logging/LoggingUtils.cs
```diff
@@ -0,0 +1,44 @@
...
+                    return new EntityId((long) dataEntityId);
+                }
+            }
+
+            return new Collections.Option<EntityId>();
+        }
+
+        public static string ExtractLoggerName(Dictionary<string, object> data)
+        {
+            if (data.ContainsKey(LoggingUtils.LoggerName))
```
samiwh: Same here

workers/unity/Assets/Playground/Scripts/Spawning/ArchetypeInitializationSystem.cs
```diff
@@ -26,6 +26,12 @@ public struct Data
 
         [Inject] private Data data;
 
+        private const string LoggerName = "ArchetypeInitializationSystem";
+        private const string ArchetypeMappingNotFound = "No corresponding archetype mapping found.";
+
+        private const string UnsupportedArchetypeType =
```
samiwh: Make it clear this is for a worker in the name

workers/unity/Assets/Gdk/Core/Logging/LoggingUtils.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                var loggerName = data[LoggerName];
+                var name = loggerName as string;
+                if (name != null)
+                {
+                    return name;
+                }
+            }
+
+            return "UnknownLogger";
```
jared-improbable: Extract this as a public string constant so user code can check against it.

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -353,5 +356,33 @@ private void FindTranslationUnits()
...
             SetComponentObject(entity, messageBuffer);
         }
+
+        private static class Errors
+        {
+            public const string UnexpectedAuthorityChangeError =
+                "Unexpected state transition during authority change.";
+
+            public const string EntityNotFoundOnAdd =
+                "Entity not found when attempting to add a component to an entity.";
```
jared-improbable: "to an entity" is redundant and can be removed.

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -353,5 +356,33 @@ private void FindTranslationUnits()
...
+
+            public const string EntityNotFoundOnRemove =
+                "Entity not found when attemting to remove a component from an entity.";
```
jared-improbable: "from an entity" is redundant and can be removed.

workers/unity/Assets/Playground/Scripts/Workers/UnityClient.cs
```diff
@@ -11,8 +11,17 @@ public class UnityClient : WorkerBase
 
         public override string GetWorkerType => WorkerType;
 
-        public UnityClient(string workerId, Vector3 origin) : base(workerId, origin)
+        private readonly ForwardingDispatcher forwardingDispatcher;
+
+        public UnityClient(string workerId, Vector3 origin) : base(workerId, origin, new ForwardingDispatcher())
+        {
+            forwardingDispatcher = (ForwardingDispatcher) View.LogDispatcher;
```
jared-improbable: How about `private readonly ForwardingDispatcher forwardingDispatcher = new ForwardingDispatcher();`
Then you can pass it along to `base` and drop the need for the cast and initialization here.

code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -116,8 +118,9 @@ namespace <#= qualifiedNamespace #>
                 if (!view.TryGetEntity(op.EntityId.Id, out entity))
                 {
                     LogDispatcher.HandleLog(LogType.Error, new LogEvent("Entity not found during OnAddComponent.")
-                        .WithField(IdType.EntityId.ToString(), op.EntityId.Id)
-                        .WithField("OpName", op.GetType().Name));
+                        .WithField(LoggingUtils.LoggerName, LoggerName)
```
jared-improbable: Possibly controversial, but how about adding an extension method `WithLoggerName` to `LoggingUtils` to reduce boilerplate a bit?

jared-improbable: Same for a `WithEntityId` method for below.

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,56 @@
...
+        public ForwardingDispatcher(LogLevel minimumLogLevel = LogLevel.Warn)
+        {
+            this.minimumLogLevel = minimumLogLevel;
+            Application.logMessageReceived += LogCallback;
+        }
+
+        private void LogCallback(string message, string stackTrace, LogType type)
+        {
+            if (type == LogType.Exception)
```
jared-improbable: A comment explaining why this is necessary will be useful.

workers/unity/Assets/Gdk/Core/Logging/LoggingDispatcher.cs
```diff
@@ -0,0 +1,12 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public class LoggingDispatcher : ILogDispatcher
```
jared-improbable: A `///` XMLDoc comment would be useful here, since this is a public class.

workers/unity/Assets/Gdk/Core/Logging/ForwardingDispatcher.cs
```diff
@@ -0,0 +1,56 @@
+using System.Collections.Generic;
+using Improbable.Worker;
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public class ForwardingDispatcher : ILogDispatcher
```
jared-improbable: A `///` XMLDoc comment would be useful here, since this is a public class.

workers/unity/Assets/Playground/Scripts/Workers/UnityClient.cs
martinzlocha: When I attempt to do that I get: "An object reference is required for the non-static field, method, or property 'UnityClient.forwardingDispatcher'. Cannot access non-static field 'forwardingDispatcher' in static context."

jared-improbable: Ah well, I must have mis-remembered how that worked.

code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
jared-improbable: Fair enough.

samiwh: We did talk about adding the helpers for those specifically though. I think that is not controversial.

PR #83, opened 2018-07-11
UTY-415: Exporting load metrics from Unity GDK.

Ivaskovic:

-------

#### Description
UnityGDK now periodically exports metrics that may be viewed in the inspector. In particular, this means that Load can also be exported, and it is computed as a function of fps.


workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
```diff
@@ -19,13 +19,34 @@ internal static class Warnings
         private readonly List<int> registeredReplicators = new List<int>();
         private readonly List<int> commandSenders = new List<int>();
 
+        private int ticks = 0;
+        private const int METRIC_TICKS_UPDATE = 200;
```
jamiebrynes7: We use Pascal case for constants. `MetricTicksUpdate`. Also, we should use time to gate metric sends rather than ticks as ticks have a variable length between them.

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
```diff
@@ -19,13 +19,34 @@ internal static class Warnings
...
+
+        public delegate Worker.Metrics MetricsComputation();
+        private MetricsComputation metricComputedEverySend;
```
jamiebrynes7: On second thought, it might be better to split this out into its own system that deals just with metrics.

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
```diff
@@ -19,13 +19,34 @@ internal static class Warnings
...
+
         protected override void OnCreateManager(int capacity)
+        {
+            OnCreateManager(capacity, delegate()
+            {
+                float fps = 1.0f / Time.deltaTime;
+                return new Worker.Metrics
+                {
+                    // Load is 0 is fps is at most 20, 1 if it is at least 30, and linear in between.
+                    Load = Mathf.Min(1.0f, Mathf.Max(0.0f, 3.0f - 0.1f * fps))
```
jamiebrynes7: We shouldn't cap the load at 1. A load of 15 and a load of 1 means very different things :)

jamiebrynes7: Oh I just realised the load is inverted here. 0 load = high fps, 1 load = low fps :)

martinzlocha: This is a lot of magic numbers, would be nice to extract into constants.

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
```diff
@@ -73,6 +94,13 @@ protected override void OnUpdate()
             {
                 view.TranslationUnits[componentTypeIndex].SendCommands(connection);
             }
+
+            ticks++;
+            if (ticks == METRIC_TICKS_UPDATE)
+            {
+                ticks = 0;
+                connection.SendMetrics(metricComputedEverySend());
```
martinzlocha: Might be nice to do some average instead of one computation every few ticks. This will have a higher variance.

martinzlocha: I completely missed the doc, will post there sorry.

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
Ivaskovic: Refactored, no longer an issue!

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
Ivaskovic: Done.

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
Ivaskovic: Done!

workers/unity/Assets/Gdk/Core/Systems/SpatialOSSendSystem.cs
Ivaskovic: Done!

workers/unity/Assets/Gdk/Core/Systems/MetricSendSystem.cs
```diff
@@ -0,0 +1,73 @@
...
+        protected override void OnCreateManager(int capacity)
+        {
+            base.OnCreateManager(capacity);
+            worker = WorkerRegistry.GetWorkerForWorld(World);
+        }
+
+        private float DefaultLoadCalculation(float fps)
+        {
+            float targetFps = Application.targetFrameRate;
+            return Mathf.Max(0.0f, (targetFps - fps) / (targetFps - 20.0f));
```
jamiebrynes7: The `20.0f` should be some multiple of `targetFps`. I.e - we say load = 1 when you are at 50% of your target FPS or similar.

workers/unity/Assets/Gdk/Core/Systems/MetricSendSystem.cs
```diff
@@ -0,0 +1,73 @@
...
+        private void AddFpsSample()
+        {
+            if (fpsMeasurements.Count == MetricConfig.MaxFpsSamples)
+            {
+                fpsMeasurements.RemoveAt(0);
+            }
+            fpsMeasurements.Add(1.0f / Time.deltaTime);
+        }
+
+        private float CalculateFps()
```
jamiebrynes7: Ordering the functions in access modifier order ( i.e. - public -> protected -> private) is a good convention (I'm not sure if the linter does this for you)

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            timeElapsedSinceUpdate += Time.deltaTime;
+            AddFpsSample();
+            if (timeElapsedSinceUpdate >= MetricConfig.TimeBetweenMetricUpdatesSecs)
+            {
+                timeElapsedSinceUpdate = 0;
+                float fps = CalculateFps();
+                var load = MetricConfig.CalculateLoad == null ?
+                    DefaultLoadCalculation(fps) : MetricConfig.CalculateLoad(fps);
+                Worker.Metrics metrics = new Worker.Metrics
```
jared-improbable: Is there a need to send the metric if nothing has changed since last time?

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
```diff
@@ -0,0 +1,76 @@
...
+                    Load = load
+                };
+                connection.SendMetrics(metrics);
+            }
+        }
+
+        private float DefaultLoadCalculation(float fps)
+        {
+            float targetFps = Application.targetFrameRate;
+            return Mathf.Max(0.0f, (targetFps - fps) / (targetFps - 20.0f));
```
jared-improbable: Why `- 20.0f`?

workers/unity/Assets/Playground/Config/MetricConfig.cs
```diff
@@ -0,0 +1,12 @@
+namespace Playground
+{
+    public delegate float CalculateLoadDelegate(float fps);
+
+    public static class MetricConfig
```
jared-improbable: I'd rather we chose a different approach than global statics, if we can help it.
More fundamental than that, however - why are we including this at all? Why isn't the default calculation built into the system enough? Where's the use case?

workers/unity/Assets/Playground/Scripts/Workers/UnityClient.cs
```diff
@@ -42,6 +42,9 @@ public override void RegisterSystems()
 
             // Client player commands system
             World.GetOrCreateManager<PlayerCommandsSystem>();
+
+            // Metric sending system
+            World.GetOrCreateManager<MetricSendSystem>();
```
jamiebrynes7: We will probably want the system on `UnityGameLogic` too :)

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
Ivaskovic: No more hardcoded value now, it was supposed to represent a value at which fps is too low. Now it's half of targetFps.

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
Ivaskovic: In practice it will always change because it's a float.

workers/unity/Assets/Playground/Scripts/Workers/UnityClient.cs
Ivaskovic: Done now!

workers/unity/Assets/Playground/Config/MetricConfig.cs
```diff
@@ -0,0 +1,12 @@
+namespace Playground
+{
+    public delegate float CalculateLoadDelegate(float fps);
```
samiwh: I think you can get rid of this if this.

workers/unity/Assets/Playground/Config/MetricConfig.cs
```diff
@@ -0,0 +1,12 @@
...
+
+    public static class MetricConfig
+    {
+        public static CalculateLoadDelegate CalculateLoad;
```
samiwh: Can probably just move these into the system.

workers/unity/Assets/Playground/Config/MetricConfig.cs
```diff
@@ -0,0 +1,12 @@
...
+
+        public static int MaxFpsSamples = 50;
```
samiwh: Can probably just move these into the system. Nothing else consumes them and this doesn't really increase visibility.

workers/unity/Assets/Playground/Config/MetricConfig.cs
Ivaskovic: W.r.t. global statics: same pattern used is currently used in `PlayerLifecycleConfig`. This is likely to be changed.

samiwh: I don't think there is any built in calculation. Either on the sdk end, on how often to send, or from unity end, on the current fps (which always surprises me).

The use case is the rare event is to check when running in the cloud if the worker is overloaded. It came up a fair bit in the last game jam.

samiwh: Deleted my comment as I'd completely misunderstood the comment I was responding to.

samiwh: +1 to no global statistics. This a system specific set of constants to do a specific job. If anything we just need some constants for the sake of not having magic numbers.

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
```diff
@@ -0,0 +1,75 @@
...
+using UnityEngine;
+using Improbable.Gdk.Core;
+
+namespace Playground
+{
+    public class MetricSendSystem : ComponentSystem
+    {
+        private WorkerBase worker;
+
+        private float timeElapsedSinceUpdate = 0.0f;
```
samiwh: Would it be easier just to write the time of the next update and check for that?

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        private List<float> fpsMeasurements = new List<float>();
```
samiwh: can be readonly

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        protected override void OnCreateManager(int capacity)
+        {
+            base.OnCreateManager(capacity);
+            worker = WorkerRegistry.GetWorkerForWorld(World);
+        }
+
+        protected override void OnUpdate()
+        {
+            if (worker.Connection == null)
```
samiwh: Can inject a worker entity with an IsConnected tag for this too. Not sure which way is better.

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
Ivaskovic: Agreed.

workers/unity/Assets/Playground/Config/MetricConfig.cs
```diff
@@ -0,0 +1,12 @@
...
+        public static int TimeBetweenMetricUpdatesSecs = 2;
```
samiwh: Probably want this to be a float to be in keeping with Unity time

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
Ivaskovic: Doesn't make too much of a difference.

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
Ivaskovic: This is consistent with the implementations of similar systems.

workers/unity/Assets/Playground/Config/MetricConfig.cs
Ivaskovic: Agree, I prefer it to be a float as well.

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            float targetFps = Application.targetFrameRate;
+            return Mathf.Max(0.0f, (targetFps - fps) / (0.5f * targetFps));
+        }
+
+        private void AddFpsSample()
+        {
+            if (fpsMeasurements.Count == MetricConfig.MaxFpsSamples)
+            {
+                fpsMeasurements.RemoveAt(0);
```
samiwh: This seems a little odd. If you specify a sample rate then surely a maximum can't be hit, in fact you wouldn't be guaranteed to hit the expected minimum. list.RemoveAt(0) always make uncomfortable.

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
Ivaskovic: Actually, this is better -- floating point addition is problematic if the numbers have vastly different orders of magnitude, so it gets less accurate as time goes by.

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
samiwh: Could we also just accumulate the total fps can keep track of the count. Or do we need the intermediate values?

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
```diff
@@ -0,0 +1,75 @@
+using System.Collections.Generic;
+using Unity.Entities;
+using UnityEngine;
+using Improbable.Gdk.Core;
+
+namespace Playground
+{
+    public class MetricSendSystem : ComponentSystem
```
samiwh: Not actually sure if this needs to be executed in a specific part of the update loop. Might be worth looking up what deltaTime will be for different systems. I suspect this is fine provided we don't put it in FixedUpdate though.

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
samiwh: It would be make it mildly more regular and be slightly neater I think.

workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
Ivaskovic: There isn't really a 'sample rate'. This is a 'sliding window' approach to computing the mean. I agree that list.RemoveAt(0) isn't great -- I guess I could use a queue instead.

samiwh: I'd suggest changing that anyway. Without a constrained frame rate this will just take a large number of samples near the beginning of the tick then max out.

Ivaskovic: I'm not entirely sure I understand the comment? The FPS estimate should rely on the more recent measurements. An alternative would be to use exponential averaging.

workers/unity/Assets/Playground/Config/MetricConfig.cs
jamiebrynes7: Given that the entire system exists in user code - I think its probably best to remove the `MetricConfig.cs` file and move the constants into the system itself. The use case for a user-assigned delegate no longer exists

PR #86, opened 2018-07-11
[Experimental][CSharp Core SDK] Blittable list and map

zeroZshadow:
#### Description
Added an initial implementation of the BlittableList and BlittableMap.
Both are simple structs with a handle to their underlying data storage, wrapping their respective IList and IDictionary interfaces.

Creation should only be used in code generation through the following example snipped
```csharp
var listHandle = HandleProvider.AllocateListHandle<int>();
var list = new BlittableList<int>(listHandle);

var mapHandle = HandleProvider.AllocateMapHandle<int, float>();
var map = new BlittableMap<int, float>(mapHandle);
```

Cleanup can only be done through:

```csharp
HandleProvider.FreeHandle(listHandle);
HandleProvider.FreeHandle(mapHandle);
```

### DIFF
--- a/workers/unity/Assets/Gdk/Core/Collections/BlittableList.cs
+++ b/workers/unity/Assets/Gdk/Core/Collections/BlittableList.cs
@@ -1,0 +1,52 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Assets.Gdk.Core.Collections
+{
+    public struct BlittableList<T> : IList<T>
+    {
+        private readonly uint handle;
+
+        internal List<T> Internal => ListProvider<T>.Get(handle);
+
+        internal BlittableList(uint handle)
+        {
+            this.handle = handle;
+        }
+
+        public static implicit operator List<T>(BlittableList<T> bl) => bl.Internal;
+
+        #region IList<T>
+        public T this[int index] {
+            get { return Internal[index]; }
+            set { Internal[index] = value; }
+        }
+
+        public int Count => Internal.Count;
+
+        public bool IsReadOnly => ((ICollection<T>)Internal).IsReadOnly;
+
+        public void Add(T item) => Internal.Add(item);
+
+        public void Insert(int index, T item) => Internal.Insert(index, item);
+
+        public void Clear() => Internal.Clear();
+
+        public bool Contains(T item) => Internal.Contains(item);
+
+        public void CopyTo(T[] array, int arrayIndex) => Internal.CopyTo(array, arrayIndex);
+
+        public int IndexOf(T item) => Internal.IndexOf(item);
+
+        public bool Remove(T item) => Internal.Remove(item);
+
+        public void RemoveAt(int index) => Internal.RemoveAt(index);
+
+        public List<T>.Enumerator GetEnumerator() => Internal.GetEnumerator();
+
+        IEnumerator<T> IEnumerable<T>.GetEnumerator() => ((IEnumerable<T>)Internal).GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator() => ((IEnumerable)Internal).GetEnumerator();
+        #endregion
+    }
+}
--- a/workers/unity/Assets/Gdk/Core/Collections/HandleProvider.cs
+++ b/workers/unity/Assets/Gdk/Core/Collections/HandleProvider.cs
@@ -23,5 +23,10 @@ namespace Assets.Gdk.Core.Collections
             freeActions.Add(handle, MapProvider<TValue, TKey>.FreeHandle);
             return handle;
         }
+
+        public static void FreeHandle(uint handle)
+        {
+            freeActions[handle](handle);
+        }
     }
 }
--- a/workers/unity/Assets/Gdk/Core/Collections/HandleProvider.cs
+++ b/workers/unity/Assets/Gdk/Core/Collections/HandleProvider.cs
@@ -1,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+
+namespace Assets.Gdk.Core.Collections
+{
+    internal static class HandleProvider
+    {
+        private static uint nextHandle = 0;
+        private static Dictionary<uint, Action<uint>> freeActions;
+
+        public static uint AllocateListHandle<TValue>()
+        {
+            var handle = nextHandle++;
+            ListProvider<TValue>.AllocateHandle(handle);
+            freeActions.Add(handle, ListProvider<TValue>.FreeHandle);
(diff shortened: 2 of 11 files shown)


workers/unity/Assets/Gdk/Core/Collections/BlittableList.cs
```diff
@@ -0,0 +1,52 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Assets.Gdk.Core.Collections
+{
+    public struct BlittableList<T> : IList<T>
```
samiwh: I don't think these should implement IList or IDictionary as they aren't meant to be new collections; they are meant to be the existing collections served in an unusual way. They will act more like the collections they are trying to mimic if you do not provide conversions that will be preferentially picked over the intended ones.

workers/unity/Assets/Gdk/Core/Collections/BlittableList.cs
```diff
@@ -0,0 +1,52 @@
...
+        public T this[int index] {
+            get { return Internal[index]; }
+            set { Internal[index] = value; }
+        }
+
+        public int Count => Internal.Count;
+
+        public bool IsReadOnly => ((ICollection<T>)Internal).IsReadOnly;
+
+        public void Add(T item) => Internal.Add(item);
```
samiwh: If you want this to be like a list then you should probably cover all the public methods of List. Currently it will only act like List as much as IList.

workers/unity/Assets/Gdk/Core/Collections/HandleProvider.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using System.Collections.Generic;
+
+namespace Assets.Gdk.Core.Collections
+{
+    internal static class HandleProvider
```
samiwh: This might end up being a fairly common name if we aren't careful. Might be worth name it something more specific.

workers/unity/Assets/Gdk/Core/Collections/HandleProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private static uint nextHandle = 0;
+        private static Dictionary<uint, Action<uint>> freeActions;
+
+        public static uint AllocateListHandle<TValue>()
```
samiwh: I would avoid the word allocate somewhere that doesn't do any allocation.

workers/unity/Assets/Gdk/Core/Collections/HandleProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            var handle = nextHandle++;
+            MapProvider<TValue, TKey>.AllocateHandle(handle);
+            freeActions.Add(handle, MapProvider<TValue, TKey>.FreeHandle);
+            return handle;
+        }
+
+        public static void FreeHandle(uint handle)
+        {
+            freeActions[handle](handle);
```
samiwh: You probably want to remove the action after this. Otherwise this will be unbounded. You know if the action isn't there that nothing needs to be freed.

workers/unity/Assets/Gdk/Core/Collections/ListProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        internal static void AllocateHandle(uint handle)
+        {
+            var item = new List<T>();
+            Storage.Add(handle, item);
+        }
+
+        internal static List<T> Get(uint handle)
+        {
+            return Storage[handle];
```
samiwh: If error checking on the free, then we should probably check here. I'd say they are about the same risk.

workers/unity/Assets/Gdk/Core/Collections/ListProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        internal static void FreeHandle(uint handle)
+        {
+            List<T> list;
+            if (!Storage.TryGetValue(handle, out list))
+            {
+                throw new ArgumentException($"ListProvider<{typeof(T).Name}> does not contain handle `{handle}`");
```
samiwh: Might be worth using the error logger here. And exception might not be helpful as this isn't something a will be trying to catch.

workers/unity/Assets/Gdk/Core/Collections/ListProvider.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+
+namespace Assets.Gdk.Core.Collections
+{
+    internal static class ListProvider<T>
```
samiwh: T isn't actually being used as a free parameter here, it's always in the form of List<T>. Similarly for the dictionary version. You could replace these both with a generic provider that just takes a single parameter.

samiwh: The only sticking point would be that you need to constrain it to `ICollection<U>` to call `Clear()`. Which can be done by adding another parameter, `U` as a sort of pseudo existential type. But because Dictionary is awkward in its definition of `T` you have to call it with `Provider<Dictionary<TKey, TValue>, KeyValuePair<TKey, TValue>.Get(handle)`.

workers/unity/Assets/Gdk/Core/Collections/MapProvider.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+
+namespace Assets.Gdk.Core.Collections
+{
+    internal static class MapProvider<TKey, TValue>
```
jared-improbable: `Map` is a C++/schema name - would `Dictionary` be more in-line with C# convention?

workers/unity/Assets/Gdk/Core/Collections/HandleProvider.cs
samiwh: In which case it should say that. Possibly call it AllocateList and return a handle.

workers/unity/Assets/Gdk/Core/Collections/HandleProvider.cs
samiwh: That doesn't really help. If this is a global then it should give a lot of context in it's name, and not be something that could clash with anything else.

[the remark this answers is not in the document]
samiwh: I think this was meant to be a stopgap. Personally I'd always rather a system owns the data, then a static interface can ask for an allocation.

PR #94, opened 2018-07-13
[CI] Automatic unity downloader

zeroZshadow:
#### Description
CI Will now automatically download and install the required unity version when ran on CI
This uses the downloader @ https://github.com/improbable/unity_downloader

#### Tests
Ran the build through CI with the new build steps

#### Documentation
Still need to add documentation on how to update to a new unity version with the tool.

#### Primary reviewers
@jared-improbable  @jessicafalk


tools/UnityPaths/UnityPaths.cs
```diff
@@ -18,7 +18,7 @@ public static class Paths
 
         // Adjust the root based on Windows or MacOS. Linux is currently unsupported.
         public static string ImprobableUnityRootPath =>
-            RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "/Applications" : Path.Combine(@"c:\", "Unity");
+            RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "/Applications/Unity" : Path.Combine(@"c:\", "Unity");
```
jessicafalk: That breaks how it's currently working locally. Why was this change made?

ci/prepare-unity.sh
```diff
@@ -0,0 +1,64 @@
...
+
+if [ -d "$IMPROBABLE_UNITY_VERSIONED" ]; then
+  echo "Unity is already installed in:" $IMPROBABLE_UNITY_VERSIONED
+  exit 0
+else
+  echo "Could not find Unity installation."
+fi
+
+UNITY_PACKAGE="github.com/improbable/unity_downloader"
+GOPATH="$(pwd)/go"
```
jessicafalk: group all variables together to directly after the `source` section. makes it easier to read
then export GOPATH
then all the checks
and then your installing unity_downloader and download unity section

ci/prepare-unity.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
```
jessicafalk: in general: no way to get this into a c# script?

ci/prepare-unity.sh
```diff
@@ -0,0 +1,64 @@
...
+set -e -u -x -o pipefail
+
+cd "$(dirname "$0")/../"
+
+source ci/includes/pinned-tools.sh
+source ci/includes/profiling.sh
+
+# Only run inside TeamCity
+if [ -z "${TEAMCITY_CAPTURE_ENV+x}" ]; then
+  exit 0
```
jessicafalk: needs an error message

ci/prepare-unity.sh
```diff
@@ -0,0 +1,64 @@
...
+
+UNITY_PINNED_VERSION=$(grep "m_EditorVersion" "workers/unity/ProjectSettings/ProjectVersion.txt" | cut -d ' ' -f2)
+UNITY_PINNED_HASH=$(cat "workers/unity/ProjectSettings/ProjectVersionHash.txt")
+IMPROBABLE_UNITY_VERSIONED="$IMPROBABLE_UNITY_ROOT/$UNITY_PINNED_VERSION"
+
+if [ -d "$IMPROBABLE_UNITY_VERSIONED" ]; then
+  echo "Unity is already installed in:" $IMPROBABLE_UNITY_VERSIONED
+  exit 0
+else
+  echo "Could not find Unity installation."
```
jessicafalk: maybe add which version of unity it was looking for

tools/UnityPaths/UnityPaths.cs
zeroZshadow: This will otherwise install all unity version at `/Applications/2018.2.0b10` `/Applications/2018.2.0f2` etc

ci/prepare-unity.sh
```diff
@@ -0,0 +1,64 @@
...
+  MODULES="linux,windows-mono,ios,android"
+  IMPROBABLE_UNITY_ROOT="/Applications/Unity"
+elif isLinux; then
+  echo "Building is not supported on linux."
+  exit 1
+fi
+
+UNITY_PINNED_VERSION=$(grep "m_EditorVersion" "workers/unity/ProjectSettings/ProjectVersion.txt" | cut -d ' ' -f2)
+UNITY_PINNED_HASH=$(cat "workers/unity/ProjectSettings/ProjectVersionHash.txt")
+IMPROBABLE_UNITY_VERSIONED="$IMPROBABLE_UNITY_ROOT/$UNITY_PINNED_VERSION"
```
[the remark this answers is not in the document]
zeroZshadow: The `Unity-` is actually deprecated after this change.
There is no need for compatibility with the SDK.

ci/prepare-unity.sh
zeroZshadow: It will be much more work for a small script that is only run by CI

ci/prepare-unity.sh
zeroZshadow: How is: "This script is only to be run on Improbable CI" ?

ci/prepare-unity.sh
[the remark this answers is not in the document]
jared-improbable: We can also update the SDK to look for Unity at the new path

ci/prepare-unity.sh
jamiebrynes7: I think it may be a good idea to (at some point) separate out scripts that are meant for use by users (codegen, build-client, etc.) and scripts that will only be run by ci (cc @samiwh).

ci/prepare-unity.sh
```diff
@@ -0,0 +1,64 @@
...
+go get -v -d "${UNITY_PACKAGE}"
+
+markEndOfBlock "Installing unity_downloader"
+
+markStartOfBlock "Download Unity"
+
+pushd "${GOPATH}/src/${UNITY_PACKAGE}"
+  go run main.go install "${UNITY_PINNED_VERSION}" "${UNITY_PINNED_HASH}" "${IMPROBABLE_UNITY_ROOT}" "--modules" "${MODULES}"
+popd
+
```
jamiebrynes7: Should clean up the `go` directory after this has run

jared-improbable: Use a `trap` to ensure it's always run, even if the command fails

workers/unity/ProjectSettings/ProjectVersionHash.txt
```diff
@@ -0,0 +1 @@
+4bc57476174c
```
jared-improbable: This is a **little** odd, though I'm not sure I have a better suggestion.

zeroZshadow: I don't have another way to reconstruct the download url sadly

jared-improbable: What I meant was the location of this is odd - we've invented a new file that looks like Unity's, but it's not.

ci/prepare-unity.sh
zeroZshadow: I don't understand why this should clean up the local go folder? We don't clean up the Library/Temp/build folder either after a run?

workers/unity/ProjectSettings/ProjectVersionHash.txt
zeroZshadow: I can change the location, but I'm not sure what would be a better place for it.
I can put in in `ci` since it is only used by the ci? And rename it to something like `UnityEditorHash.txt`
I'd rather not hardcode it into the script itself, as this makes updating it later more of a pain

jessicafalk: I would be fine with moving it to `ci` (or maybe `ci/includes`) and naming it `UnityEditorHash.txt`

ci/prepare-unity.sh
jessicafalk: This is a very good point and we should clean these up as well

zeroZshadow: As discussed there is no change required here. TC Actually cleans this up after it has completed the steps.

zeroZshadow:
rebased onto latest master to fix docs problems

PR #95, opened 2018-07-13
Feature/UTY-642 Move GDK Modules into Unity Package Manager packages

mattyoung-improbable:
#### Description
Moves UnityGDK modules into Unity Package Manager packages. Adds support for referencing UPM packages outside of the unity project.

- Moves Assets/Gdk/<Module> to Packages/<com.improbable.gdk.module>
- Adds package.json files for each module and adds them as dependencies to manifest.json
- Adds tools/CopySchema that searches UPM packages for schema files and puts them in a schema/from_gdk_packages folder in the root.

See branch https://github.com/spatialos/UnityGDK/tree/example/UTY-642-SampleGame for an example project that imports packages from another project.

#### Tests
Unit Tests included for CopySchema
#### Documentation
Will add documentation on how this will work a starter project once we've tested it more. I'm sure there are going to be some GDK dependencies that will need to be put into the packages before they're truly modular.
#### Primary reviewers


.gitignore
```diff
@@ -45,3 +45,6 @@ workers/unity/Assets/Improbable.Generated.NetworkTypes/Generated/**/*.meta
 /workers/unity/Assets/Improbable.Generated.NetworkTypes/Generated/improbable/gdk/tests/*.cs
 
 code_generator/GdkCodeGenerator/Generated/
+
+# Schema from Unity Package Manager Packages (Copied via tools/CopySchema)
+schema/from_gdk_packages
```
samiwh: Generated code doesn't car about which directory this is in does it? I thought the schema compiler used to care, but I might have just been a dream.

tools/CopySchema/Program.cs
```diff
@@ -0,0 +1,202 @@
...
+                return other.Name == Name && other.Path == Path;
+            }
+
+            public override int GetHashCode()
+            {
+                return Name.GetHashCode() * (31 + Path.GetHashCode());
+            }
+        }
+
+        public const string FromGdkPackages = "from_gdk_packages";
```
samiwh: Is this a directory name? If so could we put that in the variable name?

tools/CopySchema/Program.cs
```diff
@@ -0,0 +1,202 @@
...
+                    results.Add(new PackageInfo(dep.Name, fullPath));
+                }
+            }
+
+            return results;
+        }
+
+        public static void ParseArguments(string[] args, out string manifestFile, out string packagesRoot, out string schemaRoot)
+        {
+            if (args.Length != 2)
```
samiwh: Would it make more snse just to pull out the two array elements into two arguments? I presume this is just here so that it's easier to pass command line arguments to it.

samiwh: Especially as the two arguments have explicit names already.

.gitignore
mattyoung-improbable: Spatial cares when it generates the schema descriptor

tools/CopySchema/Program.cs
samiwh: Resolved offline.

.gitignore
samiwh: Sorry I meant it doesn't do something weird like require packages to be in the corresponding directory? If this has worked then the answer is probably no.

PR #97, opened 2018-07-16
UTY-553: Connect a physical Android device to a local spatial instance

DanailPenev:
#### Description
- A splash screen with a connect button to connect the client worker
- Connection configuration is still automatically configured on PC
- A text input field to enter an IP address on mobile
- Leaving the input field blank on an Android Studio emulator will default to emulator config
#### Tests
Game connects and runs properly on PC, physical Android device and Android emulator
#### Documentation
-
#### Primary reviewers


workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -111,7 +99,13 @@ public void Start()
             foreach (var worker in Workers)
             {
                 LoadLevel(worker);
-                worker.Connect(connectionConfig);
+
+                if (worker.GetWorkerType.Equals("UnityClient"))
```
jessicafalk: `if (worker is UnityClient)`

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,99 @@
+using Improbable.Gdk.Core;
+using Unity.Entities;
+using UnityEngine;
+using UnityEngine.UI;
+#if UNITY_ANDROID
+using Improbable.Gdk.Android;
+
```
jessicafalk: newline

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,99 @@
...
+#endif
+
+namespace Playground
+{
+    [UpdateBefore(typeof(ConnectClientSystem))]
+    [UpdateBefore(typeof(SpatialOSUpdateGroup))]
```
jessicafalk: I'm not sure if that works. I believe you can only have one type of tag at a time

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,99 @@
...
+    internal class ConnectButtonClickSystem : ComponentSystem
+    {
+        private struct Data
+        {
+            public readonly int Length;
+            public EntityArray Entity;
+            public ComponentDataArray<WorkerEntityTag> DenotesWorker;
```
jessicafalk: add `[ReadOnly]`

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                connectParam.gameObject.SetActive(false);
+            }
+        }
+
+        protected override void OnUpdate()
+        {
+            for (var i = 0; i < data.Length; i++)
+            {
+                if (!clicked)
```
jessicafalk: move that out of the loop

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,99 @@
...
+                {
+                    SetConnectionParameters(GetInputString());
+                }
+            }
+#endif
+        }
+
+        private string GetInputString()
+        {
+            return connectParam.text;
```
jessicafalk: no need for an extra method here. just inline the code

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,99 @@
...
+        [Inject] private Data data;
+
+        protected override void OnCreateManager(int capacity)
+        {
+            base.OnCreateManager(capacity);
+            worker = WorkerRegistry.GetWorkerForWorld(World);
+            connectParam = GameObject.Find("ConnectParam").GetComponent<InputField>();
+            connectButton = GameObject.Find("ConnectButton").GetComponent<Button>();
+            connectButton.onClick.AddListener(IsClicked);
+            error = GameObject.Find("ConnectionError").GetComponent<Text>();
```
jessicafalk: call it `errorField`

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,99 @@
...
+        private WorkerBase worker;
+        private Text error;
+
+        [Inject] private Data data;
+
+        protected override void OnCreateManager(int capacity)
+        {
+            base.OnCreateManager(capacity);
+            worker = WorkerRegistry.GetWorkerForWorld(World);
+            connectParam = GameObject.Find("ConnectParam").GetComponent<InputField>();
```
jessicafalk: `connectParamInputField`

workers/unity/Assets/Playground/Scripts/Player/ConnectClientSystem.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    [UpdateBefore(typeof(SpatialOSUpdateGroup))]
+    internal class ConnectClientSystem : ComponentSystem
+    {
+        private struct Data
+        {
+            public readonly int Length;
+            public EntityArray Entity;
+            public ComponentDataArray<WorkerEntityTag> DenotesWorkerEntity;
+            public ComponentDataArray<ConnectButtonClicked> DenotesClickedConnectButton;
```
jessicafalk: same here: add [ReadOnly] where possible

workers/unity/Assets/Playground/Scripts/Player/PlayerComponentTypes.cs
```diff
@@ -0,0 +1,8 @@
+using Unity.Entities;
```
jessicafalk: would rename that script to `ConnectButtonClicked`, if that's the only thing that's stored in this file

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
DanailPenev: Rider complains if there is no newline. Should I just overrule it?

jessicafalk: just saw that we did the same thing for other definitions.. slightly weird ,but leave it for now

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+
+        private void IsClicked()
+        {
+            clicked = true;
+#if UNITY_ANDROID
+            if (Application.isMobilePlatform)
+            {
+                if (DeviceInfo.IsAndroidStudioEmulator() && connectParamInputField.text.Equals(""))
```
jessicafalk: use `String.Empty` for empty strings

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                worker.ConnectionConfig = ReceptionistConfig.CreateConnectionConfigForPhysicalAndroid(param);
+            }
+
+            // TODO: UTY-558 else -> cloud connection
+        }
+
+        private static bool IsIpAddress(string param)
+        {
+            return param.Split('.').Length == 4;
```
jessicafalk: use regex to check if it's an ip address: https://stackoverflow.com/questions/5284147/validating-ipv4-addresses-with-regexp

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
```diff
@@ -0,0 +1,93 @@
...
+#endif
+        }
+
+
+        private void SetConnectionParameters(string param)
+        {
+            if (IsIpAddress(param))
+            {
+                worker.ConnectionConfig = ReceptionistConfig.CreateConnectionConfigForPhysicalAndroid(param);
+            }
```
jessicafalk: we should show some kind of error, if it's not an ip address

workers/unity/Assets/Playground/Scripts/Player/ConnectClientSystem.cs
```diff
@@ -0,0 +1,57 @@
...
+            [ReadOnly] public ComponentDataArray<WorkerEntityTag> DenotesWorkerEntity;
+            [ReadOnly] public ComponentDataArray<ConnectButtonClicked> DenotesClickedConnectButton;
+        }
+
+        [Inject] private Data data;
+
+        private WorkerBase worker;
+        private GameObject panel;
+        private Button button;
+        private Text error;
```
jessicafalk: rename the variables here too :)

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
DanailPenev: I was thinking about doing this but it is generally advised ( https://www.perlmonks.org/index.pl?node_id=221512 ) not do to that. Maybe in our case it doesn't matter because we can assume they will be only entering it in human-readable format?

jessicafalk: I think it's fine, because with the current check you are also checking for 4 dots, which the page also pointed out as not being correct

jessicafalk: the other option would be to use: https://msdn.microsoft.com/en-us/library/system.net.ipaddress.tryparse(v=vs.110).aspx

though that seems like a bit of an overhead right now

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -111,7 +99,13 @@ public void Start()
             foreach (var worker in Workers)
             {
                 LoadLevel(worker);
-                worker.Connect(connectionConfig);
+
+                if (worker is UnityClient)
```
samiwh: This will need a comment.

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
samiwh: Turn off the warning in rider.
Pu a newline above the #if though.

DanailPenev: Resolved offline. Leaving it like that to be consistent with other files.

workers/unity/Assets/Playground/Scripts/Player/ConnectButtonClickSystem.cs
DanailPenev: Resolved.

DanailPenev: Resolved in 17bf200

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -84,16 +72,20 @@ public void Awake()
                     ? WorkerRegistry.CreateWorker<UnityClient>($"{workerType}-{Guid.NewGuid()}", new Vector3(0, 0, 0))
                     : WorkerRegistry.CreateWorker(workerType, workerId, new Vector3(0, 0, 0));
 
+                if (worker == null)
```
jessicafalk: we don't need that null check anymore

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -84,16 +72,20 @@ public void Awake()
...
-                connectionConfig = ConnectionUtility.CreateConnectionConfigFromCommandLine(commandLineArgs);
+                worker.ConnectionConfig = ConnectionUtility.CreateConnectionConfigFromCommandLine(commandLineArgs);
             }
 
             if (World.AllWorlds.Count <= 0)
             {
-                throw new InvalidConfigurationException(
-                    "No worlds have been created, due to invalid worker types being specified. Check the config in" +
-                    "Improbable -> Configure editor workers.");
+                Debug.LogError(
```
jessicafalk: don't use Debug.LogError, but the exception

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -108,15 +100,13 @@ public void Start()
...
-                    worker.Connect(connectionConfig);
-                }
-                catch (ConnectionFailedException exception)
-                {
-                    worker.View.LogDispatcher.HandleLog(LogType.Error, new LogEvent(exception.Message)
-                        .WithField("Reason", exception.Reason));
+                    continue;
                 }
+
+                worker.Connect(worker.ConnectionConfig);
```
jessicafalk: add the try-catch back in

PR #100, opened 2018-07-16
Add Mac steps to setup guide

anne-edwards:
#### Description
Added the Mac prerequisites to `setup-and-installing.md`.

Since only the prerequisites are different between operating systems, I've kept it as one doc and just added the Mac prerequisites into both the short version and the longer version.

#### Tests
n/a
#### Documentation
n/a - already docs :)
#### Primary reviewers
@jessicafalk @ElleEss

### DIFF
--- a/docs/setup-and-installing.md
+++ b/docs/setup-and-installing.md
@@ -4,8 +4,6 @@
 
 # Set up and get started with the SpatialOS Unity GDK
 
-**Note:** The Unity GDK currently runs on Windows only but a MacOS version is to come.
-
 ## Short version (for SpatialOS local deployment only)
 
 1. Set up your machine by installing: 
--- a/docs/setup-and-installing.md
+++ b/docs/setup-and-installing.md
@@ -8,7 +8,10 @@
 
 ## Short version (for SpatialOS local deployment only)
 
-1. Set up your machine by installing: [Unity 2018.2.0b10](https://unity3d.com/unity/beta-download), [Visual Studio 2017](https://www.visualstudio.com/downloads/), [SpatialOS](https://console.improbable.io/installer/download/stable/latest/win), [ReSharper](https://www.jetbrains.com/resharper/) (optional), and [ReSharper Command Line Tools](https://www.jetbrains.com/resharper/download/index.html#section=resharper-clt) (optional).
+1. Set up your machine by installing: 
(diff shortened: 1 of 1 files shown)


zeroZshadow:
This might need an extra note that we need `.NET Core 2.1.301` minimum. The version that visual studio 2017 installs is not good enough.

jessicafalk:
nope, 2.1.300 should be fine. I tested that
Visual Studio only provides 2.0

docs/setup-and-installing.md
```diff
@@ -4,11 +4,12 @@
...
 
-1. Set up your machine by installing: [Unity 2018.2.0b10](https://unity3d.com/unity/beta-download), [Visual Studio 2017](https://www.visualstudio.com/downloads/), [SpatialOS](https://console.improbable.io/installer/download/stable/latest/win), [ReSharper](https://www.jetbrains.com/resharper/) (optional), and [ReSharper Command Line Tools](https://www.jetbrains.com/resharper/download/index.html#section=resharper-clt) (optional).
+1. Set up your machine by installing: 
+* (For Windows) [Unity 2018.2.0b10](https://unity3d.com/unity/beta-download), [Visual Studio 2017](https://www.visualstudio.com/downloads/), [.NET Core 2.1](https://www.microsoft.com/net/download/), [SpatialOS](https://console.improbable.io/installer/download/stable/latest/win), [ReSharper](https://www.jetbrains.com/resharper/) (optional), and [ReSharper Command Line Tools](https://www.jetbrains.com/resharper/download/index.html#section=resharper-clt) (optional)
+* (For Mac) [Unity 2018.2.0b10](https://unity3d.com/unity/beta-download), [Rider](https://www.jetbrains.com/rider/), [.NET Core 2.1](https://www.microsoft.com/net/download/), [SpatialOS](https://console.improbable.io/installer/download/stable/latest/win)
```
jessicafalk: Rider should be optional as it is not a free software

jared-improbable: There's also Visual Studio for Mac, which Unity supports and is free.

anne-edwards: Is it similar to what we've said about ReSharper for Windows? I.e. is Rider optional unless you want to contribute to the repo, in which case you need it because of our linting rules?

jessicafalk: kind of, yes

anne-edwards:
@jessicafalk I've updated to clarify that Rider is optional (but necessary if you want to lint), and that you can use VS instead. Please approve if this is all good now :)

PR #101, opened 2018-07-16
Simplest version of moving to Improbadoc

bethaitman:
This is a PR to demonstrate the simplest version of porting the docs to use Improbadoc, our internal docs tool, so that we can publish these docs on the main docs.improbable.io site. 

Main changes: 

- Added a toc.md as a table of contents.
- Links:
   - Change all links from relative links (eg `../reactive-components.md`) to using `{{urlRoot}}` variable (eg   `{{urlRoot}}/content/reactive-components.md`). 
   - For links _not_ within the docs directory, change these links to absolute links.
- Assets
   - All asset names must be lowercase.
   - Use {{assetRoot}} to link to assets.
   - All assets need to be in assets folder.
- File renames
   - Renamed README to index - I think it didn't like the capital letters.
 - Tidying up:
    - Made line breaks after headings, and heading levels, consistent.
    - Fixed some formatting.

I've also had to add four empty files, in order to get linting to pass (they're required). I'm talking to Infra to get this linting rule removed.


docs/index.md
```diff
@@ -0,0 +1,48 @@
+# SpatialOS Unity GDK documentation
```
jessicafalk: if we rename it to index.md, it won't be shown by github as the main page of the docs folder

jamiebrynes7: I think we can write a pre-processing tool to convert Github format docs to Improbadoc format docs

bethaitman: That sounds like it could be a good solution if you want to keep the GitHub docs browseable, though I guess it's another thing to maintain.

If you wanted to try putting something together, I've written up the steps you'd need to run through to get something running (and therefore the steps a tool would need to do): https://docs.google.com/document/d/1nCWButxwioEk4y7SlB76u4jDYkID4sXvAQzoDfmLCfM/edit#

PR #105, opened 2018-07-16
Upgrade to 2018.2.0f2

zeroZshadow:
#### Description
Upgraded GDK to 2018.2.0f2
Small extras:
Disable unused VR module
Enable Audio module to fix warning about AudioListener

#### Tests
Ran builds on CI
Locally build UnityClient and LogicWorker

#### Documentation
A separate PR with documented steps on how this upgrade was done will be created before the ticket is signed off.

#### Primary reviewers
@oblm @jessicafalk


workers/unity/Packages/manifest.json
```diff
@@ -4,10 +4,10 @@
     "com.unity.entities": "0.0.12-preview.8",
     "com.unity.incrementalcompiler": "0.0.42-preview.14",
     "com.unity.package-manager-ui": "1.9.9",
+    "com.unity.modules.audio": "1.0.0",
```
jessicafalk: actually, why do we need the audio listener?

workers/unity/Packages/manifest.json
```diff
@@ -4,10 +4,10 @@
...
     "com.unity.modules.imgui": "1.0.0",
```
jessicafalk: can we get rid of imgui and ui as well? think it was also a dependency

zeroZshadow: All of these are for the playground game

PR #108, opened 2018-07-17
Change line endings to be platform specific.

martinzlocha:

-------

#### Description
Unix line endings should be used instead of dos so that there are no warnings on mac.

#### Tests
Run codegen, no dos line endings in codegened files were found.

#### Documentation
N/A

#### Primary reviewers

### DIFF
--- a/workers/unity/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingularTranslation.cs
+++ b/workers/unity/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingularTranslation.cs
@@ -18,6 +18,8 @@ namespace Generated.Improbable.Gdk.Tests
     {
         public class Translation : ComponentTranslation, IDispatcherCallbacks<global::Improbable.Gdk.Tests.ExhaustiveBlittableSingular>
         {
+            private const string LoggerName = "ExhaustiveBlittableSingular.Translation";
+        
             public override ComponentType TargetComponentType => targetComponentType;
             private static readonly ComponentType targetComponentType = typeof(SpatialOSExhaustiveBlittableSingular);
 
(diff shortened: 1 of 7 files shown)


jessicafalk:
Test this also on a Windows machine. Just ran it and now I'm getting the warnings on windows ;)

martinzlocha:
@jessicafalk The code is now using System.Environment.NewLine, please have a look. I think it throws warnings if you codegen with an open editor but when you reopen it OK and you can now codegen without closing.

jamiebrynes7:
@martinzlocha You'll want to check in any generated code that changed too

martinzlocha:
@jamiebrynes7 Thats not needed in this case because this change doesn't affect the generated code which we have checked in.

jamiebrynes7:
Oh my, somewhere some generated code that should be checked in.. hasn't been :/

jessicafalk:
do we want to add that generated code back in in this PR or a separate one?

martinzlocha:
screenshot_12

I was just looking at the generated code and I realized that if we use System.Environment.NewLine we have those two lines end with CRLF but all others are only LF. Also the issue with using System.Environment.NewLine in committed code is that every time someone commits on different OS it will ask him to commit the code generated files. Even though its not the nicest of solutions I think we should be using "\n" everywhere.

@jessicafalk I will add the codegened files here.

martinzlocha:
Tested on mac and everything seems to work fine.

samiwh:
@jessicafalk @martinzlocha normally I'd argue the generated code should be added back in with a separate PR as this is a codegen change so it would make this review easier. However given this is the one place where the exhaustive generated code will act differently from the other generated code (checking it in changes the line endings) I think it's fine to do it as a driveby.

martinzlocha:
@zeroZshadow Yes I should change the name of the PR. I realized that changing all line endings to unix is not practical and this is a better solution. I will change the pr name.

zeroZshadow:
However because it is now platform specific, the tests will constantly show up as changed as some of us work on unix systems and the others on windows.

martinzlocha:
@zeroZshadow Just tested this on Mac and after codegening there have been no changes picked up by git so I think it should be fine. If something arises though we can have a look again and convert everything to just /n

PR #116, opened 2018-07-19
UTY-375: Add ComponentUpdated when a component is added.

martinzlocha:

-------

#### Description
We should add ComponentUpdated when a new component is added.

Example generated translation code:

```csharp
public partial class Position
{
    public class Translation : ComponentTranslation, IDispatcherCallbacks<global::Improbable.Position>
    {
        private readonly Dictionary<Unity.Entities.Entity, SpatialOSPosition.Update> possibleUpdates 
            = new Dictionary<Unity.Entities.Entity, SpatialOSPosition.Update>();


        public void OnAddComponent(AddComponentOp<global::Improbable.Position> op)
        {
            var spatialOSPositionUpdate = new SpatialOSPosition.Update();

            var spatialOSPosition = new SpatialOSPosition();
            spatialOSPosition.Coords = global::Generated.Improbable.Coordinates.ToNative(data.coords);
            
            if (view.HasComponent<ComponentRemoved<SpatialOSPosition>>(entity))
            {
                ...
            }
            else if (!view.HasComponent<ComponentAdded<SpatialOSPosition>>(entity))
            {
                possibleUpdates.Add(entity, spatialOSPositionUpdate);
            }
        }

        public void OnAuthorityChange(AuthorityChangeOp op)
        {
            ...
            
            if (op.Authority == Authority.Authoritative)
            {
                possibleUpdates.Remove(entity);
            }
        }

        public override void PostReceive()
        {

            foreach (var update in possibleUpdates)
            {
                view.AddComponentsUpdated(update.Key, update.Value, UpdatesPool);
            }

            possibleUpdates.Clear();
        }
    }
}
```
```csharp
public class SpatialOSReceiveSystem : ComponentSystem
{
    {
        if (worker.Connection == null)
        {
            return;
        }

        do
        {
            using (var opList = worker.Connection.GetOpList(0))
            {
                dispatcher.Process(opList);
            }
        }
        while (inCriticalSection);

        foreach (var translationUnit in view.TranslationUnits.Values)
        {
            translationUnit.PostReceive();
        }
    }
}
```

#### Tests
Run code, nothing broke.

#### Documentation
Will add documentation in a separate PR.

#### Primary reviewers

### DIFF
--- a/code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
+++ b/code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
@@ -125,11 +125,14 @@ namespace <#= qualifiedNamespace #>
                 }
 <# if(fieldDetailsList.Count > 0) { #>
                 var data = op.Data.Get().Value;
+                var <#= componentDetails.CamelCaseTypeName#>Update = new <#= componentDetails.TypeName #>.Update();
 <# } #>
 
                 var <#= componentDetails.CamelCaseTypeName #> = new <#= componentDetails.TypeName#>();
 <# foreach (var fieldDetails in fieldDetailsList) { #>
                 <#= componentDetails.CamelCaseTypeName#>.<#= fieldDetails.PascalCaseName #> = <#= UnityTypeMappings.GetNativeTypeMethod(fieldDetails.RawFieldDefinition, "data." + fieldDetails.CamelCaseName, enumSet) #>;
+                <#= componentDetails.CamelCaseTypeName#>Update.<#= fieldDetails.PascalCaseName #> = new Option<<#= fieldDetails.Type #>>(<#= UnityTypeMappings.GetNativeTypeMethod(fieldDetails.RawFieldDefinition, "data." + fieldDetails.CamelCaseName, enumSet) #>);
+
 <# } #>
                 <#= componentDetails.CamelCaseTypeName#>.DirtyBit = false;
 
@@ -147,6 +150,9 @@ namespace <#= qualifiedNamespace #>
                 else if (!view.HasComponent<ComponentAdded<<#= componentDetails.TypeName #>>>(entity))
                 {
                     view.AddComponent(entity, new ComponentAdded<<#= componentDetails.TypeName #>>());
+<# if(fieldDetailsList.Count > 0) { #>
+                    view.AddComponentsUpdated(entity, <#= componentDetails.CamelCaseTypeName#>Update, UpdatesPool);
+<# } #>
                 }
                 else
                 {
--- a/code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
+++ b/code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
@@ -86,6 +86,8 @@ namespace <#= qualifiedNamespace #>
                 new ComponentPool<ComponentsUpdated<<#= componentDetails.TypeName #>.Update>>(
                     () => new ComponentsUpdated<<#= componentDetails.TypeName #>.Update>(),
                     (component) => component.Buffer.Clear());
+                    
+            private readonly Dictionary<Unity.Entities.Entity, <#= componentDetails.TypeName #>.Update> possibleUpdates = new Dictionary<Unity.Entities.Entity, <#= componentDetails.TypeName #>.Update>();
 <# } #>
 
             public Translation(MutableView view) : base(view)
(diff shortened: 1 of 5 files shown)


code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -147,6 +150,9 @@ namespace <#= qualifiedNamespace #>
                 else if (!view.HasComponent<ComponentAdded<<#= componentDetails.TypeName #>>>(entity))
                 {
                     view.AddComponent(entity, new ComponentAdded<<#= componentDetails.TypeName #>>());
+<# if(fieldDetailsList.Count > 0) { #>
+                    view.AddComponentsUpdated(entity, <#= componentDetails.CamelCaseTypeName#>Update, UpdatesPool);
```
samiwh: If authority is also added this will appear as two component updates with the same data.

jamiebrynes7: Is this true? @martinzlocha Could we get a full file for the generated code?

martinzlocha: Yup, thats true, gonna write a possible solution to avoid this.

martinzlocha: I have updated the codegen so that it does not generate duplicate updates.

code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -86,6 +86,8 @@ namespace <#= qualifiedNamespace #>
                 new ComponentPool<ComponentsUpdated<<#= componentDetails.TypeName #>.Update>>(
                     () => new ComponentsUpdated<<#= componentDetails.TypeName #>.Update>(),
                     (component) => component.Buffer.Clear());
+                    
+            private readonly Dictionary<Unity.Entities.Entity, <#= componentDetails.TypeName #>.Update> possibleUpdates = new Dictionary<Unity.Entities.Entity, <#= componentDetails.TypeName #>.Update>();
```
jared-improbable: Should these sorts of collections be lazily initialized? Are they guaranteed to be used 100% of the time?
If not, could we use the pattern `private asdf Blah { get => blah ?? ((blah = new asdf() )) }` ?

martinzlocha: This will always be used, when you receive OnComponentUpdate you don't know whether you will get authority and if you don't you have to add a ComponentUpdated so I always store the value in a dictionary.

martinzlocha: This would involve generating a system per class and registering it. Do you think this would be fine? Don't think it would be possible to have one global system because it needs to be aware of all of the fields of the component.

samiwh: I wouldn't generate a system for each one, that makes it hard to see what systems are running on a world and makes it hard to talk about what systems are part of core. I'd make one system which does that and make it call out to something generated. It might also be that the benefits don't outweigh the dis-benefits however I suspect that this is worth trying. Might be worth catching up with @jamiebrynes7 about whether or not the core is going to contain an easy to use AddUpdate method of a given component type when the core changes go in.

jamiebrynes7: An `AddUpdate` method as currently exists in the `MutableView`? Probably not - will pull that out into generated code as much as possible

samiwh: It would have to go in generated code, but the question is will it be accessible. We'd also need a helper method for turn data to update (which you likely want for users anyway). If you have a single system doing this then it would likely have to iterate over whatever the translation equivalent is and call a function which matches new components and adds an update to the entity.

samiwh: It might be best to just use the book keeping in the current core and do this in the new one. Or just go straight to the new one.

martinzlocha: I have implemented a version using group filters and added it to the design doc. Please have a look: https://brevi.link/reactive-component-addition

martinzlocha:
Closing this in favor of a better implementation.

PR #117, opened 2018-07-19
UTY-476: Log error if users attempt to respond to a command request twice.

martinzlocha:

-------

#### Description
If two systems each inject for command requests and attempt to respond to command requests, we currently respond to a command requests twice. In this case one of the two responses is dropped by the bridge. We want to avoid the situation where something goes wrong and the user can't tell why and we want to help users debug their issues through meaningful error messages.

#### Tests
Send two responses to a command, got an error warning.

#### Documentation
N/A

#### Primary reviewers

### DIFF
--- a/code_generator/GdkCodeGenerator/Templates/UnityCommandPayloadGenerator.tt
+++ b/code_generator/GdkCodeGenerator/Templates/UnityCommandPayloadGenerator.tt
@@ -9,6 +9,7 @@
 
 using Improbable.Gdk.Core;
 using System.Collections.Generic;
+using UnityEngine;
 
 namespace <#= qualifiedNamespace #>
 {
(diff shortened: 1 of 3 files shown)


JonasImprobable:
Did you regenerate the generated testfiles? (`schema\improbable\gdk\tests`)

code_generator/GdkCodeGenerator/Templates/UnityCommandPayloadGenerator.tt
```diff
@@ -44,6 +45,17 @@ namespace <#= qualifiedNamespace #>
 
                 public void Send<#= commandDetails.CommandName #>Response(<#= commandDetails.ResponseType #> response)
                 {
+                    if (Translation.<#= commandDetails.CommandName #>ResponsesToSend.Contains(RequestId))
```
JonasImprobable: This approach here would detect if we attempt to send duplicate command responses during the same frame but it wouldn't be able to detect duplicate command responses after the first response was sent, can it? I.e. send a response in frame 1, send a duplicate response in frame 2.

martinzlocha:
After discussion with @JonasImprobable we agreed that this solution wont work if you store the command and respond in a different tick. Further discussed other solutions with @samiwh and there seems no elegant solution to do this without unbounded memory so will close this for the time being.

code_generator/GdkCodeGenerator/Templates/UnityCommandPayloadGenerator.tt
mattyoung-improbable: Presumably the components that allow sending responses get cleaned up at the end of the frame they're responded on? So the user would have to jump through some hoops to keep a reference around for a frame in order to hit this issue?

martinzlocha: Yup they won't get injected again but we already have the playerlifecycle system which stores it for later so I dont think it will be uncommon.

PR #118, opened 2018-07-19
UTY-713: Add docs for logging.

martinzlocha:

-------

#### Description
Public logging docs.

#### Tests
N/A

#### Documentation
N/A

#### Primary reviewers

### DIFF
--- a/docs/content/logging.md
+++ b/docs/content/logging.md
@@ -4,28 +4,28 @@
 
 ##  Logging
 
-The Unity GDK uses a custom ILogDispatcher interface instead of UnityEngine.Debug, which gives more flexibility to handle logs separately in different workers and gives more context when handling the logs. There are two provided implementations of this interface: 
+The Unity GDK uses a custom `ILogDispatcher` interface instead of `UnityEngine.Debug`, which gives more flexibility to handle logs separately in different workers and gives more context when handling the logs. There are two provided implementations of this interface: 
 
-*  LoggingDispatcher, which simply logs to the Unity console
-*  ForwardingDispatcher, which logs to the Unity console and forwards it to the SpatialOS Console
+*  `LoggingDispatcher`, which simply logs to the Unity console
+*  `ForwardingDispatcher`, which logs to the Unity console and forwards it to the SpatialOS Console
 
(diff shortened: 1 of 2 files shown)


docs/content/logging.md
```diff
@@ -0,0 +1,65 @@
+**Warning:** The [pre-alpha](https://docs.improbable.io/reference/latest/shared/release-policy#maturity-stages) release is for evaluation purposes only, with limited documentation - see the guidance on [Recommended use](../../README.md#recommended-use).
+
+-----
+
+##  Logging
+
+The Unity GDK uses a custom ILogDispatcher interface instead of UnityEngine.Debug, which gives more flexibility to handle logs separately in different workers and gives more context when handling the logs. There are two provided implementations of this interface: 
+
+*  LoggingDispatcher, which simply logs to the Unity console
```
jessicafalk: only realised now, but we also have punctuation at the end of a list, I believe

anne-edwards:
@jessicafalk we do? 🤔 (I just removed it!) 

I don't mind much either way to be honest - if Laura suggested/agreed to full stops at the end of bulleted lists, then that's fine :)

martinzlocha:
@jessicafalk Added the full stops :)

jessicafalk:
hm.. so I'm not sure anymore. I checked a few docs and on some there are and on some there aren't. I think Laura was in favor of this, but not completely sure. As you are the tech writer in charge now, I'll leave the decision to you whether there should be full stops or not :)

samiwh:
We have a new process for dealing with docs PRs. https://brevi.link/unity-gdk-wip-docs

anne-edwards:
Good point @samiwh.

I think all this means for this particular PR is that @martinzlocha should add the edit status at the bottom of the doc (I think the edit status should be "Full review"). And then make a note of any further work required if necessary. TBH I haven't been very involved in the process of creating that doc, but I believe what needs adding is:

[//]: # (Editorial review status:XXXXXX)

anne-edwards:
As for full stops at ends of lists - I would say nope, leave them out. But it's a _very_ minor point :)

PR #120, opened 2018-07-19
UTY-590: SpatialOSBehaviourLibrary

MPeti01:
#### Description
Added the SpatialOSBehaviourLibrary, a class to be used for reflectively collecting Reader and Writer members in a MonoBehaviour tagged with [Require] and injecting them (once we *have* Readers and Writers that is). Partly inspired by the corresponding implementation in the SDK, partly by common sense and sanity where the legacy code was very legacy code.
#### Tests
Included a set of tests to cover the newly added functionality, except the code paths that can't easily tested without actually having the creation of readers and writers implemented - marked those with TODO.
#### Documentation
In-code documentation.
#### Primary reviewers
@JonasImprobable @firtina-improbable


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/MemberReflectionUtil.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Improbable.Gdk.Core
+{
+    class MemberReflectionUtil
```
JonasImprobable: Give this class an access modifier, e.g. `public`.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderInterfaceAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    /// Marks an interface as the Reader Interface for use by visualizers.
```
JonasImprobable: "Visualizers" is not a known term in the GDK. ;) How about just "Marks an interface as a Reader interface."?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/RequireAttribute.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core
+{
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
```
JonasImprobable: Please add a short summary here.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+using Unity.Entities;
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    /// Retrieves Reader and Writer fields from MonoBehaviours and handles injection into them.
+    /// </summary>
+    public class SpatialOSBehaviourLibrary {
+        // Stores maps from Spatial component IDs to MemberAdapters to reflectively set Readers/Writers for each SpatialOSBehaviour
```
JonasImprobable: Full stop at the end of the sentence.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+            = new Dictionary<Type, Dictionary<uint, IMemberAdapter>>();
+
+        // List of component IDs for which a reader/writer is required for each SpatialOSBehaviour
+        private readonly Dictionary<Type, List<uint>> componentReaderIdsForBehaviours = new Dictionary<Type, List<uint>>();
+        private readonly Dictionary<Type, List<uint>> componentWriterIdsForBehaviours = new Dictionary<Type, List<uint>>();
+
+        // Helper objects for querying members with the [Require] attribute
+        private readonly MemberReflectionUtil reflectionUtil = new MemberReflectionUtil(typeof(RequireAttribute));
+
+        private readonly HashSet<Type> badTypes = new HashSet<Type>();
```
JonasImprobable: How about `invlalidMonoBehaviourTypes` instead to make it more self-explanatory?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        private readonly ILogDispatcher logger;
+        private const string LoggerName = "SpatialOSBehaviourLibrary";
+
+        private const string BadRequiredMemberWarning
+            = "[Require] attribute found on member that is not Reader or Writer or incorrectly generated, ignoring!";
+
+        private const string MultipleReadersWritersRequiredError
+            = "MonoBehaviour found requesting more than one view (Reader or Writer) to the same component, " +
```
JonasImprobable: Don't mention the word "view" here as that might confuse people. Just say `MonoBehaviour found requesting more than one Reader or Writer for the same component...`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+                DeInject(spatialOSBehaviour, readerWriterComponentId);
+            }
+
+            foreach (var readerWriterComponentId in componentWriterIdsForBehaviours[spatialOSBehaviour.GetType()])
+            {
+                DeInject(spatialOSBehaviour, readerWriterComponentId);
+            }
+        }
+
+        public IEnumerable<uint> GetRequiredReaderComponentIds(Type behaviourType)
```
jamiebrynes7: Is there a reason this returns an `IEnumerable` and not a `List` ?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                // Figure out if reader or writer
+                // Get component ID
+                // Store in data structures
+                Type requiredType = adapter.TypeOfMember;
+                var componentIdAttribute =
+                    (ComponentIdAttribute) Attribute.GetCustomAttribute(requiredType, typeof(ComponentIdAttribute), false);
+                if (componentIdAttribute == null)
+                {
+                    logger.HandleLog(LogType.Warning, new LogEvent(BadRequiredMemberWarning)
```
jamiebrynes7: This warning doesn't exactly correspond to the condition under which it is fired.

This will fire if the underlying type injected doesn't have a component id attribute - this doesn't tell you whether the underlying type is a reader or writer at all.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+
+                var componentId = componentIdAttribute.Id;
+                if (componentIdsToAdapters.ContainsKey(componentId))
+                {
+                    logger.HandleLog(LogType.Error, new LogEvent(MultipleReadersWritersRequiredError)
+                        .WithField(LoggingUtils.LoggerName, LoggerName)
+                        .WithField("MonoBehaviour", behaviourType.Name)
+                        .WithField("ComponentID", componentId)
+                        .WithField("RequiredType", requiredType.Name));
+                    badTypes.Add(behaviourType);
```
jamiebrynes7: Shouldn't this break this iteration of the loop? Currently we fire the error and then overwrite the previous adapter

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/MemberReflectionUtil.cs
```diff
@@ -0,0 +1,57 @@
...
+            for (int i = 0; i < members.Count; i++)
+            {
+                if (Attribute.IsDefined(members[i], attributeType, false))
+                {
+                    resultList.Add(adapterFactory(members[i]));
+                }
+            }
+        }
+
+        private static FieldInfoAdapter CreateFieldAdapter(FieldInfo member)
```
samiwh: Why make this a function?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: +1 - use concrete types unless there is a good reason not to

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/MemberReflectionUtil.cs
```diff
@@ -0,0 +1,57 @@
...
+            this.attributeType = attributeType;
+        }
+
+        /// <returns>
+        ///     A MemberAdapter list of properties and fields declared in given type that match at least one of given attributes.
+        /// </returns>
+        internal List<IMemberAdapter> GetMembersWithMatchingAttributes(Type targetType)
+        {
+            List<IMemberAdapter> result = new List<IMemberAdapter>();
+            ProcessMembers(result, targetType.GetProperties(PropertyFlags), CreatePropertyAdapter);
```
samiwh: This might be worth unpacking. This is currently a lot harder to read than the 10 lines or so that you'd get if you unwrap it.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: I was trying to figure out how to communicate that I'm talking about more than one [Reader or Writer], rather than [more than one reader] or [more than one writer], ie you can't even have a reader and a writer. Any suggestions on how to make that clear?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/FieldInfoAdapter.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Reflection;
+
+namespace Improbable.Gdk.Core
+{
+    public class FieldInfoAdapter : IMemberAdapter
```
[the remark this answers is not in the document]
samiwh: I'd also make them internal. Unless there is a reason to expose these to users.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ComponentIdAttribute.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    /// Associates a unique component Id with a specific interface.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Interface)]
+    public class ComponentIdAttribute : Attribute
```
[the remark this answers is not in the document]
MPeti01: Yes

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: Will change

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/PropertyInfoAdapter.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            get { return member; }
+        }
+
+        public PropertyInfoAdapter(PropertyInfo member)
+        {
+            this.member = member;
+        }
+
+        public void SetValue(object obj, object value)
```
samiwh: These can be specialised. E.g. we know that obj is always going to be a monobehaviour

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: It doesn't really matter because whenever we would use the adapter afterwards, we first check and see that it's an invalid type and avoid touching it. But probably would be less confusing if we had a break, so I'll add that

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/IMemberAdapter.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Reflection;
+
+namespace Improbable.Gdk.Core
+{
+    /// <returns>
+    ///     A handle on a field/property of a type obtained through reflection, for getting/setting it on instances conveniently.
+    /// </returns>
+    public interface IMemberAdapter
```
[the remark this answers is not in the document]
MPeti01: I'm not sure what the alternative is that you're suggesting, is it just to have the types more specific, setting IComponentReaders on MonoBehaviours? If that's it then yes, that should work fine too

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/RequireAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core
+{
+    /// <returns>
+    ///     Marks fields/properties of MonoBehaviours that require a Reader or Writer to be injected into them.
+    /// </returns>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
```
samiwh: Do we need to be able to inject properties at all? If we only do fields then users can still expose these outside of the class. I don't think I've ever seen this done with a property. It also get's us into the world of trying to set values to properties that might do something odd with the `value` field. This will likely always work for autoproerties, but is breakable elsewhere.

MPeti01: Since it was supported in the SDK like this, I assumed it should be brought over to make it as easy as possible to move SDK code to the GDK.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+        // Helper objects for querying members with the [Require] attribute
+        private readonly MemberReflectionUtil reflectionUtil = new MemberReflectionUtil(typeof(RequireAttribute));
+
+        private readonly HashSet<Type> badTypes = new HashSet<Type>();
+
+        private readonly ILogDispatcher logger;
+        private const string LoggerName = "SpatialOSBehaviourLibrary";
+
+        private const string BadRequiredMemberWarning
+            = "[Require] attribute found on member that is not Reader or Writer or incorrectly generated, ignoring!";
```
samiwh: Is this ignoring the invalid object or the monobehaviour?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/IComponentReader.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Improbable.Gdk.Core
+{
+    public interface IComponentReader
```
MPeti01: I just noticed I left this in, this should not actually exist like this. We'll probably have some supertype once we implement the readers/writers, but the component ID needs to be in the type attribute, because we need access to it when we have a reference to the Type object, not an instance of the type. Probably unnecessary to have this property.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+            memberAdapter.SetValue(spatialOSBehaviour, readerWriter);
+        }
+
+        private void DeInject(MonoBehaviour spatialOSBehaviour, uint componentId)
+        {
+            var memberAdapter = adapterCache[spatialOSBehaviour.GetType()][componentId];
+            memberAdapter.SetValue(spatialOSBehaviour, null);
+        }
+
+        private void EnsureLoaded(Type behaviourType)
```
samiwh: Wouldn't you want to iterate over all monobehavious once on worker startup?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/MemberReflectionUtil.cs
MPeti01: Good point

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: The idea is that all reader and writer interfaces have to be generated with the component ID attribute as well, so if something doesn't have it, it must be something else or a malformed reader/writer or something.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+        private const string MultipleReadersWritersRequiredError
+            = "MonoBehaviour found requesting more than one view (Reader or Writer) to the same component, " +
+            "this is invalid and will not be enabled!";
+
+        public SpatialOSBehaviourLibrary(ILogDispatcher logger)
+        {
+            this.logger = logger;
+        }
+
+        public void InjectAllReadersWriters(MonoBehaviour spatialOSBehaviour)
```
samiwh: Where would this be called?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+using System.Collections.Generic;
+using Unity.Entities;
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    /// Retrieves Reader and Writer fields from MonoBehaviours and handles injection into them.
+    /// </summary>
+    public class SpatialOSBehaviourLibrary {
```
samiwh: This seems like it mixes worker wide stuff with gameobject specific stuff. What do we expect to own instances of this?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: Read through the code below. Looks like we ignore the whole monobehaviour (which seems reasonable). Please add that to the message.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterFactory.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core
+{
+    public static class ReaderWriterFactory
+    {
+        public static object CreateReaderWriter(uint componentId)
+        {
+            throw new NotImplementedException();
```
firtina-improbable: Wondering if we should target an experimental branch for `monobehaviour stuff` rather than `master` if we're going to be merging in `NotImplementedException`s. Thoughts?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        private void EnsureLoaded(Type behaviourType)
+        {
+            if (adapterCache.ContainsKey(behaviourType))
+            {
+                return;
+            }
+
+            var adapters = reflectionUtil.GetMembersWithMatchingAttributes(behaviourType);
```
samiwh: we can probably just pull this out into a function. This class is mostly just a wrapper around a single function and that function doesn't have all that much code in ti.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: By the GameObjectRepresentation ECS component, which will keep track of which MonoBehaviours on the corresponding GameObjects require which readers/writers and when those are available.

MPeti01: By methods on the GameObjectRepresentation ECS component, which will be keeping track of which readers/writers are needed by the behaviours on its gameobject, and when those are all available (components exist and authority for writers)

MPeti01: See also the 3rd design doc for readers/writers, this is sketched there pretty accurately

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: This was initially a static class made non-static for testability. It's probably going to be created by a dispatcher ECS system that calls into the GameObjectRepresentation ECS component's methods, so the same instance will be shared between all gameobjects.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/PropertyInfoAdapter.cs
JonasImprobable: I actually kinda like the general nature of the current implementation. An argument can be made that this class could in theory be used for arbitrary classes. We would probably also need to call the class `MonoBehaviourPropertyInfoAdapter` otherwise which feels overkill. Just my two cents.

MPeti01:
@jamiebrynes7 
The Utils class was grabbed from the SDK as is, probably a good point to move it in IF we don't take the route of going through all MonoBehaviours on init. That would require a lot more reflection magic that I'd keep all separate in a ReflectionUtils class (dropping the Member from the name).

The IMemberAdapter is useful because it refers to a member of an unknown type to assign to it. Unless you have any other view of how the `Inject()` method should do its job? Plus it provides a nice clean way to handle reflection on properties and fields the same way.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/RequireAttribute.cs
JonasImprobable: Let's avoid breaking changes for now and think about modifications later.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+    /// <summary>
+    /// Retrieves Reader and Writer fields from MonoBehaviours and handles injection into them.
+    /// </summary>
+    public class SpatialOSBehaviourLibrary {
+        // Stores maps from Spatial component IDs to MemberAdapters to reflectively set Readers/Writers for each SpatialOSBehaviour.
+        private readonly Dictionary<Type, Dictionary<uint, IMemberAdapter>> adapterCache
+            = new Dictionary<Type, Dictionary<uint, IMemberAdapter>>();
+
+        // List of component IDs for which a reader/writer is required for each SpatialOSBehaviour
+        private readonly Dictionary<Type, List<uint>> componentReaderIdsForBehaviours = new Dictionary<Type, List<uint>>();
```
JonasImprobable: Can the lists still contain duplicates? If not, let's change it to hashsets to be safe. ;)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
jamiebrynes7: I think the problem is this error is used in two places for two actually different errors:

1. A Reader/Writer doesn't have a `[ComponentId]` attribute
2. The thing marked as `[Require]` is not a Reader/Writer

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterFactory.cs
MPeti01: As long as we're not making anything nonfunctional actually run I'd say it's okay, especially since it shouldn't take a massively long time? And even though other changes in master shouldn't affect us much, when they do it's nice to keep up with them and not have huge merge conflicts at the end.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: They can't so good point

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+using Unity.Entities;
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    /// Retrieves Reader and Writer fields from MonoBehaviours and handles injection into them.
+    /// </summary>
+    public class SpatialOSBehaviourLibrary {
+        // Stores maps from Spatial component IDs to MemberAdapters to reflectively set Readers/Writers for each SpatialOSBehaviour.
```
JonasImprobable: I prefer to keep comments to a minimum especially if the intention of the field can be guessed without the comment. Additional comments represent a potential maintenance burden.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+        private readonly Dictionary<Type, Dictionary<uint, IMemberAdapter>> adapterCache
+            = new Dictionary<Type, Dictionary<uint, IMemberAdapter>>();
+
+        // List of component IDs for which a reader/writer is required for each SpatialOSBehaviour
```
JonasImprobable: Please remove this comment as it doesn't tell the user anything new.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,166 @@
...
+        private readonly Dictionary<Type, List<uint>> componentReaderIdsForBehaviours = new Dictionary<Type, List<uint>>();
+        private readonly Dictionary<Type, List<uint>> componentWriterIdsForBehaviours = new Dictionary<Type, List<uint>>();
+
+        // Helper objects for querying members with the [Require] attribute
```
JonasImprobable: Please remove this comment. The next line is self-explanatory enough.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
JonasImprobable: Could this be a hashset?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/PropertyInfoAdapter.cs
MPeti01: Also right now I'm not sure if the readers are actually going to have a supertype and if so what, so keeping them general helps not force that decision before other PRs are in

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/RequireAttribute.cs
samiwh: How well did the sdk support it though? If I created a property that isn't an auto property what would happen?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterFactory.cs
firtina-improbable: It's still good to prevent committing partial functionality into master IMO, it should be trivial to retarget onto the branch

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/MemberReflectionUtil.cs
```diff
@@ -0,0 +1,57 @@
...
+            ProcessMembers(result, targetType.GetFields(FieldFlags), CreateFieldAdapter);
```
JonasImprobable: Aren't the field flags already contained in the property flags? I'm not sure how the `private const BindingFlags PropertyFlags = FieldFlags | BindingFlags.SetProperty | BindingFlags.GetProperty;` part works but it might make sense to remove the `FieldFlags` from the `PropertyFlags`.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: If you iterate over it regularly then List is always better. Use a hashset if you are are mostly checking for values being contained.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/FieldInfoAdapter.cs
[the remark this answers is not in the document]
MPeti01: @jamiebrynes7 I'm adding class level comments, but the functions seem self explanatory to me? I don't know what I'd add other than "sets the thing", "gets the type" which is not exactly useful...

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterFactory.cs
JonasImprobable: Let's have the feature branch regardless just to follow a clean approach. It is indeed weird to have code that throws `NotImplementedExceptions` checked into master. Same for the placeholder unit tests. Master should be a shining oasis sort of thing. ;)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: I'd struggle to work out what the next line is without reading the code.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: I'd go with Jonas's suggestion. View is an overloaded term (and not quite right here).

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: I don't think the actual code in that doc was the thing that was signed off. It was the issues raised. But more importantly, where would you inject all the readers and writers? Wouldn't you inject only the ones that become available?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/IMemberAdapter.cs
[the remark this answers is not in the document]
jared-improbable: I believe this approach was carried over from the Unity SDK?
I think having these is probably over-engineering.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/PropertyInfoAdapter.cs
jared-improbable: Let's not over-design something on the chance that it *may* be used at some point. Solve exactly the problem you intend to solve now/

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: This will convolute the control flow slightly but I'll make it so that there's a separate error message if there is no Reader/Writer marks on the required thing, or if there's Reader/Writer but no ComponentId. The latter really shouldn't happen though.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/RequireAttribute.cs
jared-improbable: It only makes it easier if properties were used as injection targets. Why not start with the simplest baseline (fields) and then add to it later, if it's absolutely necessary.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: Same comment as above. Are you iterating over it or sporadically checking if a value exists?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: As far as I know the plan is to keep track of which / how many are still not ready, and when everything is ready only then inject everything and enable the monobehaviour. I.e. when a new component is added or authority is received and the GameObjectReference sees that this was the last remaining piece, it calls this function.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: I'd still get rid of refelction utils for the reasons mentioned elsewhere, however this doesn't need to be static. If a core system sets up the internal stuff for accepting readers and writers then it can own a single instance of something which finds all monobehaviours. Then you have a testable class that doesn't do reflection multiple times.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: I think GameObjectReference will need to iterate over it once for each MonoBehaviour on each GameObject, so I'll keep it a list then.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/PropertyInfoAdapter.cs
samiwh: If we start doing lots of arbitrary reflection then arbitrary utils classes makes sense. While it's just one specific class it doesn't though.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/MemberReflectionUtil.cs
MPeti01: Indeed that seems strange, I'll try to figure out... (Code was taken as is from SDK) But it's definitely not duplicating adapters, whatever the reason is, because the tests would have caught that.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: How does it do the entity specific stuff then? Like the inject functions.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/RequireAttribute.cs
MPeti01: Looking through the SDK implementation and the C# reflection documentation, I'm pretty sure it would call your custom setter to inject the reader/writer, and this is what this code will end up doing as well.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/MemberReflectionUtil.cs
MPeti01: Yeah those flags make absolutely no sense, will fix them

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/IMemberAdapter.cs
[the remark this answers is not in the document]
MPeti01: It's useful for handling fields and properties together. Perhaps not an important thing to have, but the SDK supported it and since we have this implementation it's not a hard thing to have that supported right away.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: Actually no, we only ignore the whole monobehaviour if it requires several readers/writers to the same component type. If it requires something nonsensical, we just ignore that member. I'll clarify that.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: Discussed offline - there is going to be one lazily initialised library per worker so it's OK as is

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: Probably iterating so keeping it a list

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: Discussed offline - as far as I understand this will be created in an ECS system, passed to each GameObjectRepresentation ECS component (one per entity), and that component will do the bookkeeping of what can be activated when, calling into the library to 1) know what components it's looking for, 2) do the injections when everything is ready. (Then it enables the monobehaviour.)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
JonasImprobable: Required readers/writers in a monobehaviour shall be either all injected or all null at all times as signed off in the design doc.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,198 @@
...
+        private const BindingFlags FieldFlags = BindingFlags.Instance | BindingFlags.NonPublic |
+            BindingFlags.Public | BindingFlags.DeclaredOnly;
+
+        private const BindingFlags PropertyFlags =
+            FieldFlags | BindingFlags.SetProperty | BindingFlags.GetProperty;
+
+        private List<IMemberAdapter> GetMembersWithMatchingAttributes(Type targetType)
+        {
+            List<IMemberAdapter> adapters = new List<IMemberAdapter>();
+            foreach (var property in targetType.GetProperties(PropertyFlags))
```
jared-improbable: I think a quick survey of AAA, Spilt Milk and Klang about whether or not they actually inject into Properties is in order. If they don't, we can drop all of this extra machinery, interfaces and classes that try to normalize field/property handling.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ComponentIdAttribute.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core
```
jared-improbable: Shouldn't all of this be in a feature module rather than the Core?

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/SpatialOSBehaviourLibraryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        public void SpatialOSBehaviourLibrary_injects_Reader()
+        {
+            // TODO - Needs ReaderWriterFactory implemented first!
+            Assert.Pass();
+        }
+
+        [Test]
+        public void SpatialOSBehaviourLibrary_injects_two_Writers()
+        {
+            // TODO - Needs ReaderWriterFactory implemented first!
```
jared-improbable: How about asserting that it throws `NotImplementedException` so tests can't be forgotten?

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/SpatialOSBehaviourLibraryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+        }
+
+        [WriterInterface]
+        [ComponentId(ComponentId2)]
+        private interface DummyWriter2
+        {
+        }
+
+        private class SingleReaderBehaviour : MonoBehaviour
```
jared-improbable: How about a test for a Reader + Writer?

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/SpatialOSBehaviourLibraryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        private class TwoWritersBehaviour : MonoBehaviour
+        {
+            [Require] public DummyWriter1 Writer1;
+            [Require] public DummyWriter2 Writer2;
+        }
+
+        private class MultipleReadersOfSameType : MonoBehaviour
+        {
+            [Require] public DummyReader Reader1;
+            [Require] public DummyReader Reader2;
```
jared-improbable: If we insist that we do need support for injecting properties, then we need tests for them as well.

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/SpatialOSBehaviourLibraryTests.cs
MPeti01: Good point now that we are merging to a feature branch and don't need all tests to pass.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: What should be the process with this PR for the while? It's kind of annoying if merging it has to wait for that, when we are trying to complete the monobehaviour feature set. Maybe insist on having an answer before we merge the feature branch into master?

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/SpatialOSBehaviourLibraryTests.cs
MPeti01: Do you mean a Reader + Writer for the same component (which should give an error just like two readers of the same type), or two separate components? If it's separate, I thought it was really not that different from two different readers. If it's the same, I could see an argument for that.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ComponentIdAttribute.cs
JonasImprobable: Our decision for now is to put the monobehaviour workflow into core. We can still separate it out later.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: I don't think that is what we discussed in the meetings. That reintroduces old race conditions.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterFactory.cs
samiwh: +1. Don't merge this into master. However make it an experimental branch as it needs to have PRs to add code to it.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ComponentIdAttribute.cs
samiwh: We will almost certainly move this out as nothing needs to be in core.
I actually think we should revisit the decision to have anything in core.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: I think leave it out and add it in later if we need it. It's not something I've ever seen used and seems brittle.

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/SpatialOSBehaviourLibraryTests.cs
MPeti01: Done

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: That way round doesn't make sense. If it requires two readers to the same type that should be perfectly well defined. We could warn a user if we think there is literally no use case for it, but it's asking for a well defined thing.
If a monobehaviour requires something that isn't a reader or writer that is an error and means either they meant to use a different attribute or that they expect different behaviour. In that case we should insist on fixing the monobehaviour before running it.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ComponentIdAttribute.cs
JonasImprobable: If we want to move everything out of core then that should be done in a separate PR at a different time. The decision in the meeting was made just to prevent deadlocks like this.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
JonasImprobable: This was 100% discussed. Do we really need to start recording every meeting now? What are the race conditions you talked about? Also, these issues could have been pointed out way earlier. We had the design doc and the production meeting where this was mentioned.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ComponentIdAttribute.cs
MPeti01: We can do it easily just before merging the final state of the feature branch into master. Right now there's monobehaviour things in different PRs / repos to-be-made-prs all in the Core namespace, changing that would be a hassle

jessicafalk: I would prefer to do it earlier, but agree that it doesn't have to be done in this pr

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: The change required to this code is really minimal though, we'd literally just need to make the private Inject()/DeInject() methods public, and could remove the InjectAll() / DeInjectAll() methods if they are not used. I'm not sure about the discussion but I don't think it should block this PR anyway.

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/SpatialOSBehaviourLibraryTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            // TODO - Needs ReaderWriterFactory implemented to pass!
+            var behaviour = testGameObject.AddComponent<SingleReaderBehaviour>();
+            library.InjectAllReadersWriters(behaviour);
+            Assert.NotNull(behaviour.Reader);
+        }
+
+        [Test]
+        public void SpatialOSBehaviourLibrary_injects_two_Writers()
+        {
+            // TODO - Needs ReaderWriterFactory implemented to pass!
```
jessicafalk: does this mean it fails currently? then CI will fail as well

jessicafalk:
the experimental branch doesn't seem to be set up correctly. it should not be possible to merge into it without soemone accepting the pr and it passing CI

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/SpatialOSBehaviourLibraryTests.cs
```diff
@@ -85,7 +85,8 @@ public void TearDown()
         [Test]
         public void SpatialOSBehaviourLibrary_injects_Reader()
         {
-            // TODO - Needs ReaderWriterFactory implemented to pass!
+            // TODO - Needs ReaderWriterFactory implemented to pass properly!
```
jessicafalk: this looks like something that could easily be forgotten... would prefer to have a ticket for that. we should not have TODOs in the code that are not associated with a ticket.

Also not a big fan of `Assert.Pass()` as it throws exceptions and therefore the whole thing could end up in a weird state. Just comment out the lines below and run an `Assert.IsTrue(true)` or remove the code completely and add it to the ticket

MPeti01: Done and done: https://improbableio.atlassian.net/browse/UTY-732

MPeti01:
Production confirmed that Bossa and SpiltMilk didn't appear to use Readers/Writers with properties at all, so removed now.

Seems ready for merging to me now. @jessicafalk @JonasImprobable ?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,187 @@
...
+            componentWriterIdsForBehaviours[behaviourType] = writerIds;
+            foreach (var adapter in adapters)
+            {
+                // Figure out if reader or writer
+                // Get component ID
+                // Store in data structures
+                Type requiredType = adapter.FieldType;
+                var isReader = Attribute.IsDefined(requiredType, typeof(ReaderInterfaceAttribute), false);
+                var isWriter = Attribute.IsDefined(requiredType, typeof(WriterInterfaceAttribute), false);
+                if (!isReader && !isWriter)
```
firtina-improbable: Is `isReader && isWriter` possible?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        private void EnsureLoaded(Type behaviourType)
+        {
+            if (adapterCache.ContainsKey(behaviourType))
+            {
+                return;
+            }
+
+            var adapters = GetMembersWithMatchingAttributes(behaviourType);
```
firtina-improbable: These are no longer adapters, so the variable name of `adapters` is not as suitable

firtina-improbable: See other `adapter`~ variable names in the PR

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,187 @@
...
+            var componentIdsToAdapters = new Dictionary<uint, FieldInfo>();
+            var readerIds = new List<uint>();
```
firtina-improbable: Are these `ids of the readers`, or `component ids to be used for readers`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,187 @@
...
+            var writerIds = new List<uint>();
+            adapterCache[behaviourType] = componentIdsToAdapters;
+            componentReaderIdsForBehaviours[behaviourType] = readerIds;
```
firtina-improbable: I recommend assigning these after the foreach block, because it requires going back and forth to see what is actually being edited when you push to `readerIds`

firtina-improbable: Or `writerIds`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: No (as in: nothing should be codegenned like that)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: The latter. Will rename to readerComponentIds

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,191 @@
...
+        private readonly Dictionary<Type, List<uint>> componentWriterIdsForBehaviours =
+            new Dictionary<Type, List<uint>>();
+
+        private readonly HashSet<Type> invalidMonoBehaviourTypes = new HashSet<Type>();
+
+        private readonly ILogDispatcher logger;
+        private const string LoggerName = "SpatialOSBehaviourLibrary";
+
+        private const string BadRequiredMemberWarning
+            = "[Require] attribute found on member that is not Reader or Writer or incorrectly generated, ignoring this member!";
```
JonasImprobable: It might make sense to change the wording to something more formal, e.g. change `... generated, ignoring this member!` to `... generated. This member will be ignored.`. Also, leave out the `incorrectly generated` part. Just say that it's neither a reader nor writer.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        private const string MultipleReadersWritersRequiredError
```
JonasImprobable: Change to `... component. This MonoBehaviour will not be enabled.`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,191 @@
...
+            = "MonoBehaviour found requesting more than one Reader or Writer for the same component, " +
+            "this is invalid and will not be enabled!";
+
+        private const string MalformedReaderOrWriter
```
JonasImprobable: Remove the "!"

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -0,0 +1,191 @@
...
+            = "Reader or Writer found without a Component ID attribute, this is invalid!";
+
+        public SpatialOSBehaviourLibrary(ILogDispatcher logger)
+        {
+            this.logger = logger;
+        }
+
+        public void InjectAllReadersWriters(MonoBehaviour spatialOSBehaviour)
+        {
+            EnsureLoaded(spatialOSBehaviour.GetType());
```
JonasImprobable: `spatialOSBehaviour.GetType()` can be cached since this is used over and over, i.e. `var spatialOSBehaviourType = spatialOSBehaviour.GetType();`.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
JonasImprobable: Also, how about changing the name "Bad..." to something else, like `InvalidRequiredMemberWarning`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
jessicafalk: was there a conclusion to that?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/RequireAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    ///     Marks fields/properties of MonoBehaviours that require a Reader or Writer to be injected into them.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
```
jared-improbable: `AttributeTargets.Property` is no longer allowed, right?

PR #121, opened 2018-07-19
UTY-587: GameObjectComponentDispatchers

JonasImprobable:

-------

#### Description
- Codegen GameObjectTranslation classes that contain logic for invoking dispatcher callbacks in a typed way.
- Introduce a GameObjectDispatcherCallbacksSystem for invoking GameObjectTranslation methods.

Here is a sample codegened translation file:
```csharp
public partial class Position
    {
        public class GameObjectComponentDispatcher : GameObjectComponentDispatcherBase
        {
            public override ComponentType[] ComponentAddedComponentTypes => new ComponentType[]
            {
                typeof(ComponentAdded<SpatialOSPosition>), typeof(GameObjectReference)
            };

            public override ComponentType[] ComponentRemovedComponentTypes => new ComponentType[]
            {
                typeof(ComponentRemoved<SpatialOSPosition>), typeof(GameObjectReference)
            };

            public override ComponentType[] AuthoritiesChangedComponentTypes => new ComponentType[]
            {
                typeof(AuthoritiesChanged<SpatialOSPosition>), typeof(GameObjectReference)
            };

            public override ComponentType[] ComponentsUpdatedComponentTypes => new ComponentType[]
            {
                typeof(ComponentsUpdated<SpatialOSPosition.Update>), typeof(GameObjectReference)
            };

            public override ComponentType[][] EventsReceivedComponentTypeArrays => new ComponentType[][]
            {
            };

            public override ComponentType[][] CommandRequestsComponentTypeArrays => new ComponentType[][]
            {
            };

            public override void InvokeOnAddComponentLifecycleCallbacks(GameObjectDispatcherSystem gameObjectDispatcherSystem)
            {
                var entities = ComponentAddedComponentGroup.GetEntityArray();
                {
                    var spatialOSBehaviourManager = gameObjectDispatcherSystem.GetSpatialOSBehaviourManager(entities[i].Index);
                    spatialOSBehaviourManager.OnAddComponent(54);
                }
            }

            public override void InvokeOnRemoveComponentLifecycleCallbacks(GameObjectDispatcherSystem gameObjectDispatcherSystem)
            {
                var entities = ComponentRemovedComponentGroup.GetEntityArray();
                {
                    var spatialOSBehaviourManager = gameObjectDispatcherSystem.GetSpatialOSBehaviourManager(entities[i].Index);
                    spatialOSBehaviourManager.OnRemoveComponent(54);
                }
            }

            public override void InvokeOnAuthorityChangeLifecycleCallbacks(GameObjectDispatcherSystem gameObjectDispatcherSystem)
            {
                var authoritiesChangedTags = AuthoritiesChangedComponentGroup.GetComponentArray<AuthoritiesChanged<SpatialOSPosition>>();
                var entities = AuthoritiesChangedComponentGroup.GetEntityArray();
                {
                    var spatialOSBehaviourManager = gameObjectDispatcherSystem.GetSpatialOSBehaviourManager(entities[i].Index);
                    for (var j = 0; j < authoritiesChangedTags[i].Buffer.Count; j++)
                    {
                        spatialOSBehaviourManager.OnAuthorityChange(54, authoritiesChangedTags[i].Buffer[j]);
                    }
                }
            }

            public override void InvokeOnAuthorityChangeUserCallbacks(GameObjectDispatcherSystem gameObjectDispatcherSystem)
            {
            }

            public override void InvokeOnComponentUpdateUserCallbacks(GameObjectDispatcherSystem gameObjectDispatcherSystem)
            {
            }

            public override void InvokeOnEventUserCallbacks(GameObjectDispatcherSystem gameObjectDispatcherSystem)
            {
            }

            public override void InvokeOnCommandRequestUserCallbacks(GameObjectDispatcherSystem gameObjectDispatcherSystem)
            {
            }
        }
    }
```
#### Tests
Unit tests and manual tests.
#### Documentation
None
#### Primary reviewers
@samiwh @firtina-improbable


workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -332,6 +335,18 @@ private void FindTranslationUnits()
             }
         }
 
+        private void FindGameObjectTranslations()
+        {
+            var monoBehaviourTranslationTypes = AppDomain.CurrentDomain.GetAssemblies().SelectMany(assembly => assembly.GetTypes())
```
firtina-improbable: Do you mean `gameobjectTranslationTypes`?

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -19,6 +19,8 @@ public class MutableView : IDisposable
         public readonly Dictionary<int, ComponentTranslation> TranslationUnits =
             new Dictionary<int, ComponentTranslation>();
 
+        public readonly HashSet<GameObjectTranslationBase> MonoBehaviourTranslations = new HashSet<GameObjectTranslationBase>();
```
firtina-improbable: Do you mean `GameObjectTranslations`?

firtina-improbable: And ctrl-f `monobehaviourtranslation` in other parts of PR

code_generator/GdkCodeGenerator/Templates/UnityGameObjectTranslationGenerator.tt
```diff
@@ -0,0 +1,137 @@
...
+            };
+
+            public override ComponentType[] ComponentsUpdatedComponentTypes => new ComponentType[]
+            {
+<# if (fieldDetailsList.Count > 0) { #>
+                typeof(ComponentsUpdated<<#= componentDetails.TypeName #>.Update>), typeof(GameObjectReference)
+<# } #>
+            };
+
+            public override List<ComponentType[]> EventsReceivedComponentTypeArrays => new List<ComponentType[]>
```
firtina-improbable: Why not a `ComponentType[][]`?

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
JonasImprobable: Yes, I forgot to rename this across the board. Thanks for pointing out. Fixed.

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
JonasImprobable: Fixed.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectTranslationGenerator.tt
JonasImprobable: Nice catch. Fixed.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
```diff
@@ -0,0 +1,69 @@
...
+                monoBehaviourTranslation.CommandRequestsComponentGroups = new ComponentGroup[monoBehaviourTranslation.CommandRequestsComponentTypeArrays.Length];
+                for (var i = 0; i < monoBehaviourTranslation.CommandRequestsComponentTypeArrays.Length; i++)
+                {
+                    monoBehaviourTranslation.CommandRequestsComponentGroups[i] =
+                        GetComponentGroup(monoBehaviourTranslation.CommandRequestsComponentTypeArrays[i]);
+                }
+            }
+        }
+
+        protected override void OnUpdate()
```
firtina-improbable: I am unsure if this update will tick if there is nothing in the world to inject...

One idea I had is to either not make it a system, or to create a meta-world for each worker to tick manually and have the system utilise injections rather than component groups within monobehaviourtranslations... what I'm asking may be a slight redesign so I understand if it can be resolved in other ways too; so feel free to grab me for a chat :)

JonasImprobable:
@jamiebrynes7 The current implementation still has "Translations" in the name which probably doesn't make sense anymore with our ComponentTranslations rework. Happy to rename this to something more sensible.

jamiebrynes7:
@JonasImprobable `GameObjectDispatcher` ?

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectTranslationBase.cs
```diff
@@ -0,0 +1,28 @@
+using Unity.Entities;
+
+namespace Improbable.Gdk.Core
+{
+    public abstract class GameObjectTranslationBase
```
samiwh: Best to call this something other than translation. It doesn't look like this is translating anything, just defining groups and doing some piping.

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
```diff
@@ -45,6 +48,7 @@ public MutableView(World world, ILogDispatcher logDispatcher)
                 typeof(Action<Entity, ComponentType, object>), entityManager, setComponentObjectMethodInfo);
 
             FindTranslationUnits();
+            FindGameObjectTranslations();
```
samiwh: Is there any reason this call or the function needs to be in the mutable view? It looks like this would be better suited to being in the `GameObjectDispatcherCallbacksSystem`.

workers/unity/Assets/Gdk/Core/Tests/Editmode/GameObjectRepresentation/GameObjectDispatcherCallbacksSystemTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        [Test]
+        public void OnCreateManager_initializes_GameObjectTranslation_ComponentGroups()
+        {
+            var world = worker.World;
+
+            foreach (var gameObjectTranslation in worker.View.GameObjectTranslations)
+            {
+                Assert.IsNull(gameObjectTranslation.ComponentAddedComponentGroup);
```
samiwh: This looks like it's just testing that a default constructed reference is null.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
```diff
@@ -0,0 +1,70 @@
+using Unity.Entities;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    ///     Gathers incoming dispatcher ops and invokes callbacks on relevant GameObjects.
+    /// </summary>
+    [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
```
samiwh: This should probably be done earlier. I'd suggest after the initialisation group and before the update group.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
```diff
@@ -0,0 +1,70 @@
...
+    public class GameObjectDispatcherCallbacksSystem : ComponentSystem
+    {
+        private MutableView view;
```
samiwh: If we move the function that finds the generated classes we can remove this reference.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectReference.cs
```diff
@@ -1,9 +1,37 @@
...
+
+        public void OnRemoveComponent(uint componentId)
+        {
+        }
+
+        public void OnAuthorityChange(uint componentId, Authority authority)
+        {
+        }
+
+        public void OnComponentUpdate(uint componentId, object update)
```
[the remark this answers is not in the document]
samiwh: Whatever invokes the callbacks is allowed to know about readers, so that can use the actual types.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            var worker = WorkerRegistry.GetWorkerForWorld(World);
+            view = worker.View;
+
+            GenerateComponentGroups();
+        }
+
+        private void GenerateComponentGroups()
+        {
+            foreach (var gameObjectTranslation in view.GameObjectTranslations)
```
jamiebrynes7: If this is the only place these are used - find them in here rather the Mutable View

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
```diff
@@ -0,0 +1,70 @@
...
+                gameObjectTranslation.AuthoritiesChangedComponentGroup =
+                    GetComponentGroup(gameObjectTranslation.AuthoritiesChangedComponentTypes);
+                if (gameObjectTranslation.ComponentsUpdatedComponentTypes.Length > 0)
+                {
+                    gameObjectTranslation.ComponentsUpdatedComponentGroup =
+                        GetComponentGroup(gameObjectTranslation.ComponentsUpdatedComponentTypes);
+                }
+
+                gameObjectTranslation.EventsReceivedComponentGroups =
+                    new ComponentGroup[gameObjectTranslation.EventsReceivedComponentTypeArrays.Length];
```
jamiebrynes7: Wrap this with an `if` as well?

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
```diff
@@ -0,0 +1,70 @@
...
+                for (var i = 0; i < gameObjectTranslation.EventsReceivedComponentTypeArrays.Length; i++)
+                {
+                    gameObjectTranslation.EventsReceivedComponentGroups[i] =
+                        GetComponentGroup(gameObjectTranslation.EventsReceivedComponentTypeArrays[i]);
+                }
+
+                gameObjectTranslation.CommandRequestsComponentGroups =
+                    new ComponentGroup[gameObjectTranslation.CommandRequestsComponentTypeArrays.Length];
+                for (var i = 0; i < gameObjectTranslation.CommandRequestsComponentTypeArrays.Length; i++)
```
jamiebrynes7: Similarly here - wrap with an if

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
```diff
@@ -0,0 +1,70 @@
...
+                        GetComponentGroup(gameObjectTranslation.CommandRequestsComponentTypeArrays[i]);
+                }
+            }
+        }
+
+        protected override void OnUpdate()
+        {
+            foreach (var gameObjectTranslation in view.GameObjectTranslations)
+            {
+                gameObjectTranslation.InvokeOnAddComponentCallbacks();
```
jamiebrynes7: Are we concerned about the fact that callbacks are technically called out of order here? (Compared to how the ops are received)

JonasImprobable: Yeah, this is a limitation of the GDK Core since by the time the reactive components are added to the ecs entities, you have no information about the order at which the ops came in anymore. This might lead to breaks in legacy projects.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
JonasImprobable: Added a new update group between `EntityInitialisationGroup` and `SpatialOSUpdateGroup`.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
JonasImprobable: Moved the GameObjectTranslationsDictionary into the system.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
JonasImprobable: Moved the GameObjectTranslationsDictionary into the system.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
JonasImprobable: Fixed!

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
JonasImprobable: Fixed!

workers/unity/Assets/Gdk/Core/MutableView/MutableView.cs
JonasImprobable: Moved this into the `GameObjectDispatcherCallbacksSystem`.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
jamiebrynes7: Gotcha, we will need to make sure this is communicated properly :)

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectReference.cs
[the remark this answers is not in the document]
MPeti01: @samiwh The idea is for GameObjectReference to be not codegenned, though. We could work around that by exposing a method to get a list of ReaderWriters for a given component ID maybe? Then the translation classes would query GameObjectReference for that list (which would exist internally anyway), cast it to the specific type it knows them to be and call the specific type callbacks itself.

MPeti01:
Alternative to Translation name: something like Relayer?

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectReference.cs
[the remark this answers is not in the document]
samiwh: Might be easier to have the readers and writers exposed to something that the codegen code can use. Have codegened helper methods that match against all entities that have the relevant codegened components and a handle to the readers and writers of a given type. Then directly call the callbacks on the reader/writer.

Using object isn't something we can keep.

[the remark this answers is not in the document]
JonasImprobable: Ok, I have two ideas here:

1. Let the `GameObjectReference` callbacks not contain payloads, i.e. change `public void OnComponentUpdate(uint componentId, object update)` to `public void OnComponentUpdate(uint componentId)`. Readers/Writers do not get the update, they just get notified that an update happened. Readers/Writers fetch the update payload themselves using `EntityManager.GetComponentObject<specific codegened type>(entity)`.

2. Have codgenerated extension methods for the GameObjectReference class for each component. Things will get very ugly when you have multiple components with the same name but in different namespaces.

3. Generate a type specific GameObjectReference class for every component and add multiple of those to entities. The sad part here is that we still need to keep a general component unspecific GameObjectReference to handle the injection/lifecycle.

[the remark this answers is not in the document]
MPeti01: I like the first one. I can't see any serious downside other than the fact that the code looks a bit surprising to read? Which could be mitigated by comments above the GameObjectReference methods.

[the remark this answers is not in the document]
JonasImprobable: Option 2 looks like a no-go. Option 1 is my favourite right now because it's the simplest and fastest to implement. Also, we'd need to implement Option 1 as a prerequisite for implementing Option 3.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectDispatcherCallbacksSystem.cs
JonasImprobable: No, all callbacks are invoked after the data model was updated. By the time, the callbacks for an update for A is fired all component values are up-to-date.

workers/unity/Assets/Gdk/Core/Tests/Editmode/GameObjectRepresentation/GameObjectDispatcherCallbacksSystemTests.cs
JonasImprobable: removed those checks.

workers/unity/Assets/Gdk/Core/GameObjectRepresentation/GameObjectTranslationBase.cs
JonasImprobable: Renamed to `GameObjectComponentDispatcher`.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectReference.cs
```diff
@@ -2,8 +2,35 @@
...
 namespace Improbable.Gdk.Core
 {
+    /// <summary>
+    ///     Invokes dispatcher callbacks on GameObjects that represent entities.
+    /// </summary>
     public class GameObjectReference : Component
     {
         public GameObject GameObject;
+
+        public void OnAddComponent(uint componentId)
```
samiwh: Will this be a system calling this which then has to call back out to a system to find the relevant data? If so can we cut out the middle man and just call whatever this calls?

JonasImprobable: Re: First question: Yes, that's the current plan. See Option 1 in my comment above.
Re: Second question: That would be Option 3 (at least I don't see any other way to do this without doing Option 3) which I find less preferable. See comment above.

samiwh: Not this would ultimately get rid of the game object reference. The thing I'm questioning is having systems call into Components which call back into systems.

Instead you could just have the original codegen function call all of that.

sorry I didn't see the options as github hides them when the code changes.
Incidentally you can do option 2 by fully qualifying all names. If this is being done in codegen that isn't actually a problem. Although I don't think I'd recommend it as a solution.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectReference.cs
```diff
@@ -2,8 +2,35 @@
...
+        {
+        }
+
+        public void OnRemoveComponent(uint componentId)
+        {
+        }
+
+        public void OnAuthorityChange(uint componentId)
```
MPeti01: Can we pass the Authority itself as well here? That shouldn't have boxing issues, and the GameObjectReference itself will need it for injection logic. Alternatively the GameObjectReference could go through the Entity like the ReaderWriter itself will have to, but just passing would be more convenient and readable I think.

JonasImprobable: Yes, we can do that. That would make sense.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectReference.cs
MPeti01: I'm not sure what your proposed solution is, surely you'd need some component to keep track of what readers/writers could be available and handle injection and activation? That involves storing data about what spatialOS components we already have, what behaviours need what components etc. Unless we read through what we need and what we have and check a massive amount of info whenever components are added/removed...

I suppose after activation the readerwriters themselves could be stored pretty much anywhere, but even there I don't see any simple options that would be a net improvement on the system overall, especially when the only improvement seems to be just one less level of indirection here?

JonasImprobable: Maybe, @samiwh's idea here is have something like this:
```
// codegenerated classes, i.e. GameObjectComponentDispatcher
public override void InvokeOnComponentUpdateCallbacks()
{
    var gameObjectReferences = ComponentsUpdatedComponentGroup.GetComponentArray<GameObjectReference>();
    for (var i = 0; i < gameObjectReferences.Length; i++)
    {
        var readersWriters = gameObjectReferences[i].GetReadersWriters(54 /* the componentId */);
        readersWriters.foreach(readerWriter => readerWriter.OnComponentUpdate(update /* concrete update type */);)
    }
}
```

JonasImprobable: Discussed offline: We decided to go ahead with the store managed data in a system approach.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
```diff
@@ -0,0 +1,89 @@
...
+                    continue;
+                }
+
+                var gameObject = addedEntitiesData.GameObjectReferences[i].GameObject;
+                gameObjectDispatcherSystem.EntityIndexToSpatialOSBehaviourManager[entityIndex] = new SpatialOSBehaviourManager(gameObject);
+            }
+
+            for (var i = 0; i < removedEntitiesData.Length; i++)
+            {
+                var entityIndex = addedEntitiesData.Entities[i].Index;
```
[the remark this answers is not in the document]
JonasImprobable: Ha! Thanks for spotting. 👍

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -0,0 +1,106 @@
...
+{
+    /// <summary>
+    ///     Gathers incoming dispatcher ops and invokes callbacks on relevant GameObjects.
+    /// </summary>
+    [UpdateInGroup(typeof(SpatialOSReceiveGroup.GameObjectReceiveGroup))]
+    public class GameObjectDispatcherSystem : ComponentSystem
+    {
+        public readonly Dictionary<int, SpatialOSBehaviourManager> EntityIndexToSpatialOSBehaviourManager = new Dictionary<int, SpatialOSBehaviourManager>();
+
+        public readonly HashSet<GameObjectComponentDispatcherBase> GameObjectComponentDispatchers =
```
jamiebrynes7: Why is this a `HashSet`? If we are iterating over it every loop, we want that to be as fast as possible, should be a list I think

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -0,0 +1,106 @@
...
+            GenerateComponentGroups();
+        }
+
+        private void FindGameObjectComponentDispatchers()
+        {
+            var gameObjectComponentDispatcherTypes = AppDomain.CurrentDomain.GetAssemblies()
+                .SelectMany(assembly => assembly.GetTypes())
+                .Where(type => typeof(GameObjectComponentDispatcherBase).IsAssignableFrom(type) && !type.IsAbstract).ToList();
+
+            foreach (var gameObjectComponentDispatcherType in gameObjectComponentDispatcherTypes)
```
jamiebrynes7: If you switch `GameObjectComponentDispatchers` to a List you can use LINQ:

```GameObjectComponentDispatchers = gameObjectComponentDispatcherTypes.Select(type => (GameObjectComponentDispatcherBase)Activator.CreateInstance(type)).ToList();```

File: workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs [hunk 20]
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    ///     Gathers incoming dispatcher ops and invokes callbacks on relevant GameObjects.
+    /// </summary>
+    [UpdateInGroup(typeof(SpatialOSReceiveGroup.GameObjectReceiveGroup))]
+    public class GameObjectDispatcherSystem : ComponentSystem
+    {
+        public readonly Dictionary<int, SpatialOSBehaviourManager> EntityIndexToSpatialOSBehaviourManager = new Dictionary<int, SpatialOSBehaviourManager>();
```
jamiebrynes7: Given that we iterate over the values of this dictionary every frame - is there a more efficient manner of storing this data?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        protected override void OnUpdate()
+        {
+            foreach (var gameObjectComponentDispatcher in GameObjectComponentDispatchers)
+            {
+                gameObjectComponentDispatcher.InvokeOnAddComponentLifecycleCallbacks(this);
+                gameObjectComponentDispatcher.InvokeOnRemoveComponentLifecycleCallbacks(this);
+                gameObjectComponentDispatcher.InvokeOnAuthorityChangeLifecycleCallbacks(this);
+
+                foreach (var spatialOSBehaviourManager in EntityIndexToSpatialOSBehaviourManager.Values)
```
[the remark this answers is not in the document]
jamiebrynes7: Yeah - it looks like we will invoke `EnableSpatialBehaviours` many more times than is necessary, unless I am misunderstanding the intent of this code

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectReferenceHandle.cs
```diff
@@ -1,6 +1,6 @@
 using Unity.Entities;
 
-namespace Playground
+namespace Improbable.Gdk.Core
```
jamiebrynes7: Please put these in a more specific namespace - applies to most files in this PR

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    public class SpatialOSBehaviourManager
+    {
+        public GameObject GameObject;
+        private Dictionary<uint, List<MonoBehaviour>> componentIdToReaderSpatialOSBehaviours;
+        private Dictionary<uint, List<MonoBehaviour>> componentIdToWriterSpatialOSBehaviours;
+        private Dictionary<MonoBehaviour, int> numUnsatisfiedComponents;
+        private List<MonoBehaviour> spatialOsBehavioursToEnable;
+        private List<MonoBehaviour> spatialOsBehavioursToDisable;
+        private Dictionary<uint, Dictionary<MonoBehaviour, ReaderWriterImpl>> readersWriters;
```
jamiebrynes7: Do we want to initialise these here or are we planning on doing that in the constructor?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
```diff
@@ -0,0 +1,89 @@
...
+        [Inject] private RemovedEntitiesData removedEntitiesData;
+
+        private GameObjectDispatcherSystem gameObjectDispatcherSystem;
+        private ILogDispatcher logDispatcher;
+
+        protected override void OnCreateManager(int capacity)
+        {
+            base.OnCreateManager(capacity);
+
+            gameObjectDispatcherSystem = World.GetExistingManager<GameObjectDispatcherSystem>();
```
jamiebrynes7: This has a dependency on the order in which these systems are defined in the world setup - is that intentional?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
```diff
@@ -0,0 +1,89 @@
...
+                if (gameObjectDispatcherSystem.EntityIndexToSpatialOSBehaviourManager.ContainsKey(entityIndex))
+                {
+                    logDispatcher.HandleLog(LogType.Error, new LogEvent(
+                            "Entity already has a SpatialOSBehaviourManager.")
+                        .WithField("EntityIndex", entityIndex));
+                    continue;
+                }
+
+                var gameObject = addedEntitiesData.GameObjectReferences[i].GameObject;
+                gameObjectDispatcherSystem.EntityIndexToSpatialOSBehaviourManager[entityIndex] = new SpatialOSBehaviourManager(gameObject);
```
jamiebrynes7: Can combine the two lines here

workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
```diff
@@ -84,11 +84,14 @@ protected override void OnUpdate()
...
                 var gameObjectReference = new GameObjectReference { GameObject = gameObject };
 
+                var requiresSpatialOSBehaviourManagerComponent = new RequiresSpatialOSBehaviourManager();
+
                 entityGameObjectCache[entity.Index] = gameObject;
                 var gameObjectReferenceHandleComponent = new GameObjectReferenceHandle();
 
-                PostUpdateCommands.AddComponent(addedEntitiesData.Entities[i], gameObjectReferenceHandleComponent);
                 viewCommandBuffer.AddComponent(entity, gameObjectReference);
+                PostUpdateCommands.AddComponent(addedEntitiesData.Entities[i], requiresSpatialOSBehaviourManagerComponent);
```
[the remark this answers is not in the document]
JonasImprobable: Yeah, that would be less error prone indeed but rather than having `NeedSomeSetupAndDeleteTheTagAfterwards` component tags, we have `AddMeAndKeppMeForeverToShowThatSetupIsNotRequiredAnymore` tags which bloat up the entity over time. Let me think about this.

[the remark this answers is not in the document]
firtina-improbable: Yes I'm unsure of the costs of long-persisting components on entities

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
JonasImprobable: Yes, you are right. The code in its current state doesn't make sense. What you posted is how the code should have looked like.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
JonasImprobable: Good spot. Changed to a list.

tools/DocsLinter/Tests/LinkCheckingTests.cs
```diff
@@ -21,8 +21,9 @@ private static RemoteLink GetUrlForStatusCode(int statusCode)
         }
 
         [OneTimeSetUp]
-        public void IgnoreSslErrors() {
-            System.Net.ServicePointManager.ServerCertificateValidationCallback += 
+        public void IgnoreSslErrors()
```
jared-improbable: Could these drive-by formats (and the s->z rename) be split into a different PR?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
JonasImprobable: This is super elegant. I love it. I'm using `GameObjectComponentDispatchers.AddRange()` now because `GameObjectComponentDispatchers` can stay readonly that way.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
MPeti01: Isn't it more readable as it is though? And I'd expect a decent compiler to avoid performance overhead here as well.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
JonasImprobable: The caches will be populated in the constructor. Overall, this file is more a placeholder file (hence the lack of any actual implementation) and we will make a proper pass over it later.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
JonasImprobable: Very interesting. You are right. The reason I'm using `GetExistingManager` here is because I got trolled before when I ran a worker instance where I intentionally didn't want the `GameObjectDispatcherSystem` running. Reverting to `GetOrCreateManager()`.

tools/DocsLinter/Tests/LinkCheckingTests.cs
JonasImprobable: Yeah, that was the linter's doing. I will revert the drive-by changes.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
JonasImprobable: Done!

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
JonasImprobable: I will now also use a separate list to quickly iterate over the SpatialOSBehaviourManagers.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <summary>
+    ///     Invokes dispatcher callbacks on GameObjects that represent entities.
+    /// </summary>
+    public class SpatialOSBehaviourManager
+    {
+        public GameObject GameObject;
+        private Dictionary<uint, List<MonoBehaviour>> componentIdToReaderSpatialOSBehaviours;
+        private Dictionary<uint, List<MonoBehaviour>> componentIdToWriterSpatialOSBehaviours;
+        private Dictionary<MonoBehaviour, int> numUnsatisfiedComponents;
+        private List<MonoBehaviour> spatialOsBehavioursToEnable;
```
samiwh: s/Os/OS/g

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                var entityIndex = addedEntitiesData.Entities[i].Index;
+                var spatialOSBehaviourManager = new SpatialOSBehaviourManager(addedEntitiesData.GameObjectReferences[i].GameObject);
+                gameObjectDispatcherSystem.AddSpatialOSBehaviourManager(entityIndex, spatialOSBehaviourManager);
+            }
+
+            for (var i = 0; i < removedEntitiesData.Length; i++)
+            {
+                var entityIndex = removedEntitiesData.Entities[i].Index;
+                gameObjectDispatcherSystem.RemoveSpatialOSBehaviourManager(entityIndex);
```
samiwh: Need to remove the system state data too.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+            foreach (var gameObjectComponentDispatcher in GameObjectComponentDispatchers)
+            {
+                gameObjectComponentDispatcher.InvokeOnAddComponentLifecycleCallbacks(this);
+                gameObjectComponentDispatcher.InvokeOnRemoveComponentLifecycleCallbacks(this);
+                gameObjectComponentDispatcher.InvokeOnAuthorityChangeLifecycleCallbacks(this);
+            }
+
+            foreach (var spatialOSBehaviourManager in spatialOSBehaviourManagers)
+            {
+                    spatialOSBehaviourManager.EnableSpatialOSBehaviours();
```
samiwh: If the component is removed I think we don't actually want to receive callbacks. That will create a null ref if they try to read the component during the update.

samiwh: Similarly this will cause callbacks on writers to give callbacks for your previous frame's updates, which might trigger sending another update, thinking the authority is still there.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+                }
+            }
+        }
+
+        protected override void OnUpdate()
+        {
+            foreach (var gameObjectComponentDispatcher in GameObjectComponentDispatchers)
+            {
+                gameObjectComponentDispatcher.InvokeOnAddComponentLifecycleCallbacks(this);
+                gameObjectComponentDispatcher.InvokeOnRemoveComponentLifecycleCallbacks(this);
```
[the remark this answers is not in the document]
MPeti01: As written it's a problem if it is on the entity, then gets removed and readded in the same tick right? Because then we initially call the BehaviourManager's AddComponent() method with a component that's already on it?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            foreach (var spatialOSBehaviourManager in spatialOSBehaviourManagers)
+            {
+                    spatialOSBehaviourManager.EnableSpatialOSBehaviours();
+            }
+
+            foreach (var gameObjectComponentDispatcher in GameObjectComponentDispatchers)
+            {
+                gameObjectComponentDispatcher.InvokeOnAuthorityChangeUserCallbacks(this);
+                gameObjectComponentDispatcher.InvokeOnComponentUpdateUserCallbacks(this);
```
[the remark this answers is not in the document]
samiwh: We might even need to keep and ordering list on each entity. A list of enums saying what arrived in what order.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
MPeti01: It's removed by the GameObjectInitializationSystem - I know because I asked Jonas the same question yesterday, so maybe there should be a comment to clarify that?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
samiwh: Or we could squash. Or we could assume auth loss comes before updates and auth gained comes after

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
samiwh: It's best to rework this to not assume anything about ordering. Use the component added and removed components to denote an entity that needs to be checked, then explicitly check if the component is there or not probably.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
JonasImprobable: Oh shoot. Missed that. Sorry!

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
JonasImprobable: The GDK core should ensure that a component can't be simultaneously added or removed.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
JonasImprobable: I'm not sure if I'm following. This is not the place in the code where user updates would be triggered. If a component is removed or authority is lost during a frame, the assumption is that the GDK core would not give you the respective reactive component tags.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
JonasImprobable: Under which circumstances would you want e.g. component update callbacks to fire before authority callbacks? Agree though that the overall approach here looks fishy e.g. for situations where user code depends on an event being processed before component field updates.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
JonasImprobable: The components are all removed by the system that created them.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+            if (entityIndexToSpatialOSBehaviourManager.ContainsKey(entityIndex))
+            {
+                throw new SpatialOSBehaviourManagerAlreadyExistsException($"SpatialOSBehaviourManager already exists for entityIndex {entityIndex}.");
+            }
+
+            entityIndexToSpatialOSBehaviourManager[entityIndex] = spatialOSBehaviourManager;
+            spatialOSBehaviourManagers.Add(spatialOSBehaviourManager);
+        }
+
+        public void RemoveSpatialOSBehaviourManager(int entityIndex)
```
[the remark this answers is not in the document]
JonasImprobable: Made function internal.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            // inject + enable all spatialOsBehavioursToEnable
+        }
+
+        public void DisableSpatialOSBehaviours()
+        {
+            // disable + deinject all spatialOsBehavioursToDisable
+        }
+
+        public void OnAddComponent(uint componentId)
```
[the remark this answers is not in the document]
JonasImprobable: Renamed to `AddComponent()`, `RemoveComponent()`, `ChangeAuthority()`.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
JonasImprobable: I made the methods public again so we can call them in unit tests.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
JonasImprobable: Ok, after thinking about this a little more, here is what I'm currently think:
- If we want to preserve the order of the incoming ops, that would require a refactor of the GDK Core as the GDK Core currently does not surface that information. A such, those changes shall not be in scope of this PR.
- Assuming that the information about the order of incoming ops can not be recovered given the current limitations of GDK Core, what we have here is as good as any other possible alternative.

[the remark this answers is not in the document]
jamiebrynes7: We definitely can surface that info - but I agree it is out of scope for this PR

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
jared-improbable: Why not use the `InternalsVisibleTo` assembly attribute to expose these to tests?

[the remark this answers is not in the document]
JonasImprobable: Made methods internal and added an assemblyinfo file.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
samiwh: You want the updates before the authority callbacks if you gained authority. Doing it the other way round is wrong.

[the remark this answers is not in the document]
samiwh: This stuff doesn't need any fancy things in the core. I mentioned as an extension you could want to preserve orderings in more complete ways but even without that you need to order the updates and auth callbacks correctly. As it currently stands this code is wrong.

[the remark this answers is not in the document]
JonasImprobable: Re: "You want the updates before the authority callbacks if you gained authority": Why? Wouldn't you want to maybe do some sort of setup upon gaining authority that subsequent component updates depend on?
Re: "I mentioned as an extension you could want to preserve orderings in more complete ways but even without that you need to order the updates and auth callbacks correctly": How would I do that?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
samiwh: The comment was about the ordering in which you enable update and disable monobehaviour callbacks. If a component has been removed you probably want to disable the stuff before giving callbacks on it. If you have two readers for different types on a single monobehaviour, then one could try to read the data from the other during a callback. If one has been removed and the other has a callback the current ordering will cause that to null ref. Which breaks the design that an enabled monobehaviour is good to go.

To fix this you want to disable things before giving callbacks.

However you also want to make sure you aren't doing things like trigger update callbacks on writers. So you can't just enable everything before the start and naively proceed. You probably want to enable behaviours with writers without invoking their callbacks for updates on that component.

So the overall order likely needs to be
Disable stuff
Enable stuff without writers
Deal with auth lost and auth loss immanent
Deal with component updates
Deal with auth gained
Enable stuff with writers

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManagerInitializationSystem.cs
samiwh: This needs to be explicitly run before that system then.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
samiwh: For the first question: No you don't want to be receiving updates from other workers during a block where you expect to have authority. The way a user expects to interpret update callbacks on writers is as something that worker sent itself. You probably want to get the current state of the component, and you might want to access the updates that have been received before the auth change (for predictive code etc) but you don't want to be told you received an update after the auth change. That would cause users to have to write all their writer callback code having to check if they just gained authority or had it for a while :S.
The ECS layer will give you the ability to access the updates that were received before hand if you want them. You could also do it by having a companion behaviour, which only has a reader, so it gets activated earlier and will receive the component update before the auth callback. But that is a special case where a user wants to do something clever, so it can be a slightly more awkward flow. The standard case is that once you have authority you don't care what happened before.

For the second point: That was the extension that was meant for another time. E.g. an list of enums per component. So I agree with your assessment that this is out of scope for now. More something to think about for the future.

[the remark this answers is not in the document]
JonasImprobable: Re: first point: Not sure if I understand your point but if we don't want to receive updates on authoritative workers then that's solved in the GDK Core, as in the GDK Core should or shouldn't emit the reactive component tags. Assuming that reactive component tags are not emitted (afaik they currently are which I'm fine with), the execution order won't matter as there are no tags to look for.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
JonasImprobable: This is a very good point and the more I look at the current code the more flawed the execution order looks. Agreed that we need to disable everything first. I still don't understand why we need to distinguish our behaviour for readers and writers though. 
- An authority gained callback can be triggered for both readers and writers after the enabling phase.
- Authority loss iminent notifications can be received by both readers and writers in the same way.
- Re: Component updates: As mentioned in my other post, this matter should be solved on the GDK Core level.
- Authority loss callbacks are "automatically" not fired on writers assuming that they get disabled earlier in the frame.

JonasImprobable: Created https://improbableio.atlassian.net/browse/UTY-830

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
samiwh: No, you still want to be able to find what the updates were for some cases. You just don't want them as callbacks.So in this case we can't assume the gdk core is removing them.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
samiwh: It would be confusing to receive auth added callbacks on writers. As a user I'd expect that a writer is only enabled when I have authority. If I saw I could get an auth callback after OnEnable I'd be confused about when I'm allowed to send updates. On a writer enabled behaviour you might be getting auth callbacks for auth loss imminent. If you got a callback with something else in it then users need to write code to call that too and would likely be considered a bug.

You're right auth loss imminent won't break things, so my execution order is probably wrong. As an aside we might need to be more careful about it as we add things (e.g. readonlyers or abouttoloseauthwriters). But we don't have those yet.

Unfortunately you can't assume that about the core. Updates do have to be handled more carefully :(.

Yes, the disabling of stuff earlier will solve that.

PR #123, opened 2018-07-19
UTY-622 Allow `spatial local worker launch` and reconnections

firtina-improbable:

-------

#### Description
Comes with some driveby fixes and features:
- Can launch UnityGameLogic
- Fixed ArgumentException for bad worker type in config params
- Added unique connection ID param to be passed into UnityGameLogic worker during startup (because we cannot have the cloud deployment be affected)
- Moved launch log files to logs folder

#### Tests
Tested locally using:
```
spatial local worker launch UnityGameLogic default
spatial local worker launch UnityClient default
```

#### Documentation
Needs to be added when documentation for launch arguments are considered (will create a ticket if it doesn't exist)

#### Primary reviewers
@jessicafalk @zeroZshadow

### DIFF
--- a/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
+++ b/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
@@ -66,7 +66,7 @@ namespace Playground
                         string.Empty);
                 var useDynamicWorkerId = CommandLineUtility.GetCommandLineValue(commandLineArgs,
                     RuntimeConfigNames.UseDynamicWorkerId,
-                    "false") == "true";
+                    false);
 
                 // because the launcher does not pass in the worker type as an argument
                 var worker = workerType.Equals(string.Empty)
@@ -75,7 +75,7 @@ namespace Playground
 
                 if (useDynamicWorkerId)
                 {
-                    worker.AllowDynamicId = true;
+                    worker.UseDynamicId = true;
                 }
 
                 Workers.Add(worker);
--- a/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
+++ b/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
@@ -64,19 +64,13 @@ namespace Playground
                 var workerId =
                     CommandLineUtility.GetCommandLineValue(commandLineArgs, RuntimeConfigNames.WorkerId,
                         string.Empty);
-                var forceDynamicWorkerId = CommandLineUtility.GetCommandLineValue(commandLineArgs,
-                    RuntimeConfigNames.ForceDynamicWorkerId,
-                    false);
 
                 // because the launcher does not pass in the worker type as an argument
                 var worker = workerType.Equals(string.Empty)
-                    ? WorkerRegistry.CreateWorker<UnityClient>($"{workerType}-{Guid.NewGuid()}", new Vector3(0, 0, 0))
+                    ? WorkerRegistry.CreateWorker<UnityClient>(
+                        workerId: null, // The worker id for the UnityClient will be auto-generated.
+                        origin: new Vector3(0, 0, 0))
                     : WorkerRegistry.CreateWorker(workerType, workerId, new Vector3(0, 0, 0));
-
-                if (forceDynamicWorkerId)
-                {
-                    worker.UseDynamicId = true;
-                }
 
                 Workers.Add(worker);
 
--- a/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
+++ b/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
@@ -64,11 +64,19 @@ namespace Playground
                 var workerId =
                     CommandLineUtility.GetCommandLineValue(commandLineArgs, RuntimeConfigNames.WorkerId,
                         string.Empty);
+                var useUniqueConnectionId = CommandLineUtility.GetCommandLineValue(commandLineArgs,
+                    RuntimeConfigNames.UseUniqueConnectionId,
+                    "false") == "true";
 
                 // because the launcher does not pass in the worker type as an argument
                 var worker = workerType.Equals(string.Empty)
                     ? WorkerRegistry.CreateWorker<UnityClient>($"{workerType}-{Guid.NewGuid()}", new Vector3(0, 0, 0))
                     : WorkerRegistry.CreateWorker(workerType, workerId, new Vector3(0, 0, 0));
+
+                if (useUniqueConnectionId)
+                {
+                    worker.RequiresUniqueWorkerIdPerConnection = true;
+                }
 
                 Workers.Add(worker);
 
--- a/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
+++ b/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
@@ -64,8 +64,8 @@ namespace Playground
                 var workerId =
                     CommandLineUtility.GetCommandLineValue(commandLineArgs, RuntimeConfigNames.WorkerId,
                         string.Empty);
-                var useDynamicWorkerId = CommandLineUtility.GetCommandLineValue(commandLineArgs,
-                    RuntimeConfigNames.UseDynamicWorkerId,
+                var forceDynamicWorkerId = CommandLineUtility.GetCommandLineValue(commandLineArgs,
+                    RuntimeConfigNames.ForceDynamicWorkerId,
                     false);
 
                 // because the launcher does not pass in the worker type as an argument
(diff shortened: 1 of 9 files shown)


workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -64,12 +64,20 @@ public void Awake()
...
+                    "false") == "true";
 
                 // because the launcher does not pass in the worker type as an argument
                 var worker = workerType.Equals(string.Empty)
                     ? WorkerRegistry.CreateWorker<UnityClient>($"{workerType}-{Guid.NewGuid()}", new Vector3(0, 0, 0))
                     : WorkerRegistry.CreateWorker(workerType, workerId, new Vector3(0, 0, 0));
 
+                if (useUniqueConnectionId)
+                {
+                    worker.RequiresUniqueWorkerIdPerConnection = true;
```
martinzlocha: I don't think this is the best variable name, all connections require a unique id because if you try to use the same one then you can't reconnect so it might mislead people. Maybe something like: `AllowDynamicId`

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerBase.cs
```diff
@@ -49,11 +49,25 @@ public void Dispose()
...
         public virtual void Connect(ConnectionConfig config)
         {
             if (config is ReceptionistConfig)
             {
-                Connection = ConnectionUtility.ConnectToSpatial((ReceptionistConfig) config, GetWorkerType, WorkerId);
+                string workerIdForConnection;
+
+                if (RequiresUniqueWorkerIdPerConnection)
+                {
+                    workerIdForConnection = $"{WorkerId}-{Guid.NewGuid()}";
```
martinzlocha: If you do this then `connection.GetWorkerId()` and `WorkerId` will have different values because you never set the `WorkerId`, I think this is very unexpected behaviour.

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
firtina-improbable: Aye, I like the suggestion

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerBase.cs
firtina-improbable: `WorkerId` is currently only used privately within `WorkerBase`, so I'm thinking of making the whole property a private string instead. 

And we you want to get "the worker id" we just get it from the connection.

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
firtina-improbable: Also moving the declaration to top of class

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerBase.cs
martinzlocha: Why wouldn't we rather use: $"{GetWorkerType}-{Guid.NewGuid()}" and then set that to WorkerId? This way we can keep the WorkerId field and the change shouldn't be major.

firtina-improbable: Yes that is a better idea :D

workers/unity/Assets/Playground/Scripts/Workers/UnityClient.cs
```diff
@@ -16,6 +16,8 @@ public class UnityClient : WorkerBase
         public UnityClient(string workerId, Vector3 origin) : base(workerId, origin, new ForwardingDispatcher())
         {
             forwardingDispatcher = (ForwardingDispatcher) View.LogDispatcher;
+
```
martinzlocha: Nit: Extra line not needed.

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -64,12 +64,20 @@ public void Awake()
                 var workerId =
                     CommandLineUtility.GetCommandLineValue(commandLineArgs, RuntimeConfigNames.WorkerId,
                         string.Empty);
+                var useDynamicWorkerId = CommandLineUtility.GetCommandLineValue(commandLineArgs,
```
jared-improbable: It's kind of odd to be dealing with `string` rather than `bool`.

workers/unity/Assets/Playground/Scripts/Workers/UnityGameLogic.cs
```diff
@@ -17,6 +17,10 @@ public UnityGameLogic(string workerId, Vector3 origin) : base(workerId, origin,
         {
             PlayerLifecycleConfig.CreatePlayerEntityTemplate = PlayerTemplate.CreatePlayerEntityTemplate;
             forwardingDispatcher = (ForwardingDispatcher) View.LogDispatcher;
+
+#if UNITY_EDITOR
+            AllowDynamicId = true;
```
jared-improbable: How about `AllowDynamicId = Application.isEditor` instead?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerBase.cs
```diff
@@ -21,6 +21,8 @@ public abstract class WorkerBase : IDisposable
 
         public EntityGameObjectLinker EntityGameObjectLinker { get; private set; }
 
+        public bool AllowDynamicId = false;
```
jared-improbable: Rename from to match the flags' names?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerRegistry.cs
```diff
@@ -65,7 +65,7 @@ public static WorkerBase CreateWorker(string workerType, string workerId, Vector
             Func<string, Vector3, WorkerBase> createWorker;
             if (!WorkerTypeToInitializationFunction.TryGetValue(workerType, out createWorker))
             {
-                throw new ArgumentException("No worker found for worker type '{0}'", workerType);
+                throw new ArgumentException(string.Format("No worker found for worker type '{0}'", workerType), nameof(workerType));
```
jared-improbable: Use interpolation string?

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
firtina-improbable: I agree, will take a look within CommandLineUtility

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerBase.cs
firtina-improbable: What do you think of this version, to replace `Allow` with `Use`: 

`WorkerBase.UseDynamicId` to match `useDynamicWorkerId`

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerBase.cs
```diff
@@ -53,7 +55,15 @@ public virtual void Connect(ConnectionConfig config)
         {
             if (config is ReceptionistConfig)
             {
-                Connection = ConnectionUtility.ConnectToSpatial((ReceptionistConfig) config, GetWorkerType, WorkerId);
+                string workerIdForConnection;
```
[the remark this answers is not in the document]
firtina-improbable: Good catch! Was missed in previous commits, removing.

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -64,12 +64,20 @@ public void Awake()
...
+                    false);
 
                 // because the launcher does not pass in the worker type as an argument
                 var worker = workerType.Equals(string.Empty)
                     ? WorkerRegistry.CreateWorker<UnityClient>($"{workerType}-{Guid.NewGuid()}", new Vector3(0, 0, 0))
                     : WorkerRegistry.CreateWorker(workerType, workerId, new Vector3(0, 0, 0));
 
+                if (useDynamicWorkerId)
+                {
+                    worker.UseDynamicId = true;
```
jared-improbable: `worker.UseDynamicId = useDynamicWorkerId`? :)
Or just collapse the assignment to the lookup above.

firtina-improbable: I want the `useDynamicWorkerId` to override only if it's true. If it's false, it should not make `worker.UseDynamicId` to false. I should rename `use` to `forceUse` perhaps and add a comment around the if block

firtina-improbable:
Re-tested local and cloud deployments and they work without problems for both server and client workers ✅

workers/unity/spatialos.UnityClient.worker.json
```diff
@@ -46,7 +46,7 @@
           "+projectName",
           "${IMPROBABLE_PROJECT_NAME}",
           "-logfile",
-          "external-default-unityclient.log"
+          "../../logs/external-default-unityclient.log"
```
jessicafalk: nice!

PR #125, opened 2018-07-20
Check build result and throw when failed.

zeroZshadow:
#### Description
Add check to build result instead of ignoring failure

#### Tests
N/A

#### Documentation
N/A

#### Primary reviewers
N/A

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.legacy/Editor/BuildSystem/WorkerBuilder.cs
+++ b/workers/unity/Packages/com.improbable.gdk.legacy/Editor/BuildSystem/WorkerBuilder.cs
@@ -7,6 +7,7 @@ using Improbable.Gdk.Legacy.BuildSystem.Configuration;
 using Improbable.Gdk.Legacy.BuildSystem.Util;
 using UnityEditor;
 using UnityEditor.Build;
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.legacy/Editor/BuildSystem/WorkerBuilder.cs
```diff
@@ -188,7 +189,11 @@ internal static IEnumerable<BuildTarget> GetUnityBuildTargets(SpatialBuildPlatfo
                     locationPathName = workerBuildData.BuildScratchDirectory
                 };
 
-                BuildPipeline.BuildPlayer(buildPlayerOptions);
+                var result = BuildPipeline.BuildPlayer(buildPlayerOptions);
+                if (result.summary.result != BuildResult.Succeeded)
+                {
+                    throw new BuildFailedException($"Build failed for {workerPlatform}");
```
jessicafalk: do we have info on why it failed and can add it to the exception?

zeroZshadow: The Build itself prints all errors into the console.

PR #128, opened 2018-07-23
[Experimental][CSharp Core SDK] Blittable list and map

zeroZshadow:
#### Description
Add Wrappers for List, Dictionary, and string to use in codegen for schema list,map, and strings.
This is a quick version to allow @jamiebrynes7 to make all components blittable.

This still needs to be discussed and improved upon later.
IList and IDictionary was implemented just for ease of use now.
Will need to be re-considered.

#### Primary reviewers
@jamiebrynes7

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Collections/BlittableList.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Collections/BlittableList.cs
@@ -1,0 +1,91 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Improbable.Gdk.Core.Collections
+{
+    public struct BlittableList<TValue> : IList<TValue>
+    {
+        private readonly uint handle;
+
+        internal List<TValue> Internal => CollectionProvider<List<TValue>, TValue>.Get(handle);
+
+        internal BlittableList(uint handle)
+        {
+            this.handle = handle;
+        }
+
+        public static implicit operator List<TValue>(BlittableList<TValue> bl)
+        {
+            return bl.Internal;
+        }
+
+        #region IList<T>
+
+        public TValue this[int index]
+        {
+            get => Internal[index];
+            set => Internal[index] = value;
+        }
+
+        public int Count => Internal.Count;
+
+        public bool IsReadOnly => ((ICollection<TValue>) Internal).IsReadOnly;
+
+        public void Add(TValue item)
+        {
+            Internal.Add(item);
+        }
+
+        public void Insert(int index, TValue item)
+        {
+            Internal.Insert(index, item);
+        }
+
+        public void Clear()
+        {
+            Internal.Clear();
+        }
+
+        public bool Contains(TValue item)
+        {
+            return Internal.Contains(item);
+        }
+
+        public void CopyTo(TValue[] array, int arrayIndex)
+        {
+            Internal.CopyTo(array, arrayIndex);
+        }
+
+        public int IndexOf(TValue item)
+        {
+            return Internal.IndexOf(item);
+        }
+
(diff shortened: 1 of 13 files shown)


zeroZshadow:
Still needs an addition to allow you to provide an initial List/Dictionary/string, for the created handle.

workers/unity/Packages/com.improbable.gdk.core/Collections/BlittableList.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            this.handle = handle;
+        }
+
+        public static implicit operator List<TValue>(BlittableList<TValue> bl)
+        {
+            return bl.Internal;
+        }
+
+        #region IList<T>
```
samiwh: I'm still very against this unless there is some reason we need it. We want it to be a drop in replacement for List not a new IList.

workers/unity/Packages/com.improbable.gdk.core/Collections/BlittableMap.cs
```diff
@@ -0,0 +1,101 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Improbable.Gdk.Core.Collections
+{
+    public struct BlittableMap<TKey, TValue> : IDictionary<TKey, TValue>
```
samiwh: What was the reason for keeping Map instead of dictionary?

workers/unity/Packages/com.improbable.gdk.core/Collections/BlittableString.cs
```diff
@@ -0,0 +1,24 @@
...
+        private readonly uint handle;
+
+        internal string Internal => StringProvider.Get(handle);
+
+        internal BlittableString(uint handle)
+        {
+            this.handle = handle;
+        }
+
+        public void Set(string value)
```
samiwh: Unless this is the only thing you need you'll always have to assign it to a local string variable to use it (except for passing it to functions)

workers/unity/Packages/com.improbable.gdk.core/Collections/CollectionProvider.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+using System.Collections.Generic;
+
+namespace Improbable.Gdk.Core.Collections
+{
+    internal static class CollectionProvider<TProvider, TValue> where TProvider : ICollection<TValue>, new()
```
samiwh: Should this be called TProvider if it's a collection?

workers/unity/Packages/com.improbable.gdk.core/Collections/BlittableMap.cs
zeroZshadow: Because this represents a schema Map.

workers/unity/Packages/com.improbable.gdk.core/Collections/CollectionProvider.cs
zeroZshadow: 👍

workers/unity/Packages/com.improbable.gdk.core/Collections/BlittableMap.cs
samiwh: But it's a blittable version of the Dictionary class. Unless we mean it to be a new IDictionary which I am against.

zeroZshadow: If you want, I can remove the IDictionary, and just straight implement ICollection and still keep the same Add functions etc as Dictionary has?
Same for the List

PR #129, opened 2018-07-23
[Core C# SDK] UTY-697 Part 1: Add Serialization Methods to Types and Components

jamiebrynes7:
#### Description
Commits are logical pieces of work. Will be easier to review by viewing commit by commit.

Laundry List of Changes:
1. All components are blittable (soon ™️) - preemptively made this change in anticipation since I'm already here.
2. Added a schema function mapping: This maps from built in type to Schema functions (Get value from an object, Get the count of a certain type from an object, Add value to an object).
3. Added a ton of logic to the field details: a lot of this was to refactor detail OUT of the template files since its arguably easier to read in the `cs` file.
   1. At instantiation time, the `UnityFieldDetails` will create a detailed `FieldTypeInfo` struct - this consists of 4 nullable fields - one for singular types, list, option, and map respectively. Only one of this will be non-null at any given time.
   2. Inside each `TypeInfo` we have 4 pieces of info - is this a primitive type, is this an enum type, the FQN typename, and a set of serialization functions for schema.
   3. This info is used in each of the generated serialization functions to implement the correct behaviour 😅 
4. Added 2 static serialization methods to each generated type (not components) - these are simply `Serialize` and `Deserialize` which sit in a static class. I.e. -`TypeName.Serialization.Serialize()`
5. Added 3 static serialization methods to each component - `Serialize` (for an sending component updates - currently we don't diff), `Deserialize` (for an `AddComponentOp`), and `ApplyUpdate` which both applies the update to the component data and returns the `ComponentUpdate<T>` struct.
6. Removed the generation of the old C# SDK code - just for convenience! 

Will post some example serialization code below.

#### Tests
The generated code appears to compile - will be able to check exactly when the `Core` compiles with further fixes. 
#### Documentation
N/A
#### Primary reviewers
@zeroZshadow


jamiebrynes7:
Generated Transform code: 

```
// ===========
// DO NOT EDIT - this file is automatically regenerated.
// ===========

using Unity.Entities;
using Improbable.Gdk.Core;

namespace Generated.Improbable.Transform
{ 
    public struct SpatialOSTransform : IComponentData, ISpatialComponentData
    {
        public BlittableBool DirtyBit { get; set; }

        private global::Generated.Improbable.Transform.Location location;

        public global::Generated.Improbable.Transform.Location Location
        {
            get { return location; }
            set
            {
                DirtyBit = true;
                location = value;
            }
        }

        private global::Generated.Improbable.Transform.Quaternion rotation;

        public global::Generated.Improbable.Transform.Quaternion Rotation
        {
            get { return rotation; }
            set
            {
                DirtyBit = true;
                rotation = value;
            }
        }

        private uint tick;

        public uint Tick
        {
            get { return tick; }
            set
            {
                DirtyBit = true;
                tick = value;
            }
        }

        public static class Serialization 
        {
            public static void Serialize(SpatialOSTransform component, global::Improbable.Worker.Core.SchemaObject obj)
            {
                global::Generated.Improbable.Transform.Location.Serialization.Serialize(component.Location, obj.AddObject(1));
                global::Generated.Improbable.Transform.Quaternion.Serialization.Serialize(component.Rotation, obj.AddObject(2));
                obj.AddUint32(3, component.Tick);
            }

            public static SpatialOSTransform Deserialize(global::Improbable.Worker.Core.SchemaObject obj)
            {
                var component = new SpatialOSTransform();
                component.Location = global::Generated.Improbable.Transform.Location.Serialization.Deserialize(obj.GetObject(1));
                component.Rotation = global::Generated.Improbable.Transform.Quaternion.Serialization.Deserialize(obj.GetObject(2));
                component.Tick = obj.GetUint32(3);
                return component;
            }

            public static SpatialOSTransform.Update ApplyUpdate(ref SpatialOSTransform component, global::Improbable.Worker.Core.SchemaObject obj)
            {
                var update = new SpatialOSTransform.Update();
                if (obj.GetObjectCount(1) == 1) 
                {
                    var value = global::Generated.Improbable.Transform.Location.Serialization.Deserialize(obj.GetObject(1));
                    update.Location = new Option<global::Generated.Improbable.Transform.Location>(value);
                    component.Location = value;
                }
                if (obj.GetObjectCount(2) == 1) 
                {
                    var value = global::Generated.Improbable.Transform.Quaternion.Serialization.Deserialize(obj.GetObject(2));
                    update.Rotation = new Option<global::Generated.Improbable.Transform.Quaternion>(value);
                    component.Rotation = value;
                }
                if (obj.GetUint32Count(3) == 1) 
                {
                    var value = obj.GetUint32(3);
                    update.Tick = new Option<uint>(value);
                    component.Tick = value;
                }
             
                return update;
            }
        }

        public struct Update : ISpatialComponentUpdate
        {
            public Option<global::Generated.Improbable.Transform.Location> Location;
            public Option<global::Generated.Improbable.Transform.Quaternion> Rotation;
            public Option<uint> Tick;
        }
    }
}
```

code_generator/GdkCodeGenerator/src/Generation/Model/SchemaFunctionMappings.cs
```diff
@@ -0,0 +1,69 @@
...
+using Improbable.CodeGeneration.Model;
+
+namespace Improbable.Gdk.CodeGenerator
+{
+    public class SchemaFunctionMappings
+    {
+        public static Dictionary<string, string> PrimitiveTypeToAddSchemaFunction = new Dictionary<string, string>
+        {
+            { BuiltInTypeConstants.builtInBool, "AddBool"},
+            { BuiltInTypeConstants.builtInBytes, "AddBytes" },
```
samiwh: Why is bytes a primative but not string?

jamiebrynes7: It is (see below) - this is a schema primitive (maybe built-in is a better name)

code_generator/GdkCodeGenerator/Templates/UnityComponentGenerator.cs
```diff
@@ -0,0 +1 @@
+ErrorGeneratingOutput
```
jessicafalk: what is that?

code_generator/GdkCodeGenerator/src/Generation/Model/SchemaFunctionMappings.cs
samiwh: Discussed offline - should possible change the name. These are proto 2 primitive types (also SchemObject getters) rather than Schemalang primitives.

code_generator/GdkCodeGenerator/Templates/UnityComponentGenerator.cs
jamiebrynes7: An accident - removed in a later commit

jessicafalk: just saw it 👍

code_generator/GdkCodeGenerator/src/Generation/Model/Details/UnityFieldDetails.cs
```diff
@@ -1,42 +1,171 @@
...
         {
             PascalCaseName = Formatting.SnakeCaseToCapitalisedCamelCase(rawFieldDefinition.name);
             CamelCaseName = Formatting.SnakeCaseToCamelCase(rawFieldDefinition.name);
             var packagePrefix = UnityTypeMappings.PackagePrefix;
             RawFieldDefinition = rawFieldDefinition;
+            FieldNumber = rawFieldDefinition.Number;
 
             if (rawFieldDefinition.IsOption())
             {
                 var containedType = GetTypeFromTypeReference(rawFieldDefinition.optionType.valueType, packagePrefix);
```
jessicafalk: move `rawFieldDefinition.optionType.valueType` into a variable to avoid having to use it everywhere ;)

jessicafalk:
do we still need the dirty bit?

jamiebrynes7:
For now yes - implementing a different replication system is something to do later for sure (simple in codegen, less simple in impl)

PR #131, opened 2018-07-24
fixing shell scripts

jessicafalk:

-------

#### Description
yeah, scripts were completely broken.  Fixing it and actually running codegen

#### Tests
ran it locally

#### Documentation
no docs needed

### DIFF
--- a/tools/RunUnity/Program.cs
+++ b/tools/RunUnity/Program.cs
@@ -34,6 +34,7 @@ namespace RunUnity
                 }
 
                 var path = Path.Combine(Paths.TryGetUnityPath(), location);
+                Console.WriteLine($"Found Unity in {path}");
                 var arguments = string.Join(' ', args);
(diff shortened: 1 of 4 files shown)


tools/RunUnity/Program.cs
```diff
@@ -34,6 +34,7 @@ static void Main(string[] args)
                 }
 
                 var path = Path.Combine(Paths.TryGetUnityPath(), location);
+                Console.WriteLine($"Found Unity in {path}");
```
jessicafalk: added this as I got trolled for roughly an hour by it picking the wrong Unity version...

workers/unity/Packages/manifest.json
```diff
@@ -14,7 +14,6 @@
   },
   "testables": [
     "com.unity.collections",
-    "com.unity.entities",
```
zeroZshadow: Drive by change?

PR #135, opened 2018-07-24
UTY-708: Add Scoring to the Playground

Ivaskovic:

-------

#### Description
Scoring is added to Playground. There is an UI element that displays the score and players get points for shooting cubes (as well as when cubes collide).

Key changes:
- The `Cube` prefab now has a `MonoBehaviour` that registers collisions.
- These collisions are represented as ECS components that are added to the colliding entities (that also contain `Launchable`s).
- Every `Launchable` might contain the ID of the 'most recent player' entity associated with it (in order to determine who gets points).
- There is a system that handles these collision events (based on the temporary `CollisionComponent`s).


workers/unity/ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +0,0 @@
-m_EditorVersion: 2018.2.0b10
```
martinzlocha: Was this intentional?

workers/unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -52,6 +52,7 @@ PlayerSettings:
   m_StackTraceTypes: 010000000100000001000000010000000100000001000000
   iosShowActivityIndicatorOnLoading: -1
   androidShowActivityIndicatorOnLoading: -1
+  tizenShowActivityIndicatorOnLoading: -1
```
martinzlocha: There are many changes in this file which relate to tizen, might be better to revert this file.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,68 @@
...
+                    continue;
+                }
+                if (first_launcher != 0 && second_launcher != 0 && first_launcher != second_launcher)
+                {
+                    first.MostRecentLauncher = 0;
+                    second.MostRecentLauncher = 0;
+                    launchableData.Launchable[i] = first;
+                    collision.other = second;
+                    launchableData.Collision[i] = collision;
+                    continue;
```
martinzlocha: Would it be possible to use else if instead of continue in these statements?

schema/playground/launcher.schema
```diff
@@ -20,19 +21,31 @@ type LaunchMeCommandRequest {
...
+    command ScoreIncreaseResponse score_increase(ScoreIncreaseRequest);
 }
 
 component Launchable {
     id = 12006;
+    int64 most_recent_launcher = 1;
     command LaunchMeCommandResponse launch_me(LaunchMeCommandRequest);
 }
 
+
```
jamiebrynes7: Nit: Extra newline

schema/playground/launcher.schema
```diff
@@ -20,19 +21,31 @@ type LaunchMeCommandRequest {
...
+    float amount_increase = 2;
+}
+
+type ScoreIncreaseResponse {}
+
 component Launcher {
     id = 12005;
     float energy_left = 1;
     float recharge_time_left = 3;
+    float score = 4;
```
jamiebrynes7: Should probably refactor this to its own `Score` component

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        var component = GetComponent<SpatialOSComponent>();
+        if (component)
+        {
+            view = WorkerRegistry.GetWorkerForWorld(component.World).View;
+            ownLaunchable = view.GetComponent<SpatialOSLaunchable>(component.Entity);
+        }
+        else
+        {
+            Debug.LogError("Could not retrieve SpatialOSComponent from game object.");
```
martinzlocha: We arent using Debug.LogError anymore, use the LoggingDispatcher.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,52 @@
...
+    public void OnCollisionEnter(Collision col)
+    {
+        if (col.gameObject && col.gameObject.tag == "Cube")
+        {
+            var component = col.gameObject.GetComponent<SpatialOSComponent>();
+            if (component != null && view.HasComponent(component.Entity, typeof(SpatialOSLaunchable)))
+            {
+                SpatialOSLaunchable otherEntity = view.GetComponent<SpatialOSLaunchable>(component.Entity);
+                var entityManager = World.Active.GetExistingManager<EntityManager>();
+                Entity collisionEventEntity = entityManager.CreateEntity(typeof(SpatialOSLaunchable), typeof(CollisionComponent),
```
martinzlocha: Why are you creating a new entity, could the component be added to an entity which already exists? Also if this is required when do the entities get cleaned up?

workers/unity/Assets/Playground/Scripts/Player/PlayerCommandsSystem.cs
```diff
@@ -90,8 +90,10 @@ protected override void OnUpdate()
                     EntityToLaunch = component.SpatialEntityId,
                     ImpactPoint = impactPoint,
                     LaunchDirection = launchDirection,
-                    LaunchEnergy = command == PlayerCommand.LaunchLarge ? LargeEnergy : SmallEnergy
+                    LaunchEnergy = command == PlayerCommand.LaunchLarge ? LargeEnergy : SmallEnergy,
+                    Launcher = playerId
                 });
+
```
martinzlocha: Nit: Extra whitespace

schema/playground/launcher.schema
```diff
@@ -20,19 +21,31 @@ type LaunchMeCommandRequest {
...
     command LaunchCommandResponse launch_entity(LaunchCommandRequest);
+    command ScoreIncreaseResponse score_increase(ScoreIncreaseRequest);
 }
 
 component Launchable {
     id = 12006;
+    int64 most_recent_launcher = 1;
```
jamiebrynes7: Small driveby - there is actually an `EntityId` schema type that you can use (and should probably use). It looks like the original schema was using `int64`. 

Can you change these to `EntityId` where appropriate - it shouldn't affect the code generation (right now)

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
martinzlocha: Oh I just realized that you can't do that given that you might not have access to a View. Ok ignore that comment.

workers/unity/Assets/Playground/Scripts/Player/ProcessScoresSystem.cs
```diff
@@ -0,0 +1,38 @@
...
+    [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
+    class ProcessScoresSystem : ComponentSystem
+    {
+        private struct ScoringData
+        {
+            public readonly int Length;
+            public ComponentDataArray<SpatialOSLauncher> Launcher;
+
+            [ReadOnly]
+            public ComponentArray<CommandRequests<Generated.Playground.Launcher.ScoreIncrease.Request>> CommandRequests;
```
jessicafalk: do we need the whole namespace here

workers/unity/Assets/Playground/Scripts/Player/ProcessScoresSystem.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        [Inject] private ScoringData scoringData;
+
+        protected override void OnUpdate()
+        {
+            // Process all requests for updating player scores.
+            // Increase the respective players' scores.
```
jessicafalk: I don't think the comment is needed, the system is nice and short. :)

workers/unity/Assets/Playground/Scripts/Player/ProcessScoresSystem.cs
```diff
@@ -0,0 +1,38 @@
...
+            for (int i = 0; i < scoringData.Length; i++)
+            {
+                var launcher = scoringData.Launcher[i];
+                foreach (var request in scoringData.CommandRequests[i].Buffer)
+                {
+                    var info = request.RawRequest;
+                    launcher.Score += info.AmountIncrease;
```
jessicafalk: merge the two lines into one

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: Indeed it can, it's two different ways of approaching it.
The entities get cleaned up because of the [RemoveAtEndOfTick] annotation of CollisionComponent.

workers/unity/ProjectSettings/UnityConnectSettings.asset
```diff
@@ -3,7 +3,7 @@
 --- !u!310 &1
 UnityConnectSettings:
   m_ObjectHideFlags: 0
-  m_Enabled: 0
```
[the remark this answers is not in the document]
Ivaskovic: Whoops! Point taken.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public void OnCollisionEnter(Collision col)
+    {
+        if (col.gameObject && col.gameObject.tag == "Cube")
+        {
+            var component = col.gameObject.GetComponent<SpatialOSComponent>();
+            if (component != null && view.HasComponent(component.Entity, typeof(SpatialOSLaunchable)))
+            {
+                SpatialOSLaunchable otherEntity = view.GetComponent<SpatialOSLaunchable>(component.Entity);
```
jessicafalk: would not call this `otherEntity`, this is a component, right?

workers/unity/Assets/Playground/Scripts/Player/ProcessScoresSystem.cs
Ivaskovic: My personal preference is to keep it -- but I agree it's best to be consistent and remove it for simple systems.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,52 @@
...
+                var entityManager = World.Active.GetExistingManager<EntityManager>();
```
jessicafalk: do we need to get the entity manager at every collision? i.e. store it when running `Start`

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: Point taken, forgot to rename it when I changed the logic slightly.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
jamiebrynes7: The `RemoveAtEndOfTick` doesn't clean up the entity - just the component

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: Good point!

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: Will do!

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
jamiebrynes7: We shouldn't be creating instances of `SpatialOS*` components in user code - instead, add the information required in the `CollisionComponent`

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
Ivaskovic: I tend to prefer not using continue, break and similar, though this seemed fairly natural. Will change it.

workers/unity/ProjectSettings/ProjectVersion.txt
Ivaskovic: Included by accident!

workers/unity/ProjectSettings/ProjectSettings.asset
Ivaskovic: Included by accident!

workers/unity/Assets/Playground/Scripts/Workers/UnityClient.cs
```diff
@@ -45,6 +45,10 @@ public override void RegisterSystems()
 
             // Metric sending system
             World.GetOrCreateManager<MetricSendSystem>();
+
+            // Score update systems
+            World.GetOrCreateManager<ProcessScoresSystem>();
```
jamiebrynes7: Why is this on both the client and the game logic workers? Only one of them will receive the command (the game logic since its authoritative over the `Launchable` component). 

The pattern should be closer to: 
Collision happens -> send command to increment score -> command received by the worker authoritative over the launchable component, update the score on the component -> client listens for updates and updates the score UI locally.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: Actually, this check is not needed (it shouldn't ever fail), I think?

jamiebrynes7: It could if someone puts the `ScoreCollision` component on a prefab that is not instantiated by our systems - granted that's a user error, but the check is probably still worth having

workers/unity/Assets/Playground/Scripts/Player/ProcessScoresSystem.cs
Ivaskovic: Unfortunately, we do :(

workers/unity/Assets/Playground/Scripts/Player/ProcessScoresSystem.cs
Ivaskovic: Will do!

schema/playground/launcher.schema
```diff
@@ -11,6 +11,7 @@ type LaunchCommandRequest {
     improbable.Vector3f impact_point = 2;
     improbable.Vector3f launch_direction = 3;
     float launch_energy = 4;
+    int64 launcher = 5;
```
samiwh: What does this represent?

schema/playground/launcher.schema
```diff
@@ -20,19 +21,31 @@ type LaunchMeCommandRequest {
     improbable.Vector3f impact_point = 1;
     improbable.Vector3f launch_direction = 2;
     float launch_energy = 3;
+    int64 launcher = 4;
 }
 
 type LaunchMeCommandResponse {}
 
+type ScoreIncreaseRequest {
+    int64 launcher_to_score = 1;
```
samiwh: Not sure what this means.

schema/playground/launcher.schema
```diff
@@ -20,19 +21,31 @@ type LaunchMeCommandRequest {
...
+    float amount_increase = 2;
```
samiwh: Minor: amount_increase sounds a bit odd. Amount or points might be better.

schema/playground/launcher.schema
samiwh: +1

samiwh: Also in a different file. A component is meant to represent some set of behaviour on that entity. Score and launch are quite different although one consumes the other.

schema/playground/launcher.schema
```diff
@@ -20,19 +21,31 @@ type LaunchMeCommandRequest {
...
+
+type ScoreIncreaseResponse {}
+
 component Launcher {
     id = 12005;
     float energy_left = 1;
     float recharge_time_left = 3;
+    float score = 4;
     command LaunchCommandResponse launch_entity(LaunchCommandRequest);
+    command ScoreIncreaseResponse score_increase(ScoreIncreaseRequest);
```
samiwh: Minor: Sounds better as `increase_score`. You are sending a commands rather than state.

workers/unity/Assets/Playground/Resources/Prefabs/Cube.prefab
```diff
@@ -8,32 +8,33 @@ Prefab:
     m_TransformParent: {fileID: 0}
     m_Modifications: []
     m_RemovedComponents: []
-  m_ParentPrefab: {fileID: 0}
+  m_SourcePrefab: {fileID: 0}
```
samiwh: Please put in the description what has changed on the prefabs. If too much stuff then say which prefabs and that reviewers will have to check them out.

samiwh: A diff is always useful though. Screenshots can be useful here.

workers/unity/Assets/Playground/Scripts/Player/ProcessScoresSystem.cs
samiwh: Can put a using at the top if it affects readability.

workers/unity/Assets/Playground/Scripts/Player/ProcessScoresSystem.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Playground
+{
+    [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
+    class ProcessScoresSystem : ComponentSystem
+    {
+        private struct ScoringData
+        {
+            public readonly int Length;
+            public ComponentDataArray<SpatialOSLauncher> Launcher;
```
samiwh: Might want to add that you need authority too. There is a race on receiving a command request and then losing authority. Although really we should handle that in core if we don't, as that can only troll users.

samiwh: However there is still an argument for adding it to specify intent.

workers/unity/Assets/Playground/Scripts/Player/ProcessLaunchCommandSystem.cs
```diff
@@ -62,7 +64,8 @@ protected override void OnUpdate()
                     {
                         ImpactPoint = info.ImpactPoint,
                         LaunchDirection = info.LaunchDirection,
-                        LaunchEnergy = energy
+                        LaunchEnergy = energy,
+                        Launcher = info.Launcher
```
samiwh: Not a bug introduced in this PR but it looks like this is missing the tag for authority on the injection group.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,59 @@
...
+        Debug.Log(WorkerRegistry.GetWorkerForWorld(GetComponent<SpatialOSComponent>().World).WorkerId);
+        if (entityManager == null)
+        {
+            return;
+        }
+
+        if (col.gameObject && col.gameObject.tag == "Cube")
+        {
+            var component = col.gameObject.GetComponent<SpatialOSComponent>();
+            if (component != null && view.HasComponent(component.Entity, typeof(SpatialOSLaunchable)))
```
samiwh: This doesn't need the view. It can just use the entity registry. I think we can remove the dependency on the view entirely.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        var component = GetComponent<SpatialOSComponent>();
+        if (component)
+        {
+            view = WorkerRegistry.GetWorkerForWorld(component.World).View;
+            entityManager = component.World.GetExistingManager<EntityManager>();
+        }
+        else
+        {
+            Debug.LogError("Could not retrieve SpatialOSComponent from game object.");
```
samiwh: We should make a ticket for exposing the logger in the monobehaviours.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,59 @@
...
+    public SpatialOSLaunchable own;
+    public SpatialOSLaunchable other;
+}
+
+public class ScoreCollision : MonoBehaviour
+{
+    private MutableView view;
+    private EntityManager entityManager;
+
+    public void Start()
```
samiwh: Better to run this in OnEnable. Right now Start will work fine, but if you pool the prefabs this will break.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,69 @@
...
+            [ReadOnly] public ComponentDataArray<CommandRequestSender<SpatialOSLauncher>> Sender;
+        }
+
+        [Inject] private LaunchableData launchableData;
+
+        protected override void OnUpdate()
+        {
+            for (var i = 0; i < launchableData.Length; i++)
+            {
+                Debug.Log("there are collisions");
```
martinzlocha: Is this meant to be here?

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,59 @@
+using Generated.Playground;
+using Improbable.Gdk.Core;
+using Unity.Entities;
+using UnityEngine;
+
+[RemoveAtEndOfTick]
+public struct CollisionComponent : IComponentData
```
samiwh: Define this in the components directory relevant to this feature rather than with the monobehaviour.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+        else
+        {
+            Debug.LogError("Could not retrieve SpatialOSComponent from game object.");
+        }
+    }
+
+    public void OnCollisionEnter(Collision col)
+    {
+        Debug.Log(WorkerRegistry.GetWorkerForWorld(GetComponent<SpatialOSComponent>().World).WorkerId);
```
samiwh: Was this meant to be left in.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,69 @@
...
+                var collision = launchableData.Collision[i];
+                var first = collision.own;
+                var second = collision.other;
+                var first_launcher = first.MostRecentLauncher;
+                var second_launcher = second.MostRecentLauncher;
+                if (first_launcher == 0 && second_launcher == 0)
+                {
```
samiwh: What's going on here?

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,69 @@
...
+                }
+                else if (first_launcher != 0 && second_launcher == 0)
+                {
+                    second.MostRecentLauncher = first_launcher;
+                    collision.other = second;
+                    launchableData.Collision[i] = collision;
+                }
+                else if (first_launcher == second_launcher)
+                {
+                    launchableData.Sender[i].SendScoreIncreaseRequest(first_launcher,
```
samiwh: Should this increase the score? Wouldn't this mean no new cubes are controlled?

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
martinzlocha: I mean yes this approach does work but its much more heavy weight, you are creating an entity with 3 components but you could just be adding a component to an existing entity.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
Ivaskovic: Handling different kinds of collisions, mostly determining whose score increases. But I agree that this warrants some comments.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
+    internal class CollisionProcessSystem : ComponentSystem
+    {
+        private struct LaunchableData
+        {
+            public readonly int Length;
+            public ComponentDataArray<CollisionComponent> Collision;
+            public EntityArray Entities;
+            [ReadOnly] public ComponentDataArray<CommandRequestSender<SpatialOSLauncher>> Sender;
```
samiwh: I'm not a little confused about what this entity is. I though at first it was a spatial one, but now it looks like it's not. In which case this sender won't be matched against anything.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: +1

Ivaskovic: +1

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: ACK

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: Oops, well spotted.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,59 @@
+using Generated.Playground;
+using Improbable.Gdk.Core;
+using Unity.Entities;
+using UnityEngine;
+
+[RemoveAtEndOfTick]
+public struct CollisionComponent : IComponentData
+{
+    public SpatialOSLaunchable own;
```
samiwh: If this is for an event entity should this be own and other? (Contrary to my other comment from when I thought this was being read on the entity themselves).
If this is being read on the spatial entity then this is fine.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: ACK

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: Update: can't really do this in OnEnable

jessicafalk: why?

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,47 @@
...
+        if (entityManager == null)
+        {
+            return;
+        }
+
+        if (col.gameObject && col.gameObject.tag == "Cube")
+        {
+            var component = GetComponent<SpatialOSComponent>();
+            if (component != null && view.HasComponent(component.Entity, typeof(SpatialOSLaunchable))
+                && !view.HasComponent(component.Entity, typeof(Playground.CollisionComponent)))
```
jessicafalk: you only use the `MutableView` for `HasComponent`. Remove the `MutableView` completely from this script and use the `entityManager` directly. If you look at the code, we only use it as a wrapper and is really not necessary.

workers/unity/Assets/Playground/Scripts/Player/ProcessLaunchCommandSystem.cs
```diff
@@ -33,6 +33,8 @@ private struct LaunchableData
             public ComponentArray<CommandRequests<Generated.Playground.Launchable.LaunchMe.Request>> CommandRequests;
 
             [ReadOnly] public ComponentArray<Rigidbody> Rigidbody;
+
```
jessicafalk: no need for new line

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,76 @@
...
+                var first_launchable = EntityManager.GetComponentData<SpatialOSLaunchable>(first);
+                var second_launchable = EntityManager.GetComponentData<SpatialOSLaunchable>(second);
+                var first_launcher = first_launchable.MostRecentLauncher;
+                var second_launcher = second_launchable.MostRecentLauncher;
+                if (first_launcher == second_launcher)
+                {
+                    if (first_launcher != 0)
+                    {
+                        data.Sender[i].SendIncreaseScoreRequest(first_launcher,
+                            new Generated.Playground.ScoreIncreaseRequest()
```
jessicafalk: do we need the whole namespace or can we just write `new ScoreIncreaseRequest()`

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: The component may not exist at that stage (and hence it's impossible to get the EntityManager). Observed this in practice.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: Good point! Doing so right now.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
Ivaskovic: We need it. Not sure why.

workers/unity/Assets/Playground/Scripts/Player/ProcessLaunchCommandSystem.cs
Ivaskovic: ACK

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
jessicafalk: you could initialize  lazily instead of during Start or OnEnable. that would also get rid of reassigning `component` during each collision

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
+    internal class CollisionProcessSystem : ComponentSystem
+    {
+        private struct Data
+        {
+            public readonly int Length;
+            public ComponentDataArray<CollisionComponent> Collision;
+            public EntityArray Entities;
+            public ComponentDataArray<SpatialOSLaunchable> Launchable;
```
jessicafalk: nit: personally prefer to have them ordered by `readonly`, type of array, .. etc.
so don't have a `ComponentDataArray`, `EntityArray `, `ComponentDataArray`. makes it a bit harder to read

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,76 @@
...
+        public Entity otherEntity;
+    }
+
+    [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
+    internal class CollisionProcessSystem : ComponentSystem
+    {
+        private struct Data
+        {
+            public readonly int Length;
+            public ComponentDataArray<CollisionComponent> Collision;
```
jessicafalk: can this be readonly?

schema/playground/launcher.schema
jessicafalk: in that case, could we make it more obvious? :)

Ivaskovic: Next commit is doing mostly renaming :)

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: True, though it means we make the check over and over again. It's not a huge change. though.

jessicafalk:
it looks like it still has the wrong project version? https://github.com/spatialos/UnityGDK/blob/feature/UTY-708-scoring/workers/unity/ProjectSettings/ProjectVersion.txt

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
jessicafalk: you already make the check `if (entityManager ==null)` already over and over again. just put the logic in there instead of just returning

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
Ivaskovic: We don't seem to have an agreed way of ordering them. This seemed OK me, as `CollisionComponent` objects are the only ones that the collision monobehaviour injects.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
Ivaskovic: It can indeed.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
jessicafalk: found out why: if you just use `new ScoreIncreaseRequest()`, it will pick the wrong code-generated struct, not our wrapper. in that case it's fine

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -5,35 +5,24 @@
...
-            view = WorkerRegistry.GetWorkerForWorld(component.World).View;
-            entityManager = component.World.GetExistingManager<EntityManager>();
-        }
-        else
+        if (!component)
         {
-            Debug.LogError("Could not retrieve SpatialOSComponent from game object.");
+            Debug.LogError("Could not get SpatialOSComponent.");
+            return;
         }
```
jessicafalk: move lines 12 - 17 also into the `entityManager == null` part

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    public void OnCollisionEnter(Collision col)
+    {
+        var component = GetComponent<SpatialOSComponent>();
+        if (!component)
+        {
+            Debug.LogError("Could not get SpatialOSComponent.");
+            return;
+        }
+        if (entityManager == null)
```
jessicafalk: comment seems to have disappeared, so trying it again: move lines 12 - 17 into this if-statement

Ivaskovic: 👍

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,37 @@
...
+            }
+            entityManager = component.World.GetExistingManager<EntityManager>();
+        }
+        if (col.gameObject && col.gameObject.tag == "Cube")
+        {
+            if (entityManager.HasComponent(component.Entity, typeof(SpatialOSLaunchable))
+                && !entityManager.HasComponent(component.Entity, typeof(Playground.CollisionComponent)))
+            {
+                entityManager.AddComponent(component.Entity, typeof(Playground.CollisionComponent));
+                entityManager.SetComponentData(component.Entity, new Playground.CollisionComponent
```
jessicafalk: use ` entityManager.AddComponentData(component.Entity, new Playground.ColissionComponent{ ... } )` instead of set and add

schema/playground/launcher.schema
jessicafalk: launcher is about shooting something, right? (I find the name very confusing)
 this component is mainly about shooting and `energy_left` directly defines whether you are able to shoot or not.

The score is a result of shooting, but is not as tightly coupled as the shooting energy. So I agree that this should be refactored into a `Score` component

schema/playground/launcher.schema
```diff
@@ -7,10 +7,11 @@ import "improbable/vector3.schema";
 
 // Command sent by clients to request launching a Launchable.
 type LaunchCommandRequest {
-    int64 entity_to_launch = 1;
+    EntityId entity_to_launch = 1;
     improbable.Vector3f impact_point = 2;
     improbable.Vector3f launch_direction = 3;
     float launch_energy = 4;
+    EntityId launcher_responsible = 5;
```
jessicafalk: `launcher_responsible` still doesn't really say what it is imo. how about just `player` / `launching_player`?

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,76 @@
...
+        public Entity otherEntity;
+    }
+
+    [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
+    internal class CollisionProcessSystem : ComponentSystem
+    {
+        private struct Data
+        {
+            public readonly int Length;
+            public readonly ComponentDataArray<CollisionComponent> Collision;
```
jessicafalk: that's the wrong tag. it should be `[ReadOnly] public ...`

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+        if (col.gameObject && col.gameObject.tag == "Cube")
+        {
+            if (entityManager.HasComponent(component.Entity, typeof(SpatialOSLaunchable))
+                && !entityManager.HasComponent(component.Entity, typeof(Playground.CollisionComponent)))
+            {
+                entityManager.AddComponentData(component.Entity, new Playground.CollisionComponent
+                {
+                    ownEntity = component.Entity,
+                    otherEntity = col.gameObject.GetComponent<SpatialOSComponent>().Entity,
```
jessicafalk: just realized that this could fail. I think, we need to check whether the other object also has a `SpatialOSComponent`

schema/playground/launcher.schema
Ivaskovic: I agree about the naming being confusing.

And sure enough, I'll refactor it, then.

schema/playground/launcher.schema
```diff
@@ -7,10 +7,11 @@ import "improbable/vector3.schema";
 
 // Command sent by clients to request launching a Launchable.
 type LaunchCommandRequest {
-    int64 entity_to_launch = 1;
+    EntityId entity_to_launch = 1;
     improbable.Vector3f impact_point = 2;
     improbable.Vector3f launch_direction = 3;
     float launch_energy = 4;
+    EntityId player = 5;
```
samiwh: This is only sent from a player entity top the player entity. So this field isn't needed.

schema/playground/launcher.schema
```diff
@@ -20,19 +21,28 @@ type LaunchMeCommandRequest {
     improbable.Vector3f impact_point = 1;
     improbable.Vector3f launch_direction = 2;
     float launch_energy = 3;
+    EntityId player = 4;
```
samiwh: What is this?

samiwh: A component should tell you more or less everything you need to know about some game system.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -0,0 +1,40 @@
...
+    private SpatialOSComponent component;
+
+    public void OnCollisionEnter(Collision col)
+    {
+        if (entityManager == null)
+        {
+            component = GetComponent<SpatialOSComponent>();
+            if (!component)
+            {
+                Debug.LogError("Could not get SpatialOSComponent.");
```
samiwh: Use the logger rather than Debug.Log

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
+    internal class CollisionProcessSystem : ComponentSystem
+    {
+        private struct Data
+        {
+            public readonly int Length;
+            public EntityArray Entities;
+            public ComponentDataArray<SpatialOSLaunchable> Launchable;
```
samiwh: Having both this and the command sender is generally an anti-pattern. You want each system to have exactly the data it needs, both to prevent performance loss and, more importantly, to make it easy to tell at a glance what the system does. You don't want to include more stuff than you need and work out which bits to write to later if you can avoid it.

samiwh: The way to avoid this here is for the collision detection code to work out what is going on and add the relevant component to each entity. Then the presents of a specific component tells you whether you want to write to state or send a command. This could also be done in a system that runs before this if you wanted to keep the responsibility of each system / monobehaviour as limited as possible.

schema/playground/launcher.schema
Ivaskovic: This was already discussed (with Jessica). I like `player` or `originating_player` (with a preference for the latter because it's shorter). I don't think it's too important.

samiwh: But I have no idea what it is by looking at it without reading through a load of systems. Which means it's not a good name in this context.

samiwh: The alternative is that you add comments for what the component is for.

schema/playground/launcher.schema
samiwh: If this command is now being sent to other entities then this probably needs to be redesigned.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: Potentially problematic if we don't have access to a view. I'll keep it as is right now.

schema/playground/launcher.schema
Ivaskovic: Not exactly, it is sent to a Cube entity: it is needed to keep track of the most recent player that shot the cube.

PR #143, opened 2018-07-26
Adding support for IL2CPP to GDK

gkassabli:
This adds assembly linker instructions to not strip classes needed for AOT and instantiation calls to methods that should be declared for AOT

-------

Adding support for IL2CPP in GDK.
Enable IL2CPP and ensure that iOS-compiled client starts correctly
@jessicafalk

### DIFF
--- a/workers/unity/Assets/link.xml
+++ b/workers/unity/Assets/link.xml
@@ -1,0 +1,5 @@
+<linker>
+	<assembly fullname="Improbable.Gdk.Core" preserve="all"/>
+	<assembly fullname="Improbable.Gdk.TransformSynchronization" preserve="all"/>
+	<assembly fullname="Improbable.Gdk.PlayerLifecycle" preserve="all"/>
+</linker>
--- a/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
@@ -7,8 +7,8 @@ namespace Improbable.Gdk.Core
     ///     Any component with this attribute will be removed from all entities by the CleanReactiveComponentSystem
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
-    [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false, Inherited = false)]
-    public class RemoveAtEndOfTickAttribute : Attribute
+    [AttributeUsage(AttributeTargets.Struct)]
+    public class RemoveAtEndOfTick : Attribute
     {
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
@@ -4,11 +4,13 @@ using Unity.Entities;
 namespace Improbable.Gdk.Core
 {
     /// <summary>
-    ///     Any component with this attribute will be removed from all entities by the CleanReactiveComponentSystem
+    ///     Any component that inheriting this interrace will be removed from all entities
+    ///     by the CleanReactiveComponentSystem.
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
-    [AttributeUsage(AttributeTargets.Struct)]
-    public class RemoveAtEndOfTickAttribute : Attribute
+
+    public interface RemoveAtEndOfTick
     {
+        void RemoveComponent(EntityCommandBuffer commands, Entity entity);
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
@@ -10,6 +10,5 @@ namespace Improbable.Gdk.Core
     [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false, Inherited = false)]
     public class RemoveAtEndOfTick : Attribute
     {
-        
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
@@ -8,7 +8,7 @@ namespace Improbable.Gdk.Core
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
     [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false, Inherited = false)]
-    public class RemoveAtEndOfTick : Attribute
+    public class RemoveAtEndOfTickAttribute : Attribute
     {
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
@@ -8,7 +8,7 @@ namespace Improbable.Gdk.Core
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
     [AttributeUsage(AttributeTargets.Struct)]
-    public class RemoveAtEndOfTick : Attribute
+    public class RemoveAtEndOfTickAtrribute : Attribute
     {
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
@@ -4,12 +4,12 @@ using Unity.Entities;
 namespace Improbable.Gdk.Core
 {
     /// <summary>
-    ///     Any component which inherits this interface will be removed from all entities by the CleanReactiveComponentSystem.
+    ///     Any component with this attribute will be removed from all entities by the CleanReactiveComponentSystem
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
-
-    public interface IRemoveableComponent
+    [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false, Inherited = false)]
+    public class RemoveAtEndOfTick : Attribute
     {
-        void RemoveComponent(EntityCommandBuffer commands, Entity entity);
+        
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
@@ -4,12 +4,11 @@ using Unity.Entities;
 namespace Improbable.Gdk.Core
 {
     /// <summary>
-    ///     Any component that inheriting this interrace will be removed from all entities
-    ///     by the CleanReactiveComponentSystem.
+    ///     Any component which inherits this interface will be removed from all entities by the CleanReactiveComponentSystem.
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
 
-    public interface RemoveAtEndOfTick
+    public interface IRemoveableComponent
     {
         void RemoveComponent(EntityCommandBuffer commands, Entity entity);
     }
--- a/workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
@@ -7,7 +7,7 @@ namespace Improbable.Gdk.Core
     ///     This component is automatically added to an entity upon its creation and automatically removed at the end of the
     ///     same frame.
     /// </summary>
-    public struct NewlyAddedSpatialOSEntity : IComponentData, RemoveAtEndOfTick
+    public struct NewlyAddedSpatialOSEntity : IComponentData, IRemoveableComponent
     {
         public void RemoveComponent(EntityCommandBuffer commands, Entity entity)
         {
--- a/workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
@@ -7,7 +7,7 @@ namespace Improbable.Gdk.Core
     ///     This component is automatically added to an entity upon its creation and automatically removed at the end of the
     ///     same frame.
     /// </summary>
-    [RemoveAtEndOfTick]
+    [RemoveAtEndOfTickAttribute]
     public struct NewlyAddedSpatialOSEntity : IComponentData
     {
     }
--- a/workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
@@ -7,11 +7,8 @@ namespace Improbable.Gdk.Core
     ///     This component is automatically added to an entity upon its creation and automatically removed at the end of the
     ///     same frame.
     /// </summary>
-    public struct NewlyAddedSpatialOSEntity : IComponentData, IRemoveableComponent
+    [RemoveAtEndOfTick]
+    public struct NewlyAddedSpatialOSEntity : IComponentData
     {
-        public void RemoveComponent(EntityCommandBuffer commands, Entity entity)
-        {
-            commands.RemoveComponent<NewlyAddedSpatialOSEntity>(entity);
-        }
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
@@ -7,8 +7,11 @@ namespace Improbable.Gdk.Core
     ///     This component is automatically added to an entity upon its creation and automatically removed at the end of the
     ///     same frame.
     /// </summary>
-    [RemoveAtEndOfTick]
-    public struct NewlyAddedSpatialOSEntity : IComponentData
+    public struct NewlyAddedSpatialOSEntity : IComponentData, RemoveAtEndOfTick
     {
+        public void RemoveComponent(EntityCommandBuffer commands, Entity entity)
+        {
+            commands.RemoveComponent<NewlyAddedSpatialOSEntity>(entity);
+        }
     }
 }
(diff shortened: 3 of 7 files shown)


gkassabli:
Original PR with reviews https://github.com/spatialos/UnityGDK/pull/140

jessicafalk:
as mentioned in the other PR: I would prefer to have it in master, but would be fine with accepting this for now to unblock iOS. We will however need to create a ticket to fix the actual problems with IL2CPP

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -87,6 +87,21 @@ private void AddRemoveComponentAction<T>()
                 }
             });
         }
+        
+        public void UsedOnlyForAotCodeGeneration()
```
gkassabli: This might be usable as private. I'll check if it's possible

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -87,6 +87,21 @@ private void AddRemoveComponentAction<T>()
...
+        {
+            // IL2CPP needs those lines for AOT instantiation of methods
+            AddRemoveComponentAction<OnConnected>();
```
gkassabli: This is error-prone if we need to instantiate every method that AOT needs. Unfortunately, IL2CPP can only be used with AOT, which in turn, needs instantiation of these methods.
I'll try to experiment to see what we can do here to make this more

workers/unity/Assets/link.xml
```diff
@@ -0,0 +1,20 @@
+<linker>
+	<assembly fullname="Improbable.Gdk.Core">
```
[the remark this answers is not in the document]
jessicafalk: we will have to actually add some documentation on our side to make sure that users are aware that we are doing this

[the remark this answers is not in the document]
gkassabli: What would be a good way to document this?

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -87,6 +87,18 @@ private void AddRemoveComponentAction<T>()
                 }
             });
         }
+        
+        private void UsedOnlyForAotCodeGeneration()
+        {
+            // IL2CPP needs those lines for AOT instantiation of methods
+            AddRemoveComponentAction<OnConnected>();
+            AddRemoveComponentAction<OnDisconnected>();
+            AddRemoveComponentAction<NewlyAddedSpatialOSEntity>();
```
gkassabli: It doesn't look like there's generic way to declare such methods for AOT.
Not sure if there's any other way to address this issue (some kind of code generation may be).

workers/unity/Assets/link.xml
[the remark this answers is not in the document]
jessicafalk: if you look at our repo, we have a docs folder. Create an internal google docs where you would add the documentation for this feature and then share it with the team so that we can review it before it will be added to the pr

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
jessicafalk: that is unfortunate... maybe we need to refactor this system or, as you said, find a way to codegenerate it

jessicafalk: one idea jamie and I had is to hook into Unity's build system. Before actually running the build, use reflection to find all tags that have the `RemoveAtEndOfTick`attribute and generate this method. not sure how nice that will be, but would probably be the best solution unless we do a redesign.

relevant docs:
https://docs.unity3d.com/Manual/BuildPlayerPipeline.html
https://docs.unity3d.com/ScriptReference/Build.IPreprocessBuildWithReport.html
https://docs.unity3d.com/ScriptReference/Build.IPreprocessBuildWithReport.OnPreprocessBuild.html

workers/unity/Assets/link.xml
```diff
@@ -0,0 +1,20 @@
...
+		<type fullname="Improbable.Gdk.Core.WorldCommandsTranslation" preserve="all"/>
```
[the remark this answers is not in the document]
jessicafalk: that makes sense. would be interesting to do some benchmarking at some point to see how much less efficient this would acutally be

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
jamiebrynes7: That is assuming that you can generate files that are then included in the builds step

workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTick.cs
```diff
@@ -4,11 +4,13 @@
 namespace Improbable.Gdk.Core
 {
     /// <summary>
-    ///     Any component with this attribute will be removed from all entities by the CleanReactiveComponentSystem
+    ///     Any component that inheriting this interrace will be removed from all entities
```
gkassabli: typo: interface

workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
```diff
@@ -7,8 +7,11 @@ namespace Improbable.Gdk.Core
...
     ///     same frame.
     /// </summary>
-    [RemoveAtEndOfTick]
-    public struct NewlyAddedSpatialOSEntity : IComponentData
+    public struct NewlyAddedSpatialOSEntity : IComponentData, RemoveAtEndOfTick
     {
+        public void RemoveComponent(EntityCommandBuffer commands, Entity entity)
+        {
+            commands.RemoveComponent<NewlyAddedSpatialOSEntity>(entity);
+        }
```
gkassabli: This doesn't look very nice, and the implementation is effectively boilerplate with component type inserted.
Unfortunately, as EntityCommandBuffer RemoveComponent method is generic, we have to explicitly declare it somewhere for the AOT compiler.
This approach allows us to keep compile-time safety, where compiler will complain if RemoveComponent isn't implemented
I am open to other suggestions!

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -65,29 +57,31 @@ private void GenerateComponentGroups()
...
-            var componentGroup = GetComponentGroup(ComponentType.ReadOnly<T>());
+            var componentGroup = GetComponentGroup(type);
             removeComponentActions.Add(() =>
             {
                 if (componentGroup.IsEmptyIgnoreFilter)
                 {
                     return;
                 }
 
+                RemoveAtEndOfTick typeInstance = (RemoveAtEndOfTick)Activator.CreateInstance(type.GetManagedType());
```
gkassabli: Would be nice if we don't have to instantiate a component just to call RemoveComponent, but as a c# noob, I am not sure how we can force declaring static method when inheriting from an interface

workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTick.cs
```diff
@@ -4,11 +4,13 @@
...
+    ///     by the CleanReactiveComponentSystem.
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
-    [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false, Inherited = false)]
-    public class RemoveAtEndOfTick : Attribute
+
+    public interface RemoveAtEndOfTick
```
samiwh: nitpick - the naming convention is IRemoveAtEndOfTick. But if it's an interface instead of an attribute we might want to rename it.

workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
```diff
@@ -7,8 +7,11 @@ namespace Improbable.Gdk.Core
     ///     This component is automatically added to an entity upon its creation and automatically removed at the end of the
     ///     same frame.
     /// </summary>
-    [RemoveAtEndOfTick]
-    public struct NewlyAddedSpatialOSEntity : IComponentData
+    public struct NewlyAddedSpatialOSEntity : IComponentData, RemoveAtEndOfTick
     {
+        public void RemoveComponent(EntityCommandBuffer commands, Entity entity)
```
samiwh: public override void

workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
samiwh: Agreed. It doesn't make me feel warm an fuzzy but it's the best solution I can think of that allows us to keep the feature. The question then is how much do we get from this feature and is the boiletplate too annoying. I think this might be fine.

Annoying C# doesn't let us inherit functionality in structs or we could have made the interface generic.

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
samiwh: This line should go outside of the lambda function. We don't want to be using activator in a tight loop.

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -65,29 +57,31 @@ private void GenerateComponentGroups()
...
+                    AddRemoveComponentAction(type);
                 }
             }
         }
 
-        private void AddRemoveComponentAction<T>()
+        private void AddRemoveComponentAction(ComponentType type)
         {
-            var componentGroup = GetComponentGroup(ComponentType.ReadOnly<T>());
+            var componentGroup = GetComponentGroup(type);
```
samiwh: We still want this to be readonly.

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -38,23 +38,15 @@ private void GenerateComponentGroups()
...
-                    if (type.GetCustomAttribute<RemoveAtEndOfTick>(false) == null)
-                    {
-                        continue;
-                    }
-
-                    if (!typeof(IComponentData).IsAssignableFrom(type)
-                        && !typeof(ISharedComponentData).IsAssignableFrom(type))
+                    if (!typeof(RemoveAtEndOfTick).IsAssignableFrom(type) ||
+                        (!typeof(IComponentData).IsAssignableFrom(type)
+                         && !typeof(ISharedComponentData).IsAssignableFrom(type)))
```
samiwh: also need to check for abstract

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
samiwh: However I think this will also cause a boxing allocation, which is further reason against this in total.

workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
samiwh: Mentioned further down - I'd forgotten that this will create boxing allocations on each access, which might sink this idea.

samiwh: Actually I'm wrong we can still use this but it will require some funky reflection code. If we make a static function then we can find that via reflection when we find the types in the cleanup system. We can store each of those in an action and call them at runtime. Not exactly pretty but it would work. That doesn't allocate on removal but it does place a larger burden on creating temporary components (although you can give good error messages as we can diagnose the missing function perfectly).

workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTick.cs
samiwh: Why split over two lines?
Also grammar point - "Any component which inherits this interface"

workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
samiwh: Another possibility is users registering an action and type in the clean system at startup.

samiwh: Another option is we don't inject anything and we only use injections to get a list of entity then use the manger directly.

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
jamiebrynes7: You can't have a static method on an interface 😢 Its something I've lamented about in the past.

workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
gkassabli: Apparently, if you implement interface method, you can't mark it as virtual

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -65,27 +58,26 @@ private void GenerateComponentGroups()
...
+                componentGroups.Add(GetComponentGroup(ComponentType.ReadOnly(type)));
+            }
 
+            foreach (var componentGroup in componentGroups)
+            {
                 var entityArray = componentGroup.GetEntityArray();
                 for (var i = 0; i < entityArray.Length; ++i)
                 {
-                    PostUpdateCommands.RemoveComponent<T>(entityArray[i]);
+                    EntityManager.RemoveComponent(entityArray[i], componentGroup.Types[0]);
```
gkassabli: I am curious how viable is this approach?
As this is running in InternalSpatialOSCleanGroup, this system is going to be in the last group running.
Will extracting all component groups and removing them after that cause cache invalidation after all operations are completed?
Or do we need to separate removal to another system and inject components that needs removal there?

jamiebrynes7: Ah yes - this will invalidate the injected data.

gkassabli: I am not very familiar with the internals - what is the good approach that will allow us not to invalidate it

jamiebrynes7: Using the `PostUpdateCommands` is the canonical way - but the method you want is inexplicably not on it. You can copy the `EntityArray` to a list and directly use the `EntityManager`

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -65,27 +58,27 @@ private void GenerateComponentGroups()
...
                 {
-                    PostUpdateCommands.RemoveComponent<T>(entityArray[i]);
+                    componentToRemove.Add(Tuple.Create<Entity, Type>(entityArray[i], type));
                 }
-            });
+            }
+
+            foreach (Tuple<Entity, Type> entity in componentToRemove)
+            {
+                EntityManager.RemoveComponent(entity.Item1, entity.Item2);
```
gkassabli: If I understood correctly, ComponentGroup is invalidated when components is removed.
This code grabs all the entities to the list first while all componentGroups are still valid and removes them one by one afterwards

jamiebrynes7: Yup that _should_ work as far as I understand it.

You can use C#7 tuples I think to cut down the verbosity: 
`var componentToRemove = new List<(Entity, Type)>();` 
and
`componentToRemove.Add((entityArray[i], type));` 

Will need to check the performance of this code however, cc @samiwh

gkassabli: Checked this in profiler and it looks like CleanReactiveComponentSystem roughly takes the same time 2-3ms

workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTick.cs
```diff
@@ -4,11 +4,12 @@
...
 {
     /// <summary>
-    ///     Any component with this attribute will be removed from all entities by the CleanReactiveComponentSystem
+    ///     Any component which inherits this interface will be removed from all entities by the CleanReactiveComponentSystem.
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
-    [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false, Inherited = false)]
-    public class RemoveAtEndOfTick : Attribute
+
+    public interface IRemoveableComponent
```
jared-improbable: *Removable

workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
jared-improbable: Wasn't one of the points of this to remove the need for reflection? :)

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -97,11 +90,11 @@ protected override void OnUpdate()
             {
                 translationUnit.CleanUpComponents(ref commandBuffer);
             }
-
+
```
jared-improbable: Nit: blank spaces

workers/unity/ProjectSettings/UnityConnectSettings.asset
```diff
@@ -3,7 +3,7 @@
 --- !u!310 &1
 UnityConnectSettings:
   m_ObjectHideFlags: 0
-  m_Enabled: 0
+  m_Enabled: 1
```
jared-improbable: Did you mean to commit this change?

workers/unity/Packages/com.improbable.gdk.core/Components/NewlyAddedSpatialOSEntity.cs
gkassabli: Yeah, this is no longer needed, I'll remove it

workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
```diff
@@ -8,7 +8,7 @@ namespace Improbable.Gdk.Core
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
     [AttributeUsage(AttributeTargets.Struct)]
-    public class RemoveAtEndOfTickAttribute : Attribute
+    public class RemoveAtEndOfTick : Attribute
```
jessicafalk: the naming convention afaik is to add `Attribute` to the end of a class inheriting from `Attribute`

PR #150, opened 2018-07-27
C# 7 refactor + Logging

zeroZshadow:
#### Description
Refactor code to use the new C# 7 functionality unity provides
Small refactor on Logging to allow de-duplication of logger name everywhere.

#### Primary reviewers
@jamiebrynes7 @martinzlocha

### DIFF
--- a/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
+++ b/workers/unity/Assets/Playground/Scripts/Bootstrap.cs
@@ -14,13 +14,10 @@ namespace Playground
     public class Bootstrap : MonoBehaviour
     {
         public GameObject Level;
-
-        private const int TargetFrameRate = -1; // Turns off VSync
-
         public const string LoggerName = "Bootstrap";
 
+        private const int TargetFrameRate = -1; // Turns off VSync
         private static readonly List<WorkerBase> Workers = new List<WorkerBase>();
-
         private static ConnectionConfig connectionConfig;
 
         public void Awake()
@@ -31,9 +28,9 @@ namespace Playground
             PlayerLoopManager.RegisterDomainUnload(DomainUnloadShutdown, 10000); // Clean up worlds and player loop
 
             Application.targetFrameRate = TargetFrameRate;
+#if UNITY_EDITOR
             if (Application.isEditor)
             {
-#if UNITY_EDITOR
                 var workerConfigurations =
                     AssetDatabase.LoadAssetAtPath<ScriptableWorkerConfiguration>(ScriptableWorkerConfiguration
                         .AssetPath);
@@ -49,11 +46,13 @@ namespace Playground
                     Workers.Add(worker);
                 }
 
-                connectionConfig = new ReceptionistConfig();
-                connectionConfig.UseExternalIp = workerConfigurations.UseExternalIp;
-#endif
+                connectionConfig = new ReceptionistConfig
+                {
+                    UseExternalIp = workerConfigurations.UseExternalIp
+                };
             }
             else
+#endif
             {
                 var commandLineArguments = System.Environment.GetCommandLineArgs();
                 Debug.LogFormat("Command line {0}", string.Join(" ", commandLineArguments.ToArray()));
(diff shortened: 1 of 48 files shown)


workers/unity/Packages/com.improbable.gdk.core/Logging/Logger.cs
```diff
@@ -0,0 +1,21 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public class Logger : ILogger
+    {
+        private readonly ILogDispatcher dispatcher;
```
samiwh: We are we adding our a wrapper around our own logger and using another similarly named interface?

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -33,13 +33,14 @@ private void GenerateComponentGroups()
...
+                foreach (var componentType in translationUnit.CleanUpComponentTypes)
                 {
-                    translationUnit.CleanUpComponentGroups.Add(GetComponentGroup(componentType));
+                    var componentGroup = GetComponentGroup(componentType);
+                    translationUnit.CleanUpComponentGroups.Add(componentGroup);
                 }
             }
 
-            MethodInfo addRemoveComponentActionMethod =
+            var addRemoveComponentActionMethod =
```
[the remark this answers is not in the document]
zeroZshadow: If it was a value that was used throughout a longer scope, I can see why you might not want to use `var`. But with a short scope like this, I see no reason for it to be explicitly typed?

zeroZshadow:
This refactor doesn't have to be merged in. Its something I did on Friday to enjoy the air con, as an experiment to see what the new C# 7 stuff could help us with.

workers/unity/Packages/com.improbable.gdk.core/Logging/Logger.cs
zeroZshadow: I added this as an attempt to get rid of the pattern where every log would have to pass in the class it was logging for. I'm open for suggested alternatives.

MPeti01: Why couldn't the wrapper also implement ILogDispatcher?

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -31,9 +28,9 @@ public void Awake()
             PlayerLoopManager.RegisterDomainUnload(DomainUnloadShutdown, 10000); // Clean up worlds and player loop
 
             Application.targetFrameRate = TargetFrameRate;
+#if UNITY_EDITOR
```
martinzlocha: Doesn't this have to be inside of the if statement? Wont the code be invalid if UNITY_EDITOR is false?

workers/unity/Packages/com.improbable.gdk.core/Logging/ILogger.cs
```diff
@@ -0,0 +1,9 @@
+using UnityEngine;
```
martinzlocha: Don't think we should have yet another logger interface. Instead we could consider making changes to ILogDispatcher

workers/unity/Packages/com.improbable.gdk.core/Logging/Logger.cs
zeroZshadow: Maybe the `ILogger` interface was a bit much, I could just make it a `Logger` directly? Its simply meant to be a tool to de-duplicate the class field passing.

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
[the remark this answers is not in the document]
jamiebrynes7: I'm fine with this being `var` personally. I think C# tooling is sufficiently good that a healthy use of `var` is fine.

workers/unity/Packages/manifest.json
```diff
@@ -5,10 +5,10 @@
     "com.improbable.gdk.playerlifecycle": "file:com.improbable.gdk.playerlifecycle",
     "com.improbable.gdk.testutils": "file:com.improbable.gdk.testutils",
     "com.improbable.gdk.transformsynchronization": "file:com.improbable.gdk.transformsynchronization",
-    "com.unity.burst": "0.2.4-preview.18",
+    "com.unity.burst": "0.2.4-preview.21",
```
jared-improbable: This could be split out into a separate PR.

zeroZshadow:
Will redo this in separate bits when back from vacation

PR #153, opened 2018-07-30
Move Monobehaviour features to the Core.Monobehaviours namespace

MPeti01:

-------

#### Description
Just moved everything involving Monobehaviours from the Improbable.Gdk.Core namespace to Improbable.Gdk.Core.Monobehaviours. Also replaced the placeholder ReaderWriterImpl with the actual type, IReaderInternal, and made a few classes internal so that no public method on a public class returned it.

#### Tests
Not relevant
#### Documentation
Not relevant
#### Primary reviewers
@JonasImprobable

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
@@ -2,17 +2,13 @@ using System.Collections.Generic;
 using Improbable.Worker;
 using UnityEngine;
 
-namespace Improbable.Gdk.Core
+namespace Improbable.Gdk.Core.MonoBehaviours
 {
-    public class ReaderWriterImpl
-    {
-        // This is placeholder code.
-    }
 
     /// <summary>
     ///     Invokes dispatcher callbacks on GameObjects that represent entities.
     /// </summary>
-    public class SpatialOSBehaviourManager
+    internal class SpatialOSBehaviourManager
     {
         public GameObject GameObject;
         private Dictionary<uint, List<MonoBehaviour>> componentIdToReaderSpatialOSBehaviours;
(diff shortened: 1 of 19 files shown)


MPeti01:
A problem came up: Currently, a reference to the EntityGameObjectLinker, of which there should be one per worker, is stored in the WorkerBase. This breaks the design of Monobehaviour support being a feature module. This could be solved by e.g. adding a static class similar to WorkerRegistry for mapping from workers or worlds to EntityGameObjectLinkers, but since it's not obvious and wasn't actually planned for this sprint, talking with @JonasImprobable we decided to put it on hold until it's time for cleanup / refactoring things like this.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -2,32 +2,28 @@
...
-    public class ReaderWriterImpl
-    {
-        // This is placeholder code.
-    }
 
     /// <summary>
     ///     Invokes dispatcher callbacks on GameObjects that represent entities.
     /// </summary>
-    public class SpatialOSBehaviourManager
+    internal class SpatialOSBehaviourManager
```
jared-improbable: Why are these changes mixed up with the namespace move? That should be its own PR.

MPeti01: Fair point. The original intention was to just remove the placeholder ReaderWriterImpl from SpatialOSBehaviourManager because that was also a 100% obvious refactor, but since the replacing IReaderInternal is internal, that cascaded into making a few additional things internal. Anyway that will probably end up being done in a PR about the SpatialOSBehaviourManager now that this one is postponed.

MPeti01:
Opening a new PR to actually do it

PR #156, opened 2018-07-30
UTY-473: Fix Dotnet Path for Mac TC Agents

jamiebrynes7:
#### Description
Paths are borked on the mac TC agents. We know exactly where dotnet is though.
#### Tests
Ran the CI build step. Dotnet actually ran but the build failed for other reasons (similar to Mac crashes I imagine)
#### Documentation
N/A
#### Primary reviewers

### DIFF
--- a/ci/includes/pinned-tools.sh
+++ b/ci/includes/pinned-tools.sh
@@ -23,6 +23,13 @@ function cleanUnity() {
 }
 
 
+# Ensure for the Mac TC agents that dotnet is on the path.
+if isMacOS; then
+  if ! which dotnet; then
+    export PATH="${PATH}:/usr/local/share/dotnet/"
+  fi
+fi
+
 # Print the .NETCore version to aid debugging,
 # as well as ensuring that later calls to the tool don't print the welcome message on first run.
 dotnet --version


ci/includes/pinned-tools.sh
```diff
@@ -25,11 +25,17 @@ function cleanUnity() {
 
 # Print the .NETCore version to aid debugging,
 # as well as ensuring that later calls to the tool don't print the welcome message on first run.
-dotnet --version
+if isMacOS; then
+  export DOTNET="/usr/local/share/dotnet/dotnet"
```
zeroZshadow: Wouldn't it be easier to add this to path on macos when including this file?

jared-improbable: Keep in mind this won't just be running on our CI.
How about isolating this by adding `/usr/local/share/dotnet` to PATH iff `which dotnet` returns non-zero?

zeroZshadow: The change in path would only persist during the run of the script doesnt it?

jared-improbable: Correct - my suggestion to avoid doing anything unless `dotnet` is completely unavailable on any existing PATH.

jamiebrynes7: Will make the change 👍

PR #158, opened 2018-07-31
Update setup-and-installing.md

oblm:
#### Description
List the required Unity components in setup-and-installing.md. This is part of [UTY-693](https://improbableio.atlassian.net/browse/UTY-693).
#### Tests
I installed these components and could cloud deploy.
I didn't install these components and I couldn't.
#### Documentation
This is documentation.

### DIFF
--- a/docs/setup-and-installing.md
+++ b/docs/setup-and-installing.md
@@ -7,8 +7,8 @@
 ## Short version (for SpatialOS local deployment only)
 
 1. Set up your machine by installing: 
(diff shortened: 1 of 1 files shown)


docs/setup-and-installing.md
```diff
@@ -29,7 +29,7 @@
 
     #### Windows
 
-	- [Unity 2018.2.0f2](https://unity3d.com/get-unity/download/archive)
+	- [Unity 2018.2.0f2](https://unity3d.com/get-unity/download/archive) with Linux Build Support, Mac Build Support (IL2CPP) and Windows Build Support (Mono) components selected during the installation process.
```
zeroZshadow: Windows doesn't actually have the option to install mac il2cpp, only just "mac" which is mono.
Similarly, Windows build support is not an available module as it is installed by default.

docs/setup-and-installing.md
```diff
@@ -46,7 +46,7 @@
 
     #### Mac
 
-	- [Unity 2018.2.0f2](https://unity3d.com/get-unity/download/archive)
+	- [Unity 2018.2.0f2](https://unity3d.com/get-unity/download/archive) with Linux Build Support, Mac Build Support (IL2CPP) and Windows Build Support (Mono) components selected during the installation process.
```
zeroZshadow: Mac does not need mac il2cpp, as it gets the mono variant installed by default.

PR #159, opened 2018-07-31
UTY-588: Reader/Writer injection logic (SpatialOSBehaviourManager)

MPeti01:

-------

#### Description
Implemented SpatialOSBehaviourManager, which keeps track of what Readers/Writers are available for the MonoBehaviours on a particular GameObject and when they are ready to be enabled. 

It knows what is needed from info given by the SpatialOSBehaviourLibrary, keeps track of components being added/removed and authorities changing to know when Readers and Writers are available, and when appropriate, calls into the library for injection/deinjection of Readers and Writers and enables/disabled the MonoBehaviours.

Also includes some code that glues it together with the GameObjectDispatchers calling it and the SpatialOSLibrary which it calls. Everything handles IReaderInternal objects instead of placeholder types (this will be implemented by all readers and writers). Since this is an internal type, a few classes had to be made internal since their method signatures included this. Also, the SpatialOSLibrary now returns the Readers/Writers it created in an injection call.

#### Tests
Since this class is closely coupled to the SpatialOSLibrary, most code paths cannot be tested by isolated unit tests unless we create interfaces (e.g. for the library in this case) just to be able to mock dependencies. This will be easier to test with some integration tests once we have a ReaderWriterFactory to actually create the Readers and Writers, because then we can test adding components to a GameObject and seeing that everything is injected/deinjected at the appropriate time.
#### Documentation
In-code.
#### Primary reviewers
@jamiebrynes7 @JonasImprobable


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -40,24 +41,29 @@ public SpatialOSBehaviourLibrary(ILogDispatcher logger)
...
-        public void InjectAllReadersWriters(MonoBehaviour spatialOSBehaviour)
+        public Dictionary<uint, IReaderInternal> InjectAllReadersWriters(MonoBehaviour spatialOSBehaviour, Entity entity)
         {
             var spatialOSBehaviourType = spatialOSBehaviour.GetType();
             EnsureLoaded(spatialOSBehaviourType);
+            var createdReaderWriters = new Dictionary<uint, IReaderInternal>();
             if (invalidMonoBehaviourTypes.Contains(spatialOSBehaviourType))
             {
-                return;
+                return null;
```
jamiebrynes7: Should this return an empty dictionary instead of null?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
+        private readonly Entity entity;
+        private readonly long spatialId;
+
+        private readonly Dictionary<uint, HashSet<MonoBehaviour>> componentIdToReaderOrWriterSpatialOSBehaviours
+            = new Dictionary<uint, HashSet<MonoBehaviour>>();
+
+        private readonly Dictionary<uint, HashSet<MonoBehaviour>> componentIdToWriterSpatialOSBehaviours
+            = new Dictionary<uint, HashSet<MonoBehaviour>>();
+
+        private readonly Dictionary<MonoBehaviour, int> numUnSatisfiedReadersWriters
```
jamiebrynes7: Nit: Should be `numUnsatisifedReaderWriters`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
 
-        public SpatialOSBehaviourManager(GameObject gameObject)
+        private readonly SpatialOSBehaviourLibrary behaviourLibrary;
+
+        private readonly ILogDispatcher logger;
+
+        private const string NotInterestedInComponent =
+            "SpatialOSBehaviourManager received update on component it is not interested in!";
+
+        private const string LoggerName = "GameObjectReference";
```
jamiebrynes7: Shouldn't the name be `SpatialOSBehaviourManager` ?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
+        public SpatialOSBehaviourManager(GameObject gameObject, SpatialOSBehaviourLibrary library, ILogDispatcher logger)
         {
-            // Setup caches
+            this.logger = logger;
+
+            behaviourLibrary = library;
+            var spatialComponent = gameObject.GetComponent<SpatialOSComponent>();
+            entity = spatialComponent.Entity;
+            spatialId = spatialComponent.SpatialEntityId;
+            foreach (var behaviour in gameObject.GetComponents<MonoBehaviour>())
```
jamiebrynes7: Super nit: Whitespace could be used to make this cleaner

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
jamiebrynes7: Given that we don't do any null checks on the return type when its used - my guess would be yes?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
+
+                var writerIds = library.GetRequiredWriterComponentIds(behaviour.GetType());
+                foreach (var id in writerIds)
+                {
+                    if (!componentIdToWriterSpatialOSBehaviours.ContainsKey(id))
+                    {
+                        componentIdToWriterSpatialOSBehaviours.Add(id, new HashSet<MonoBehaviour>());
+                    }
+
+                    componentIdToWriterSpatialOSBehaviours[id].Add(behaviour);
```
jamiebrynes7: Do we need to add it to the `componentIdToReaderOrWriterSpatialOSBehaviours` dictionary too?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
+            }
+
+            spatialOSBehavioursToDisable.Clear();
         }
 
         public void AddComponent(uint componentId)
         {
-            // Mark readers as to be enabled
+            var writerWanted = componentIdToWriterSpatialOSBehaviours.ContainsKey(componentId);
+            var readerWanted = !writerWanted && componentIdToReaderOrWriterSpatialOSBehaviours.ContainsKey(componentId);
```
jamiebrynes7: Nit: `isWriterWanted` and `isReaderWanted`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
+        private readonly Dictionary<uint, HashSet<IReaderInternal>> componentIdToCreatedReadersWriters =
+            new Dictionary<uint, HashSet<IReaderInternal>>();
 
-        public SpatialOSBehaviourManager(GameObject gameObject)
+        private readonly SpatialOSBehaviourLibrary behaviourLibrary;
+
+        private readonly ILogDispatcher logger;
+
+        private const string NotInterestedInComponent =
+            "SpatialOSBehaviourManager received update on component it is not interested in!";
```
jamiebrynes7: I'd be careful about talking about being interested in a component as that's overloading what we use to describe workers being interested in a component.

I'd probably rephrase this to use different terminology

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
+            {
+                MarkComponentReady(spatialOSBehaviour);
+            }
         }
 
         public void RemoveComponent(uint componentId)
         {
-            // Mark readers as to be disabled
+            var writerWanted = componentIdToWriterSpatialOSBehaviours.ContainsKey(componentId);
+            var readerWanted = !writerWanted && componentIdToReaderOrWriterSpatialOSBehaviours.ContainsKey(componentId);
```
jamiebrynes7: Similar comment as above

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
+                MarkComponentNotReady(spatialOSBehaviour);
+            }
         }
 
         public void ChangeAuthority(uint componentId, Authority authority)
         {
-            // Mark writers as to be enabled/disabled
+            var writerWanted = componentIdToWriterSpatialOSBehaviours.ContainsKey(componentId);
+            var readerWanted = !writerWanted && componentIdToReaderOrWriterSpatialOSBehaviours.ContainsKey(componentId);
+            if (!readerWanted && !writerWanted)
```
jamiebrynes7: Similar comment as above

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
+                {
+                    MarkComponentNotReady(spatialOSBehaviour);
+                }
+            }
+        }
+
+        private void MarkComponentReady(MonoBehaviour spatialOSBehaviour)
+        {
+            // Inject all Readers/Writers at once when all requirements are met
+            if (--numUnSatisfiedReadersWriters[spatialOSBehaviour] == 0)
```
jamiebrynes7: I'm not sure that this will set the value back - also from a legibility POV it would be nice to split this into multiple lines

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,238 @@
...
+                {
+                    spatialOSBehavioursToDisable.Remove(spatialOSBehaviour);
+                }
+            }
+        }
+
+        private void MarkComponentNotReady(MonoBehaviour spatialOSBehaviour)
+        {
+            // De-inject all Readers/Writers at once when a single requirement is not met
+            if (numUnSatisfiedReadersWriters[spatialOSBehaviour]++ == 0)
```
jamiebrynes7: Similar comment to above.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: Good point, will change

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: It has a slightly different meaning to me, though neither is really incorrect. I read "ReadersWriters" as "readers or writers", referring to the things the user is actually asking for, while a "ReaderWriters" would be talking about the implementing "ReaderWriter" objects that implement the reader and writer interfaces. Since the class is about handling what the user is asking for, I'd keep it the way it is. Or should I explicitly make it ReadersOrWriters?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: Correct, missed a rename there

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: Actually I have ReaderOrWriter in another name anyway so I'll rename it to numUnsatisfiedReadersOrWriters. (Also just noticed it's capitalized as UnSatisfied atm, I might as well change that too since it's one word...)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: I'll check the functionality (I'm pretty sure it should work though, given that it compiled at all) and split it

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
jamiebrynes7: Yeah I was referring to the `Unsatisfied` bit - didn't notice Reader vs Readers

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
jamiebrynes7: Yeah it will compile - but as far as I understand, it will read the value in the dictionary, subtract, and do the comparison. 

But it won't set the new value back in the dictionary - meaning that if there are 2 reader or writers that condition can never be true 😉

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: Yes it should have been like that, but after looking at it more I'm actually very confused about why I even wanted to iterate over that set of behaviours at any point. The only place I'm doing that is when I add/remove components, but that is incorrect since writers should be marked as ready based on authority change, not adding/removing components... So I'll change this by making it just store readers in that map (also makes isReaderWanted checks cleaner).

MPeti01:
While making the requested changes I realised that the class _will_ be called for components it is not interested in as normal operation, so I removed the error checking. (And removed the logger since it's not used anymore.)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,202 @@
...
+
+            spatialOSBehavioursToEnable.Clear();
         }
 
         public void DisableSpatialOSBehaviours()
         {
-            // disable + deinject all spatialOSBehavioursToDisable
+            foreach (var behaviour in spatialOSBehavioursToDisable)
+            {
+                behaviour.enabled = false;
```
jared-improbable: User `OnDisable` code could throw an exception - double-check that it's not propagated up to this level or you'll need to protect against it.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,202 @@
...
+                    var reader = idToReaderWriter.Value;
+                    if (!componentIdToCreatedReadersWriters.ContainsKey(id))
+                    {
+                        componentIdToCreatedReadersWriters[id] = new HashSet<IReaderInternal>();
+                    }
+
+                    componentIdToCreatedReadersWriters[id].Add(reader);
+                }
+
+                behaviour.enabled = true;
```
jared-improbable: User `OnEnable` code could throw an exception - double-check that it's not propagated up to this level or you'll need to protect against it.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,202 @@
...
         }
 
         public void AddComponent(uint componentId)
         {
-            // Mark readers as to be enabled
+            // Mark reader components ready in relevant SpatialOSBehaviours
+            var relevantReaderSpatialOSBehaviours = componentIdToReaderSpatialOSBehaviours[componentId];
+            foreach (var spatialOSBehaviour in relevantReaderSpatialOSBehaviours)
+            {
+                MarkComponentReady(spatialOSBehaviour);
```
jared-improbable: Why not refactor `MarkComponentReady` and `MarkComponentNotReady` to accept an `IEnumerable<>` so the loops happen in the methods rather than at the call site?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,202 @@
...
-        private Dictionary<uint, List<MonoBehaviour>> componentIdToReaderSpatialOSBehaviours;
-        private Dictionary<uint, List<MonoBehaviour>> componentIdToWriterSpatialOSBehaviours;
-        private Dictionary<MonoBehaviour, int> numUnsatisfiedComponents;
-        private List<MonoBehaviour> spatialOSBehavioursToEnable;
-        private List<MonoBehaviour> spatialOSBehavioursToDisable;
-        private Dictionary<uint, Dictionary<MonoBehaviour, ReaderWriterImpl>> readersWriters;
+        private readonly Entity entity;
+        private readonly long spatialId;
+
+        private readonly Dictionary<uint, HashSet<MonoBehaviour>> componentIdToReaderSpatialOSBehaviours
```
jared-improbable: These variables are quite verbose - my eyes glaze over every time I read them. Is there some abbreviation that could be applied here and below?
(Also, should be `behavior`)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
```
jared-improbable: Should be `Behavior`, we use US spellings. 
cc @ElleEss

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,202 @@
...
+            = new Dictionary<MonoBehaviour, int>();
+
+        private readonly Dictionary<MonoBehaviour, Dictionary<uint, IReaderInternal>> behaviourToCreatedReadersWriters
+            = new Dictionary<MonoBehaviour, Dictionary<uint, IReaderInternal>>();
+
+        private readonly HashSet<MonoBehaviour> spatialOSBehavioursToEnable = new HashSet<MonoBehaviour>();
+        private readonly HashSet<MonoBehaviour> spatialOSBehavioursToDisable = new HashSet<MonoBehaviour>();
 
-        public SpatialOSBehaviourManager(GameObject gameObject)
+        private readonly Dictionary<uint, HashSet<IReaderInternal>> componentIdToCreatedReadersWriters =
```
jared-improbable: There's a lot of bookkeeping and cross-referencing going on here. Consider refactoring some of this into structs to reduce to number of raw mappings that you have here.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,202 @@
...
+                    componentIdToWriterSpatialOSBehaviours[id].Add(behaviour);
+                }
+
+                numUnsatisfiedReadersOrWriters[behaviour] = readerIds.Count + writerIds.Count;
+            }
         }
 
-        public List<ReaderWriterImpl> GetReadersWriters(uint componentId)
+
+        public HashSet<IReaderInternal> GetReadersWriters(uint componentId)
```
jared-improbable: What does this function add over just doing the lookup directly?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,202 @@
...
+                // Mark writer components not ready in relevant SpatialOSBehaviours
+                var relevantWriterSpatialOSBehaviours = componentIdToWriterSpatialOSBehaviours[componentId];
+                foreach (var spatialOSBehaviour in relevantWriterSpatialOSBehaviours)
+                {
+                    MarkComponentNotReady(spatialOSBehaviour);
+                }
+            }
+        }
+
+        private void MarkComponentReady(MonoBehaviour spatialOSBehaviour)
```
jared-improbable: This could just be shorted to `behavior` or even `b` and it'd be more readable and not lose any meaning.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: I'll change them, in some places it's definitely a tradeoff of verbose and expressive vs short and unclear, though. Check back to see how you like whatever I come up with

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: The name derives from Unity's MonoBehaviours though, so an exception seems sensible in this case?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: This is just to have a public interface rather than being tied to a particular way of storing internally

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -40,24 +41,29 @@ public SpatialOSBehaviourLibrary(ILogDispatcher logger)
...
             }
 
             foreach (var readerWriterComponentId in componentReaderIdsForBehaviours[spatialOSBehaviourType])
             {
-                Inject(spatialOSBehaviour, readerWriterComponentId);
+                var readerWriter = Inject(spatialOSBehaviour, readerWriterComponentId, entity);
+                createdReaderWriters[readerWriterComponentId] = readerWriter;
             }
 
             foreach (var readerWriterComponentId in componentWriterIdsForBehaviours[spatialOSBehaviourType])
```
zeroZshadow: What happens here if you require both a reader and a writer of the same component?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,202 @@
...
+        private readonly HashSet<MonoBehaviour> spatialOSBehavioursToEnable = new HashSet<MonoBehaviour>();
+        private readonly HashSet<MonoBehaviour> spatialOSBehavioursToDisable = new HashSet<MonoBehaviour>();
 
-        public SpatialOSBehaviourManager(GameObject gameObject)
+        private readonly Dictionary<uint, HashSet<IReaderInternal>> componentIdToCreatedReadersWriters =
+            new Dictionary<uint, HashSet<IReaderInternal>>();
+
+        private readonly SpatialOSBehaviourLibrary behaviourLibrary;
+
+        public SpatialOSBehaviourManager(GameObject gameObject, SpatialOSBehaviourLibrary library)
```
zeroZshadow: If this takes `SpatialOSComponent` instead of `GameObject` then you ensure only spatialos gameobjects can be used for this.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
samiwh: I think behaviour is the one English spelling that we use, because MonoBehaviour uses it :P

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
samiwh: I think behaviour is the one English spelling that we use, because MonoBehaviour uses it :P
If not in docs it would probably make sense to keep this consistent in code.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
zeroZshadow: Wasn't the deal that we inject on all behaviours first, and then enable them all ?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,202 @@
...
+                {
+                    MarkComponentNotReady(spatialOSBehaviour);
+                }
+            }
+        }
+
+        private void MarkComponentReady(MonoBehaviour spatialOSBehaviour)
+        {
+            // Inject all Readers/Writers at once when all requirements are met
+            numUnsatisfiedReadersOrWriters[spatialOSBehaviour]--;
```
zeroZshadow: Does this actually decrease the value stored?

MPeti01: Yes (see link on Jamie's comment above)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: This was also asked by Sami when the SpatialOSBehaviourLibrary was written. My understanding is that either no or not for this first iteration. @JonasImprobable ?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: That component is added to any GameObject added to an Entity by the EntityGameObjectLinker.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: That is invalid because you are meant to use Writers in place of Readers as well, so you get an error when you try to inject (see from line 155) and the MonoBehaviour will never be enabled. (This matches the SDK behavior btw.)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
zeroZshadow: I see no reason not too, its a tiny change to fix this.

MPeti01: I pinged @JonasImprobable on slack, waiting for reply before changing this but I agree it's easy to do.

MPeti01:
Feedback on less verbose names in last commit appreciated, might be less expressive now but hopefully actually readable. (@jared-improbable )

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,227 @@
...
-            // inject + enable all spatialOSBehavioursToEnable
+            foreach (var behaviour in behavioursToEnable)
+            {
+                var dict = behaviourLibrary.InjectAllReadersWriters(behaviour, entity);
+                behaviourToReadersWriters[behaviour] = dict;
+                foreach (var idToReaderWriter in dict)
+                {
+                    var id = idToReaderWriter.Key;
+                    var reader = idToReaderWriter.Value;
+                    if (!compIdToReadersWriters.ContainsKey(id))
```
jared-improbable: Use `TryGet` pattern here, too.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,227 @@
...
+                {
+                    behaviour.enabled = true;
+                }
+                catch (Exception e)
+                {
+                    logger.HandleLog(LogType.Error,
+                        new LogEvent("Exception thrown in OnEnable() method of MonoBehaviour.")
+                            .WithField(LoggingUtils.LoggerName, LoggerName)
+                            .WithField(LoggingUtils.EntityId, spatialId)
+                            .WithField("Exception", e.ToString()));
```
jared-improbable: Can you pass the full exception object to the logger, rather than the `ToString`?

MPeti01:
Exceptions now handled using the new exception logging method, which is being merged from master to the monobehaviour feature branch right now. So it doesn't compile on this PR branch, but it will once it's merged to the shared feature branch.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
jessicafalk: it's been a while since the last design meeting, so I might be wrong, but did we not decide to allow this behaviour?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -40,24 +41,29 @@ public SpatialOSBehaviourLibrary(ILogDispatcher logger)
...
-        public void InjectAllReadersWriters(MonoBehaviour spatialOSBehaviour)
+        public Dictionary<uint, IReaderInternal> InjectAllReadersWriters(MonoBehaviour spatialOSBehaviour, Entity entity)
         {
             var spatialOSBehaviourType = spatialOSBehaviour.GetType();
             EnsureLoaded(spatialOSBehaviourType);
+            var createdReaderWriters = new Dictionary<uint, IReaderInternal>();
             if (invalidMonoBehaviourTypes.Contains(spatialOSBehaviourType))
             {
-                return;
+                return createdReaderWriters;
```
jessicafalk: what exactly will be stored in `invalidMonoBehaviourTypes`? so far it doesn't seem to be used

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,228 @@
...
+                }
+
+                var writerIds = library.GetRequiredWriterComponentIds(behaviour.GetType());
+                foreach (var id in writerIds)
+                {
+                    if (!behavioursRequiringWriterTypes.TryGetValue(id, out var behavioursForWriter))
+                    {
+                        behavioursForWriter = new HashSet<MonoBehaviour>();
+                        behavioursRequiringWriterTypes[id] = behavioursForWriter;
+                    }
```
jamiebrynes7: This pattern shows up a few times - you can turn this into a generic helper method I think: 

```
        private TValue GetOrCreateValue<TKey, TValue>(Dictionary<TKey, TValue> dictionary, TKey key) where TValue : new()
        {
            if (!dictionary.TryGetValue(key, out var value))
            {
                value = new TValue();
                dictionary[key] = value;
            }

            return value;
        }
```

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,228 @@
...
+                if (numUnsatisfiedReadersOrWriters[behaviour] == 0)
+                {
+                    // Schedule deactivation
+                    if (behaviour.enabled)
+                    {
+                        behavioursToDisable.Add(behaviour);
+                    }
+                    else
+                    {
+                        behavioursToEnable.Remove(behaviour);
```
jamiebrynes7: Are we guaranteed that `behavioursToEnable` will contain `behaviour` ?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
MPeti01: It is used, types requiring multiple readers/writers (or both a reader and writer) of the same component get added when loaded (in `EnsureLoaded()`)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: The code assumes that no other system interferes with enabling/disabling these MonoBehaviours, so if something has all Readers/Writers ready but is not enabled, it must have been marked for enabling in this tick. That is something I am allowed to assume, right?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: Done

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: Turns out I completely misunderstood the question, and you're right, I will change this

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,221 @@
...
+        private readonly Dictionary<MonoBehaviour, int> numUnsatisfiedReadersOrWriters
+            = new Dictionary<MonoBehaviour, int>();
+
+        private readonly Dictionary<MonoBehaviour, Dictionary<uint, IReaderInternal>> behaviourToReadersWriters
+            = new Dictionary<MonoBehaviour, Dictionary<uint, IReaderInternal>>();
+
+        private readonly Dictionary<uint, HashSet<IReaderInternal>> compIdToReadersWriters =
+            new Dictionary<uint, HashSet<IReaderInternal>>();
+
+        private readonly HashSet<MonoBehaviour> behavioursToEnable = new HashSet<MonoBehaviour>();
```
JonasImprobable: Could this be a list so that iterating over them is cheaper?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,221 @@
...
+        private readonly HashSet<MonoBehaviour> behavioursToDisable = new HashSet<MonoBehaviour>();
 
-        public SpatialOSBehaviourManager(GameObject gameObject)
+        private readonly SpatialOSBehaviourLibrary behaviourLibrary;
+
+        private readonly ILogDispatcher logger;
+
+        private const string LoggerName = "SpatialOSBehaviourManager";
```
JonasImprobable: I'm personally in favour of getting the loggername programmatically, i.e. using `this.GetType().ToString()` or something similar rather than saving a constant.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,221 @@
...
         {
-            // inject + enable all spatialOSBehavioursToEnable
+            foreach (var behaviour in behavioursToEnable)
+            {
+                var dict = behaviourLibrary.InjectAllReadersWriters(behaviour, entity);
+                behaviourToReadersWriters[behaviour] = dict;
+                foreach (var idToReaderWriter in dict)
+                {
+                    var id = idToReaderWriter.Key;
+                    var reader = idToReaderWriter.Value;
```
JonasImprobable: Nit: Rename `reader` to `readerWriter`.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: It needs to be easy to remove an object from it as well in the case when one requirement is met in the tick, so a behaviour is marked to be enabled, then another requirement is lost, so it has to be unmarked.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,221 @@
...
+            }
+            else if (authority == Authority.NotAuthoritative)
+            {
+                // Mark writer components not ready in relevant SpatialOSBehaviours
+                var relevantWriterSpatialOSBehaviours = behavioursRequiringWriterTypes[componentId];
+                MarkComponentsNotReady(relevantWriterSpatialOSBehaviours);
+            }
+        }
+
+        private void MarkComponentsReady(IEnumerable<MonoBehaviour> behaviours)
```
JonasImprobable: The method name might be misleading as people might confuse "marking a requirement as satisfied" with "marking a monobehaviour to be enabled as all requirements are satisfied". How about renaming this method to `SatisfyReaderWriterRequirement()`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
JonasImprobable: Yes, fair point.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
JonasImprobable: Yeah, we backtracked on the design meeting decision here. The design meeting decision was indeed to allow for multiple readers/writers per component per monobehaviour. Sorry for the backtracking here. I didn't communicate that properly. I can't remember the specifics anymore but something related to not being able to clearly identify a reader/writer field in a behaviour when given a just component id made the original semantics really hard to implement. Now looking at the code again, I can't see/remember anymore what the problem was so I'm happy with us reverting back to the initial design meeting decision. At the same time, this is not a matter that I feel very strongly about so I'd be fine with either outcome (no restrictions vs. at most one reader/writer per monobehaviour).

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,221 @@
...
+            // Inject all Readers/Writers at once when all requirements are met
+            foreach (var behaviour in behaviours)
+            {
+                numUnsatisfiedReadersOrWriters[behaviour]--;
+                if (numUnsatisfiedReadersOrWriters[behaviour] == 0)
+                {
+                    // Schedule activation
+                    if (!behaviour.enabled)
+                    {
+                        behavioursToEnable.Add(behaviour);
```
JonasImprobable: Discussed offline: This behaviour looks suspicious. Let's check for whether a behaviour is scheduled to be enabled rather than whether it's already enabled.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
MPeti01: That would be nice, but I don't see a way to do that without making it a non-static field, and I'd rather not store additional data on each object without a good reason.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourLibrary.cs
```diff
@@ -9,19 +11,17 @@ namespace Improbable.Gdk.Core
...
+        private readonly Dictionary<Type, Dictionary<uint, FieldInfo[]>> fieldInfoCache
+            = new Dictionary<Type, Dictionary<uint, FieldInfo[]>>();
 
         private readonly Dictionary<Type, List<uint>> componentReaderIdsForBehaviours =
             new Dictionary<Type, List<uint>>();
 
         private readonly Dictionary<Type, List<uint>> componentWriterIdsForBehaviours =
             new Dictionary<Type, List<uint>>();
 
-        private readonly HashSet<Type> invalidMonoBehaviourTypes = new HashSet<Type>();
```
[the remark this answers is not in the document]
MPeti01: We had previously decided that in that case the invalid required type is simply ignored, logging an error, without affecting anything. (This could easily be changed of course, but seems reasonable to me.)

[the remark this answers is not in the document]
JonasImprobable: Ok nice, sounds good. :)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
JonasImprobable: My idea was that rather than declaring a const string here, we can just get the loggername every time we want to send a log, i.e. ```
logger.HandleLog(LogType.Error, new LogEvent(MultipleReadersWritersRequiredError).WithField(LoggingUtils.LoggerName, GetType());
```

File: workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs [hunk 34]
```diff
@@ -1,60 +1,214 @@
...
+                {
+                    if (!behavioursToEnable.Contains(behaviour))
+                    {
+                        // Schedule activation
+                        behavioursToEnable.Add(behaviour);
+                    }
+                    else
+                    {
+                        // Must be enabled already, so we were going to disable it - let's not
+                        behavioursToDisable.Remove(behaviour);
```
JonasImprobable: Are you sure the code here is correct? You are saying "If the behaviour was previously already scheduled to be enabled, we need to remove it from the list of behaviours to be disabled". Shouldn't this rather be:
```
if (numUnsatisfiedReadersOrWriters[behaviour] == 0)
{
    if (!behavioursToEnable.Contains(behaviour))
    {
        behavioursToEnable.Add(behaviour);
    }
    else
    {
        LogError("THis shouldn't happen");
    }

    if (!behavioursToDisable.Contains(behaviour))
    {
        behavioursToDisable.Remove(behaviour);
    }
}
```

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSBehaviourManager.cs
```diff
@@ -1,60 +1,214 @@
...
-            // Mark readers as to be disabled
+            // Mark reader components not ready in relevant SpatialOSBehaviours
+            var relevantReaderSpatialOSBehaviours = behavioursRequiringReaderTypes[componentId];
+            MarkComponentRequirementUnsatisfied(relevantReaderSpatialOSBehaviours);
         }
 
         public void ChangeAuthority(uint componentId, Authority authority)
         {
-            // Mark writers as to be enabled/disabled
+            if (authority == Authority.Authoritative)
```
JonasImprobable: What about the third case when you transition from authoritative to authoritylossimminent?

MPeti01: Discussed offline, this is OK since it's for lifecycle, not user callbacks.

PR #174, opened 2018-08-03
Removing functionality from the Mutable View

jessicafalk:

-------

#### Description
The first of many PRs to get rid of the mutable view. This one simply removes all the wrapper methods that we have and don't provide any additional functionality. Any method that contains a bit more logic is still in there until I decide on how to exactly factor this out. Right now, we use the EntityManager from the view, this will also change in one of the next PRs

#### Tests
Tested that local deployment still works.

#### Documentation
no docs 

#### Primary reviewers
@jamiebrynes7

### DIFF
--- a/workers/unity/Assets/Playground/Scripts/Player/PlayerCommandsSystem.cs
+++ b/workers/unity/Assets/Playground/Scripts/Player/PlayerCommandsSystem.cs
@@ -80,7 +80,7 @@ namespace Playground
             var playerId = playerData.SpatialEntity[0].EntityId;
 
             var component = rigidBody.gameObject.GetComponent<SpatialOSComponent>();
-            if (component != null && view.EntityManager.HasComponent(component.Entity, typeof(SpatialOSLaunchable)))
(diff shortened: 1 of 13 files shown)


workers/unity/Assets/Playground/Scripts/Player/PlayerCommandsSystem.cs
```diff
@@ -80,7 +80,7 @@ protected override void OnUpdate()
             var playerId = playerData.SpatialEntity[0].EntityId;
 
             var component = rigidBody.gameObject.GetComponent<SpatialOSComponent>();
-            if (component != null && view.HasComponent(component.Entity, typeof(SpatialOSLaunchable)))
+            if (component != null && view.EntityManager.HasComponent(component.Entity, typeof(SpatialOSLaunchable)))
```
jamiebrynes7: This can just access the `EntityManager` on the system directly

jessicafalk:
just changed the base to an experimental branch to be able to create smaller PRs and play around with the stuff before actually merging the new worker abstraction in

PR #175, opened 2018-08-03
UTY-792 Implement writer stubs and replace IReaderInternal interface …

firtina-improbable:
…with IReaderWriterInternal tag

-------

#### Description
Since we have codegen we do not need the IReaderWriterInternal interface to contain anything,
#### Tests
Existing tests succeed except SpatialOSBehaviourLibrary_injects_two_Writers
This test will pass using the stubs defined in this PR.
#### Documentation
Stub-only, actual implementation will have documentation
#### Primary reviewers
@MPeti01 @mattyoung-improbable


code_generator/GdkCodeGenerator/Templates/UnityReaderWriterGenerator.tt
```diff
@@ -23,44 +23,48 @@ namespace <#= qualifiedNamespace #>
...
+            public IReaderWriterInternal CreateReaderWriter(Entity entity, EntityManager entityManager, ILogDispatcher logDispatcher)
             {
                 return new ReaderWriterImpl(entity, entityManager, logDispatcher);
             }
         }
 
         [ReaderInterface]
         [ComponentId(<#= unityComponentDefinition.Id #>)]
-        public interface Reader
+        public interface Reader : IReader<<#= componentDetails.TypeName #><# if(fieldDetailsList.Count > 0){ #>, <#= componentDetails.TypeName #>.Update<# } #>>
```
jamiebrynes7: Fine for this PR - but I wonder if there's a more elegant way of writing this

code_generator/GdkCodeGenerator/Templates/UnityReaderWriterGenerator.tt
```diff
@@ -23,44 +23,48 @@ namespace <#= qualifiedNamespace #>
...
         {
         }
 
-        public class ReaderWriterImpl : Reader, IReaderInternal
+        [WriterInterface]
```
jamiebrynes7: Isn't this tag redundant? Given the interface name?

code_generator/GdkCodeGenerator/Templates/UnityReaderWriterGenerator.tt
```diff
@@ -23,44 +23,48 @@ namespace <#= qualifiedNamespace #>
...
-            
-            void IReaderInternal.OnEvent(int eventIndex)
+            public event GameObjectDelegates.AuthorityChanged AuthorityChanged
             {
-                throw new System.NotImplementedException();
+                add => throw new System.NotImplementedException();
+                remove => throw new System.NotImplementedException();
             }
+<# if (fieldDetailsList.Count > 0) #>
+<# { #>
```
jamiebrynes7: Put this in the same `<# #>` block as the if - it makes it marginally easier to read imo

workers/unity/Assets/Generated/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Improbable.Gdk.Core.EditmodeTests")]
```
jamiebrynes7: Hmmm - we should rethink our test organisation if `Generated` needs to be visible to `Core.EditmodeTests`

code_generator/GdkCodeGenerator/Templates/UnityReaderWriterGenerator.tt
MPeti01: The attribute marks it as a component-specific writer interface and is picked up by the SpatialOSLibrary, which identifies required readers and writers by looking for these tags via reflection.

jamiebrynes7: You can check whether the type implements the interface?

code_generator/GdkCodeGenerator/Templates/UnityReaderWriterGenerator.tt
firtina-improbable: I have an idea :)

workers/unity/Assets/Generated/AssemblyInfo.cs
firtina-improbable: I think this should be resolved when/if MonoBehaviour support is out of core.

Other options:
- creating another test assembly for "things that depend on generated code"
- generating code into the test assembly? 🤔 - this sounds slightly worse

code_generator/GdkCodeGenerator/Templates/UnityReaderWriterGenerator.tt
MPeti01: That could work too, but could get a bit unintuitive with Writers being Readers as well, and currently there is no general "Writer" superinterface. The decision was basically having an empty tag interface with slightly confusing details vs attribute tags, and the version with attribute tags was already written at that point :)

jamiebrynes7: It feels like a code smell to be - but we can revisit this after the MVP is in

workers/unity/Assets/Generated/AssemblyInfo.cs
jamiebrynes7: Long term - creating a complete test project for codegen is the most attractive option for me. Once FPL lands this could tie in quite nicely

MPeti01: It might make sense to have test components which have a good variety of features they use, then we could use those in tests instead of actual feature components.  It's also annoying atm that we don't even have any component in Core or the Playground with both events and commands on it atm, this could also help with that... (Though something like this might already exist in generated/gdk/tests in fact? I haven't looked much at those)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Readers/IReaderInternal.cs
```diff
@@ -6,11 +6,7 @@ namespace Improbable.Gdk.Core.MonoBehaviours
     ///     The interface that is used to signal a reader that a message has been received from the SpatialOS Dispatcher
     ///     about the component that the reader is connected to.
     /// </summary>
-    internal interface IReaderInternal
+    internal interface IReaderWriterInternal
```
MPeti01: The filename should change as well, and the in-code documentation above should be updated

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Readers/IReader.cs
```diff
@@ -7,15 +7,21 @@ namespace Improbable.Gdk.Core.MonoBehaviours
     ///     callbacks for updates, events and commands for component instances.
     /// </summary>
     /// <typeparam name="TSpatialComponentData">The data type for the SpatialOS component.</typeparam>
-    /// <typeparam name="TComponentUpdate">The update type for the SpatialOS component.</typeparam>
-    public interface IReader<TSpatialComponentData, TComponentUpdate>
+    public interface IReader<TSpatialComponentData>
```
MPeti01: I should've asked this in a previous PR, but is anything actually going to refer to the readers via this interface?

firtina-improbable: Yes, when the users use a `SomeComponent.Reader` interface they will be allowed to access only the methods declared here.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Readers/IReaderInternal.cs
firtina-improbable: good catch :)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Readers/IReader.cs
MPeti01: Ie they'll be referring to it as SomeComponent.Reader, not IReader<Stuff>. But that's fine, no need to redeclare similar interfaces when you can just put it in a generic supertype, and also lets users see the similarities/relations I guess.

PR #194, opened 2018-08-09
[Core C# SDK] New Worker Package

jamiebrynes7:
New package upgrade to fix a small issue with component loopback.


jamiebrynes7:
Just the components - laying the groundwork

PR #200, opened 2018-08-10
UTY 845 snapshot generator upgrade

zeroZshadow:

-------

#### Description
Upgrade snapshot generator to work with new EntityBuilder
Drive by fixes in code generation templates

#### Tests
None, I don't even know if it runs

#### Documentation
Added more comments, split out functionality in well named functions

### DIFF
--- a/workers/unity/Assets/Playground/Config/ArchetypeConfig.cs
+++ b/workers/unity/Assets/Playground/Config/ArchetypeConfig.cs
@@ -9,7 +9,6 @@ namespace Playground
     {
         public const string CharacterArchetype = "Character";
         public const string CubeArchetype = "Cube";
-        public const string PlayerCreator = "PlayerCreator";
 
         public static readonly Dictionary<string, Dictionary<string, ComponentType[]>>
             WorkerTypeToArchetypeNameToComponentTypes = new Dictionary<string, Dictionary<string, ComponentType[]>>
--- a/workers/unity/Assets/Playground/Config/ArchetypeConfig.cs
+++ b/workers/unity/Assets/Playground/Config/ArchetypeConfig.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using Improbable.Gdk.TransformSynchronization;
 using Unity.Entities;
@@ -8,6 +9,7 @@ namespace Playground
     {
         public const string CharacterArchetype = "Character";
         public const string CubeArchetype = "Cube";
+        public const string PlayerCreator = "PlayerCreator";
 
         public static readonly Dictionary<string, Dictionary<string, ComponentType[]>>
             WorkerTypeToArchetypeNameToComponentTypes = new Dictionary<string, Dictionary<string, ComponentType[]>>
--- a/workers/unity/Assets/Playground/Editor/SnapshotGenerator/SnapshotGenerator.cs
+++ b/workers/unity/Assets/Playground/Editor/SnapshotGenerator/SnapshotGenerator.cs
@@ -26,12 +26,12 @@ namespace Playground.Editor.SnapshotGenerator
             var snapshot = CreateSnapshot(arguments.NumberEntities);
 
             Debug.Log($"Writing snapshot to: {arguments.OutputPath}");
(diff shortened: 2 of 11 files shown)


workers/unity/Assets/Playground/Config/ArchetypeConfig.cs
```diff
@@ -8,6 +9,7 @@ public static class ArchetypeConfig
     {
         public const string CharacterArchetype = "Character";
         public const string CubeArchetype = "Cube";
+        public const string PlayerCreator = "PlayerCreator";
```
jamiebrynes7: This doesn't belong here I don't think.

There aren't any special components we put on the PlayerCreator

workers/unity/Assets/Playground/Editor/SnapshotGenerator/SnapshotGenerator.cs
```diff
@@ -16,112 +14,99 @@ internal static class SnapshotGenerator
...
 
         public static void Generate(Arguments arguments)
         {
             Debug.Log("Generating snapshot.");
-            var entities = CreateEntities(arguments.WorldLength, arguments.NumberEntities);
-            WriteSnapshot(entities, arguments.OutputPath);
+            var snapshot = CreateSnapshot(arguments.NumberEntities);
+
+            Debug.Log($"Writing snapshot to: {arguments.OutputPath}");
+            snapshot.Serialize(arguments.OutputPath);
```
jamiebrynes7: Instead of serialize - what about `WriteOut`? Given that strictly all the serialisation is done before that function call.

workers/unity/Packages/com.improbable.gdk.core/Utility/SnapshotBuilder.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System.Collections.Generic;
+using Improbable.Worker;
+using Improbable.Worker.Core;
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public class SnapshotBuilder
```
jamiebrynes7: Given this class doesn't actually follow the Builder pattern - can we rename this?

workers/unity/Packages/com.improbable.gdk.core/Utility/SnapshotBuilder.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        private readonly Dictionary<EntityId, Entity> entities = new Dictionary<EntityId, Entity>();
+
+        public int Count => entities.Count;
+
+        public void AddEntity(Entity entity)
```
jamiebrynes7: Maybe over-engineering this - but this probably wants an API for a specific entity with a specific entity id.

Given like a spawner entity where you expect it to always be id 1

workers/unity/Assets/Playground/Editor/SnapshotGenerator/SnapshotGenerator.cs
```diff
@@ -16,112 +14,99 @@ internal static class SnapshotGenerator
...
 
-            for (var i = -gridLength + 1; i <= gridLength - 1; i += 2)
+            for (var x = -gridLength + 1; x <= gridLength - 1; x += 2)
             {
-                for (var j = -gridLength + 1; j <= gridLength - 1; j += 2)
+                for (var z = -gridLength + 1; z <= gridLength - 1; z += 2)
                 {
-                    if (i == 0 && j == 0)
+                    // Leave the centre empty
+                    if (x == 0 && z == 0)
```
jamiebrynes7: What's the purpose of this given that we enforce the grid is even numbered.

jamiebrynes7: This condition can never be hit I think

workers/unity/Assets/Playground/Editor/SnapshotGenerator/SnapshotGenerator.cs
zeroZshadow: `WriteToDisk`

PR #202, opened 2018-08-10
UTY-793: Sending events

pmernyei:

-------

#### Description
Added support for sending events from Writers.

Also added tests for events - only for receiving events for now, because to test sending we would need to create a MutableView, a LoggingDispatcher, a Translation, an EventSender etc. and all of that is messy and subject to change anyway. Made a ticket to add unit tests for event sending once the Core rework is in, and another one to add event-based MonoBehaviour functionality to the Playground soon.

#### Tests
Tests included for receiving events. 
#### Documentation
Implementation work
#### Primary reviewers
@firtina-improbable @zeroZshadow

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/Readers/EventTests.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/Readers/EventTests.cs
@@ -1,8 +1,11 @@
 ﻿using System;
+using System.Text.RegularExpressions;
 using Generated.Improbable.Gdk.Tests.BlittableTypes;
 using Generated.Improbable.Gdk.Tests.ComponentsWithNoFields;
 using NUnit.Framework;
 using Unity.Entities;
+using UnityEngine;
+using UnityEngine.TestTools;
 using ComponentWithEvents = Generated.Improbable.Gdk.Tests.ComponentsWithNoFields.ComponentWithNoFieldsWithEvents;
 
 namespace Improbable.Gdk.Core.EditmodeTests.MonoBehaviours.Readers
@@ -10,20 +13,20 @@ namespace Improbable.Gdk.Core.EditmodeTests.MonoBehaviours.Readers
     [TestFixture]
     internal class EventTests
     {
-        protected ComponentWithEvents.Reader ReaderPublic;
-        protected ComponentWithEvents.ReaderWriterImpl ReaderWriterInternal;
-        protected EntityManager EntityManager;
-        protected Entity Entity;
+        private ComponentWithEvents.Reader readerPublic;
+        private ComponentWithEvents.ReaderWriterImpl readerWriterInternal;
+        private EntityManager entityManager;
+        private Entity entity;
         private World world;
 
         [SetUp]
         public void SetUp()
         {
             world = new World("test-world");
-            EntityManager = world.GetOrCreateManager<EntityManager>();
-            Entity = EntityManager.CreateEntity(typeof(SpatialOSBlittableComponent));
-            ReaderWriterInternal = new ComponentWithEvents.ReaderWriterImpl(Entity, EntityManager, new LoggingDispatcher());
-            ReaderPublic = ReaderWriterInternal;
+            entityManager = world.GetOrCreateManager<EntityManager>();
+            entity = entityManager.CreateEntity(ComponentType.Create<SpatialOSBlittableComponent>());
+            readerWriterInternal = new ComponentWithEvents.ReaderWriterImpl(entity, entityManager, new LoggingDispatcher());
+            readerPublic = readerWriterInternal;
         }
 
         [TearDown]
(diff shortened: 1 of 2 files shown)


workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/Readers/EventTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        protected EntityManager EntityManager;
+        protected Entity Entity;
+        private World world;
+
+        [SetUp]
+        public void SetUp()
+        {
+            world = new World("test-world");
+            EntityManager = world.GetOrCreateManager<EntityManager>();
+            Entity = EntityManager.CreateEntity(typeof(SpatialOSBlittableComponent));
```
zeroZshadow: use `ComponentType.Create<>` instead of `typeof`

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/Readers/EventTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        public void All_event_callbacks_invoked_if_one_throws_exceptiom()
+        {
+            bool firstCallbackInvoked = false;
+            ReaderPublic.OnEvt += (ev => firstCallbackInvoked = true);
+            ReaderWriterInternal.OnEvtEvent(new EvtEvent());
+            bool secondCallbackInvoked = false;
+            ReaderPublic.OnEvt += (ev =>
+            {
+                secondCallbackInvoked = true;
+                throw new Exception("no u");
```
zeroZshadow: make this a proper exception

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/Readers/EventTests.cs
```diff
@@ -0,0 +1,65 @@
...
+            ReaderWriterInternal.OnEvtEvent(new EvtEvent());
+            Assert.IsTrue(callbackInvoked);
+        }
+
+        [Test]
+        public void All_event_callbacks_invoked_if_one_throws_exceptiom()
+        {
+            bool firstCallbackInvoked = false;
+            ReaderPublic.OnEvt += (ev => firstCallbackInvoked = true);
+            ReaderWriterInternal.OnEvtEvent(new EvtEvent());
```
firtina-improbable: If you do `OnEvtEvent` here `firstCallbackInvoked` will be true and the end assertion will not be valid for this test case (similar with the other non-final `OnEvtEvent` calls

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/Readers/EventTests.cs
```diff
@@ -0,0 +1,65 @@
...
+            bool secondCallbackInvoked = false;
+            ReaderPublic.OnEvt += (ev =>
+            {
+                secondCallbackInvoked = true;
+                throw new Exception("no u");
+            });
+            ReaderWriterInternal.OnEvtEvent(new EvtEvent());
+            bool thirdCallbackInvoked = false;
+            ReaderPublic.OnEvt += (ev => thirdCallbackInvoked = true);
+            ReaderWriterInternal.OnEvtEvent(new EvtEvent());
```
firtina-improbable: Also need to assert that the exception is logged

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/Readers/EventTests.cs
samiwh: In create entity I think using types is fine. They do almost the same thing and the end result is the same.

zeroZshadow: I agree they do the same thing, But I think not using the implicit conversion is clearer, and allowed for readonly types. Just a good habit to get into?

samiwh: I think the implicit cast exists mostly for this case. Can you have readonly components in an archetype, as that is essentially static data (except for writing directly using the entity manager, which I guess you'd have to do as you don't have default values on an archetype)? Given that most ComponentType modifiers make no sense on an entity I'd be surprised if archetypes respected them, although a quick glance at the code doesn't reveal much.

If you can have readonly types and they are being used then I agree. Otherwise I don't think it matters which one you use, other than how easy it is to read.

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/Readers/EventTests.cs
samiwh: The side effects of logging an exception are logger dependent so you would also need to have a special test logger.

PR #210, opened 2018-08-14
UTY-934: Refactor SpatialOSBehaviourManager and -Library

MPeti01:

-------

#### Description
Refactor pass over the activation part of the MonoBehaviour workflow. Slight restructuring, renaming, fixing small issues.

Commits mostly logical pieces of work, except the first commit (split ReaderWriterStore out) should only be considered together with the later commit ("fixed ReaderWriterStore being completely broken"...).

Changes in summary:

- The SpatialOSBehaviourManager was doing two things: managing activation and storing the Readers/Writers. This was split in two with appropriate, meaningful names for each (ReaderWriterStore, MonoBehaviourActivationManager).
  - First there was one ReaderWriterStore created in total, and no per-entity bookkeeping, which clearly can't work since updates/etc should go to readers on that particular entity... This was fixed in the later commit.
- Renamed SpatialOSBehaviourLibrary to ReaderWriterInjector. This is the best we could think of with Firtina, but we're open for any better suggestions! (The class uses reflection to find fields with [Require] tags given a MonoBehaviour type, caching the info about the field, and performs the actual injection using that info.)
- Noticed that we were using a List to store ActivationManagers and removing from that list in O(N) time when an Entity is removed from the worker. Removed that list and instead iterating over entries in the Entity index -> ActivationManager map.

#### Tests
Unit tests still pass, Playground still works.
#### Documentation
N/A
#### Primary reviewers
@firtina-improbable @jamiebrynes7

### DIFF
--- a/code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
+++ b/code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
@@ -59,7 +59,7 @@ namespace <#= qualifiedNamespace #>
 
             private const uint componentId = <#= unityComponentDefinition.Id #>;
 
-            public override void InvokeOnAddComponentLifecycleMethods(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
+            public override void MarkComponentsAddedForActivation(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
                 var entities = ComponentAddedComponentGroup.GetEntityArray();
                 for (var i = 0; i < entities.Length; i++)
@@ -69,7 +69,7 @@ namespace <#= qualifiedNamespace #>
                 }
             }
 
-            public override void InvokeOnRemoveComponentLifecycleMethods(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
+            public override void MarkComponentsRemovedForDeactivation(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
                 var entities = ComponentRemovedComponentGroup.GetEntityArray();
                 for (var i = 0; i < entities.Length; i++)
@@ -79,7 +79,7 @@ namespace <#= qualifiedNamespace #>
                 }
             }
 
-            public override void InvokeOnAuthorityChangeLifecycleMethods(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
+            public override void MarkAuthorityChangesForActivation(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
                 var authoritiesChangedTags = AuthoritiesChangedComponentGroup.GetComponentArray<AuthoritiesChanged<<#= componentDetails.TypeName #>>>();
                 var entities = AuthoritiesChangedComponentGroup.GetEntityArray();
--- a/code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
+++ b/code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
@@ -62,16 +62,16 @@ namespace <#= qualifiedNamespace #>
                 return <#= unityComponentDefinition.Id #>;
             }
 
-            public override void InvokeOnAuthorityChangeLifecycleCallbacks(Dictionary<int, SpatialOSBehaviourManager> entityIndexToManagers)
+            public override void InvokeOnAuthorityChangeLifecycleCallbacks(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
                 var authoritiesChangedTags = AuthoritiesChangedComponentGroup.GetComponentArray<AuthoritiesChanged<<#= componentDetails.TypeName #>>>();
                 var entities = AuthoritiesChangedComponentGroup.GetEntityArray();
                 for (var i = 0; i < entities.Length; i++)
                 {
-                    var spatialOSBehaviourManager = entityIndexToManagers[entities[i].Index];
+                    var activationManager = entityIndexToManagers[entities[i].Index];
                     for (var j = 0; j < authoritiesChangedTags[i].Buffer.Count; j++)
                     {
-                        spatialOSBehaviourManager.ChangeAuthority(<#= unityComponentDefinition.Id #>, authoritiesChangedTags[i].Buffer[j]);
+                        activationManager.ChangeAuthority(<#= unityComponentDefinition.Id #>, authoritiesChangedTags[i].Buffer[j]);
                     }
                 }
             }
--- a/code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
+++ b/code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
@@ -57,12 +57,29 @@ namespace <#= qualifiedNamespace #>
 <# } #>
             };
 
-            protected override uint GetComponentId()
+            private const uint componentId = <#= unityComponentDefinition.Id #>;
+
+            public override void InvokeOnAddComponentLifecycleMethods(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
-                return <#= unityComponentDefinition.Id #>;
+                var entities = ComponentAddedComponentGroup.GetEntityArray();
+                for (var i = 0; i < entities.Length; i++)
+                {
+                    var activationManager = entityIndexToManagers[entities[i].Index];
+                    activationManager.AddComponent(componentId);
+                }
             }
 
-            public override void InvokeOnAuthorityChangeLifecycleCallbacks(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
+            public override void InvokeOnRemoveComponentLifecycleMethods(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
+            {
+                var entities = ComponentRemovedComponentGroup.GetEntityArray();
+                for (var i = 0; i < entities.Length; i++)
+                {
+                    var activationManager = entityIndexToManagers[entities[i].Index];
+                    activationManager.RemoveComponent(componentId);
+                }
+            }
+
+            public override void InvokeOnAuthorityChangeLifecycleMethods(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
                 var authoritiesChangedTags = AuthoritiesChangedComponentGroup.GetComponentArray<AuthoritiesChanged<<#= componentDetails.TypeName #>>>();
                 var entities = AuthoritiesChangedComponentGroup.GetEntityArray();
@@ -71,26 +88,26 @@ namespace <#= qualifiedNamespace #>
                     var activationManager = entityIndexToManagers[entities[i].Index];
                     for (var j = 0; j < authoritiesChangedTags[i].Buffer.Count; j++)
                     {
-                        activationManager.ChangeAuthority(<#= unityComponentDefinition.Id #>, authoritiesChangedTags[i].Buffer[j]);
+                        activationManager.ChangeAuthority(componentId, authoritiesChangedTags[i].Buffer[j]);
                     }
                 }
             }
 
-            public override void InvokeOnComponentUpdateUserCallbacks(Dictionary<int, ReaderWriterStore> readerWriterStores)
+            public override void InvokeOnComponentUpdateCallbacks(Dictionary<int, ReaderWriterStore> entityIdToReaderWriterStore)
             {
 <# if (fieldDetailsList.Count > 0) { #>
                 var entities = ComponentsUpdatedComponentGroup.GetEntityArray();
                 var updateLists = ComponentsUpdatedComponentGroup.GetComponentArray<ComponentsUpdated<<#= componentDetails.TypeName #>.Update>>();
                 for (var i = 0; i < entities.Length; i++)
                 {
-                    var readerWriterStore = readerWriterStores[entities[i].Index];
-                    if (!readerWriterStore.TryGetReaderWritersForComponent(<#= unityComponentDefinition.Id #>, out var readers))
+                    var readerWriterStore = entityIdToReaderWriterStore[entities[i].Index];
+                    if (!readerWriterStore.TryGetReaderWritersForComponent(componentId, out var readers))
                     {
                         continue;
                     }
 
                     var updateList = updateLists[i];
-                    foreach (var reader in readers.OfType<<#= componentDetails.ComponentName #>.ReaderWriterImpl>())
+                    foreach (ReaderWriterImpl reader in readers)
                     {
                         foreach (var update in updateList.Buffer)
                         {
@@ -101,7 +118,7 @@ namespace <#= qualifiedNamespace #>
 <# } #>
             }
 
-            public override void InvokeOnEventUserCallbacks(Dictionary<int, ReaderWriterStore> readerWriterStores)
+            public override void InvokeOnEventCallbacks(Dictionary<int, ReaderWriterStore> entityIdToReaderWriterStore)
             {
 <# for (var j = 0; j < eventDetailsList.Count; j++) { #>
                 {
@@ -109,15 +126,15 @@ namespace <#= qualifiedNamespace #>
                     var eventLists = EventsReceivedComponentGroups[<#= j #>].GetComponentArray<EventsReceived<<#= eventDetailsList[j].EventName #>Event>>();
                     for (var i = 0; i < entities.Length; i++)
                     {
-                        var readerWriterStore = readerWriterStores[entities[i].Index];
-                        if (!readerWriterStore.TryGetReaderWritersForComponent(<#= unityComponentDefinition.Id #>, out var readers))
+                        var readerWriterStore = entityIdToReaderWriterStore[entities[i].Index];
+                        if (!readerWriterStore.TryGetReaderWritersForComponent(componentId, out var readers))
                         {
                             continue;
                         }
 
                         var eventList = eventLists[i];
 
-                        foreach (var reader in readers.OfType<<#= componentDetails.ComponentName #>.ReaderWriterImpl>())
+                        foreach (ReaderWriterImpl reader in readers)
                         {
                             foreach (var e in eventList.Buffer)
                             {
@@ -129,7 +146,7 @@ namespace <#= qualifiedNamespace #>
 <# } #>
             }
 
-            public override void InvokeOnCommandRequestUserCallbacks(Dictionary<int, ReaderWriterStore> readerWriterStores)
+            public override void InvokeOnCommandRequestCallbacks(Dictionary<int, ReaderWriterStore> entityIdToReaderWriterStore)
             {
 <# for (var j = 0; j < commandDetailsList.Count; j++) { #>
                 {
@@ -137,14 +154,14 @@ namespace <#= qualifiedNamespace #>
                     var commandLists = CommandRequestsComponentGroups[<#= j #>].GetComponentArray<CommandRequests<<#= commandDetailsList[j].CommandName #>.Request>>();
                     for (var i = 0; i < entities.Length; i++)
                     {
-                        var readerWriterStore = readerWriterStores[entities[i].Index];
-                        if (!readerWriterStore.TryGetReaderWritersForComponent(<#= unityComponentDefinition.Id #>, out var readers))
+                        var readerWriterStore = entityIdToReaderWriterStore[entities[i].Index];
+                        if (!readerWriterStore.TryGetReaderWritersForComponent(componentId, out var readers))
                         {
                             continue;
                         }
 
                         var commandList = commandLists[i];
-                        foreach (var reader in readers.OfType<<#= componentDetails.ComponentName #>.ReaderWriterImpl>())
+                        foreach (ReaderWriterImpl reader in readers)
                         {
                             foreach (var req in commandList.Buffer)
                             {
@@ -156,20 +173,20 @@ namespace <#= qualifiedNamespace #>
 <# } #>
             }
 
-            public override void InvokeOnAuthorityChangeUserCallbacks(Dictionary<int, ReaderWriterStore> readerWriterStores)
+            public override void InvokeOnAuthorityChangeCallbacks(Dictionary<int, ReaderWriterStore> entityIdToReaderWriterStore)
             {
                 var entities = AuthoritiesChangedComponentGroup.GetEntityArray();
                 var authChangeLists = AuthoritiesChangedComponentGroup.GetComponentArray<AuthoritiesChanged<<#= componentDetails.TypeName #>>>();
                 for (var i = 0; i < entities.Length; i++)
                 {
-                    var readerWriterStore = readerWriterStores[entities[i].Index];
-                    if (!readerWriterStore.TryGetReaderWritersForComponent(<#= unityComponentDefinition.Id #>, out var readers))
+                    var readerWriterStore = entityIdToReaderWriterStore[entities[i].Index];
+                    if (!readerWriterStore.TryGetReaderWritersForComponent(componentId, out var readers))
                     {
                         continue;
                     }
 
                     var authChanges = authChangeLists[i];
-                    foreach (var reader in readers.OfType<<#= componentDetails.ComponentName #>.ReaderWriterImpl>())
+                    foreach (ReaderWriterImpl reader in readers)
                     {
                         foreach (var auth in authChanges.Buffer)
                         {
--- a/code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
+++ b/code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
@@ -61,6 +61,11 @@ namespace <#= qualifiedNamespace #>
 
             public override void MarkComponentsAddedForActivation(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
+                if (ComponentAddedComponentGroup.IsEmptyIgnoreFilter)
+                {
+                    return;
+                }
+
                 var entities = ComponentAddedComponentGroup.GetEntityArray();
                 for (var i = 0; i < entities.Length; i++)
                 {
@@ -71,6 +76,11 @@ namespace <#= qualifiedNamespace #>
 
             public override void MarkComponentsRemovedForDeactivation(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
+                if (ComponentRemovedComponentGroup.IsEmptyIgnoreFilter)
+                {
+                    return;
+                }
+
                 var entities = ComponentRemovedComponentGroup.GetEntityArray();
                 for (var i = 0; i < entities.Length; i++)
                 {
@@ -81,6 +91,11 @@ namespace <#= qualifiedNamespace #>
 
             public override void MarkAuthorityChangesForActivation(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
+                if (AuthoritiesChangedComponentGroup.IsEmptyIgnoreFilter)
+                {
+                    return;
+                }
+
                 var authoritiesChangedTags = AuthoritiesChangedComponentGroup.GetComponentArray<AuthoritiesChanged<<#= componentDetails.TypeName #>>>();
                 var entities = AuthoritiesChangedComponentGroup.GetEntityArray();
                 for (var i = 0; i < entities.Length; i++)
@@ -96,6 +111,11 @@ namespace <#= qualifiedNamespace #>
             public override void InvokeOnComponentUpdateCallbacks(Dictionary<int, ReaderWriterStore> entityIdToReaderWriterStore)
             {
 <# if (fieldDetailsList.Count > 0) { #>
+                if (ComponentsUpdatedComponentGroup.IsEmptyIgnoreFilter)
+                {
+                    return;
+                }
+
                 var entities = ComponentsUpdatedComponentGroup.GetEntityArray();
                 var updateLists = ComponentsUpdatedComponentGroup.GetComponentArray<ComponentsUpdated<<#= componentDetails.TypeName #>.Update>>();
                 for (var i = 0; i < entities.Length; i++)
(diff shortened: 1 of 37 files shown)


code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -60,21 +62,35 @@ namespace <#= qualifiedNamespace #>
...
-                    if (!manager.TryGetReadersWriters(<#= unityComponentDefinition.Id #>, out var readers))
+                    var readerWriterStore = readerWriterStores[entities[i].Index];
+                    if (!readerWriterStore.TryGetReaderWritersForComponent(<#= unityComponentDefinition.Id #>, out var readers))
                     {
                         continue;
                     }
 
                     var updateList = updateLists[i];
-                    foreach (<#= componentDetails.ComponentName #>.ReaderWriterImpl reader in readers)
+                    foreach (var reader in readers.OfType<<#= componentDetails.ComponentName #>.ReaderWriterImpl>())
```
firtina-improbable: Do you mean `Cast`? https://stackoverflow.com/questions/4015930/when-to-use-cast-and-oftype-in-linq

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -85,23 +101,23 @@ namespace <#= qualifiedNamespace #>
 <# } #>
             }
 
-            public override void InvokeOnEventUserCallbacks(GameObjectDispatcherSystem gameObjectDispatcherSystem)
+            public override void InvokeOnEventUserCallbacks(Dictionary<int, ReaderWriterStore> readerWriterStores)
```
firtina-improbable: The variable name should make it easy to infer that it's a dictionary of "component id" to "readerwriterstore"

firtina-improbable: I mean "entity id" ;)

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -85,23 +101,23 @@ namespace <#= qualifiedNamespace #>
...
+                        var readerWriterStore = readerWriterStores[entities[i].Index];
+                        if (!readerWriterStore.TryGetReaderWritersForComponent(<#= unityComponentDefinition.Id #>, out var readers))
                         {
                             continue;
                         }
 
                         var eventList = eventLists[i];
 
-                        foreach (<#= componentDetails.ComponentName #>.ReaderWriterImpl reader in readers)
+                        foreach (var reader in readers.OfType<<#= componentDetails.ComponentName #>.ReaderWriterImpl>())
```
firtina-improbable: `Cast`?

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -85,23 +101,23 @@ namespace <#= qualifiedNamespace #>
...
 <# for (var j = 0; j < eventDetailsList.Count; j++) { #>
                 {
                     var entities = EventsReceivedComponentGroups[<#= j #>].GetEntityArray();
                     var eventLists = EventsReceivedComponentGroups[<#= j #>].GetComponentArray<EventsReceived<<#= eventDetailsList[j].EventName #>Event>>();
                     for (var i = 0; i < entities.Length; i++)
                     {
-                        var manager = gameObjectDispatcherSystem.GetSpatialOSBehaviourManager(entities[i].Index);
-                        if (!manager.TryGetReadersWriters(<#= unityComponentDefinition.Id #>, out var readers))
+                        var readerWriterStore = readerWriterStores[entities[i].Index];
+                        if (!readerWriterStore.TryGetReaderWritersForComponent(<#= unityComponentDefinition.Id #>, out var readers))
```
firtina-improbable: Should the `TryGetReaderWriters` be doing the casting at this point, instead of the `OfType` / `Cast` below?

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
firtina-improbable: May be invalidated by comment above

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -7,9 +7,10 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
 {
     /// <summary>
     ///     Keeps track of Reader/Writer availability for SpatialOSBehaviours on a particular GameObject and decides when
-    ///     a SpatialOSBehaviour should be enabled, calling into the SpatialOSBehaviourLibrary for injection.
+    ///     a SpatialOSBehaviour should be enabled, calling into the SpatialOSBehaviourLibrary for injection, storing
+    ///     the created Readers/Writers in the given ReaderWriterStore.
     /// </summary>
-    internal class SpatialOSBehaviourManager
+    internal class MonoBehaviourActivationManager
```
firtina-improbable: I think this name can be better, something in between `SpatialOSBehaviour` and `MonoBehaviour`. It deals with a special kind of `MonoBehaviour`s.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -11,50 +12,27 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
...
-        }
-
-        internal void RemoveSpatialOSBehaviourManager(int entityIndex)
+        internal void RemoveActivationManagerAndReaderWriterStore(int entityIndex)
         {
-            if (!entityIndexToSpatialOSBehaviourManager.ContainsKey(entityIndex))
+            if (!entityIndexToActivationManager.ContainsKey(entityIndex))
             {
-                throw new SpatialOSBehaviourManagerNotFoundException($"SpatialOSBehaviourManager not found for entityIndex {entityIndex}.");
+                throw new ActivationManagerNotFoundException($"SpatialOSBehaviourManager not found for entityIndex {entityIndex}.");
```
firtina-improbable: Exception message needs to be modified too

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -23,71 +24,62 @@ internal class SpatialOSBehaviourManager
...
 
                 numUnsatisfiedReadersOrWriters[behaviour] = readerIds.Count + writerIds.Count;
 
                 behaviour.enabled = false;
             }
         }
 
-
-        public bool TryGetReadersWriters(uint componentId, out HashSet<IReaderWriterInternal> readers)
+        private void AddBehaviourForIds(MonoBehaviour behaviour, List<uint> componentIds,
```
firtina-improbable: `...ForComponentIds` perhaps?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -23,71 +24,62 @@ internal class SpatialOSBehaviourManager
...
+            Dictionary<uint, HashSet<MonoBehaviour>> idsToBehaviours)
```
firtina-improbable: `componentIdsToBehaviours` ?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -23,71 +24,62 @@ internal class SpatialOSBehaviourManager
...
-                foreach (var idToReaderWriterList in dict)
-                {
-                    var id = idToReaderWriterList.Key;
-                    var readerWriterList = idToReaderWriterList.Value;
-                    foreach (var readerWriter in readerWriterList)
-                    {
-                        GetOrCreateValue(compIdToReadersWriters, id).Add(readerWriter);
-                    }
-                }
+                var idsToReaderWriterLists = injector.InjectAllReadersWriters(behaviour, entity);
```
firtina-improbable: Which sort of `id`s?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterStore.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Collections.Generic;
+using UnityEngine;
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    internal class ReaderWriterStore
+    {
+        private readonly Dictionary<MonoBehaviour, Dictionary<uint, IReaderWriterInternal[]>> behaviourToReadersWriters
```
firtina-improbable: behaviour to `?` to readerwriters

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterStore.cs
```diff
@@ -0,0 +1,55 @@
...
+            = new Dictionary<MonoBehaviour, Dictionary<uint, IReaderWriterInternal[]>>();
+
+        private readonly Dictionary<uint, HashSet<IReaderWriterInternal>> compIdToReadersWriters =
```
firtina-improbable: `comp` => long form please

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterStore.cs
```diff
@@ -0,0 +1,55 @@
...
+            new Dictionary<uint, HashSet<IReaderWriterInternal>>();
+
+        public void AddReaderWritersForBehaviour(MonoBehaviour behaviour,
+            Dictionary<uint, IReaderWriterInternal[]> idsToReaderWriterLists)
```
firtina-improbable: which kind of `id`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
MPeti01: In the MonoBehaviour design we agreed on SpatialOSBehaviour meaning a MonoBehaviour which [Require]s things, so that did make sense, but I'd suggest for changing it because 1) that meaning is not intuitive, 2) it's long, 3) all MonoBehaviours go through it, it just finds nothing to do on the ones without [Require] statements.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
MPeti01: No, that part of the code doesn't know about the specific types.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
firtina-improbable: Makes sense, resolving this comment

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
firtina-improbable: Why not?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterStore.cs
MPeti01: Could call it something like behaviourToReadersWritersForComponents or something similar, but I feel that would be too long to be readable... I think most of these were changed when @jared-improbable suggested the original names I put in the BehaviourManager PR were too verbose, and especially here I think making this fully expressive would just return to that problem

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
samiwh: In which case please use regular explicit cast. Cast will return an IEnumerable which will allocate on foreach. While probably not a huge problem it is unnecessary.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
MPeti01: If we wanted that, the store would have to allocate a new HashSet<CorrectType> right? As it is, it just returns the internal collection it has already allocated

firtina-improbable: I see what you mean... well, it depends on whether you would need to use `.OfType` or `.Cast` every time you use this method. Then the cost can still be simplified to this method call.

I can think of quite an elaborate way of resolving the "storing of the right type value" but I'm unsure if it'd be worth it so I'll spare the details :D

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -60,21 +62,35 @@ namespace <#= qualifiedNamespace #>
...
+            public override void InvokeOnAuthorityChangeLifecycleCallbacks(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
+            {
+                var authoritiesChangedTags = AuthoritiesChangedComponentGroup.GetComponentArray<AuthoritiesChanged<<#= componentDetails.TypeName #>>>();
+                var entities = AuthoritiesChangedComponentGroup.GetEntityArray();
+                for (var i = 0; i < entities.Length; i++)
+                {
+                    var activationManager = entityIndexToManagers[entities[i].Index];
+                    for (var j = 0; j < authoritiesChangedTags[i].Buffer.Count; j++)
+                    {
+                        activationManager.ChangeAuthority(<#= unityComponentDefinition.Id #>, authoritiesChangedTags[i].Buffer[j]);
```
samiwh: Preference for GetComponentId(). Eventually we should replace the function with a field though.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/ReaderWriterInjector.cs
```diff
@@ -10,7 +10,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
     /// <summary>
     ///     Retrieves Reader and Writer fields from MonoBehaviours and handles injection into them.
     /// </summary>
-    internal class SpatialOSBehaviourLibrary
+    internal class ReaderWriterInjector
```
samiwh: The problem with ReaderWriter is that we are going to introduce more for commands. Are those also going to be handled here or elsewhere.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/ReaderWriterInjector.cs
```diff
@@ -22,7 +22,7 @@ internal class SpatialOSBehaviourLibrary
         private readonly ILogDispatcher logger;
         private readonly ReaderWriterFactory readerWriterFactory;
 
-        private const string LoggerName = "SpatialOSBehaviourLibrary";
+        private const string LoggerName = "ReaderWriterInjector";
```
samiwh: Can use nameof(ReaderWriterInjector)

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -60,21 +62,35 @@ namespace <#= qualifiedNamespace #>
                 return <#= unityComponentDefinition.Id #>;
             }
 
-            public override void InvokeOnComponentUpdateUserCallbacks(GameObjectDispatcherSystem gameObjectDispatcherSystem)
+            public override void InvokeOnAuthorityChangeLifecycleCallbacks(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
```
samiwh: Potential drive by - these aren't callbacks. The things we are calling user callbacks are callbacks and the things we call lifecycle callbacks are just methods.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/ReaderWriterInjector.cs
MPeti01: That's something I'm planning to think about right now, no plans made yet... But yes, this Injector should almost certainly be generalized to handle those as well, good point

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
MPeti01: It's a function because it's used by the methods implemented in the abstract superclass too. Not sure that's actually good though, might be more readable to have all the method bodies in one place and then we could make this a field too

MPeti01: I decided to move the 2 methods implemented in the superclass down to the subclasses, moving the method into a const field.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterStore.cs
firtina-improbable: Understood, not a big issue, resolving

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequireTagInjector.cs
```diff
@@ -10,7 +10,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
     /// <summary>
     ///     Retrieves Reader and Writer fields from MonoBehaviours and handles injection into them.
     /// </summary>
-    internal class SpatialOSBehaviourLibrary
+    internal class RequireTagInjector
```
samiwh: As it doesn't inject require tags this name could be better although I think you get the basic idea from it.

MPeti01: Can't really think of any more correct name that is not also annoyingly unclear/vague... :/

firtina-improbable: `RequiredFieldInjector`
`RequiredFieldResolver`
`SpatialAccessorInjector` / `Resolver`
`RequireResolver`
`MonoBehaviourInjection`
`MonoBehaviourInjectionResolver`
`RequirementResolver`
`SpatialInjector`
`SpatialBehaviourInjector`

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -55,26 +57,57 @@ namespace <#= qualifiedNamespace #>
 <# } #>
             };
 
-            protected override uint GetComponentId()
+            private const uint componentId = <#= unityComponentDefinition.Id #>;
+
+            public override void MarkComponentsAddedForActivation(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
+            {
+                var entities = ComponentAddedComponentGroup.GetEntityArray();
```
zeroZshadow: Check if the group is empty first. `ComponentAddedComponentGroup.IsEmptyIgnoreFilter` if true, just early out.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -55,26 +57,57 @@ namespace <#= qualifiedNamespace #>
...
+                for (var i = 0; i < entities.Length; i++)
+                {
+                    var activationManager = entityIndexToManagers[entities[i].Index];
+                    activationManager.AddComponent(componentId);
+                }
+            }
+
+            public override void MarkComponentsRemovedForDeactivation(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
+            {
+                var entities = ComponentRemovedComponentGroup.GetEntityArray();
```
zeroZshadow: Same group check

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -55,26 +57,57 @@ namespace <#= qualifiedNamespace #>
...
+                {
+                    var activationManager = entityIndexToManagers[entities[i].Index];
+                    activationManager.RemoveComponent(componentId);
+                }
+            }
+
+            public override void MarkAuthorityChangesForActivation(Dictionary<int, MonoBehaviourActivationManager> entityIndexToManagers)
             {
-                return <#= unityComponentDefinition.Id #>;
+                var authoritiesChangedTags = AuthoritiesChangedComponentGroup.GetComponentArray<AuthoritiesChanged<<#= componentDetails.TypeName #>>>();
```
zeroZshadow: group empty check

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -11,50 +12,27 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
...
 
-            entityIndexToSpatialOSBehaviourManager[entityIndex] = spatialOSBehaviourManager;
-            spatialOSBehaviourManagers.Add(spatialOSBehaviourManager);
-        }
-
-        internal void RemoveSpatialOSBehaviourManager(int entityIndex)
+        internal void RemoveActivationManagerAndReaderWriterStore(int entityIndex)
         {
-            if (!entityIndexToSpatialOSBehaviourManager.ContainsKey(entityIndex))
+            if (!entityIndexToActivationManager.ContainsKey(entityIndex))
```
zeroZshadow: Use a TryGet here instead

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -23,71 +24,64 @@ internal class SpatialOSBehaviourManager
...
             spatialId = spatialComponent.SpatialEntityId;
 
             foreach (var behaviour in gameObject.GetComponents<MonoBehaviour>())
             {
-                var readerIds = library.GetRequiredReaderComponentIds(behaviour.GetType());
-                foreach (var id in readerIds)
-                {
-                    GetOrCreateValue(behavioursRequiringReaderTypes, id).Add(behaviour);
-                }
+                var readerIds = injector.GetRequiredReaderComponentIds(behaviour.GetType());
```
zeroZshadow: cache the `GetType`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -23,71 +24,64 @@ internal class SpatialOSBehaviourManager
...
+                AddBehaviourForComponentIds(behaviour, readerIds, behavioursRequiringReaderTypes);
```
zeroZshadow: Why do this for behaviours that don't have readers or writers? For the similar reason, why store them in `numUnsatisfiedReadersOrWriters`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManagerInitializationSystem.cs
```diff
@@ -30,34 +31,26 @@ public struct RemovedEntitiesData
...
 
         protected override void OnUpdate()
         {
             for (var i = 0; i < addedEntitiesData.Length; i++)
             {
                 var entityIndex = addedEntitiesData.Entities[i].Index;
-                var spatialOSBehaviourManager = new SpatialOSBehaviourManager(
-                    addedEntitiesData.GameObjectReferences[i].GameObject, behaviourLibrary, logger);
-                gameObjectDispatcherSystem.AddSpatialOSBehaviourManager(entityIndex, spatialOSBehaviourManager);
+                gameObjectDispatcherSystem.CreateActivationManagerAndReaderWriterStore(addedEntitiesData.Entities[i]);
```
zeroZshadow: cache the `addedEntitiesData.Entities[i]`, we reading it twice now

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -41,41 +38,36 @@ public SpatialOSBehaviourLibrary(EntityManager entityManager, ILogDispatcher log
...
-            foreach (var componentId in componentReaderIdsForBehaviours[behaviourType])
+            var createdReaderWriters = new Dictionary<uint, IReaderWriterInternal[]>();
+            foreach (var idToFields in fieldInfoCache[behaviourType])
             {
-                List<IReaderWriterInternal> readerWritersForComp = new List<IReaderWriterInternal>();
-                createdReaderWriters[componentId] = readerWritersForComp;
-                Inject(behaviour, componentId, entity, readerWritersForComp);
+                var id = idToFields.Key;
+                var fields = idToFields.Value;
+                createdReaderWriters[id] = fields.Select(field => Inject(behaviour, id, entity, field)).ToArray();
```
zeroZshadow: This smells like allocations

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -41,41 +38,36 @@ public SpatialOSBehaviourLibrary(EntityManager entityManager, ILogDispatcher log
...
-            foreach (var readerWriterComponentId in componentReaderIdsForBehaviours[behaviourType])
-            {
-                DeInject(behaviour, readerWriterComponentId);
-            }
-
-            foreach (var readerWriterComponentId in componentWriterIdsForBehaviours[behaviourType])
+            foreach (var fieldsToComponents in fieldInfoCache[behaviourType])
             {
-                DeInject(behaviour, readerWriterComponentId);
+                var componentId = fieldsToComponents.Key;
```
zeroZshadow: unused

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterStore.cs
```diff
@@ -0,0 +1,55 @@
...
+            foreach (var idToReaderWriterList in componentIdsToReaderWriterLists)
+            {
+                var id = idToReaderWriterList.Key;
+                var readerWriterList = idToReaderWriterList.Value;
+                foreach (var readerWriter in readerWriterList)
+                {
+                    if (!componentIdToReadersWriters.TryGetValue(id, out var readersWriters))
+                    {
+                        readersWriters = new HashSet<IReaderWriterInternal>();
+                        componentIdToReadersWriters[id] = readersWriters;
```
zeroZshadow: use `componentIdToReadersWriters.Add` here for clarity

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterStore.cs
```diff
@@ -0,0 +1,55 @@
...
+        private readonly Dictionary<MonoBehaviour, Dictionary<uint, IReaderWriterInternal[]>> behaviourToReadersWriters
+            = new Dictionary<MonoBehaviour, Dictionary<uint, IReaderWriterInternal[]>>();
+
+        private readonly Dictionary<uint, HashSet<IReaderWriterInternal>> componentIdToReadersWriters =
+            new Dictionary<uint, HashSet<IReaderWriterInternal>>();
+
+        public void AddReaderWritersForBehaviour(MonoBehaviour behaviour,
+            Dictionary<uint, IReaderWriterInternal[]> componentIdsToReaderWriterLists)
+        {
+            behaviourToReadersWriters[behaviour] = componentIdsToReaderWriterLists;
```
zeroZshadow: Is this allowed to overwrite if something is already there? if not, use `behaviourToReadersWriters.Add`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReaderWriterStore.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public void RemoveReaderWritersForBehaviour(MonoBehaviour behaviour)
+        {
+            foreach (var idToReaderWriterList in behaviourToReadersWriters[behaviour])
+            {
+                var id = idToReaderWriterList.Key;
+                var readerWriterList = idToReaderWriterList.Value;
+                foreach (var readerWriter in readerWriterList)
+                {
+                    componentIdToReadersWriters[id].Remove(readerWriter);
```
zeroZshadow: cache `componentIdToReadersWriters[id]`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
MPeti01: Actually the manager is not used afterwards anymore, I just removed getting the manager afterwards

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -47,15 +47,23 @@ internal class MonoBehaviourActivationManager
...
+                }
+
+                if (writerIds.Count > 0)
+                {
+                    AddBehaviourForComponentIds(behaviour, writerIds, behavioursRequiringWriterTypes);
+                }
 
-                numUnsatisfiedReadersOrWriters[behaviour] = readerIds.Count + writerIds.Count;
-                if (numUnsatisfiedReadersOrWriters[behaviour] > 0)
+                if (readerIds.Count > 0 || writerIds.Count > 0)
```
zeroZshadow: Move this check to line 53

PR #214, opened 2018-08-14
New Worker Abstraction (V2)

jessicafalk:

-------

#### Description
second take on a new worker abstraction. Instead of putting wrappers all around it, we remove absolutely everything that is not necessary.
The worker is now a simple class storing some data regarding connection, origin, .. etc, which makes it possible to use it either in monobehaviors or in systems (using the `WorkerConfig`component)
It is tighly coupled to the connection i.e. we connect, when the worker is created, and upon disposing the worker, we will disconnect.

Removed loads of things
moved the translation unit stuff into its own file to make it easier to delete everything later (instead of moving it into the worker which was the initial plan)

This PR is huge: reading it in commit order should make your life a lot easier

#### Tests
ran it locally. it worked ™️

#### Documentation
will come in the next pr

#### Primary reviewers
@jamiebrynes7 @samiwh


code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -135,19 +135,19 @@ namespace <#= qualifiedNamespace #>
                 <#= componentDetails.CamelCaseTypeName#>.DirtyBit = false;
 
 <# if (componentDetails.IsBlittable) { #>
-                worker.EntityManager.AddComponentData(entity, <#= componentDetails.CamelCaseTypeName#>);
+                EntityManager.AddComponentData(entity, <#= componentDetails.CamelCaseTypeName#>);
 <# } else { #>
-                 worker.EntityManager.SetComponentObject(entity, <#= componentDetails.CamelCaseTypeName #>);
+                 EntityManager.SetComponentObject(entity, <#= componentDetails.CamelCaseTypeName #>);
```
martinzlocha: Nit: Extra space.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -0,0 +1,89 @@
...
+            return worker;
+        }
+
+        public static void Disconnect(Worker worker)
+        {
+            OnDisconnect(worker);
+            worker.Dispose();
+        }
+
+        public void Dispose()
```
martinzlocha: Might be better to make this private. Currently its possible to call `worker.Dispose()` or `Worker.Disconnect(worker)`, but only in one case will you get OnDisconnect called. Also this way it better mirrors the initializer.

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
```diff
@@ -1,33 +1,41 @@
...
-            worker = WorkerRegistry.GetWorkerForWorld(World);
+        private TranslationUnityRegistry translationUnityRegistry;
 
+        protected override void OnCreateManager(int capacity)
+        {
+            base.OnCreateManager(capacity);
+            translationUnityRegistry = TranslationUnityRegistry.WorldToTranslationUnit[World];
             dispatcher = new Dispatcher();
-            SetupDispatcherHandlers();
+            SetupDispatcherHandlers();
```
martinzlocha: Nit: Extra whitespace.

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
```diff
@@ -1,33 +1,41 @@
...
+        private Worker worker;
+
         protected override void OnUpdate()
         {
-            if (worker.Connection == null)
-            {
-                return;
-            }
-
+            worker = data.WorkerConfigs[0].Worker;
```
martinzlocha: Even though it hasn't been our standard so far can we check that the array only contains 1 entry.

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
```diff
@@ -40,18 +48,52 @@ protected override void OnUpdate()
...
+            worker.EntityMapping.Add(entityId, entity);
         }
 
         private void OnRemoveEntity(RemoveEntityOp op)
         {
-            worker.RemoveEntity(op.EntityId.Id);
+            var entityId = op.EntityId.Id;
+            if (!worker.TryGetEntity(entityId, out var entity))
+            {
+                worker.LogDispatcher.HandleLog(LogType.Error, new LogEvent("Tried to delete an entity but there is no entity associated with that EntityId.")
```
martinzlocha: Duplicate error strings, could extract into a separate variable.

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
```diff
@@ -40,18 +48,52 @@ protected override void OnUpdate()
...
+                    .WithField(LoggingUtils.LoggerName, LoggerName)
+                    .WithField(LoggingUtils.EntityId, entityId));
+                return;
+            }
+
+            EntityManager.DestroyEntity(entity);
+            worker.EntityMapping.Remove(entityId);
         }
 
         private void OnDisconnect(DisconnectOp op)
```
martinzlocha: Should this also call worker.OnDisconnect so that the monobehaviours get a callback as well?

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
```diff
@@ -60,8 +102,8 @@ private void SetupDispatcherHandlers()
             dispatcher.OnRemoveEntity(OnRemoveEntity);
             dispatcher.OnDisconnect(OnDisconnect);
             dispatcher.OnCriticalSection(op => { inCriticalSection = op.InCriticalSection; });
-
-            foreach (var translationUnit in worker.TranslationUnits.Values)
+
```
martinzlocha: Nit: Extra whitespace.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
```diff
@@ -42,5 +42,10 @@ public EntityGameObjectLinker(World world, ILogDispatcher logDispatcher)
             spatialOSComponent.SpatialEntityId = spatialEntityId;
             spatialOSComponent.World = world;
         }
+
+        public void Dispose()
```
martinzlocha: Is this needed given that its never used?

workers/unity/Assets/Playground/Scripts/Cubes/CubeMovementSystem.cs
```diff
@@ -19,20 +20,25 @@ public struct Data
...
 
+        [Inject] private WorkerData workerData;
+
+        private static Vector3 speed = new Vector3(2, 0, 0);
+
         protected override void OnUpdate()
         {
+            if (workerData.Length == 0)
+            {
+                new LoggingDispatcher().HandleLog(LogType.Error, new LogEvent("This system should not have been run without a worker entity"));
```
martinzlocha: Can this use the default dispatcher?

workers/unity/Assets/Playground/Config/ArchetypeConfig.cs
```diff
@@ -13,15 +13,15 @@ public static class ArchetypeConfig
             WorkerTypeToArchetypeNameToComponentTypes = new Dictionary<string, Dictionary<string, ComponentType[]>>
             {
                 {
-                    UnityGameLogic.WorkerType,
+                    "UnityGameLogic",
```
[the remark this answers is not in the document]
jamiebrynes7: Agree entirely - we should define a set of constants _somewhere_ for these as they will be used in multiple places

workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
```diff
@@ -32,50 +32,68 @@ public struct RemovedEntitiesData
...
 
         protected override void OnUpdate()
         {
+            if (workerData.Length == 0)
+            {
+                new LoggingDispatcher().HandleLog(LogType.Error, new LogEvent("This system should not have been run without a worker entity"));
+            }
+
+            var worker = workerData.WorkerConfigs[0].Worker;
+            if (viewCommandBuffer == null)
```
zeroZshadow: This is bad, is there no way to do this in `OnCreateManager` ?

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -17,64 +13,35 @@ public class Bootstrap : MonoBehaviour
...
-                connectionConfig.UseExternalIp = workerConfigurations.UseExternalIp;
-#endif
+                    WorkerType = "UnityClient",
+                };
+                CreateWorker(config, Vector3.zero);
+                config = new ReceptionistConfig
+                {
+                    WorkerType = "UnityGameLogic",
+                };
+                CreateWorker(config, new Vector3(10, 0, 0));
```
jamiebrynes7: Inline the creation of the configs here -

`CreateWorker(new ReceptionistConfig("UnityClient"), Vector3.zero)`

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -140,9 +82,19 @@ public static void DomainUnloadShutdown()
             ScriptBehaviourUpdateOrder.UpdatePlayerLoop();
         }
 
-        private void LoadLevel(WorkerBase worker)
+        private void CreateWorker(ConnectionConfig config, Vector3 origin)
         {
-            Instantiate(Level, worker.Origin, Quaternion.identity);
+            if (config.WorkerType.Equals("UnityGameLogic"))
+            {
+                PlayerLifecycleConfig.CreatePlayerEntityTemplate = PlayerTemplate.CreatePlayerEntityTemplate;
```
jamiebrynes7: Why does this need to be conditional? If its set on the client, does it matter? 

Related - does this even belong here?

workers/unity/Assets/Playground/Scripts/Cubes/CubeMovementSystem.cs
```diff
@@ -19,20 +20,25 @@ public struct Data
...
+            }
+
+            var origin = workerData.WorkerConfigs[0].Worker.Origin;
+
```
jamiebrynes7: This boilerplate makes me sad

workers/unity/Assets/Playground/Scripts/DisconnectSystem.cs
```diff
@@ -11,17 +11,20 @@ internal class DisconnectSystem : ComponentSystem
...
         }
 
         [Inject] private DisconnectData data;
 
         protected override void OnUpdate()
         {
-            Debug.LogWarningFormat("Diconnected from SpatialOS with reason: \"{0}\"",
-                data.DisconnectMessage[0].ReasonForDisconnect);
+            data.Configs[0].Worker.LogDispatcher.HandleLog(
+                LogType.Warning,
```
jamiebrynes7: Does this need to be a warning? Disconnecting is not abnormal flow

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -16,18 +17,18 @@ public class CleanReactiveComponentsSystem : ComponentSystem
         // Here to prevent adding an action for the same type multiple times
         private readonly HashSet<Type> typesToRemove = new HashSet<Type>();
 
-        private WorkerBase worker;
-
+        private TranslationUnityRegistry translationUnityRegistry;
+
```
zeroZshadow: Whitespace 🏳️

workers/unity/Assets/Playground/Scripts/EntityInitialization/ArchetypeInitializationSystem.cs
```diff
@@ -26,58 +26,64 @@ public struct Data
...
+                viewCommandBuffer = new ViewCommandBuffer(worker.LogDispatcher);
+            }
             for (var i = 0; i < data.Length; i++)
             {
                 var archetypeName = data.ArchetypeComponents[i].ArchetypeName;
                 var entity = data.Entities[i];
 
                 ComponentType[] componentTypesToAdd;
-                if (!ArchetypeConfig.WorkerTypeToArchetypeNameToComponentTypes[workerType]
+                if (!ArchetypeConfig.WorkerTypeToArchetypeNameToComponentTypes[worker.WorkerType]
```
jamiebrynes7: Driveby - use `out var`

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
zeroZshadow: Whitespace ⚪️

workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
```diff
@@ -32,50 +32,68 @@ public struct RemovedEntitiesData
             public SubtractiveComponent<GameObjectReference> NoGameObjectReference;
         }
 
+        public struct WorkerData
```
jamiebrynes7: It might be worth pulling this out into core given that its use everywhere as a common data struct.

Then no need to redefine in every system

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
zeroZshadow: Whitespace ⬜️

workers/unity/Packages/com.improbable.gdk.core/Systems/CustomSpatialOSSendSystem.cs
```diff
@@ -13,18 +14,22 @@ public abstract class CustomSpatialOSSendSystem<T> : ComponentSystem where T : I
...
         {
             base.OnCreateManager(capacity);
 
-            worker = WorkerRegistry.GetWorkerForWorld(World);
-
             spatialOSSendSystem = World.GetOrCreateManager<SpatialOSSendSystem>();
             if (!spatialOSSendSystem.TryRegisterCustomReplicationSystem(typeof(T)))
             {
-                worker.LogDispatcher.HandleLog(LogType.Error, new LogEvent(CustomReplicationSystemAlreadyExists)
+                new LoggingDispatcher().HandleLog(LogType.Error, new LogEvent(CustomReplicationSystemAlreadyExists)
```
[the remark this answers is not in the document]
jamiebrynes7: You could add it as another generic on the system - but that doesn't feel good necessarily.

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
jamiebrynes7: Yup - magic strings should be const

workers/unity/Packages/com.improbable.gdk.core/Systems/CustomSpatialOSSendSystem.cs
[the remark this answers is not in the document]
jessicafalk: one idea that came up in between would be to let the systems own the dispatcher instead of the worker as this is more aligned to how logging dispatchers in other projects. However, in this case it is actually possible (after realizing that you actually have access to the injected groups from the very start) and I'll change that

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
jessicafalk: can't make that private as it's inheriting from `IDisposable`, but I can move the call into `Dispose`

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
jessicafalk: yeah, so far I only checked the length, if necessary (because we have multiple injection groups in a system). If we only have one injection group, then that system will only run if there are any worker entities, but I can add a check for whether there is only one entry

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
jessicafalk: it was only set for the gamelogic worker in the previous code, simply keeping the same logic here. 
The overall idea would be to factor it out into it's own monobehavior/gameobject, but I didn't want to make the PR even bigger

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
jessicafalk: I think that the `Worker.OnDisconnect` should only be called from the worker class. we don't actually disconnect here, but only later. if we call it later, we might end up calling it twice.
once here and a second time during the actual disconnect.

The alternative would be to only call it here and not when we dispose  / disconnect

workers/unity/Assets/Playground/Scripts/Cubes/CubeMovementSystem.cs
```diff
@@ -19,20 +20,25 @@ public struct Data
 
         [Inject] private Data data;
 
-        private static Vector3 speed = new Vector3(2, 0, 0);
-
-        private Vector3 origin;
-
-        protected override void OnCreateManager(int capacity)
+        public struct WorkerData
```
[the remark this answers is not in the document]
jessicafalk: it doesn't seem like we have a convention for that. After checking a few files we usually make them `public`. Would prefer to leave it like this to have it the same everywhere and then create a new PR to make it private everywhere. Otherwise this PR will just become even messier :P

workers/unity/Assets/Playground/Scripts/Cubes/CubeMovementSystem.cs
jessicafalk: discussed offline: we don't have the worker here, so we need to create a logging dispatcher

workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
jessicafalk: found out that we actually can do it. Thought we don't have access to the injected groups in `OnCreateManager`. will revert that change

workers/unity/Assets/Playground/Scripts/Cubes/CubeMovementSystem.cs
jessicafalk: me too :(
maybe we need to create a `SpatialSystem` after all that takes care of that boilerplate code.

workers/unity/Assets/Playground/Scripts/DisconnectSystem.cs
jessicafalk: I guess it depends on why you got disconnected. I simply converted it from `LogWarning` to our dispatcher, but can change it to Log if preferred. I don't have a strong opinion about this

jessicafalk:
just as a note for everyone: The worker class will be changed to inherit from `ScriptpableBehaviourManager ` to make the injection bit easier.
but before that, I'll rebase on top of the experimental core csharp branch and then create a new PR with all these changes in it

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        public readonly Vector3 Origin;
+        public readonly string WorkerType;
+        public readonly string WorkerId;
+        public readonly World World;
+        public readonly Connection Connection;
+        public readonly ILogDispatcher LogDispatcher;
+
+        // callbacks for monobehaviour
+        public static Action<Worker> OnConnect;
```
samiwh: Make this an event rather than an action. (Backed by a list of actions)

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            World = new World(WorkerId);
+            var entityManager = World.GetOrCreateManager<EntityManager>();
+
+            var workerConfig = new WorkerConfig
+            {
+                Worker = this
+            };
+
+            FindTranslationUnits();
```
samiwh: Why is this here?

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -0,0 +1,112 @@
...
+            ConnectionUtility.Disconnect(Connection);
+            Connection.Dispose();
+        }
+        
+        
+        // can be deleted soonish
+        public readonly Dictionary<int, ComponentTranslation> TranslationUnits =
+            new Dictionary<int, ComponentTranslation>();
+        
+        public Action<Entity, long> AddAllCommandRequestSenders;
```
samiwh: This should be in a core system on a worker not part of the worker itself.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        public static Worker Connect(ConnectionConfig config, ILogDispatcher logDispatcher, Vector3 origin)
+        {
+            return new Worker(config, logDispatcher, origin);
+        }
+
+        public static void Disconnect(Worker worker)
+        {
+            worker.Dispose();
```
samiwh: This probably isn't what we want. We might want a tick to run your cleanup code with a disconnect message.

workers/unity/Assets/Playground/Scripts/Cubes/CubeMovementSystem.cs
samiwh: This kind of error message isn't very useful to a user. Can be explicit about what the error is but we should probably also add something which suggests that it's because they might have started the worker incorrectly.

jessicafalk:
closing it in favor of the newest worker abstraction PR

PR #215, opened 2018-08-15
UTY-941 playground core

zeroZshadow:
#### Description
Fixed compiler errors until they were gone
Removed unused classes and struct

This will be rebased once world command work, and then retested

#### Tests
Commented out the WorldCommands bits and ran a local deployment
UnityGameLogic worker shows cubes that move around


workers/unity/Assets/Playground/Scripts/Workers/UnityClient.cs
```diff
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
```
martinzlocha: Nothing else changed in the fine. Is this import needed?

jamiebrynes7: Yeah - it was missing before 😬

PR #224, opened 2018-08-16
UTY-424: Enable Use of Data Field in a Component

jamiebrynes7:
#### Description
This PR contains a bugfix for the code generation that previously preventing the following schema from compiling:
```
package demogame.npc;

enum Color {
    BLUE = 0;
    CYAN = 1;
    GRAY = 2;
    GREEN = 3;
    MAGENTA = 4;
    RED = 5;
    YELLOW = 6;
}

type ColorData {
    Color color = 1;
}

component NpcColor {
    id = 12522;

    data ColorData;
    event ColorData change_npc_color;
}
```

We had filtered the types defined in the component as their "data definition" as these were not being used in the actual code generation. However, if the data type was externally defined (in the same file) - it wouldn't generate `ColorData` at all.

This PR improves the filtering by doing a more stringent check that enables the generated code from the above schema file to compile while still filtering out auto-generated type definitions (from the schema compiler).

#### Tests
Ran codegen on the above schema - it compiled.


code_generator/GdkCodeGenerator/src/Generation/SingleGenerationJob.cs
```diff
@@ -135,5 +134,31 @@ protected override void RunImpl()
...
+            var filteredTypes = schemaFile.TypeDefinitions.Where(type => componentDataTypes.All(componentData =>
+                type.QualifiedName != componentData.TypeName ||
+                !SourceReferenceEquals(type.SourceReference, componentData.sourceReference)));
+
+            typesToGenerate.AddRange(filteredTypes);
+        }
+
+        private bool SourceReferenceEquals(SourceReferenceRaw sourceRef1, SourceReferenceRaw sourceRef2)
+        {
+            return sourceRef1.column == sourceRef2.column && sourceRef1.line == sourceRef2.line;
```
firtina-improbable: 😅

code_generator/GdkCodeGenerator/src/Generation/SingleGenerationJob.cs
```diff
@@ -135,5 +134,31 @@ protected override void RunImpl()
...
+            // From inspection of the JSON AST you can observe that a type definition is auto-generated if the following
+            // conditions are true:
+            //     1. The FQN type names are the same .
+            //     2. The source references are the same.
+            // Using this information, we can effectively filter out auto-generated types.
+            var filteredTypes = schemaFile.TypeDefinitions.Where(type => componentDataTypes.All(componentData =>
+                type.QualifiedName != componentData.TypeName ||
+                !SourceReferenceEquals(type.SourceReference, componentData.sourceReference)));
+
+            typesToGenerate.AddRange(filteredTypes);
```
firtina-improbable: why not return the list instead of modifying within method?

zeroZshadow: as discussed, not a problem

jamiebrynes7: 😬

PR #225, opened 2018-08-16
Preprocess prefabs before entering playmode or building

firtina-improbable:

-------

#### Description
SpatialOS GDK for Unity takes ownership of MonoBehaviour activation if they [Require] Readers/Writers/Command Senders/Command Response Handlers. Let’s call these Spatial Behaviours.

This is because they should not be receiving OnEnable callbacks if their reader/writers are null.
When a prefab is instantiated, all of the active MonoBehaviours will be enabled immediately. This results in undesired behaviour if any of these MonoBehaviours are Spatial Behaviours. For example if they would like to, within OnEnable, immediately fetch SpatialOS component info through a Reader, the Reader will be null.

##### Solution
Spatial Behaviours should be disabled before a prefab that may contain them can be instantiated.
We propose an editor script to resolve this.

#### Tests
Tested locally
- by enabling behaviours, hitting play: they get disabled ✅ 
- by enabling behaviours and doing a build: they get disabled ✅ 

#### Primary reviewers
@zeroZshadow @MPeti01

### DIFF
--- a/workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
+++ b/workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
@@ -43,7 +43,8 @@ namespace Playground.Editor
                 .Select(AssetDatabase.GUIDToAssetPath)
                 .Where(assetPath => assetPath.Contains("Resources"))
                 .Select(AssetDatabase.LoadAssetAtPath<GameObject>)
-                .Where(prefabObject => prefabObject != null);
+                .Where(prefabObject => prefabObject != null)
+                .ToArray();
 
             foreach (var prefabObject in allPrefabObjectsInResources)
             {
--- a/workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
+++ b/workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
@@ -1,0 +1,95 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Improbable.Gdk.Core.GameObjectRepresentation;
+using UnityEditor;
+using UnityEditor.Build;
+using UnityEditor.Build.Reporting;
+using UnityEngine;
+using Object = UnityEngine.Object;
+
+namespace Playground.Editor
+{
+    [InitializeOnLoad]
+    public class PrefabPreprocessor : IPreprocessBuildWithReport
+    {
+        // Needed for IPreprocessBuildWithReport
+        public int callbackOrder => 0;
+
+        static PrefabPreprocessor()
+        {
+            EditorApplication.playModeStateChanged += OnPlaymodeStateChanged;
+        }
+
+        void IPreprocessBuildWithReport.OnPreprocessBuild(BuildReport report)
+        {
+            PreprocessPrefabs();
+        }
+
+        private static void OnPlaymodeStateChanged(PlayModeStateChange playModeStateChange)
+        {
+            if (playModeStateChange == PlayModeStateChange.ExitingEditMode)
+            {
+                PreprocessPrefabs();
+            }
+        }
+
+        private static void PreprocessPrefabs()
+        {
+            var assetsUpdated = false;
+
+            var allPrefabObjectsInResources = AssetDatabase.FindAssets("t:Prefab")
+                .Select(AssetDatabase.GUIDToAssetPath)
+                .Where(assetPath => assetPath.Contains("Resources"))
+                .Select(AssetDatabase.LoadAssetAtPath<GameObject>)
+                .Where(prefabObject => prefabObject != null);
+
+            foreach (var prefabObject in allPrefabObjectsInResources)
+            {
+                var prefabWasFixed = false;
+
+                foreach (var monoBehaviour in prefabObject
+                    .GetComponents<MonoBehaviour>()
+                    .Where(DoesBehaviourNeedFixing))
+                {
+                    assetsUpdated = true;
+                    prefabWasFixed = true;
+
+                    Undo.RecordObject(monoBehaviour, "Disable MonoBehaviours with [Require] fields on prefabs.");
+
+                    monoBehaviour.enabled = false;
+                }
+
+                if (prefabWasFixed)
+                {
+                    EditorUtility.SetDirty(prefabObject);
+                }
+            }
+
+            if (assetsUpdated)
+            {
+                AssetDatabase.SaveAssets();
+            }
+        }
+
+        private static bool IsBehaviourEnabledInEditor(Object obj)
+        {
+            return !ReferenceEquals(obj, null)
+                && EditorUtility.GetObjectEnabled(obj) == 1;
+        }
+
+        private static bool DoesBehaviourRequireReadersOrWriters(Type targetType)
+        {
+            return targetType
+                .GetFields(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public)
+                .Any(field => Attribute.IsDefined(field, typeof(RequireAttribute), false));
+        }
+
+        private static bool DoesBehaviourNeedFixing(MonoBehaviour monoBehaviour)
+        {
+            return IsBehaviourEnabledInEditor(monoBehaviour) &&
+                DoesBehaviourRequireReadersOrWriters(monoBehaviour.GetType());
+        }
+    }
+}
--- a/workers/unity/Assets/Playground/Editor/Prefabs.meta
+++ b/workers/unity/Assets/Playground/Editor/Prefabs.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 7c8c192344e570c4bb12a3cbd67ff8eb
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs.meta
+++ b/workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs.meta
@@ -1,0 +1,11 @@
+fileFormatVersion: 2
+guid: a7222603732cbf649915f0a7e26894af
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/workers/unity/Assets/Playground/Resources/Prefabs/Spinner.prefab
+++ b/workers/unity/Assets/Playground/Resources/Prefabs/Spinner.prefab
@@ -106,7 +106,7 @@ MonoBehaviour:
   m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 1198430049189278}
-  m_Enabled: 1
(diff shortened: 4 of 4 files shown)


workers/unity/Assets/Playground/Editor/Improbable.Playground.Editor.asmdef
```diff
@@ -3,12 +3,13 @@
...
+        "Improbable.Gdk.Core"
     ],
     "optionalUnityReferences": [],
     "includePlatforms": [
         "Editor"
     ],
     "excludePlatforms": [],
     "allowUnsafeCode": false
-}
+}
```
firtina-improbable: Do we really need to fix newlines for asmdef files?

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        private static bool RequiresReadersOrWriters(Type targetType)
+        {
+            return targetType
+                .GetFields(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public)
+                .Any(field => Attribute.IsDefined(field, typeof(RequireAttribute), false));
+        }
+
+        private static bool BehaviourNeedsFixing(MonoBehaviour monoBehaviour)
```
firtina-improbable: TODO: rename these methods

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            if (playModeStateChange == PlayModeStateChange.ExitingEditMode)
+            {
+                PreprocessPrefabs();
+            }
+        }
+
+        private static void PreprocessPrefabs()
+        {
+            if (!DoPrefabsNeedFixing())
```
zeroZshadow: Both functions do the same finds and load all assets, might want to do this once and reuse it?

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -0,0 +1,122 @@
...
+                var prefabNeedsFixing = false;
+
+                foreach (var monoBehaviour in prefabObject
+                    .GetComponents<MonoBehaviour>()
+                    .Where(BehaviourNeedsFixing))
+                {
+                    if (!prefabNeedsFixing)
+                    {
+                        prefabNeedsFixing = true;
+                        prefabsToFix.Add(prefabObject);
```
zeroZshadow: Why do we use this list only for marking things as dirty? and fix everything anyways

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -0,0 +1,122 @@
...
+            prefabsToFix.ForEach(EditorUtility.SetDirty);
+        }
+
+        private static bool IsBehaviourEnabledInEditor(Object obj)
+        {
+            return !ReferenceEquals(obj, null)
+                && EditorUtility.GetObjectEnabled(obj) == 1;
+        }
+
+        private static bool RequiresReadersOrWriters(Type targetType)
```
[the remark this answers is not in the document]
MPeti01: The RequiredFieldInjector does this but it also does a lot more with it (find component IDs, separate readers and writers, get and cache FieldInfo objects...) With how short it is here, I don't think it's worth the trouble to find a way to share

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
callumlawson: +1
Should make it both faster, and a bit easier to read. 
Should be able to do "Where(Test),Do(Fix)" without getting and looping over things twice

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
firtina-improbable: we need to mark all these prefabs as dirty only once, but each prefab can contain multiple behaviours to fix

samiwh: But you already have to load all monobehaviours into memory. In the case where a prefab doesn't need fixing you also have to iterate over all monobehaviours on that prefab.

So it will be strictly less work and I think less code to fix all monobehaviours during the first iteration.

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -36,39 +36,13 @@ private static void PlaymodeStateChanged(PlayModeStateChange playModeStateChange
...
-        private static bool PrefabNeedsFixing(GameObject prefabObject)
-        {
-            return prefabObject != null
-                && prefabObject.GetComponents<MonoBehaviour>()
-                    .Any(BehaviourNeedsFixing);
-        }
-
-        private static void FixPrefabs()
         {
             var prefabsToFix = new List<GameObject>();
```
zeroZshadow: Can this be made a `HashSet<GameObject>` that you just always add the prefab to if a monobehaviour on it needs to be disabled?
This saves some statekeeping and will ensure it is not duplicated.

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            EditorApplication.playModeStateChanged += PlaymodeStateChanged;
+        }
+
+        void IPreprocessBuildWithReport.OnPreprocessBuild(BuildReport report)
+        {
+            PreprocessPrefabs();
+        }
+
+        private static void PlaymodeStateChanged(PlayModeStateChange playModeStateChange)
```
jamiebrynes7: Nit: Can we name this `OnPlaymodeStateChanged`?

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -0,0 +1,102 @@
...
+                    }
+
+                    monoBehavioursToDisable.Add(monoBehaviour);
+                }
+            }
+
+            if (monoBehavioursToDisable.Count > 0)
+            {
+                Undo.RecordObjects(monoBehavioursToDisable.Cast<Object>().ToArray(),
+                    "Disable monobehaviours with [Require] fields on prefabs");
```
jared-improbable: *MonoBehaviours

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
firtina-improbable: Removed the list, marking it dirty as it's being processed now

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
firtina-improbable: Done

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
firtina-improbable: Fixed

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
firtina-improbable: Simplified greatly

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -0,0 +1,95 @@
...
+                var prefabWasFixed = false;
+
+                foreach (var monoBehaviour in prefabObject
+                    .GetComponents<MonoBehaviour>()
+                    .Where(DoesBehaviourNeedFixing))
+                {
+                    assetsUpdated = true;
+                    prefabWasFixed = true;
+
+                    Undo.RecordObject(monoBehaviour, "Disable MonoBehaviours with [Require] fields on prefabs.");
```
jared-improbable: Would it be better to group these operations together as one using `Undo.IncrementCurrentGroup`? Or is that unnecessary?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/FlashOnCollision.cs
```diff
@@ -15,22 +15,22 @@ public class FlashOnCollision : MonoBehaviour
...
 
-    private static MaterialPropertyBlock basicMaterial = new MaterialPropertyBlock();
-    private static MaterialPropertyBlock flashingMaterial = new MaterialPropertyBlock();
+    private static MaterialPropertyBlock basicMaterial;
+    private static MaterialPropertyBlock flashingMaterial;
 
     [RuntimeInitializeOnLoadMethod]
     public static void SetupColors()
     {
+        basicMaterial = new MaterialPropertyBlock();
```
MPeti01: Just a note, this will probably have some (easy-to-resolve) merge conflicts with my Playground PR. I think your PR is more likely to be merged in first though, so I'll just deal with it when I merge.

firtina-improbable: FYI #259 will also conflict with both

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -0,0 +1,95 @@
...
+
+namespace Playground.Editor
+{
+    [InitializeOnLoad]
+    public class PrefabPreprocessor : IPreprocessBuildWithReport
+    {
+        // Needed for IPreprocessBuildWithReport
+        public int callbackOrder => 0;
+
+        static PrefabPreprocessor()
```
jessicafalk: I think we always add the access modifiers to methods

firtina-improbable: Access modifiers not allowed for static constructors

firtina-improbable: Nor for explicit interface implementation below (I'll make explicit interface implementations a thing again!)

firtina-improbable:
Rebased onto master

workers/unity/Assets/Playground/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        private static void PreprocessPrefabs()
+        {
+            var assetsUpdated = false;
+
+            var allPrefabObjectsInResources = AssetDatabase.FindAssets("t:Prefab")
+                .Select(AssetDatabase.GUIDToAssetPath)
+                .Where(assetPath => assetPath.Contains("Resources"))
+                .Select(AssetDatabase.LoadAssetAtPath<GameObject>)
+                .Where(prefabObject => prefabObject != null);
```
samiwh: Minor point - might be worth adding ToArray at the end of this to prevent future work doing unexpected things. If left as an IEnumerable it will actually evaluate the entire LINQ expression every time the variable is accessed. Has caught people out before when trying to augment these kinds of scripts.

firtina-improbable: 👍

PR #228, opened 2018-08-16
UTY-959: Injection groundwork for MonoBehaviour commands

MPeti01:

-------

#### Description
To implement Command sender, request handlers and response handlers, we need to be able inject [Require]d things other than ReaderWriters. This involved:
* Generalizing the `ReaderWriterStore`, `MonoBehaviourActivationManager`, `RequiredFieldInjector`, `ReaderWriterFactory`, `ReaderWriterCreators` - replacing the `IReaderWriter` interface they're handling with `IInjectable` and renaming them appropriately
* Add `InjectableId` to let `IInjectables` be identified with more than the component ID so that users of the `InjectableStore` can query for just the `ReaderWriter`s, just the `CommandSender`s etc for a particular component
* Replace `[ReaderInterface]` and `[WriterInterface]` with a tag for `[InjectionCondition]`, taking an enum: can inject always, needs component present or needs authority on component
* Handle Injectables not requiring conditions (component present/authoritative) appropriately
* (EDIT) Added code generated stubs for command senders, request handlers and response handlers
#### Tests
Implementation rework, same tests still apply, tweaked them where needed. All tests are green and the Playground works fine.
#### Documentation
Class summary comments. Also driveby updated a few of these that have been outdated for a while.
#### Primary reviewers
@firtina-improbable @jessicafalk


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/IInjectableCreator.cs
```diff
@@ -0,0 +1,12 @@
+using Unity.Entities;
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    /// <summary>
+    ///     Interface for objects creating a particular type of IInjectable, to be used by the InjectableFactory.
+    /// </summary>
+    internal interface IReaderWriterCreator
```
firtina-improbable: I think it's this PR that should continue the work within `IInjectableCreator.cs` a tiny bit more and rename `IReaderWriterCreator` and accept a `InjectableType` parameter, implementing the `ReaderWriter` part and `NotImplementedException` with `// TODO {ticketid}` for the rest.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -58,6 +58,7 @@ namespace <#= qualifiedNamespace #>
             };
 
             private const uint componentId = <#= unityComponentDefinition.Id #>;
+            private static readonly InjectableId injectableId = new InjectableId(InjectableType.ReaderWriter, componentId);
```
firtina-improbable: `readerWriterInjectableId`

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -120,8 +121,8 @@ namespace <#= qualifiedNamespace #>
                 var updateLists = ComponentsUpdatedComponentGroup.GetComponentArray<ComponentsUpdated<<#= componentDetails.TypeName #>.Update>>();
                 for (var i = 0; i < entities.Length; i++)
                 {
-                    var readerWriterStore = entityIdToReaderWriterStore[entities[i].Index];
-                    if (!readerWriterStore.TryGetReaderWritersForComponent(componentId, out var readers))
+                    var readerWriterStore = entityIdToInjectableStore[entities[i].Index];
```
firtina-improbable: `injectableStore`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectableStore.cs
```diff
@@ -0,0 +1,61 @@
...
+                foreach (var injectable in injectables)
+                {
+                    allInjectablesForId.Remove(injectable);
+                }
+            }
+
+            behaviourToInjectables.Remove(behaviour);
+        }
+
+        public bool TryGetReaderWritersForComponent(InjectableId injectableId, out HashSet<IInjectable> injectables)
```
firtina-improbable: `TryGetInjetablesForComponent`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/IInjectableCreator.cs
MPeti01: Yes I missed some renames there, will change. The intention was to have separate Creators for each specific type though, ie one for the ReaderWriter, one for each command's sender etc. To have one per component we'd have to add some additional parameters beside the InjectableType too to specify the command index, but having separate Creators for each is straightforward to do.

firtina-improbable: Could you pass in the `InjectableId` as a parameter to the `CreateInjector` function and then the creator method could be as sophisticated as necessary

firtina-improbable: Could something like this work for example

```
            public IInjectable CreateInjectable(Entity entity, InjectableId injectableId, EntityManager entityManager, ILogDispatcher logDispatcher)
            {
                switch(injectableId.type) {
                    case Reader: return new ReaderWriterImpl(entity, entityManager, logDispatcher);
                    default: NotImplementedYet();
                }
           }
```

MPeti01: Yes, and then Creators would have an attribute to associate them with a component ID rather than an injectable ID still, that could work. Is there a good reason to do that though? It feels cleaner to have just one for each and not pass into them info (part of which was actually used for identifying the appropriate creator in the first place..)

firtina-improbable: It depends whether we'd like to increase the complexity in the factory vs next to the accessors. I am unsure how cleaner it would be to do the former, can you show what the code could look like that way?

MPeti01: The factory literally wouldn't have to change at all, since it's just mapping from InjectableId to Creator, and the InjectableIds identify the specific type uniquely (including different command senders on same component etc). 

So the Factory code would look as it does now, and we'd have for each component a ReaderWriterCreator and for each command 3 separate Creators (Sender, RequestHandler, ResponseHandler), all with oneliner implementations like the current ReaderWriterCreator.

firtina-improbable: Doesn't sound too bad 👍 can you create stubs in this PR?

MPeti01: Good idea

MPeti01:
I moved all Injectable types (3 command handlers per command, readers, writers) to `ComponentName.Accessors.Reader`, `ComponentName.Accessors.CommandNameSender` etc. The idea is that we would collectively call all of these Accessors (including in user facing docs), and this disambiguates these command handlers from the ECS ones. 

I'm not entirely satisfied though, especially with having to write `[Require] ComponentName.Accessors.Reader` instead of just `[Require] ComponentName.Reader`, parses much less nicely in my mind... Open for suggestions!

JonasImprobable:
What I don't like about `Accessors` is that the name doesn't make it obvious that the sub-classes are for monobehaviours. How about renaming it to something that makes the intent more clear, e.g. `MonoBehaviourAccessors` or `MonoBehaviourInjectables`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/IInjectableCreator.cs
```diff
@@ -0,0 +1,12 @@
+using Unity.Entities;
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    /// <summary>
+    ///     Interface for objects creating a particular type of IInjectable, to be used by the InjectableFactory.
```
JonasImprobable: Glad to see the XMLDoc comments. Good stuff! 👍

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectableFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    /// <summary>
+    ///     Creates IInjectables given an InjectableId and an Entity with which it should be associated.
+    /// </summary>
+    public class InjectableFactory
+    {
+        private const string InjectableIdAttributeNotFoundError
+            = "ReaderWriterCreator found with no Injectable ID attribute.";
```
JonasImprobable: This string doesn't look up-to-date.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectableId.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// <summary>
+    ///     Identifier for specific injectable type, to be used for finding IInjectableCreators and querying from
+    ///     the InjectableStore.
+    /// </summary>
+    public struct InjectableId
+    {
+        public readonly InjectableType type;
+        public readonly uint componentId;
+        public readonly int index;
```
JonasImprobable: What is this for? Is this the command index of a command of a component?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiresMonoBehaviourActivationManager.cs
```diff
@@ -3,7 +3,7 @@
 namespace Improbable.Gdk.Core.GameObjectRepresentation
 {
     /// <summary>
-    ///     Tag component for marking entities that require a RequiresSpatialOSBehaviourManager.
+    ///     Tag component for marking entities that require a MonoBehaviourActivationManager.
```
JonasImprobable: Nit: I know I'm a little late to the party but would it make sense to rename `MonoBehaviourActivationManager` to `MonoBehaviourLifecycleManager`? Since it does more than activating monobehaviours?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -76,16 +82,16 @@ public void DeInjectAllReadersWriters(MonoBehaviour behaviour)
             }
         }
 
-        public List<uint> GetRequiredReaderComponentIds(Type behaviourType)
+        public List<uint> GetComponentPresenceRequirements(Type behaviourType)
         {
             EnsureLoaded(behaviourType);
-            return componentReaderIdsForBehaviours[behaviourType];
+            return componentReadRequirementsForBehaviours[behaviourType];
```
JonasImprobable: There are some naming inconsistencies with `ComponentPresenceRequirements` and `ComponentReadRequirements`.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectionConditionAttribute.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    /// <summary>
+    ///     Used to denote when a specific IInjectable type is ready to be created.
+    /// </summary>
+    internal enum InjectionCondition
```
JonasImprobable: Shower thought question here: Wouldn't the injection conditions solely depend on the `InjectableType` of an injectable? I.e. A writer or command request handler will always require authority over a component. If that's the case, isn't this attribute here be a little redundant since you can infer the requirements from the InjectableType?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectableId.cs
MPeti01: Command index, didn't want to name it that in case we want to use it for any other purpose in other types later... Might not be the best way to go about this, what do you think?

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
```diff
@@ -0,0 +1,85 @@
...
+#>
+            internal class <#= commandName #>CommandSenderCreator : IInjectableCreator
+            {
+                public IInjectable CreateInjectable(Entity entity, EntityManager entityManager, ILogDispatcher logDispatcher)
+                {
+                    return new <#= commandName #>CommandSender(entity, entityManager, logDispatcher);
+                }
+            }
+
+            [InjectableId(InjectableType.CommandSender, <#= unityComponentDefinition.Id #>, <#= j #>)]
```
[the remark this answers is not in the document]
MPeti01: Got used to using `i` in the code and preferring `j` for the codegen meta loop... Might not actually have counted loops in the code here even when it's done though, as far as I can tell. Should I change it to `i`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectionConditionAttribute.cs
MPeti01: Right now no - we have a bit of a confusing mess as a result of Writers being Readers to the extent that they are just one class. Readers and Writers to the same component have the exact same `InjectableId` as a result but with different conditions... We could make them their own thing by separating them in `InjectableType`, and creating separate `InjectableCreators` (which could even just do the same thing for now), but then the dispatchers would have to query separately for readers and writers whenever they pass things for user callbacks. Maybe that's a change worth making all in all?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
MPeti01: True, the result of my indecision since I found "presence" a better description but longer so I kind of used it most the time except when the variable names were already very long... Should keep consistency though, will fix

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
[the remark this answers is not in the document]
jamiebrynes7: Oh right - its because there isn't a command index yet. I'd be careful assuming that index in the array = command index.

Once you have the Core merge you'll have the command index

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectionConditionAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        RequireNothing,
+        RequireComponentToRead,
+        RequireComponentWithAuthority
```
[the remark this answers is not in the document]
MPeti01: The specific component associated with the Injectable the tag is placed on. And true, should probably rename to `RequireComponentPresent`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectableId.cs
jamiebrynes7: Its rather unclear looking just at this object. Do we expect do have other injectable objects with extra information?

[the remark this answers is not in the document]
MPeti01: Yes, see reply above to the same line by Jonas

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiresMonoBehaviourActivationManager.cs
JonasImprobable: Ok, that sounds fair. Let's keep the current name. :)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectableId.cs
MPeti01: Not really, and we could change it if we do add extra info. It just felt weird specifying a field to a use case of a subset of these objects, but I guess it's still better than a random thing just hanging there like that?

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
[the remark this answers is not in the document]
MPeti01: Oh I was under the impression that assumption is safe. Making a note on the merge branch for this

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectableId.cs
JonasImprobable: Without understanding the implications of this decision to its full extent, my general opinion would be to not overload the use of a field unless necessary. Since that's not even the case right now and `index` is only ever used for command indices, let's make it unambiguous and rename it to `commandIndex`. That makes the intent clearer. We can also still refactor this later if it ever becomes necessary.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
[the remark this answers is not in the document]
jamiebrynes7: Actually - I think with the current impl `j != commandIndex` ever.

Because `commandIndex` starts at 1 :trollface:

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectionConditionAttribute.cs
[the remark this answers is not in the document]
jamiebrynes7: In which case - you'll need to adjust the `CommandSender` injectable. 

You don't need authority to send commands In fact, you don't even need to possess the component that the command exists in

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
```diff
@@ -0,0 +1,85 @@
...
+            internal class <#= commandName #>CommandRequestHandlerCreator : IInjectableCreator
+            {
+                public IInjectable CreateInjectable(Entity entity, EntityManager entityManager, ILogDispatcher logDispatcher)
+                {
+                    return new <#= commandName #>CommandRequestHandler(entity, entityManager, logDispatcher);
+                }
+            }
+
+            [InjectableId(InjectableType.CommandRequestHandler, <#= unityComponentDefinition.Id #>, <#= j #>)]
+            [InjectionCondition(InjectionCondition.RequireNothing)]
```
jamiebrynes7: To receive and respond to a request - you do need authority. That is - runtime will never route a command request to you if you aren't authoritative

MPeti01:
Regarding the Accessors namespace - I don't like it either, but this is actually something the users will have to type as part of the type after the `[Require]` tag so it needs to be reasonably short. I've been talking with Firtina and Sam Corcoran about ideas for those namespaces, and it seems we basically have a wide variety of pretty bad solutions to choose from... We need to have a discussion about how the ECS vs MonoBehaviour namespaces should be structured in general, because if we are serious about making the MonoBehaviours a first class workflow on the same level as ECS, the namespaces should reflect that

JonasImprobable:
Regarding `Accessors`: Ok, it looks like while we all agree that it's kinda bad but we also don't have any obviously better alternative either. In that case, I'm happy for that change to go in and maybe we'll revisit it again in the future.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
MPeti01: Indeed, I seem to have swapped the Sender and RequestHandler conditions somehow...

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/InjectionConditionAttribute.cs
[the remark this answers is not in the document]
MPeti01: Yes, seem to have mixed up the conditions for senders and req handlers somehow

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -48,22 +49,30 @@ internal class MonoBehaviourActivationManager
...
-                if (readerIds.Count + writerIds.Count > 0)
+                if (injector.HasRequiredFields(behaviourType))
                 {
-                    if (readerIds.Count > 0)
+                    var componentReadRequirements = injector.GetComponentPresenceRequirements(behaviourType);
+                    var componentAuthRequirements = injector.GetComponentAuthorityRequirements(behaviourType);
+                    var readRequirementCount = componentAuthRequirements.Count;
+                    var authRequirementCount = componentAuthRequirements.Count;
+
+                    if (readRequirementCount == 0 && authRequirementCount == 0)
```
JonasImprobable: Is this the case if a MonoBehaviour is not requiring anything at all? Is the idea here further that we pre-disable all monoBehaviours on a gameobject regardless of whether they require anything?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -88,8 +97,8 @@ public void EnableSpatialOSBehaviours()
         {
             foreach (var behaviour in behavioursToEnable)
             {
-                var componentIdsToReaderWriterLists = injector.InjectAllReadersWriters(behaviour, entity);
-                store.AddReaderWritersForBehaviour(behaviour, componentIdsToReaderWriterLists);
+                var injectablesFromIds = injector.InjectAllRequiredFields(behaviour, entity);
```
JonasImprobable: How about renaming this to `injectableIdToInjectables` or `injectedInjectables`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
MPeti01: This is the case of a MonoBehaviour requiring something, but only things that are always available (command senders / response handlers).

The idea is that for consistency we will manage preprocess and disable the MonoBehaviours if they Require anything, but in the case that all required fields are always available, we just immedieately enable them after all.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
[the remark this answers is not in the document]
JonasImprobable: 😱

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -120,14 +121,14 @@ namespace <#= qualifiedNamespace #>
                 var updateLists = ComponentsUpdatedComponentGroup.GetComponentArray<ComponentsUpdated<<#= componentDetails.TypeName #>.Update>>();
                 for (var i = 0; i < entities.Length; i++)
                 {
-                    var readerWriterStore = entityIdToReaderWriterStore[entities[i].Index];
-                    if (!readerWriterStore.TryGetReaderWritersForComponent(componentId, out var readers))
+                    var injectableStore = entityIdToInjectableStore[entities[i].Index];
+                    if (!injectableStore.TryGetInjectablesForComponent(reaederWriterInjectableId, out var readers))
```
JonasImprobable: Nit: How about renaming `readers` to `readersWriters` to stay consistent with the injectableId?

code_generator/GdkCodeGenerator/Templates/UnityReaderWriterGenerator.tt
```diff
@@ -21,111 +21,114 @@ namespace <#= qualifiedNamespace #>
...
 <# } else { #>
-        internal class ReaderWriterImpl :
-            NonBlittableReaderWriterBase<<#= componentDetails.TypeName #>, <#= componentDetails.TypeName #>.Update>, Reader, Writer
+            internal class ReaderWriterImpl :
+                NonBlittableReaderWriterBase<<#= componentDetails.TypeName #>, <#= componentDetails.TypeName #>.Update>, Reader, Writer
 <# } #>
-        {
-            public ReaderWriterImpl(Entity entity,EntityManager entityManager,ILogDispatcher logDispatcher)
-                : base(entity, entityManager, logDispatcher)
             {
```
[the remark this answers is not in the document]
MPeti01: Actually Git formatted the diff in a weird way here but the alignment seems fine to me. The bracket ends up belonging to a completely different scope than before...

code_generator/GdkCodeGenerator/src/Generation/SingleGenerationJob.cs
```diff
@@ -128,6 +129,15 @@ protected override void RunImpl()
...
                     readerWriterGenerator.Generate(component, package);
                 Content.Add(Path.Combine(relativeOutputPath, readerWriterFileName), readerWriterCode);
+
+                if (component.CommandDefinitions.Count > 0)
+                {
+                    var monobehaviourCommandHandlerFileName =
+                        Path.ChangeExtension($"{component.Name}MonoBehaviourCommandHandlers", fileExtension);
+                    var monobehaviourCommandHandlerCode =
+                        gameObjectCommandHandlersGenerator.Generate(component, package);
+                    Content.Add(Path.Combine(relativeOutputPath, monobehaviourCommandHandlerFileName), monobehaviourCommandHandlerCode);
```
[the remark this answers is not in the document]
MPeti01: Adding all the MonoBehaviour related files there, leaving other missing files to be merged from wherever it's being fixed

MPeti01:
Besides PR feedback changes, now only 1 sender, request handler and response handler is being generated for each component, since Jonas noted that this is what was actually in the design doc. This also meant that the `index` field could be removed from `InjectableId`.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -6,36 +6,37 @@
...
+    ///     are enabled , calling into the RequiredFieldInjector for injection, storing the created Injectables in the
+    ///     given ReaderWriterStore.
     /// </summary>
     internal class MonoBehaviourActivationManager
     {
         private readonly Entity entity;
         private readonly long spatialId;
 
-        private readonly Dictionary<uint, HashSet<MonoBehaviour>> behavioursRequiringReaderTypes
+        private readonly Dictionary<uint, HashSet<MonoBehaviour>> behavioursRequiringComponentsToRead
```
jamiebrynes7: Naming inconsistent with the `InjectionCondition` enum

PR #230, opened 2018-08-17
Adding support for IL2CPP to GDK

gkassabli:
Contributions: We are not currently taking public contributions - see our contributions policy. However, we are accepting issues and we do want your feedback.

Description
This adds assembly linker instructions to not strip classes needed for AOT and instantiation calls to methods that should be declared for AOT.
Original discussion in #143

Tests
Ensure iOS is built with these changes, performance doesn't regress and builds are done correctly

Documentation
https://brevi.link/il2cpp

Primary reviewers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
@@ -17,7 +17,9 @@ namespace Improbable.Gdk.Core
 
         // Here to prevent adding an action for the same type multiple times
         private readonly HashSet<Type> typesToRemove = new HashSet<Type>();
-        private readonly List<(ComponentGroup, Type)> componentGroupsToRemove = new List<(ComponentGroup, Type)>();
+        
+        private readonly List<(ComponentGroup, ComponentType)> componentGroupsToRemove = new List<(ComponentGroup, ComponentType)>();
+        private readonly List<(Entity, ComponentType)> componentsToRemove = new List<(Entity, ComponentType)>();
 
         protected override void OnCreateManager(int capacity)
         {
@@ -58,7 +60,7 @@ namespace Improbable.Gdk.Core
                         foreach (var componentType in componentCleanupHandler.CleanUpComponentTypes)
                         {
                             typesToRemove.Add(componentType.GetManagedType());
-                            componentGroupsToRemove.Add((GetComponentGroup(componentType), componentType.GetManagedType()));
+                            componentGroupsToRemove.Add((GetComponentGroup(componentType), componentType));
                         }
 
                         // Updates group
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -71,27 +60,32 @@ private void GenerateComponentGroups()
...
 
                     typesToRemove.Add(type);
-                    addRemoveComponentActionMethod.MakeGenericMethod(type).Invoke(this, null);
+                    componentGroupsToRemove.Add((GetComponentGroup(ComponentType.ReadOnly(type)), type));
                 }
             }
         }
 
-        private void AddRemoveComponentAction<T>()
+        private void RemoveComponents()
```
samiwh: Mild preference - can we move this after update. Was previously before because it was a use once function used only in CreateManager. Now it's used in update it would be nice if it was below that.
Our usual convention is actually private after protected though. Could also make these local functions. https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions

workers/unity/Packages/com.improbable.gdk.core/Attributes/RemoveAtEndOfTickAttribute.cs
```diff
@@ -8,7 +8,7 @@ namespace Improbable.Gdk.Core
     ///     Can be added to components extending <see cref="IComponentData" /> or <see cref="ISharedComponentData" />
     /// </summary>
     [AttributeUsage(AttributeTargets.Struct)]
-    public class RemoveAtEndOfTickAttribute : Attribute
+    public class RemoveAtEndOfTick : Attribute
```
samiwh: Re my previous comment. We can still call this `RemoveAtEndOfTickAttribute`, however when we annotate a class with it we can just use `[RemoveAtEndOfTick]`. This is just a C# convention.

gkassabli: Interesting, didn't know about that C# feature, will revert this change

workers/unity/Packages/com.improbable.gdk.core/Systems/CleanReactiveComponentsSystem.cs
```diff
@@ -134,5 +101,28 @@ private struct ComponentCleanup
             public ComponentGroup[] EventGroups;
             public ComponentGroup[] CommandsGroups;
         }
+
+        private void RemoveComponents()
+        {
+            var componentsToRemove = new List<(Entity, Type)>();
```
jamiebrynes7: Could this be a class variable and cleared at the end of each frame?

PR #231, opened 2018-08-17
Playground improvements + fixes

zeroZshadow:
#### Description
Movement system now keeps velocity state on cubes and filters on cubes only

ColorChange no longer triggers on TickSystem, as this produced issues. Uses real time.
Several fixes for best practise and preparation for filtering on changed.


workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -28,49 +34,51 @@ private struct Data
         {
             public readonly int Length;
             public EntityArray Entities;
-            public ComponentDataArray<SpatialOSLaunchable> Launchable;
+            [ReadOnly] public ComponentDataArray<SpatialOSLaunchable> Launchable; // Gets updated through PostUpdate
```
samiwh: Why are we doing that?

jared-improbable: Please put comments on the line above.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -28,49 +34,51 @@ private struct Data
...
+                        sender.RequestsToSend.Add(new Launcher.IncreaseScore.Request(secondLauncher,
                             new ScoreIncreaseRequest { Amount = 1 }));
+                        data.Sender[i] = sender;
+                        
+                        launchable.MostRecentLauncher = secondLauncher;
                     }
                     else
                     {
-                        launchable.MostRecentLauncher = EmptyEntityId;
+                        launchable.MostRecentLauncher = new EntityId(0);
```
jared-improbable: I'd keep `EmptyEntityId` here

workers/unity/Assets/Playground/Scripts/Cubes/CubeMovementSystem.cs
```diff
@@ -43,17 +42,21 @@ protected override void OnUpdate()
             for (var i = 0; i < data.Length; i++)
             {
                 var rigidbodyComponent = data.Rigidbody[i];
-                if (rigidbodyComponent.position.x - worker.Origin.x > 10)
+                var cubeComponent = data.Cube[i];
+
+                if (cubeComponent.Velocity.X > 0 && rigidbodyComponent.position.x - worker.Origin.x > 10)
```
samiwh: Ideally get rid of magic numbers. Although this doesn't matter too much in playground.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
```diff
@@ -8,35 +8,32 @@ public class ScoreCollision : MonoBehaviour
...
         {
-            component = GetComponent<SpatialOSComponent>();
-            if (!component)
-            {
-                Debug.LogError("Could not get SpatialOSComponent.");
-                return;
-            }
-
-            entityManager = component.World.GetExistingManager<EntityManager>();
+            Debug.LogError("Could not get SpatialOSComponent.");
```
jared-improbable: Is this useful to log? What can the user do about it?

schema/playground/cube.schema
```diff
@@ -0,0 +1,7 @@
+package playground;
+import "improbable/vector3.schema";
+
+component Cube {
+    id = 12008;
+    improbable.Vector3f velocity = 1;
```
samiwh: It isn't immediately obvious what this is. Calling it target_velocity would be a little better.

Generally a component should be the definition of some behaviour and the data definition should give you most of what you need to know about it . So having a `CubeTargetVelocity` component would probably make more sense. We could also add this to other entity types when we increase the scope of the playground.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
zeroZshadow: How so? It is the only place it is used, and there is no allocation since its a struct

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
zeroZshadow: @Ivaskovic You wrote this, why are we doing that?

Moved the comment

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
zeroZshadow: I'll remove the entire check then

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
jared-improbable: Because it looks a bit wrong to float a magic number around.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
Ivaskovic: Otherwise nothing got updated. Can't remember why, there used to be a comment that I deleted in one of the ancient commits.

workers/unity/Assets/Playground/Scripts/Cubes/ScoreCollision.cs
Ivaskovic: I'm pretty sure it used to make sense, but is not relevant anymore. Feel free to remove it.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
zeroZshadow: I'll add a `private const EntityId InvalidEntityId = new EntityId(0);`

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
Ivaskovic: Discussed this on Friday. The reason is that every collision is registered on two entities, and we want updates to happen only after `OnUpdate()`. This should have a one or two line comment briefly explaining it.

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
Ivaskovic: +1 for the `InvalidEntityId`

workers/unity/Assets/Playground/Scripts/Cubes/ProcessColorChangeSystem.cs
```diff
@@ -48,24 +48,24 @@ protected override void OnUpdate()
...
+                renderer.SetPropertyBlock(materialPropertyBlocks[lastEvent.Color]);
             }
         }
 
         private static void PopulateMaterialPropertyBlockMap(
             out Dictionary<Generated.Playground.Color, MaterialPropertyBlock> materialpropertyBlocks)
         {
-            materialpropertyBlocks = new Dictionary<Generated.Playground.Color, MaterialPropertyBlock>();
-            foreach (var keyValuePair in ColorMapping)
+            materialpropertyBlocks = new Dictionary<Generated.Playground.Color, MaterialPropertyBlock>(ColorMapping.Count);
```
Ivaskovic: In `TriggerColorChangeSystem` we have `using Color = Generated.Playground.Color`. Maybe use it here as well?

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -28,49 +33,56 @@ private struct Data
...
+                var otherOwner = otherLaunchable.MostRecentLauncher;
+
+                if (ourOwner == otherOwner)
                 {
-                    if (firstLauncher != EmptyEntityId)
+                    if (ourOwner.IsValid())
                     {
-                        data.Sender[i].RequestsToSend.Add(new Launcher.IncreaseScore.Request(
-                            firstLauncher, new Generated.Playground.ScoreIncreaseRequest { Amount = 1 }));
+                        Debug.Log("Another score increase requested");
```
martinzlocha: Did you mean to keep this or just a leftover log?

workers/unity/Assets/Playground/Scripts/Cubes/CollisionProcessSystem.cs
```diff
@@ -28,49 +33,56 @@ private struct Data
...
-                else if (secondLauncher != EmptyEntityId)
+                else if (otherOwner.IsValid())
                 {
-                    var launchable = data.Launchable[i];
-                    if (firstLauncher == EmptyEntityId)
+                    if (!ourOwner.IsValid())
                     {
-                        launchable.MostRecentLauncher = secondLauncher;
-                        data.Sender[i].RequestsToSend.Add(new Launcher.IncreaseScore.Request(secondLauncher,
+                        Debug.Log("3333 Another score increase requested");
```
martinzlocha: Same here.

workers/unity/Assets/Playground/Scripts/Cubes/TriggerColorChangeSystem.cs
```diff
@@ -37,22 +39,27 @@ protected override void OnCreateManager(int capacity)
...
         protected override void OnUpdate()
         {
-            if (World.GetExistingManager<TickSystem>().GlobalTick % 100 != 0)
+            if (Time.time < nextColorChange)
             {
                 return;
             }
 
-            var newColor = (Generated.Playground.Color) colorValues.GetValue(new Random().Next(colorValues.Length));
+            nextColorChange = Time.time + 2;
```
martinzlocha: Might be better to extract the 2 into a constant rather than having it as a magic number.

workers/unity/Assets/Playground/Scripts/Player/ProcessScoresSystem.cs
```diff
@@ -34,6 +35,7 @@ protected override void OnUpdate()
                 var playerScore = scoringData.Score[i];
                 foreach (var request in scoringData.CommandRequests[i].Requests)
                 {
+                    Debug.Log("Score increase");
```
martinzlocha: same here.

PR #232, opened 2018-08-17
UTY-945: Post-Merge Cleanups 

jamiebrynes7:
#### Description
This PR contains two changes requested during the review of the C# Core SDK - but were not essential and non-blocking.
1. Rename RawRequest and RawResponse to Payload in command structs.
2. Cache lists and maps during deserialization to avoid unnecessary function calls.
3. Driveby on unused interfaces

#### Tests
Tested the functionality of the GDK - all seems to work! 
#### Documentation
N/A
#### Primary reviewers
@samiwh @zeroZshadow (those that requested the changes)


code_generator/GdkCodeGenerator/Templates/UnityCommandPayloadGenerator.tt
```diff
@@ -85,8 +85,8 @@ namespace <#= qualifiedNamespace #>
                 public EntityId EntityId { get; }
                 public string Message { get; }
                 public StatusCode StatusCode { get; }
-                public <#= commandDetails.ResponseType #>? RawResponse { get; }
-                public <#= commandDetails.RequestType #> RawRequest { get; }
+                public <#= commandDetails.ResponseType #>? Payload { get; }
```
samiwh: Could we call this `ResponsePayload`? Or if we just call it payload then we could name the request `CorrespondingRequest`.

samiwh: Wild alternative is to call the `Reponse` `CommandResponse` and then have a `Response` as a field. I'm mostly just against using raw in the name for this case. If we can't decide on something then I'm happy to approve this though.

jamiebrynes7: I'm happy with `Payload` and `CorrespondingRequest` if you are

PR #234, opened 2018-08-20
changing which library we download for mac

jessicafalk:

-------

#### Description
Because the dylib didn't seem to load correctly for Mac.
Switched it to download the bundle instead.

#### Tests
ran it locally and the error that it can't download the worker library went away.

#### Documentation
no docs needed

#### Primary reviewers
@jamiebrynes7

### DIFF
--- a/workers/unity/Packages/manifest.json
+++ b/workers/unity/Packages/manifest.json
@@ -6,8 +6,8 @@
     "com.improbable.gdk.testutils": "file:com.improbable.gdk.testutils",
     "com.improbable.gdk.transformsynchronization": "file:com.improbable.gdk.transformsynchronization",
     "com.unity.burst": "0.2.4-preview.18",
-    "com.unity.entities": "0.0.12-preview.8",
-    "com.unity.incrementalcompiler": "0.0.42-preview.14",
+    "com.unity.entities": "0.0.12-preview.10",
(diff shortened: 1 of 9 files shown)


workers/unity/Packages/manifest.json
```diff
@@ -6,8 +6,8 @@
     "com.improbable.gdk.testutils": "file:com.improbable.gdk.testutils",
     "com.improbable.gdk.transformsynchronization": "file:com.improbable.gdk.transformsynchronization",
     "com.unity.burst": "0.2.4-preview.18",
-    "com.unity.entities": "0.0.12-preview.8",
-    "com.unity.incrementalcompiler": "0.0.42-preview.14",
+    "com.unity.entities": "0.0.12-preview.10",
```
jessicafalk: adding it into this pr, because it doesn't seem to break anything for us
BUT it looks like the mac crashes have disappeared (at least I haven't been able to reproduce them anymore in the last hour or so)

workers/unity/Packages/manifest.json
```diff
@@ -6,8 +6,8 @@
...
+    "com.unity.incrementalcompiler": "0.0.42-preview.18",
```
jamiebrynes7: Have we tested that these don't cause odd problems somewhere?

PR #235, opened 2018-08-20
Merge new core into MonoBehaviour support branch

MPeti01:

-------

#### Description
Merging new core into MonoBehaviours.
#### Tests
Tests green, Playground works fine.
#### Documentation
Not applicable
#### Primary reviewers
@zeroZshadow @jamiebrynes7


zeroZshadow:
Does this actually compile and run? I'm seeing it is missing the asmdef fixes we've had to do after the merge.

ReSharper2017.DotSettings
```diff
@@ -1,4 +1,6 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+
```
samiwh: Presumably dot settings files are a mistake

MPeti01:
If the commands groundwork gets merged before this, note that when merging that here we should make sure the codegen uses the new command indices rather than the one based on array iteration.

PR #236, opened 2018-08-20
(WIP) UTY-276: Add the ability to ack authority loss (v2).

martinzlocha:

-------

#### Description
Currently it is only possible to send authority loss imminent ack only using the C# SDK, added the functionality in the GDK as well.

Its easier to read in commit order to avoid having to scroll through a bunch of generated code.

Example generated code:

```csharp
            public override void SendAuthorityLossImminentAcknowledgement(ComponentGroup authorityLossComponentGroup, global::Improbable.Worker.Core.Connection connection)
            {
                var componentDataArray = authorityLossComponentGroup.GetComponentDataArray<AuthorityLossImminent<SpatialOSTransform>>();
                var spatialEntityIdData = authorityLossComponentGroup.GetComponentDataArray<SpatialEntityId>();
                for (int i = 0; i < componentDataArray.Length; i++)
                {
                    var component = componentDataArray[i];
                    if (componentDataArray[i].AuthorityLossAcknowledged && !component.AuthorityLossAcknowledgmentSent)
                    {
                        connection.SendAuthorityLossImminentAcknowledgement(spatialEntityIdData[i].EntityId, ComponentId);
                        component.AuthorityLossAcknowledgmentSent = true;
                        componentDataArray[i] = component;
                    }
                }
            }
```

#### Tests
#### Documentation
#### Primary reviewers

### DIFF
--- a/code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
+++ b/code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
@@ -303,8 +303,7 @@ namespace <#= qualifiedNamespace #>
                         }
 
                         entityManager.AddComponentData(entity, new AuthorityLossImminent<<#= componentDetails.TypeName #>> {
-                            AuthorityLossAcknowledged = false,
-                            AuthorityLossAcknowledgmentSent = false
+                            AuthorityLossAcknowledged = false
                         });
                         break;
                     case Authority.NotAuthoritative:
@@ -317,6 +316,11 @@ namespace <#= qualifiedNamespace #>
                         if (entityManager.HasComponent<AuthorityLossImminent<<#= componentDetails.TypeName #>>>(entity))
                         {
                             entityManager.RemoveComponent<AuthorityLossImminent<<#= componentDetails.TypeName #>>>(entity);
+                        }
+                        
+                        if (entityManager.HasComponent<AuthorityLossImminentSent<<#= componentDetails.TypeName #>>>(entity))
+                        {
+                            entityManager.RemoveComponent<AuthorityLossImminentSent<<#= componentDetails.TypeName #>>>(entity);
                         }
 
                         entityManager.RemoveComponent<Authoritative<<#= componentDetails.TypeName #>>>(entity);
@@ -470,6 +474,8 @@ namespace <#= qualifiedNamespace #>
 
         public class ComponentReplicator : ComponentReplicationHandler
         {
+            public override uint ComponentId => <#= componentDetails.ComponentId #>;
+
             public override ComponentType[] ReplicationComponentTypes => new ComponentType[] {
 <# foreach (var eventDetail in eventDetailsList) { #>
                 ComponentType.ReadOnly<EventSender.<#= eventDetail.EventName #>>(),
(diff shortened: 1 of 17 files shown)


code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -516,7 +524,7 @@ namespace <#= qualifiedNamespace #>
 
                     if (data.DirtyBit || dirtyEvents > 0)
                     {
-                        var update = new global::Improbable.Worker.Core.SchemaComponentUpdate(<#= componentDetails.ComponentId #>);
+                        var update = new global::Improbable.Worker.Core.SchemaComponentUpdate(ComponentId);
```
jamiebrynes7: Can you revert this? In-lining it avoids a function call in a potentially hot code path. 

The compiler may optimize it away, but I'm not certain

code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -484,6 +487,11 @@ namespace <#= qualifiedNamespace #>
                 ComponentType.ReadOnly<<#= qualifiedNamespace #>.<#= componentDetails.ComponentName #>.CommandResponders.<#= commandDetails.CommandName #>>(),
 <# } #>
             };
+            
+            public override ComponentType[] AuthorityLossComponentTypes => new ComponentType[] {
+                ComponentType.ReadOnly<AuthorityLossImminent<<#= componentDetails.TypeName #>>>(),
```
jamiebrynes7: This shouldn't be read only as you modify the data

code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
martinzlocha: I mean I can revert it but then we should change line 593 and 628 because its inconsistent. Also we should remove line 473 (`public override uint ComponentId => <#= componentDetails.ComponentId #>;`) where we create this variable.

jamiebrynes7: You need the variable to identify the component for use in the CustomReplicationSystem - that's really the only reason its there.

If its inconsistent then we should probably change it

code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
martinzlocha: Would the correct thing be to replace it by `typeof(AuthorityLossImminent<SpatialOSTransform>)`?

jamiebrynes7: `ComponentType.Create<AuthorityLossImminent<...>>`

code_generator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt
```diff
@@ -302,7 +302,10 @@ namespace <#= qualifiedNamespace #>
                             return;
                         }
 
-                        entityManager.AddComponentData(entity, new AuthorityLossImminent<<#= componentDetails.TypeName #>>());
+                        entityManager.AddComponentData(entity, new AuthorityLossImminent<<#= componentDetails.TypeName #>> {
+                            AuthorityLossAcknowledged = false,
+                            AuthorityLossAcknowledgmentSent = false
```
samiwh: As @jamiebrynes7 said, I'm in favour of not having more than one component for this. Ideally if you want to specify some transient other information I'd add another tag component, like `AuthorityLossAcked`. That way you can also predicate an inject group on it. I find it odd when you have a mix of components that we expect users to write to and us to write to and if a user writes to the wrong one it won't do what they expect.

samiwh:
Before this gets any further we may also want to think if any of the upcoming changes will affect this. Mostly the reactive component stuff that we'd like to get in. I suspect it won't change it much or at all, but it might do.

martinzlocha:
Will close this given that bridge v2 doesn't support authority loss and none of our code will use this.

PR #239, opened 2018-08-21
UTY-690, 691: CommandSenders and CommandRequestHandlers for MonoBehaviours

JonasImprobable:

-------

#### Description
Added codegened CommandRequestSenders and CommandRequestHandlers.
#### Tests
Simple playground scenario
#### Documentation
To be done later
#### Primary reviewers
@MPeti01 @firtina-improbable


code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
```diff
@@ -50,10 +78,44 @@ namespace <#= qualifiedNamespace #>
...
+                    this.entityManager = entityManager;
+                    this.logger = logger;
+                }
 
+<# foreach(var commandDetails in commandDetailsList) {
+           var commandRequestType = commandDetails.CommandName + ".ReceivedRequest";
+           var delegateList = commandDetails.CamelCaseName + "Delegates";
+#>
+                private readonly List<Action<<#= commandRequestType #>>> <#= delegateList #> = new List<Action<<#= commandRequestType #>>>();
+                public event Action<<#= commandRequestType #>> On<#= commandDetails.CommandName #>Request
```
MPeti01: Just to confirm, this callback signature is subject to change when sending responses comes in, right?

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
```diff
@@ -50,10 +78,44 @@ namespace <#= qualifiedNamespace #>
...
+                {
+                    add => <#= delegateList #>.Add(value);
+                    remove => <#= delegateList #>.Remove(value);
+                }
+
+                public void On<#= commandDetails.CommandName #>RequestInternal(<#= commandRequestType #> request)
```
MPeti01: As it is, this is visible to the user right? We could avoid that by having a user-facing interface type that includes the events above but not this method, that might be better

schema/playground/monobehaviour_test.schema
```diff
@@ -0,0 +1,9 @@
+package playground;
+
+type Void {}
+
+component MonoBehaviourTest
```
MPeti01: Could have a descriptive name, maybe SpinnerRotation?

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
```diff
@@ -8,33 +8,61 @@
...
 
+<# foreach(var commandDetails in commandDetailsList) { #>
+                public void Send<#= commandDetails.CommandName #>Request(EntityId entityId, <#= commandDetails.RequestType #> request)
+                {
+                    try
+                    {
+                        var ecsCommandRequestSender = entityManager.GetComponentData<CommandSenders.<#= commandDetails.CommandName #>>(entity);
+                        ecsCommandRequestSender.RequestsToSend.Add(new <#= commandDetails.CommandName #>.Request(entityId, request));
+                    }
+                    catch (Exception e)
```
samiwh: What are we expecting this to catch?

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
samiwh: Why not just make this internal? User facing interface types aren't something I'm fond of.

MPeti01: Because for some reason I forgot they are in a different assembly. That definitely makes things easier :D

JonasImprobable: Yeah, this is an oversight on my part. I will think about a way to hide this method from users.

JonasImprobable: I made the `On<CommandName>RequestInternal` method internal. That seems to do the trick.

schema/playground/monobehaviour_test.schema
JonasImprobable: Renamed component to `SpinnerRotation`.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
JonasImprobable: There is nothing that we are looking for here in particular. Sending commands should always work under normal conditions. Admittedly, I added this in to mirror the approach we have in `ReaderWriterBase`. Happy to remove this if you think it's overkill.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/ToggleRotationCommandSender.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            if (reader != null) // TODO UTY-791: Needed until prefab preprocessing is implemented, remove as part of UTY-791
+            {
+                ownEntityId = GetComponent<SpatialOSComponent>().SpatialEntityId;
+            }
+        }
+
+        private void Update()
+        {
+            if (reader.Authority != Authority.NotAuthoritative)
```
MPeti01: At this point it's worth considering if we want to have separate prefabs for server vs client side, which is something we could do with the existing PrefabConfig... I personally think this is OK for now to keep it simple for quick tests though.

JonasImprobable:
New PR to be raised again master branch. Closing this.

PR #240, opened 2018-08-21
UTY-879: Test component update callbacks on Readers in playground

MPeti01:

-------

#### Description
Made player character briefly flash red when it receives updates to the PlayerInput component, and otherwise be yellow when sprinting, white by default.

To make the visuals meaningful, changed the movement system to only send PlayerInput updates when they would actually change the state.
#### Tests
updates
#### Documentation
N/A
#### Primary reviewers
@JonasImprobable @jamiebrynes7


MPeti01:
Accidentally took Github's advice and opened PR from wrong branch as a result, can't seem to edit that part so opening a new, correct one instead...

PR #241, opened 2018-08-21
UTY-879: Testing component update callbacks on readers in Playground

MPeti01:

-------

#### Description
Made player character briefly flash red when it receives updates to the PlayerInput component, and otherwise be yellow when sprinting, white by default.

To make the visuals meaningful, changed the movement system to only send PlayerInput updates when they would actually change the state.
#### Tests
updates
#### Documentation
N/A
#### Primary reviewers
@JonasImprobable @jamiebrynes7


workers/unity/Assets/Playground/Scripts/MonoBehaviours/PlayerColor.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                reader.ComponentUpdated -= OnUpdateReceived;
+                reader.RunningUpdated -= OnRunningChanged;
+            }
+        }
+
+        private void Awake()
+        {
+            renderer = transform.Find("Shape").GetComponent<MeshRenderer>();
+            Debug.Log(renderer);
```
jamiebrynes7: Couple debug logs in this file

workers/unity/Assets/Playground/Scripts/MonoBehaviours/PlayerColor.cs
```diff
@@ -0,0 +1,106 @@
...
+            if (reader != null) // TODO remove in UTY-791
+            {
+                reader.ComponentUpdated -= OnUpdateReceived;
+                reader.RunningUpdated -= OnRunningChanged;
+            }
+        }
+
+        private void Awake()
+        {
+            renderer = transform.Find("Shape").GetComponent<MeshRenderer>();
```
jamiebrynes7: Needs null checks - both on the `Find` and the `GetComponent<>` calls.

Alternative, use null conditional operators and lose a bit of debug information

```
renderer = transform.Find("Shape")?.GetComponent<MeshRenderer>();
if (renderer == null)
{
	// Either the Find or GetComponent failed
}
```

workers/unity/Assets/Playground/Scripts/MonoBehaviours/PlayerColor.cs
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+            else
+            {
+                isRunning = false;
+                if (!isFlashing)
+                {
+                    renderer.SetPropertyBlock(basicMaterial);
+                }
+            }
```
jamiebrynes7: This block can be simplified I think: 

```
isRunning = isRunningUpdate;
if (!isFlashing)
{
	renderer.SetPropertyBlock(isRunningUpdate ? runningMaterial : basicMaterial);
}
```

workers/unity/Assets/Playground/Scripts/MonoBehaviours/PlayerColor.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        [SerializeField] private float flashTime = 0.05f;
+
+        [Require] PlayerInput.Requirables.Reader reader;
+
+        private MeshRenderer renderer;
+        private bool isRunning = false;
+        private bool isFlashing = false;
+        private float flashStartTime;
```
jamiebrynes7: Nit: Add units onto the time related variables

workers/unity/Assets/Playground/Scripts/MonoBehaviours/PlayerColor.cs
```diff
@@ -0,0 +1,106 @@
...
+            basicMaterial.SetColor("_Color", Color.white);
+
+            flashingMaterial = new MaterialPropertyBlock();
+            flashingMaterial.SetColor("_Color", Color.red);
+
+            runningMaterial = new MaterialPropertyBlock();
+            runningMaterial.SetColor("_Color", Color.yellow);
+        }
+
+        [SerializeField] private float flashTime = 0.05f;
```
jamiebrynes7: Maybe rename to `flashTimeDuration` ?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/PlayerColor.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                isFlashing = false;
+                if (isRunning)
+                {
+                    renderer.SetPropertyBlock(runningMaterial);
+                }
+                else
+                {
+                    renderer.SetPropertyBlock(basicMaterial);
+                }
```
jamiebrynes7: Use a ternary here

workers/unity/Assets/Playground/Scripts/Player/LocalPlayerInputSync.cs
```diff
@@ -34,15 +35,22 @@ protected override void OnUpdate()
...
-                {
-                    Horizontal = input.x,
-                    Vertical = input.z,
-                    Running = Input.GetKey(KeyCode.LeftShift)
-                };
+                var oldPlayerInput = playerInputData.PlayerInput[i];
 
-                playerInputData.PlayerInput[i] = newPlayerInput;
+                if (Math.Abs(oldPlayerInput.Horizontal - input.x) > 0.01f
+                    || Math.Abs(oldPlayerInput.Vertical - input.z) > 0.01f
```
jamiebrynes7: Extract `0.01f` to a private const

samiwh:
Can we do something less garish? This is pretty unpleasant to look at. As an aside the spinners should be removed before merging into core. Their fine as something for you to test locally but they aren't great as an addition to the playground, even if we want to remove it later.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/PlayerColor.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        private void Awake()
+        {
+            logger = GetComponent<SpatialOSComponent>()?.LogDispatcher;
+            renderer = transform.Find("Shape")?.GetComponent<MeshRenderer>();
+            if (logger == null)
+            {
+                Debug.LogError("PlayerColor behaviour on GameObject that is not connected to Spatial systems properly");
+            }
+            else if (renderer == null)
```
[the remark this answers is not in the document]
MPeti01: It's `else if` so that I don't have to handle the logger not existing - I could add a case where I log the renderer error to Debug.LogError but I'm not sure it's worth the extra case handling for that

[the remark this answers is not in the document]
jamiebrynes7: Maybe - the current implementation means you might need to run twice to find and fix all the errors. Its no biggie - but could be a minor annoyance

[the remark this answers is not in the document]
MPeti01: True, but it seems sufficiently unlikely for both to happen that I feel the small mental overhead of reading through two kinds of logging would be a tradeoff for the worse. It's not a strong conviction though, so say if you disagree and I'll change it

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -54,7 +54,7 @@ internal class MonoBehaviourActivationManager
                 {
                     var componentReadRequirements = injector.GetComponentPresenceRequirements(behaviourType);
                     var componentAuthRequirements = injector.GetComponentAuthorityRequirements(behaviourType);
-                    var readRequirementCount = componentAuthRequirements.Count;
+                    var readRequirementCount = componentReadRequirements.Count;
```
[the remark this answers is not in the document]
MPeti01: Yes

workers/unity/Assets/Playground/Scripts/MonoBehaviours/ProcessSpinnerColorChange.cs
```diff
@@ -0,0 +1,77 @@
...
+        if (colorReader != null && collisionsReader != null) // TODO UTY-791: Needed until prefab preprocessing is implemented, remove as part of UTY-791
+        {
+            collisionsReader.OnPlayerCollided += HandleCollisionEvent;
+            colorReader.ColorUpdated += HandleColorChange;
+        }
+    }
+
+    private void Awake()
+    {
+        renderer = gameObject.GetComponent<MeshRenderer>();
```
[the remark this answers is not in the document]
MPeti01: Ok, I'll just do local error logging in this case because then I don't have to null check the logger as well, and it's a local rendering effect that's influenced anyway

PR #244, opened 2018-08-22
UTY-830:

MPeti01:

-------

#### Description
Reworked order of enabling MonoBehaviours and calling user callbacks based on discussions with @samiwh 

Basic goals: Writers should not get authority change callbacks, Writers should not receive updates from different workers. Simple compromise we made to do this without lots of weird edge cases that could end up surprising users: Readers/Writers don't get update callbacks on the tick when they are enabled.

#### Tests
No additional tests, Playground still works
#### Documentation
N/A
#### Primary reviewers
@samiwh @JonasImprobable


code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -192,16 +234,80 @@ namespace <#= qualifiedNamespace #>
...
+            {
+                if (AuthorityLostComponentGroup.IsEmptyIgnoreFilter)
+                {
+                    return;
+                }
+
+                var entities = AuthorityLostComponentGroup.GetEntityArray();
+                var changeOpsLists = AuthorityLostComponentGroup.GetComponentDataArray<AuthorityChanges<<#= componentDetails.TypeName #>>>();
+
+                // Call once on all entities unless they flip-flopped back into the state they started in
```
jared-improbable: Semantically, wouldn't you want to invoke all expected callbacks, even if they ultimately change back? cc @samiwh @jamiebrynes7

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -192,16 +234,80 @@ namespace <#= qualifiedNamespace #>
...
+                    {
+                        foreach (Requirables.ReaderWriterImpl readerWriter in readersWriters)
+                        {
+                            readerWriter.OnAuthorityChange(Authority.NotAuthoritative);
+                        }
+                    }
+                }
+            }
+
+            private bool isFirstAuthChange(Authority authToMatch, AuthorityChanges<<#= componentDetails.TypeName #>> changeOps)
```
jared-improbable: Style: Upper-case to "IsFirstAuthChange"
This whole ignoring of multiple auth changes in a frame is dubious to me, however.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -90,21 +104,42 @@ namespace <#= qualifiedNamespace #>
...
+                    return;
+                }
+
+                var authoritiesChangedTags = AuthorityGainedComponentGroup.GetComponentDataArray<AuthorityChanges<<#= componentDetails.TypeName #>>>();
+                var entities = AuthorityGainedComponentGroup.GetEntityArray();
+                for (var i = 0; i < entities.Length; i++)
+                {
+                    var activationManager = entityIndexToManagers[entities[i].Index];
+                    // Call once except if flip-flopped back to starting state
+                    if (isFirstAuthChange(Authority.Authoritative, authoritiesChangedTags[i]))
```
jared-improbable: See comment below.

code_generator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
MPeti01: This is necessary with the approach we are taking to the order of passing ops and enabling behaviours in the MonoBehaviour layer - we are not retaining the information about the exact order of all ops received in the GDK Core, and as a result the MonoBehaviour layer performs all enables at a different time than all disables. So not squashing them could mean receiving multiple disables and then multiple enables for a component rather than alternating them.

samiwh: ^
We could invoke all callbacks if we preserve the op order in the core and provide that to the monobehaviour layer. Which we can do if it becomes essential. But that comes with it's own disadvantages of making users think the state of the worker is something other than it is. Right now there is one race we haven't dealt with when you receive authority cycles however that can be fixed by adding to this later (wasn't considered essential for the first set of corrections).

MPeti01:
Cherry picked on top of latest master.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -101,27 +103,53 @@ protected override void OnUpdate()
...
 
             foreach (var indexManagerPair in entityIndexToActivationManager)
             {
-                indexManagerPair.Value.DisableSpatialOSBehaviours();
+                indexManagerPair.Value.EnableSpatialOSBehaviours();
+            }
+
+            foreach (var gameObjectComponentDispatcher in GameObjectComponentDispatchers)
+            {
+                gameObjectComponentDispatcher.MarkAuthorityGainedForActivation(entityIndexToActivationManager);
```
JonasImprobable: Why is this done twice?

MPeti01: Fixed

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -65,13 +65,15 @@ private void GenerateComponentGroups()
...
-                    gameObjectComponentDispatcher.ComponentsUpdatedComponentGroup =
-                        GetComponentGroup(gameObjectComponentDispatcher.ComponentsUpdatedComponentTypes);
-                }
+                gameObjectComponentDispatcher.AuthorityGainedComponentGroup =
+                    GetComponentGroup(gameObjectComponentDispatcher.AuthorityGainedComponentTypes);
+                gameObjectComponentDispatcher.AuthorityLostComponentGroup =
+                    GetComponentGroup(gameObjectComponentDispatcher.AuthorityLostComponentTypes);
+                gameObjectComponentDispatcher.AuthorityLossImminentComponentGroup =
+                    GetComponentGroup(gameObjectComponentDispatcher.AuthorityLossImminentComponentTypes);
+                gameObjectComponentDispatcher.ComponentsUpdatedComponentGroup =
```
JonasImprobable: It looks like the previous `if (gameObjectComponentDispatcher.ComponentsUpdatedComponentTypes.Length > 0)` part has been removed. Is that intentional?

MPeti01: Not sure why I did that, reverting

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityGameObjectComponentDispatcherGenerator.tt
```diff
@@ -192,16 +234,80 @@ namespace <#= qualifiedNamespace #>
...
+                {
+                    if (auth != Authority.AuthorityLossImminent) // not relevant
+                    {
+                        return auth == authToMatch;
+                    }
+                }
+                return false;
+            }
+
+            public override void InvokeOnAuthorityLossImminentCallbacks(Dictionary<int, InjectableStore> entityIndexToInjectableStore)
```
JonasImprobable: Nit: Could you place this function below the `InvokeOnAuthorityLostCallbacks()` function? Basically flip the order of this function an `IsFirstAuthChange()`. That way, similarly functions are grouped together.

PR #247, opened 2018-08-22
Changes to the worker abstraction and user flow

samiwh:
Sort of in commit order (would advise reading it that way). Tried to at least get rid of some of the noise from prefabs and codegen. A lot less code has changed than the numbers suggest.

Mild annoyance - the AddCoreSystems function moves in the commit that deals with codegen. Also the typo is fixed there. Couldn't be moved before codegen is touched.

#### Description
Round of changes for the worker abstraction stuff. This should change most of the things that we wanted to and allow for future changes to be made more easily.
Such future work could include making a pure ecs worker and pulling this out into a feature module (surprisingly little would have to be done), or working with additive scenes (tested it and it sort of works), or making fake players share static terrain.

Workers are still created from the object Worker, however changes have been made to exactly how they are instantiated, and prefabs added to aid in creating them in playground. Adding ClientWorker or GameLogicWorker prefabs to a scene will cause it to try to create that worker on start.

Building workers now builds a specific scene with one of the appropriate prefabs in it. So building a client is the same as building with the ClientScene. The build config has been changed to reflect this.

Scenes added
Client scene:
image

GameLogic scene:
image

Sample scene:
image

#### Tests
Run locally in editor and multiple external workers. Run in cloud with clients via the launcher. Spatial connect will definitely not work right now as the external IP can't be used.
#### Documentation
Mostly not. However we also currently have no docs on this and it isn't the priority.


workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -10,77 +10,137 @@ namespace Improbable.Gdk.Core
...
+            return connectionFailureReason;
+        }
+
+        private void AddCoreSystems()
+        {
+            var workerSystem = World.GetOrCreateManager<WorkerSystem>();
+            workerSystem.Connection = Connection;
+            workerSystem.LogDispatcher = LogDispatcher;
+            workerSystem.Origin = Origin;
+            workerSystem.WorkerType = WorkerType;
```
jessicafalk: we are now duplicating fields between two classes and the user can modify them in the system. This could cause quite a few problems when the user suddenly modifies one of them..

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/InterpolateTransformSystem.cs
```diff
@@ -32,8 +32,7 @@ protected override void OnCreateManager(int capacity)
         {
             base.OnCreateManager(capacity);
 
-            var worker = Core.Worker.GetWorkerFromWorld(World);
-            origin = worker.Origin;
+            origin = World.GetExistingManager<WorkerSystem>().Origin;
```
jessicafalk: this will simply fail with a null pointer exception, I believe. do we not want to provide a better exception for this?

workers/unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -545,7 +545,7 @@ PlayerSettings:
   webGLCompressionFormat: 1
   webGLLinkerTarget: 0
   scriptingDefineSymbols:
-    1: UNITY_POST_PROCESSING_STACK_V2;UNITY_DISABLE_AUTOMATIC_SYSTEM_BOOTSTRAP
+    1: UNITY_POST_PROCESSING_STACK_V2;UNITY_DISABLE_AUTOMATIC_SYSTEM_BOOTSTRAP;IMPROBABLE_WORKERTYPE_UNITYCLIENT
```
jessicafalk: do we need that extra symbol?

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
```diff
@@ -0,0 +1,127 @@
...
+            yield return null;
+        }
+
+        if (workerTask.IsFaulted)
+        {
+            Debug.LogError(
+                $"Failed to create client worker with error: {workerTask.Exception.InnerException.Message}");
+            if (attempts > 0)
+            {
+                CreateWorker(attempts);
```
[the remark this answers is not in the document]
samiwh: This is just in playground. We can make the retries more complicated but there probably isn't any point.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/InterpolateTransformSystem.cs
samiwh: Can do. Can have a convention for feature modules that we check for there being a worker.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
samiwh: I think this is a fridge error. We can make them readonly to prevent that, but if a user goes out of their way to break things then things will break. Things are duplicated because one is a wrapper around the other. We can make it so that the Worker class has properties to call these on each access if we need them to be the same but it makes cleanup slightly different.

workers/unity/Assets/Playground/Scripts/PlaygroundOneTimeInitialisation.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+using Improbable.Gdk.PlayerLifecycle;
+using Unity.Entities;
+using UnityEngine;
+
+namespace Playground
+{
+    public static class PlaygroundOneTimeInitialisation
+    {
+        private static bool initialized = false;
```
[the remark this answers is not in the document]
samiwh: It doesn't need to persist across an app domain reload, just across scene transitions. Will check this though.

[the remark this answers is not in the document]
samiwh: Might be a reason to change the name.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                WorkerType = workerType,
+                WorkerId = $"{workerType}-{Guid.NewGuid()}"
+            };
+            workerTask = Worker.CreateWorkerAsync(config, new ForwardingDispatcher(), Origin);
+        }
+        else
+        {
+            var commandLineArguments = Environment.GetCommandLineArgs();
+            Debug.LogFormat("Command line {0}", string.Join(" ", commandLineArguments.ToArray()));
```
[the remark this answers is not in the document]
samiwh: Debug prints can't all be removed though. In this case there isn't a worker yet so we can't push this through the logger. We can change how loggers are created to get around this but we need work out what to do with connections.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
```diff
@@ -0,0 +1,127 @@
...
+    {
+        return deploymentList.Deployments[0].DeploymentName;
+    }
+
+    private static bool HandleQueueStatus(QueueStatus queueStatus)
+    {
+        return true;
+    }
+
+    private void CreateWorker(int attempts)
```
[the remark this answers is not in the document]
samiwh: We can't use await in a non async function.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
```diff
@@ -0,0 +1,127 @@
...
+    {
+        while (!workerTask.IsCompleted)
+        {
+            yield return null;
+        }
+
+        if (workerTask.IsFaulted)
+        {
+            Debug.LogError(
+                $"Failed to create client worker with error: {workerTask.Exception.InnerException.Message}");
```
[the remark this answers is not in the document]
samiwh: We know quite a lot about this task though and how it can end. What kind of failures aren't handled here?

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
```diff
@@ -0,0 +1,127 @@
...
+
+    private void InstantiateLevel()
+    {
+        Level.transform.position = Origin;
+        Level.SetActive(true);
+    }
+
+    private void OnDisconnected(string reason)
+    {
+        ClientWorker.LogDispatcher.HandleLog(LogType.Warning, new LogEvent($"Worker disconnected")
```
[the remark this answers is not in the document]
samiwh: It's a warning because it's more important than a regular message. Can downgrade it to log if we need to.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -10,77 +10,137 @@ namespace Improbable.Gdk.Core
...
+                }
+
+                return deploymentList;
+            }
+        }
+
+        private static string GetConnectionFailureReason(Connection connection)
+        {
+            var connectionFailureReason = "No reason found for connection failure";
+            var dispatcher = new Dispatcher();
```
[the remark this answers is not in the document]
samiwh: Missed that. Will fix.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -10,77 +10,137 @@ namespace Improbable.Gdk.Core
...
+            dispatcher.OnDisconnect(op => connectionFailureReason = op.Reason);
+            dispatcher.Process(connection.GetOpList(0));
+
+            return connectionFailureReason;
+        }
+
+        private void AddcoreSystems()
```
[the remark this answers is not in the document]
samiwh: See comment in the description. This is fixed already.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
jared-improbable: Right, but you don't need to. It will be printed to the raw log by Unity itself.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
samiwh: I don't think there are any races here but will look into making async work nicely here. Partly didn't do that as I wanted the monobehaviour to not be magical and async await in that context is weird to read if you haven't seen it before.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
```diff
@@ -0,0 +1,127 @@
...
+    public Worker ClientWorker;
+
+    protected void Start()
+    {
+        CreateWorker(ConnectionAttempts);
+    }
+
+    protected void OnApplicationQuit()
+    {
+        ClientWorker?.Dispose();
```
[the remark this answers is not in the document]
jared-improbable: I generally think it's a good idea to highlight that classes contain special resources, even if the code isn't "live" wrt its lifecycle.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
jared-improbable: My thought was for people basing their project on these examples.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/InterpolateTransformSystem.cs
samiwh: I don't think it's worth going full error handling in transforms right now though as that module is going to just be re-written.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
samiwh: Sure will add this to the drive by changes.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
jared-improbable: Sorry, I meant deadlock: https://olitee.com/2015/01/c-async-await-common-deadlock-scenario/
I know in this case it's explicitly checked, but it's easy to forget.
We'll be using `async/await` when interacting with the Platform APIs, so at some point devs will need to learn.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
jared-improbable: If there's no InnerException it will crash? :P

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
samiwh: Makes sense. In which case I agree this should be a separate task. Currently parsing what error you got and how to react to it is hard :S. Will make a ticket.

workers/unity/Assets/Playground/Config/SystemConfig.cs
```diff
@@ -64,5 +40,13 @@ public static void AddGameLogicSystems(World world)
             world.GetOrCreateManager<ProcessScoresSystem>();
             world.GetOrCreateManager<CollisionProcessSystem>();
         }
+
+        private static void AddCommonSystems(World world)
```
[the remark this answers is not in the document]
samiwh: Changed it to AddLifecycleSystems and removed the entity manager as it doesn't need to be there.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
samiwh: I mean is there a case where this can fail with no inner exception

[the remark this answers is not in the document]
jared-improbable: I don't know...do we want to find it in production?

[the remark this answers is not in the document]
samiwh: My understanding is that IsFaulted can not be called without Exception being non null. And the exception is always an AggregateException of inner exceptions that were not handled. The other option is that it was cancelled however that can't happen in this case. Although I haven't explicitly given it a non cancellable token, but can do if we are worried about that.

It seems like you can default construct aggregate exception although I thought Task doesn't do this.

Will add a check for this anyway.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
samiwh: Cool. Will take another look at this then.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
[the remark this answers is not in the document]
samiwh: No wait I didn't miss that. The new dispatcher isn't disposable.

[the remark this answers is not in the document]
jared-improbable: D'oh, it's all user-side now, isn't it...

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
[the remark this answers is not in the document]
samiwh: Going to move all this to async await so this is moot now, but noted for later in case we need to explicitly poll a task for some reason.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
samiwh: Changed it to use a constructor so it can be readonly.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Threading.Tasks;
+using Improbable.Gdk.Core;
+using Improbable.Worker.Core;
+using Playground;
+using Unity.Entities;
+using UnityEngine;
+
+public class ConnectClient : MonoBehaviour
+{
+    public Vector3 Origin = Vector3.zero;
```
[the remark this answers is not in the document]
callumlawson: That could be good! 
In the future, you can imagine it being really nice to be able to move workers around just by dragging prefabs.
We could even visualize spatial extents, should they have them.

workers/unity/ProjectSettings/ProjectSettings.asset
jamiebrynes7: They know!

workers/unity/Assets/Playground/Config/SystemConfig.cs
```diff
@@ -13,7 +13,7 @@ public static class SystemConfig
 
         public static void AddClientSystems(World world)
         {
-            AddCommonSystems(world);
+            AddLifecycelSystems(world);
```
jamiebrynes7: `Lifecycle`

workers/unity/ProjectSettings/ProjectSettings.asset
zeroZshadow: Actually the build system added that, but, it shouldn't...
This is literally unset in a `finally` block.. I have no idea how that gets there

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+        else
+        {
+            var commandLineArguments = Environment.GetCommandLineArgs();
+            Debug.LogFormat("Command line {0}", string.Join(" ", commandLineArguments.ToArray()));
+            var commandLineArgs = CommandLineUtility.ParseCommandLineArgs(commandLineArguments);
+            if (commandLineArgs.ContainsKey(RuntimeConfigNames.LoginToken))
+            {
+                var config = LocatorConfig.CreateConnectionConfigFromCommandLine(commandLineArgs);
+                config.WorkerType = workerType;
```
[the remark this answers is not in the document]
jessicafalk: no, the launcher does not pass it in as it assumes it will be a UnityClient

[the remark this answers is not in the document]
jamiebrynes7: That is enormously frustrating.

workers/unity/Assets/Playground/Config/SystemConfig.cs
```diff
@@ -64,5 +40,12 @@ public static void AddGameLogicSystems(World world)
             world.GetOrCreateManager<ProcessScoresSystem>();
             world.GetOrCreateManager<CollisionProcessSystem>();
         }
+
+        private static void AddLifecycelSystems(World world)
```
DanailPenev: AddLifecycleSystems

workers/unity/Assets/Playground/Scripts/Worker/ConnectGameLogicWorker.cs
```diff
@@ -7,74 +7,85 @@
...
+    {
+        while (attempts > 0)
         {
-            Debug.LogError(
-                $"Failed to create game logic worker with error: {workerTask.Exception.InnerException.Message}");
-            if (attempts > 0)
+            try
             {
-                CreateWorker(attempts);
+                var worker = await Worker.CreateWorkerAsync(config, dispatcher, origin);
```
callumlawson: Minor thing, I was confused by CreateWorkerAsync as I thought for a second that it was the method at the top of the file.
Tbh that's on me, but if we wanted to be extra clear, the private method in this file could be
'private ConnectGameLogicWorkerAsync" showing it's the specialised version.

workers/unity/Assets/Playground/Config/SystemConfig.cs
samiwh: Yep, two typos in there.

workers/unity/Assets/Playground/Scripts/Worker/ConnectGameLogicWorker.cs
```diff
@@ -0,0 +1,109 @@
...
+using Improbable.Gdk.Core;
+using Playground;
+using Unity.Entities;
+using UnityEngine;
+
+public class ConnectGameLogicWorker : MonoBehaviour
+{
+    public Vector3 Origin = Vector3.zero;
+    public GameObject Level;
+    public int ConnectionAttempts = 3;
```
callumlawson: Nit ConnectionAttempts -> MaxConnectionAttempts 
(Given it's not a count of the number of elapsed connection attempts)

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -10,7 +10,7 @@ namespace Improbable.Gdk.Core
 {
     public class Worker : IDisposable
     {
-        public readonly Vector3 Origin;
+        public Vector3 Origin => World.GetExistingManager<WorkerSystem>().Origin;
```
jessicafalk: I guess that was an unintended change?

samiwh: Yep, that wasn't meant to be committed.

workers/unity/Assets/Playground/Scripts/PlaygroundOneTimeInitialisation.cs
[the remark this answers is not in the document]
samiwh: We've discussed doing that a few times. We'd need to add that to all systems though. And all user systems. So we'd get rid of this at the expense of more boilerplate everywhere else.

workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerDisconnectCallbackSystem.cs
```diff
@@ -0,0 +1,27 @@
...
+            [ReadOnly] public ComponentDataArray<WorkerEntityTag> DenotesWorkerEntity;
+        }
+
+        public event Action<string> OnDisconnected;
+
+        [Inject] private Data data;
+
+        protected override void OnUpdate()
+        {
+            OnDisconnected?.Invoke(data.DisconnectedMessage[0].ReasonForDisconnect);
```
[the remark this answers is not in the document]
samiwh: The main risk is there being multiple entities

workers/unity/Assets/Playground/Scripts/Worker/ClientWorkerConnector.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    protected override void AddWorkerSystems()
+    {
+        WorkerUtils.AddClientSystems(Worker.World);
+    }
+
+    protected override string SelectDeploymentName(DeploymentList deployments)
+    {
+        // This could be replaced with a splash screen asking to select a deployment or some other user defined logic
+        return deployments.Deployments[0].DeploymentName;
```
jessicafalk: no need to override this. It does exactly the same as the `WorkerConnectorBase`

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -0,0 +1,227 @@
...
+                var origin = transform.position;
+                if (ShouldUseLocator())
+                {
+                    var config = GetLocatorConfig(workerType);
+                    ConnectionDelegate connectionDelegate = async () =>
+                        await Worker.CreateWorkerAsync(config, SelectDeploymentName, logger, origin)
+                            .ConfigureAwait(false);
+                    var worker =
+                        await ConnectWithRetries(connectionDelegate, MaxConnectionAttempts, logger, workerType);
+                    InitializeWorker(worker);
```
jessicafalk: these two lines can be moved outside of the if-clause

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -0,0 +1,227 @@
...
+                    return;
+                }
+            }
+
+            try
+            {
+                var origin = transform.position;
+                if (ShouldUseLocator())
+                {
+                    var config = GetLocatorConfig(workerType);
```
jessicafalk: if we are going to move this to Core, we need to make getting the configs overridable. with mobile for example we can't parse command line arguments, but must pass the values in differently (e.g. through a splash screen)

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        private bool ShouldUseLocator()
+        {
+            if (Application.isEditor)
+            {
+                return false;
+            }
+
+            var commandLineArguments = Environment.GetCommandLineArgs();
+            var commandLineArgs = CommandLineUtility.ParseCommandLineArgs(commandLineArguments);
```
jessicafalk: the other option would be to factor these two lines out as they are used in three places anyway and make that method `virtual`

workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
```diff
@@ -11,12 +11,31 @@ namespace Improbable.Gdk.Core
...
+            set
+            {
+                if (connection == null)
+                {
+                    handlingUnityLogMessages = true;
+                    Application.logMessageReceived += LogCallback;
+                }
+                else if (value == null)
+                {
+                    handlingUnityLogMessages = false;
```
jessicafalk: what is this bool needed for?

workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
```diff
@@ -11,12 +11,31 @@ namespace Improbable.Gdk.Core
...
+
+        public Connection Connection
+        {
+            get => connection;
+            set
+            {
+                if (connection == null)
+                {
+                    handlingUnityLogMessages = true;
+                    Application.logMessageReceived += LogCallback;
```
jessicafalk: doesn't this mean that we won't log it to the Unity console anymore, if we don't have a connection?

workers/unity/Packages/com.improbable.gdk.core/Systems/WorldCommandsSendSystem.cs
```diff
@@ -64,7 +65,7 @@ protected override void OnCreateManager(int capacity)
 
         protected override void OnUpdate()
         {
-            if (worker.Connection == null)
+            if (connection == null)
```
jessicafalk: is this even needed? When creating a worker, a connection has to be present.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -10,76 +11,137 @@ namespace Improbable.Gdk.Core
...
-            worker.Dispose();
+            World.CreateManager<WorkerSystem>(Connection, LogDispatcher, WorkerType, Origin);
+            World.GetOrCreateManager<SpatialOSSendSystem>();
+            World.GetOrCreateManager<SpatialOSReceiveSystem>();
+            World.GetOrCreateManager<CleanReactiveComponentsSystem>();
+            World.GetOrCreateManager<WorldCommandsCleanSystem>();
+            World.GetOrCreateManager<WorldCommandsSendSystem>();
+            World.GetOrCreateManager<CommandRequestTrackerSystem>();
+
+            // Monobehaviouor stuff
```
jessicafalk: typo :P

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -10,76 +11,137 @@ namespace Improbable.Gdk.Core
...
 
         public void Dispose()
         {
-            LogDispatcher.Dispose();
-            EntityMapping.Clear();
             World?.Dispose();
             World = null;
-            OnDisconnect?.Invoke(this);
-            ConnectionUtility.Disconnect(Connection);
+            LogDispatcher.Dispose();
```
jessicafalk: change to 
```
LogDispatcher?.Dispose();
LogDispatcher = null;
```

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/InterpolateTransformSystem.cs
jessicafalk: that was more a remark for all systems, just happened to comment on it in the transform system

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -10,77 +10,137 @@ namespace Improbable.Gdk.Core
...
-            World = new World(WorkerId);
-            World.CreateManager<WorkerSystem>(this);
-            var entityManager = World.GetOrCreateManager<EntityManager>();
-            WorkerEntity = entityManager.CreateEntity(typeof(OnConnected), typeof(WorkerEntityTag));
-            OnConnect?.Invoke(this);
+            World = new World(Connection.GetWorkerId());
+            AddCoreSystems();
+
+            // This isn't a core system, this is for an easy disconnect event
+            disconnectCallbackSystem = World.GetOrCreateManager<WorkerDisconnectCallbackSystem>();
```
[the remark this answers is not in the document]
jessicafalk: But we also have the `OnDisconnect event in this class. If a system wants to react to a worker disconnecting, it can just inject the `OnDisconnect` tag.
If a monobehaviour wants to react to it, it will use the event from this class. So I don't see a reason to add events to systems.

jessicafalk:
Need to take another look at it later, but I get the feeling that quite a bit of the connection logic should be moved to the `Worker`. the current flow forces the user to use Monobehaviours, which can be a good thing, but it doesn't allow them to simply use pure ECS to handle the overall flow unless they shuffle the code around themselves.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
samiwh: This isn't in core.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
samiwh: I'm not sure I follow. We should remove most of the command line stuff and replace it with something leaner, but I'm not sure what would need to be virtual here.

This was mostly copying the existing code. We can change how this is done here, but I consider that a drive by.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
samiwh: Yep forgot to do that after rewriting the connect function.

workers/unity/Packages/com.improbable.gdk.core/Systems/WorldCommandsSendSystem.cs
samiwh: No idea. This is just me fixing up the code that depends on this. I didn't add this check.

samiwh: Can add it to the drive by list.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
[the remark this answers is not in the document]
samiwh: Then we have to chain together the disconnect events. This class can't know about a disconnection without a system telling it. This is less redundancy than the alternative.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/InterpolateTransformSystem.cs
samiwh: Sure. But I don't think that is part of this PR.

workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
samiwh: I'm going to rewrite all of this though now that we've made it disposable. Previously I'd made the worker responsible for the lifecycle of the logger.

samiwh:
@jessicafalk what connection logic needs to be moved to the Worker? All the helper functions are already there? Do you mean the example connector in playground? If so then it definitely should not be in Worker.

workers/unity/Assets/Playground/Scripts/Worker/ConnectClient.cs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    protected void OnApplicationQuit()
+    {
+        ClientWorker?.Dispose();
+    }
+
+    private static string SelectDeploymentName(DeploymentList deploymentList)
+    {
+        return deploymentList.Deployments[0].DeploymentName;
```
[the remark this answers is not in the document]
samiwh: Actually we also guarantee that this will work.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -142,6 +149,7 @@ public void Dispose()
             World?.Dispose();
             World = null;
             LogDispatcher.Dispose();
```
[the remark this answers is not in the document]
samiwh: Good spot

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        private static async Task<Worker> ConnectWithRetries(ConnectionDelegate connectionDelegate, int attempts,
+            ILogDispatcher logger, string workerType)
+        {
+            while (attempts > 0)
+            {
+                try
+                {
+                    var worker = await connectionDelegate();
```
jared-improbable: `return await connectionDelegate();`

workers/unity/Assets/Playground/Scripts/Worker/ClientWorkerConnector.cs
```diff
@@ -0,0 +1,26 @@
...
+using Improbable.Worker.Core;
+using Playground;
+
+public class ClientWorkerConnector : WorkerConnectorBase
+{
+    private async void Start()
+    {
+        // Array covariance means the explicit cast isn't strictly needed but it is slightly safer
+        RequiredWorkerConnection = FindObjectsOfType<GameLogicWorkerConnector>()
+            .Select(c => (WorkerConnectorBase) c).ToArray();
```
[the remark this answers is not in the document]
jared-improbable: The comment isn't that illuminating, really.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -0,0 +1,225 @@
...
+            }
+
+            try
+            {
+                var origin = transform.position;
+                ConnectionDelegate connectionDelegate;
+                if (ShouldUseLocator())
+                {
+                    var config = GetLocatorConfig(workerType);
+                    connectionDelegate = async () =>
```
jared-improbable: These are a bit of an eyeful, but could be dealt with at a later date.

samiwh: We can make them inline functions or make them point to functions.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
samiwh: Yep can change to that.

PR #256, opened 2018-08-23
Serialization codegen fix for enum fields

MPeti01:

-------

#### Description
Fixed bug in serialization codegen where enum fields updates would not be received.
#### Tests
Fixed issue in MonoBehaviour Playgrounds feature I'm working on in other branch
#### Documentation
N/A
#### Primary reviewers
@jamiebrynes7 @samiwh


MPeti01:
Indeed, we only have an enum in the nested type test among all tests... I assume there should be one on the blittable component, and a few on the exhaustive test? I could add it in this PR but I'm not 100% sure exactly where

PR #259, opened 2018-08-23
Fixed eager MPB creation

firtina-improbable:

-------

#### Description
Not so eager now!

Unity doesn't like it when you create some types too eagerly. Especially material things.
#### Tests
Unity does not complain any more when these scripts are loaded or prefabs with these scripts are selected.

#### Primary reviewers
@MPeti01 @samiwh


JonasImprobable:
On a side note, is there no easier way to do this? Can't we just change the color of a material directly?

JonasImprobable:
Thanks for referencing the PR.

PR #263, opened 2018-08-27
Some clean up and preparation for IL2CPP

jessicafalk:

-------

#### Description
in commit order:
1. upgrading packages to get rid of obsolete warnings
2. fixing up our scripts to get rid of warnings
3. Getting rid of injection hooks. This allows us to have the default world.
4. adding the `DisableAutoCreation` attribute to each system to ensure they are not added to the default world
5. getting rid of using reflection for a generic method (important for IL2CPP) and refactoring everything around it. Also contains an example of how to use the newly introduced `IBufferElementData` by changing the `BufferedTransform` from a `Component` to a `IBufferElementData`.
6. the entities test package seems to work again. So adding it back in

Basically I've just been trying to clean up the code a bit and doing some work to allow IL2CPP without any messy hacks. Clean up itself is not really finished yet (still want to get rid of the `ViewCommandBuffer), but it's a first pass.

IL2CPP doesn't work yet with that PR, but it does get rid of some errors.

#### Tests
ran it locally. it worked
Tests are also passing

#### Documentation
no docs

#### Primary reviewers


workers/unity/Assets/Playground/Scripts/Camera/FollowCameraSystem.cs
```diff
@@ -22,6 +22,7 @@ namespace Playground
     ///     ScrollWheel: controls the orbit radius, moving the camera closer to or
     ///     farther from the player.
     /// </remarks>
+    [DisableAutoCreation]
```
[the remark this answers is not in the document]
jessicafalk: we might not want all systems to have these. It' more about letting the user decide whether to add a system automatically to the default world or not.

Though I agree that we might not want to do this. This commit is simply in here to show we could get rid of our reflection stuff. If people, prefer to do that instead of using the way that Unity currently proposes (which is, to be fair, quite cumbersome) then I'm fine with that as well ;)

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/InterpolateTransformSystem.cs
```diff
@@ -59,36 +59,38 @@ protected override void OnUpdate()
...
             {
-                var transformQueue = transformData.BufferedTransform[i].TransformUpdates;
-                if (transformQueue.Count == 0)
+                var transformQueue = transformData.BufferedTransform[i];
+                if (transformQueue.Length == 0)
                 {
                     continue;
                 }
 
+                var nextTransform = transformQueue[0].transformUpdate;
```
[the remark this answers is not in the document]
jessicafalk: ah good spot, will change that :)

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/ApplyTransformUpdatesSystem.cs
```diff
@@ -24,7 +25,13 @@ protected override void OnUpdate()
             for (var i = 0; i < transformUpdateData.Length; i++)
             {
                 var transformUpdates = transformUpdateData.TransformUpdate[i].Updates;
-                var lastTransformSnapshot = transformUpdateData.BufferedTransform[i].LastTransformSnapshot;
```
martinzlocha: What if there is nothing in the list and then we receive an update without all of the fields set? I don't think you can remove the `LastTransformSnapshot`.

jessicafalk: the `LastTransformSnapshot`initializes with the default `SpatialOSTransform`afaik, so the behaviour is exactly the same as before.

I agree that there are issues with this, but I don't think they are introduced in this PR.

jessicafalk: Though I can change the code in this system to work correctly even without `LastTransformSnapshot` :) will add a new commit

martinzlocha: Nice change :)

workers/unity/Assets/Playground/Config/ArchetypeConfig.cs
```diff
@@ -9,28 +10,32 @@ public static class ArchetypeConfig
...
+                case SystemConfig.UnityClient when archetype == CubeArchetype:
+                case SystemConfig.UnityClient when archetype == SpinnerArchetype:
+                    commandBuffer.AddBuffer<BufferedTransform>(entity);
+                    break;
+                case SystemConfig.UnityGameLogic when archetype == CharacterArchetype:
+                case SystemConfig.UnityGameLogic when archetype == CubeArchetype:
+                case SystemConfig.UnityGameLogic when archetype == SpinnerArchetype:
+                    commandBuffer.AddBuffer<BufferedTransform>(entity);
+                    break;
+                default:
```
jamiebrynes7: Given that we do the same thing for all archetype and worker types (currently) - can't we just have all cases all through to the final one?

workers/unity/Assets/Playground/Scripts/EntityInitialization/ArchetypeInitializationSystem.cs
```diff
@@ -67,47 +55,9 @@ protected override void OnUpdate()
...
-                    else
-                    {
-                        viewCommandBuffer.AddComponent(entity, componentType, componentInstance);
-                    }
-                }
-            }
-
-            viewCommandBuffer.FlushBuffer();
+                
+                ArchetypeConfig.AddComponentDataForArchetype(worker, archetypeName, PostUpdateCommands, entity);
```
jamiebrynes7: It feels strange that we call methods on a "config" directly.

workers/unity/Packages/com.improbable.gdk.legacy/Editor/BuildSystem/Util/SpatialCommand.cs
```diff
@@ -19,9 +19,6 @@ public sealed class SpatialCommand
 
         internal const string usrLocalBin = "/usr/local/bin";
 
-        private string spatialLocation;
-        private string oldSpatialLocation;
-        private string discoveredLocation;
```
jamiebrynes7: I think there's an open ticket for this which you can close if its merged in here

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -18,10 +18,6 @@ public class Bootstrap : MonoBehaviour
 
         public void Awake()
         {
-            // Taken from DefaultWorldInitalization.cs
-            SetupInjectionHooks(); // Register hybrid injection hooks
-            PlayerLoopManager.RegisterDomainUnload(DomainUnloadShutdown, 10000); // Clean up worlds and player loop
```
jamiebrynes7: Is this no longer required?

jamiebrynes7: How do we have a world per worker without the injection hooks?

workers/unity/Assets/Playground/Scripts/Camera/FollowCameraSystem.cs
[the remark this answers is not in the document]
zeroZshadow: Hopefully these can go away when unity gives better support for multiple worlds in a game.

workers/unity/Assets/Playground/Scripts/EntityInitialization/ArchetypeInitializationSystem.cs
DanailPenev: @jamiebrynes7 Don't we do the same with `SystemConfig` as well?

jessicafalk: I think we do, but we shouldn't to be fair. Need to think for better names for these

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
jessicafalk: these are just the hooks that were set up when registering the default world. As we disabled the default world, the hooks were never initialized and we had to do it on our own. Now that the default world gets created, we don't need to set them up anymore. the `DefaultWorldInitialization` takes care of that

workers/unity/Assets/Playground/Scripts/Bootstrap.cs
```diff
@@ -64,44 +60,20 @@ public void Awake()
...
-            InjectionHookSupport.RegisterHook(
-                (InjectionHook) Activator.CreateInstance(componentArrayInjectionHookType));
+            var worker = Worker.Connect(config, new ForwardingDispatcher(), origin);
+            Instantiate(Level, origin, Quaternion.identity);
+            SystemConfig.AddSystems(worker.World, config.WorkerType);
+            Workers.Add(worker);
         }
 
-        public static void DomainUnloadShutdown()
+        private void OnApplicationQuit()
```
zeroZshadow: This isn't always called and will totally leak workers

workers/unity/Assets/Playground/Scripts/EntityInitialization/ArchetypeInitializationSystem.cs
jessicafalk: actually we only use that method in one place, so might just move it directly into the system

zeroZshadow:
Its not about turning the default world on. Its no longer requiring users to turn their default work off.

jessicafalk:
So there are three paths we could take with the whole default world on/off discussion
1. leave it as it is (not a big fan of that)
2. turn on the default world. This allows users to use the default world, but will most likely introduce some gotchas for users as well. They (and we) will have to always think about adding the `DisableAutoCreation` attribute to any SpatialOS-related system. However, we stop restricting users to having to create any world manually and adding the right systems to them. They can just use the Default World for anything that is not related to SpatialOS.
3. move the injection hook initialization into Core. That way, users can't use the default world, but need to specify which systems should belong to each world. However, they at least don't have to care about the hooks anymore.

In the end, the user will either have to explicitly disable or enable a system for a world, so the question is what would we prefer?

jamiebrynes7:
@jessicafalk For the change to the ArchetypeConfig - you'll need to update `entity-checkout-process.md`

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Components/BufferedTransform.cs
```diff
@@ -1,16 +1,11 @@
 using Generated.Improbable.Transform;
+using Unity.Entities;
 using UnityEngine;
 
 namespace Improbable.Gdk.TransformSynchronization
 {
-    public class BufferedTransform : Component
+    public struct BufferedTransform : IBufferElementData
```
zeroZshadow: You might need to add `[InternalBufferCapacity(somevalue)]` to indicate how many elements we expect

jessicafalk: added it

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransfromSynchronizationSystemHelper.cs
```diff
@@ -4,6 +4,8 @@ namespace Improbable.Gdk.TransformSynchronization
 {
     public static class TransformSynchronizationSystemHelper
     {
+        public const int MaxBufferSize = 4;
```
samiwh: This shouldn't really be here as it's not a system helper. This is more feature module config.

PR #267, opened 2018-08-28
Local development workflow

gkassabli:

-------

#### Description
This PR adds support for local iOS workflow (iOS simulator) + adds code to connect to the local deployment. It also works around the crash when activating input field on iOS by allowing to use connect button with hardcoded IP address
#### Tests
Run local deployment + iOS simulator and ensure everything have started as expected
#### Documentation
To be added in a separate task


workers/unity/Packages/com.improbable.gdk.core/Config/ReceptionistConfig.cs
```diff
@@ -9,6 +9,7 @@ public class ReceptionistConfig : ConnectionConfig
 
         // Default Android emulator host IP alias
         private const string AndroidEmulatorHostDeviceIp = "10.0.2.2";
+        private const string LocalHostIp = "127.0.0.1";
```
jessicafalk: we already store that ip in the RuntimeConfigDefaults so no need to redefine it :)

DanailPenev:
The other thing that I was thinking about is that the `ReceptionistConfig`s for both platform might be better suited in their respective packages rather than the core but we can create a separate PR refactoring these.

gkassabli:
@DanailPenev I agree, but there is a strong chance we will change connection system for mobile later on. Considering that I would argue for keeping this simple until we know where the changes will be heading

workers/unity/Packages/com.improbable.gdk.core/Config/ReceptionistConfig.cs
```diff
@@ -48,5 +48,16 @@ public static ReceptionistConfig CreateConnectionConfigForPhysicalAndroid(string
...
             };
         }
+
+        public static ReceptionistConfig CreateConnectionConfigForiOS()
+        {
+            return new ReceptionistConfig
+            {
+                // By default this config only connects to local IP (127.0.0.1)
+                // Replace this code with hardcoded IP to coneect to non-local deployment
+                ReceptionistHost = RuntimeConfigNames.ReceptionistHost,
```
jessicafalk: `RuntimeConfigDefaults` not `RuntimeConfigNames`

gkassabli: Oops, wrong copypaste :)

workers/unity/Packages/com.improbable.gdk.core/Config/ReceptionistConfig.cs
```diff
@@ -48,5 +48,16 @@ public static ReceptionistConfig CreateConnectionConfigForPhysicalAndroid(string
...
             };
         }
+
+        public static ReceptionistConfig CreateConnectionConfigForiOS()
+        {
+            return new ReceptionistConfig
+            {
+                // By default this config only connects to local IP (127.0.0.1)
+                // Replace this code with hardcoded IP to coneect to non-local deployment
+                ReceptionistHost = RuntimeConfigDefaults.ReceptionistHost,
```
jessicafalk: actually, I think we can remove this line because it will be set to it by default anyway. sorry about that :)
we only need the `UseExternalIp`

gkassabli: Make sense, didn't think about the default config. Thanks for flagging

workers/unity/Packages/com.improbable.gdk.core/Config/ReceptionistConfig.cs
```diff
@@ -48,5 +48,16 @@ public static ReceptionistConfig CreateConnectionConfigForPhysicalAndroid(string
...
             };
         }
+
+        public static ReceptionistConfig CreateConnectionConfigForiOS()
+        {
+            return new ReceptionistConfig
+            {
+                // By default this config only connects to local IP (127.0.0.1)
+                // Uncomment the following line with hardcoded IP to connect to non-local deployment
+                // ReceptionistHost = "Your IP address",
```
jessicafalk: it's already merged in, but we shouldn't have commented out code like this. Also, there is no need to hardcode the IP for a non-local deployment. If you are connecting to a cloud deployment, you won't use the Receptionst flow, but the Locator flow.

gkassabli: Sorry, poor wording. This is actually a convenience shortcut for device client to be able to connect to any deployment - for now we only have workaround for input field crash.
I am happy to remove this if you think we shouldn't have the code commented out

PR #273, opened 2018-08-29
UTY-595: Upgrade to Bridge v2

jamiebrynes7:
#### Description
Upgraded to Bridge v2 in both current configurations. Also fixed a strange bug that presented itself whereby `BufferdTransform.LastTransformSnapshot` was uninitialised - so if you received an update with no location or rotation you would end up with bogus data being propagated to the interpolation system. 
#### Tests
1. Ran local deployment
2. Ran cloud deployment, connected client and ran around.
3. @DanailPenev Said he had tested mobile with bridge v2
#### Documentation
N/A - not necessary
#### Primary reviewers
@samiwh

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.transformsynchronization/Components/BufferedTransform.cs
+++ b/workers/unity/Packages/com.improbable.gdk.transformsynchronization/Components/BufferedTransform.cs
@@ -1,4 +1,5 @@
 using Generated.Improbable.Transform;
+using Improbable.Gdk.Core;
 using UnityEngine;
 
 namespace Improbable.Gdk.TransformSynchronization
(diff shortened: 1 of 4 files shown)


workers/unity/Packages/com.improbable.gdk.transformsynchronization/Components/BufferedTransform.cs
```diff
@@ -7,6 +8,7 @@ public class BufferedTransform : Component
     {
         public System.Collections.Generic.List<SpatialOSTransform> TransformUpdates;
         public SpatialOSTransform LastTransformSnapshot;
+        public BlittableBool IsInitialised;
```
samiwh: Is this actually bridge related? Is it sending us partial updates or something?

jamiebrynes7: `bridge_v2` has diffing turned on by default (as of today)

PR #275, opened 2018-08-29
Feature/uty 950 kill prepare workspace

jared-improbable:
* Remove need for prepare-workspace.
* Put `spatial build build-config` and `spatial local launch` into Unity - no more CLI!
* TODO: `cloud upload` and `cloud launch`
* No longer aggressively unzip packages every time codegen occurs, since it will fail once you've entered Play mode once.


workers/unity/Packages/com.improbable.gdk.legacy/Editor/BuildSystem/Util/SpatialCommands.cs
```diff
@@ -5,17 +5,11 @@ namespace Improbable.Gdk.Legacy.BuildSystem.Util
 {
     internal static class SpatialCommands
     {
-        #region Build Configuration
-
         internal static void GenerateBuildConfiguration()
         {
-            SpatialCommandRunner.RunSpatialCommand("build build-config", "generate build configuration");
+         //   SpatialCommandRunner.RunSpatialCommand("build build-config", "generate build configuration");
```
jared-improbable: Will fix these up tomorrow.

workers/unity/Packages/com.improbable.gdk.tools/.DownloadCoreSdk/Program.cs
```diff
@@ -13,11 +13,12 @@ private static void Main(string[] args)
...
             if (args.Length == 0)
             {
-                Console.Error.WriteLine("Specify the version of the CoreSdk to download.");
+                Console.Error.WriteLine("<path_to_spatial> <coresdk_version>");
                 Environment.Exit(1);
             }
 
-            var coreSdkVersion = args[0];
+            var spatialPath = args[0];
+            var coreSdkVersion = args[1];
```
zeroZshadow: This can throw an OutOfRangeException

workers/unity/Packages/com.improbable.gdk.tools/ShowProgressBarScope.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using UnityEditor;
+
+namespace Improbable.Gdk.Tools
+{
+    internal struct ShowProgressBarScope : IDisposable
```
zeroZshadow: This is a bit odd. You cannot actually nest scopes with this, as `ClearProgressBar` will close the one progress bar that `DisplayProgressBar` reuses.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        [MenuItem("Improbable/Spatial/Build worker configs")]
+        private static void BuildConfigMenu()
+        {
+            BuildConfig();
+        }
+
+        [MenuItem("Improbable/Spatial/Local launch")]
+        private static void LauncMenu()
```
DanailPenev: Typo in method name.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -0,0 +1,116 @@
...
+{
+    public static class LocalLaunch
+    {
+        private static readonly string
+            SpatialProjectRootDir = Path.GetFullPath(Path.Combine(Application.dataPath, "..", "..", ".."));
+
+        // Windows: The exit code is 0xc000013a when the user closes the console window, or presses Ctrl+C.
+        private const int WindowsCtrlCExitCode = -1073741510;
+
+        // Unixy: The exit code is 128 + SIGINT (2).
```
martinzlocha: Typo: "Unix"

jamiebrynes7: Its a portmanteau - "Unity" + "Unix" 😉

martinzlocha: :smooth:

workers/unity/Packages/com.improbable.gdk.tools/.DownloadCoreSdk/Program.cs
jared-improbable: Fixed the range check above.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
jared-improbable: Hehe, it was actually meant to mean "Unix-like", I'll just change it to that.

workers/unity/Packages/com.improbable.gdk.tools/ShowProgressBarScope.cs
jared-improbable: Yep, but it makes the showing/restoring of the progress window exception-proof and removes the need to do a `try/finally`. I don't mind removing it if you like, though.

zeroZshadow: I think it'll be fine.
Here is to hoping unity will someday make a better system for displaying work in progress

workers/unity/Packages/com.improbable.gdk.tools/.DownloadCoreSdk/Program.cs
```diff
@@ -11,13 +11,14 @@ internal class Program
     {
         private static void Main(string[] args)
         {
-            if (args.Length == 0)
+            if (args.Length != 2)
             {
-                Console.Error.WriteLine("Specify the version of the CoreSdk to download.");
+                Console.Error.WriteLine("<path_to_spatial> <coresdk_version>");
```
jamiebrynes7: Shouldn't this say something like `Usage: <path_to_spatial> <coresdk_version>` ?

workers/unity/Packages/com.improbable.gdk.tools/Common.cs
```diff
@@ -67,52 +81,37 @@ public static string GetThisPackagePath()
...
             {
-                Debug.LogError(args.Data);
+                var testPath = Path.Combine(path, fileName);
+                if (File.Exists(testPath))
+                {
+                    return testPath;
+                }
             }
+
+            return string.Empty;
```
jamiebrynes7: `return splitPaths.Select(path => Path.Combine(path, fileName)).FirstOrDefault(testPath => File.Exists(testPath)) ?? String.Empty;` 😉 

`FirstOrDefault` will return null if its not there - so `?? String.Empty` ensures we return a valid string.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -1,53 +1,127 @@
...
+        ///     by Unity.
+        /// </summary>
+        [MenuItem(DownloadForceMenuItem, true, DownloadForcePriority)]
+        private static bool DownloadForceValidate()
         {
-            var path = Path.Combine("build", "CoreSdk");
-            if (Directory.Exists(path))
+            var failedPlugins = new List<PluginImporter>();
+
+            // Unity never unloads native plugins. Detect them here and give a more useful error message.
```
jamiebrynes7: Afaict - this doesn't enable you to print an error message - just early out from the command executing. Are we able to have a custom message on validate failed?

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -1,53 +1,127 @@
...
+                return DownloadResult.AlreadyInstalled;
+            }
+
+            return Download();
+        }
+
+        /// <summary>
+        ///     Downloads and extracts the Core Sdk and associated libraries and tools.
+        /// </summary>
+        public static DownloadResult Download()
```
jamiebrynes7: Does this and the function above it need to be public?

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -0,0 +1,122 @@
...
+                UseShellExecute = true,
+                WorkingDirectory = SpatialProjectRootDir
+            };
+
+            var process = Process.Start(processInfo);
+
+            if (process == null)
+            {
+                return;
+            }
```
jamiebrynes7: What's would cause process to be null immediately after start?

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -0,0 +1,122 @@
...
+                    Debug.LogError($"Could not find a spatial log file in {logPath}.");
+                    return;
+                }
+
+                var content = File.ReadAllText(latestLogFile.FullName);
+                var message = $"Loaded from {latestLogFile.FullName}\n{content}";
+
+                if (WasProcessKilled(process))
+                {
+                    Debug.Log(message);
```
jamiebrynes7: Do we want to log the entire spatial log in the event you hit Ctrl + C?

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
jared-improbable: Ah yeah, this is a stale comment. I could also leave the command enabled and print an error message (what I originally did.)
I've gone back and forth between disabling and printing.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
jared-improbable: Not any more!

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
jared-improbable: I wasn't sure; we could just print the path to the file instead of the whole thing

jamiebrynes7: Yeah I think that would be preferable - it will spam your console with things you probably don't care about.

Something like: `"spatial local launch" finished. View the log here - <log_path>`

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
jamiebrynes7: Ah right - do we want to log that something went wrong?

jared-improbable: Yep, done!

PR #285, opened 2018-08-31
adding known issue about il2cpp

jessicafalk:

-------

#### Description
Added a known issue about IL2CPP on mac

#### Primary reviewers
@ElleEss


docs/known-issues.md
```diff
@@ -11,3 +11,4 @@ Known issue = any major user-facing bug or lack of user-facing feature that:
 | Issue | Date added | Ticket | Workaround? | Fixed? |
 |-------|-------------------|--------|-------------|-------|
 | The Unity editor randomly crashes on MacOS when using ECS preview packages. This is not caused by the SpatialOS GDK. | 2018/07/25 | 1064084 | Simply retry. | No|
+| IL2CPP does not currently work on Mac as it will crash upon starting the game. This is not caused by the SpatialOS GDK. | 2018/08/31 | 1076596 | Use the Mono compiler. | No|
```
jared-improbable: Is this reported to Unity?

jessicafalk: yep, created a bug report. The ticket number stands for the bug report ticket with unity

jamiebrynes7: Do we not want to give a short explanation of why it crashes? Or it that not our policy?

ElleEss: Is it worth saying in the column that 1076596 is a Unity ticket for future-self and any users to follow?

"...as it will crash upon starting the game."
->
"...as it crashes upon starting the game." 

"Will" is OK (not technically wrong) but we don't need it and the text is stronger without it - 'will' sort of reads like a prophecy.

ElleEss: "because Unity" is probs good enough?

jessicafalk: we could change the header from `Ticket` to `Unity Ticket`?

PR #288, opened 2018-08-31
UTY-835; Invalidate deinjected requirables

JonasImprobable:

-------

#### Description
- Requirables can now be disposed. Once disposed you cannot access them anymore. (Read component state, register/deregister callbacks).
- Requirables are disposed upon losing a component/authority and before the gameobject is removed.
- Removed MonoBehaviourActivationManagerInitializationSystem and merged its logic into the EntityGameObjectLinker.
- You can now Unlink gaemobjects.
#### Tests
Only manual tests for now for the sake of shipping this in time. Proper tests shall be done after the alpha release
#### Documentation
To be done in a follow-up PR.
#### Primary reviewers
@MPeti01 @firtina-improbable


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -177,5 +166,18 @@ public void CreateActivationManagerAndReaderWriterStore(Entity entity)
             var manager = new MonoBehaviourActivationManager(gameObject, injector, store, logger);
             entityToActivationManager.Add(entity, manager);
         }
+
+        internal void RemoveActivationManagerAndReaderWriterStore(Entity entity)
+        {
+            if (!entityToActivationManager.ContainsKey(entity))
```
jared-improbable: Use `TryGet` to avoid a double lookup here and below.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -105,25 +105,48 @@ public void EnableSpatialOSBehaviours()
             foreach (var behaviour in behavioursToEnable)
             {
                 behaviour.enabled = true;
```
jared-improbable: Can exceptions propagate out of the `OnEnable` method of the MonoBehaviour?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -105,25 +105,48 @@ public void EnableSpatialOSBehaviours()
...
+            foreach (var behaviour in behaviours)
             {
                 injector.DeInjectAllRequiredFields(behaviour);
                 store.RemoveInjectablesForBehaviour(behaviour);
             }
 
-            behavioursToDisable.Clear();
+            // Make a copy because the behaviours argument could potentially be the enabledBehaviours set itself.
+            var removedBehaviours = new List<MonoBehaviour>(behaviours);
+            foreach (var removedBehaviour in removedBehaviours)
```
jared-improbable: Should these be removed in reverse order?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -68,6 +68,26 @@ private IInjectable Inject(MonoBehaviour behaviour, InjectableId injectableId, E
...
+            EnsureLoaded(behaviourType);
+            foreach (var fieldsToComponents in fieldInfoCache[behaviourType])
+            {
+                var fields = fieldsToComponents.Value;
+                foreach (var field in fields)
+                {
+                    var oldField = field.GetValue(behaviour);
+                    if (!(oldField is RequirableBase))
+                    {
+                        Debug.LogError("Something is wrong " + behaviour);
```
jared-improbable: Gonna need to be more specific that that :)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -68,6 +68,26 @@ private IInjectable Inject(MonoBehaviour behaviour, InjectableId injectableId, E
...
+        {
+            var behaviourType = behaviour.GetType();
+            EnsureLoaded(behaviourType);
+            foreach (var fieldsToComponents in fieldInfoCache[behaviourType])
+            {
+                var fields = fieldsToComponents.Value;
+                foreach (var field in fields)
+                {
+                    var oldField = field.GetValue(behaviour);
+                    if (!(oldField is RequirableBase))
```
jared-improbable: Case it once and save the result so you don't have to cast again below.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/RequirableBase.cs
```diff
@@ -0,0 +1,37 @@
...
+        public RequirableBase(ILogDispatcher logDispatcher)
+        {
+            this.logDispatcher = logDispatcher;
+        }
+
+        protected bool LogErrorIfDisposed()
+        {
+            if (isDisposed)
+            {
+                logDispatcher.HandleLog(LogType.Error, new LogEvent(AccessedDisposedRequirable)
```
jared-improbable: I can imagine this becoming very spammy if this code path fires. Any ideas for how to mitigate this?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
JonasImprobable: Fixed. :)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
JonasImprobable: Thanks for catching! Fixed.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
JonasImprobable: Fixed.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/RequirableBase.cs
JonasImprobable: Yeah, I can see how this could get spammy. We could save for which [Require] field or MonoBehaviour type we've already logged an error and only log an error once per field/Behaviour but that would mean that we potentially accumulate a bit cache over time. Do you have a suggestion for how we could handle this more elegantly?

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/EntityGameObjectLinkerTests.cs
```diff
@@ -23,6 +24,7 @@ public void Setup()
         {
             world = new World("TestWorld");
             entityManager = world.GetOrCreateManager<EntityManager>();
+            world.CreateManager<WorkerSystem>(null, null, "TestWorker", Vector3.zero);
```
jessicafalk: do we need that line?

JonasImprobable: Unfortunately yes. Upon creating the `EntityGameObjectLinker`, that object will getorcreate a `GameObjectDispatcherSystem` which will attempt to call `World.GetExistingManager<WorkerSystem>()`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
```diff
@@ -43,6 +45,30 @@ public EntityGameObjectLinker(World world, ILogDispatcher logDispatcher)
...
             spatialOSComponent.World = world;
             spatialOSComponent.LogDispatcher = logDispatcher;
+
+            var gameObjectReference = new GameObjectReference { GameObject = gameObject };
+            viewCommandBuffer.AddComponent(entity, gameObjectReference);
+
+            var gameObjectReferenceHandleComponent = new GameObjectReferenceHandle();
+            entityCommandBuffer.AddComponent(entity, gameObjectReferenceHandleComponent);
+
+            gameObjectDispatcherSystem.CreateActivationManagerAndReaderWriterStore(entity, gameObject);
```
jessicafalk: I don't find it very intuitive that we call this from a system to call into a different system.  Systems should be modular by itself and there should be no need for a system to reference adifferent one (even indirect).
It would be better to factor this out into two different systems. It looks like we try to much in one system

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
```diff
@@ -11,15 +11,17 @@ public class EntityGameObjectLinker
...
         public EntityGameObjectLinker(World world, ILogDispatcher logDispatcher)
         {
             this.world = world;
             this.logDispatcher = logDispatcher;
+            gameObjectDispatcherSystem = world.GetOrCreateManager<GameObjectDispatcherSystem>();
         }
 
         public void LinkGameObjectToEntity(GameObject gameObject, Entity entity, EntityId spatialEntityId,
-            ViewCommandBuffer viewCommandBuffer)
+            EntityCommandBuffer entityCommandBuffer, ViewCommandBuffer viewCommandBuffer)
```
jessicafalk: why do we need to add two buffers?

JonasImprobable: One is for adding Componentdatas and one is for adding non-blittable components.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -27,6 +27,8 @@ internal class RequiredFieldInjector
             = "[Require] attribute found on member that is not Injectable. This member will be ignored.";
         private const string MalformedInjectable
             = "Injectable found without required attributes, this is invalid.";
+        private const string RequirableFieldDoesNotInheritRequirableBase
```
firtina-improbable: Wondering if this can be tested in an editor test

firtina-improbable: @jessicafalk can it be done similarly to how you tested the systems annotations?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/ReaderWriterBase.cs
```diff
@@ -9,16 +9,15 @@
 namespace Improbable.Gdk.Core.GameObjectRepresentation
 {
     internal abstract class ReaderWriterBase<TSpatialComponentData, TComponentUpdate>
-        : IWriter<TSpatialComponentData, TComponentUpdate>
+        : RequirableBase
```
firtina-improbable: Won't removing the `IWriter` inheritance cause problems?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
jessicafalk: that should be doable with a bit of reflection magic

workers/unity/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingularReaderWriter.cs
```diff
@@ -64,120 +64,360 @@ public ReaderWriterImpl(Entity entity, EntityManager entityManager, ILogDispatch
 
                 public event Action<BlittableBool> Field1Updated
                 {
-                    add => field1Delegates.Add(value);
-                    remove => field1Delegates.Remove(value);
+                    add
+                    {
+                        if (LogErrorIfDisposed())
```
samiwh: In future it would be really useful to order commits for this kind of thing. If there is a huge amount of stuff that is all the same then it's much easier to read if it can be isolated to a single commit, and the rest of the code in another (or several if there are multiple logical blocks)

workers/unity/Assets/Generated/Source/improbable/gdk/tests/nonblittabletypes/NonBlittableComponentMonoBehaviourCommandHandlers.cs
```diff
@@ -18,7 +18,8 @@ public partial class NonBlittableComponent
     {
         public partial class FirstCommand
         {
-            public struct RequestResponder {
+            public struct RequestResponder
```
samiwh: Not really part of this PR but why is this not in the `Requirable` class?

workers/unity/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingularReaderWriter.cs
JonasImprobable: Yeah sorry about that. This was one of the situations where changes that others made in the meantime had to be pulled in and modified over and over.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/ReaderWriterBase.cs
JonasImprobable: Stuff seems to be still running fine because the API code doesn't depend on `ReaderWriterBase` inheriting IWriter. The final generated Impl class inherits from everything since it inherits the codegened Reader/Writer interfaces which themselves ihnehrit from the general IReader/IWriter interface. I suppose this comes down to a matter of style. I could let ReaderWriterBase inherit from IReader and IWriter if we want but it's not necessary.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/RequirableBase.cs
samiwh: You can reduce duplicated error spam in the logger itself. Ultimately I think we need to revisit what is an isn't an error though.

workers/unity/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingularReaderWriter.cs
samiwh: If the changes are simple then rebasing is probably fairly straightforward.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
samiwh: I don't agree that a system should never need to reference another one, especially as there isn't a concept of system components at the moment. However I do agree that it seems like this can be factored out. It seems like that function should be owned by what ever does the linking.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -175,18 +165,30 @@ public ActivationManagerNotFoundException(string message) : base(message)
...
+
+        internal void RemoveActivationManagerAndReaderWriterStore(Entity entity)
+        {
+            if (!entityToActivationManager.TryGetValue(entity, out var activationManager))
+            {
+                throw new ActivationManagerNotFoundException($"MonoBehaviourActivationManager not found for entity {entity.Index}.");
+            }
+
+            // Disable enabled SpatialOSBehaviours and dispose leftover Requirables.
+            activationManager.Dispose();
```
samiwh: Dispose this after removing the references to it.

samiwh: Also if we have disposable objects we need to have a OnDeleteManager function.
This system probably needs one anyway.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/RequirableBase.cs
JonasImprobable: For the sake of getting this in quickly, are you guys ok with revisiting this later?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
JonasImprobable: We currently need to store a mapping of entity to ActivationManager somewhere. Currently it's stored in the `GameObjectDispatcherSystem`. At the same time, no matter, where the mapping is stored, the `GameObjectDispatcherSystem` needs to access it when ticked. We could start storing the mapping in the `EntityGameObjectLinker` which is currently owned by the `EntityInitializationSystem` that that would mean that the `GameObjectDispatcherSystem` needs to reference the `EntityInitializationSystem` to get the linker. Either way, as far as I can tell, all data is eventually owned by either a system or by an entity so you'll need to reference other systems somehow at the end of the day.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
firtina-improbable: We can do this in a polish phase, I'll make a ticket

JonasImprobable:
@firtina-improbable Yeah agreed. That's probably not how systemstate components should be used. Happy to make this nicer as in a follow-up refactoring step.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/RequirableBase.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        private readonly ILogDispatcher logDispatcher;
+        private bool isDisposed;
+
+        public RequirableBase(ILogDispatcher logDispatcher)
+        {
+            this.logDispatcher = logDispatcher;
+        }
+
+        protected bool LogErrorIfDisposed()
```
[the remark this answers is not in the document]
JonasImprobable: Good idea. Renaming.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
MPeti01: This looks very confusing for an API users should be able to access and use too. I don't have much context on these buffers - could we have the Linker own them maybe and have a flush method?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
firtina-improbable: UTY-1144

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -27,6 +27,8 @@ internal class RequiredFieldInjector
...
+            = "[Require] field element does not inherit RequirableBase. This is most likely a bug in the Unity GDK.";
```
MPeti01: Couldn't we just replace the Injectable interface with the RequirableBase class? Sounds like that would make more sense. Not sure if we should do it now or just put it in a log of "cleanup refactors to do" though because it's literally everywhere though

MPeti01: Either way I don't like the error message. Firstly we are "SpatialOS GDK for Unity" now, but also maybe merge with MalformedInjectable? They are both for messed up classes that shouldn't happen at all

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/ReaderWriterBase.cs
MPeti01: No reason why it couldn't implement the IWriter interface while also having a parent class though surely? Which would make more sense if this is where the actual method is being implemented

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
firtina-improbable: 1st comment can be part of UTY-1144 - please add to the acceptance criteria

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/ReaderWriterBase.cs
JonasImprobable: Let me just let `ReaderWriterBase` from IReader and IWriter then. Why the hell not? ;)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
JonasImprobable: I don't disagree with the idea of getting rid of the `IInjectable` interface but let's do this in another PR. This PR is kinda bloated already. :)

JonasImprobable:
@MPeti01 Re: using the `entityCommandBuffer` argument in `EntityGameObjectLinker.LinkGameObjectToEntity`: @jessicafalk 's PR should make that argument obsolete.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
```diff
@@ -43,6 +45,30 @@ public EntityGameObjectLinker(World world, ILogDispatcher logDispatcher)
             spatialOSComponent.SpatialEntityId = spatialEntityId;
             spatialOSComponent.World = world;
             spatialOSComponent.LogDispatcher = logDispatcher;
+
+            var gameObjectReference = new GameObjectReference { GameObject = gameObject };
+            viewCommandBuffer.AddComponent(entity, gameObjectReference);
+
+            var gameObjectReferenceHandleComponent = new GameObjectReferenceHandle();
+            entityCommandBuffer.AddComponent(entity, gameObjectReferenceHandleComponent);
```
jamiebrynes7: QQ: Why do we have `GameObjectReference` and `GameObjectReferenceHandle`? I missed the context on this

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -104,26 +101,49 @@ public void EnableSpatialOSBehaviours()
...
+            DisableAllSpatialOSBehavioursOfASet(behavioursToDisable);
+            behavioursToDisable.Clear();
+        }
+
+        private void DisableAllEnabledSpatialOSBehaviours()
+        {
+            DisableAllSpatialOSBehavioursOfASet(enabledBehaviours);
+        }
+
+        private void DisableAllSpatialOSBehavioursOfASet(HashSet<MonoBehaviour> behaviours)
```
jamiebrynes7: Nit: Can probably just rename to `DisableSpatialOSBehaviours`. The argument provides the context of the set

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
JonasImprobable: At this point, there is no need for the `GameObjectReference` component anymore. We just haven't managed to remove it yet but we should eventually. :)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -68,6 +70,28 @@ private IInjectable Inject(MonoBehaviour behaviour, InjectableId injectableId, E
...
+            {
+                var fields = fieldsToComponents.Value;
+                foreach (var field in fields)
+                {
+                    var requirableToBeDisposed = field.GetValue(behaviour) as RequirableBase;
+                    if (requirableToBeDisposed == null)
+                    {
+                        logger.HandleLog(LogType.Error, new LogEvent(RequirableFieldDoesNotInheritRequirableBase)
+                            .WithField("Behaviour", behaviour)
+                            .WithField("Field", field.GetValue(behaviour)));
```
jamiebrynes7: What does `field.GetValue(...)` print for the user?

JonasImprobable: This should (hopefully) call the default ToString() method of an object that was injected through the [Require] attribute.

jamiebrynes7: Wouldn't it be more useful to print the name of the field declared in the source if possible? So a user would get:

```
Field marked as required... 
Behaviour: MySpatialOSBehaviour
Field: MyIllegalRequireable
```

JonasImprobable: You are right, that's much better. Fixed. :)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
MPeti01: GameObjectReferenceHandle is removed in Jess's PR

workers/unity/Assets/Playground/Scenes/ClientScene.unity
```diff
@@ -257,7 +257,7 @@ Prefab:
     m_Modifications:
     - target: {fileID: 4555427980518638, guid: 5857884d058555d4282a1beb501099b9, type: 2}
       propertyPath: m_LocalPosition.x
-      value: 50
```
MPeti01: What is this change?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
JonasImprobable: I renamed the method to `DisableAllSpatialOSBehavioursInternal` so people don't confuse it over the other public `DisableSpatialOSBehaviours` method, intended to be called from outside.

workers/unity/Packages/com.improbable.gdk.tools/.DownloadCoreSdk/Program.cs
```diff
@@ -11,26 +11,26 @@ internal class Program
     {
         private static void Main(string[] args)
```
MPeti01: What's up with all this tooling stuff showing up as changes?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
jessicafalk: it will be great :D

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
jessicafalk: This line will be removed in my PR, so it's fine to merge it in for now

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -27,6 +27,8 @@ public class RequiredFieldInjector
             = "[Require] attribute found on member that is not Injectable. This member will be ignored.";
         private const string MalformedInjectable
             = "Injectable found without required attributes, this is invalid.";
+        private const string RequirableFieldDoesNotInheritRequirableBase
+            = "[Require] field element does not inherit RequirableBase. This is most likely a bug in the SpatialOS GDK.";
```
samiwh: I don't think user should be exposed to RequirableBase. It doesn't really mean anything to them.

PR #290, opened 2018-09-02
Make Authority<T> less generic to make its usage less error prone.

martinzlocha:

-------

#### Description
The PlayerLifecycle feature module wasn't working because it was using `Authoritative<PlayerHeartbeatServer>` instead of `Authoritative<PlayerHeartbeatServer.Component>`. This is a very easy mistake to make and takes a while to debug. We should restrict the generics which can be used with `Authoritative<T>`.

#### Tests
Ran locally, player got deleted.

#### Documentation
N/A

#### Primary reviewers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Authority/AuthorityChanges.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Authority/AuthorityChanges.cs
@@ -1,10 +1,10 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using Improbable.Worker.Core;
 using Unity.Entities;
 
 namespace Improbable.Gdk.Core
 {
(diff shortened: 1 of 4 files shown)


workers/unity/Packages/com.improbable.gdk.core/Authority/AuthorityChanges.cs
```diff
@@ -1,10 +1,10 @@
-﻿using System.Collections.Generic;
```
jessicafalk: what happened here?

martinzlocha: 🤷‍♂️  Removed that change.

martinzlocha:
Realized that I put the wrong name of the branch. Should be `bugfix/` instead of `fix/`.

PR #294, opened 2018-09-03
Move initialization into Core and adding DisableAutoCreation tag

jessicafalk:

-------

#### Description
in commit order:
1. moved injection hook stuff into core. Now you simply have to use the helper methods in playground to set everything up.
2. added `DisableAutoCreation` attributes to all core and feature module systems to allow users to also enable the default world instead of using injection hooks, if they want to

#### Tests
tested it locally.

#### Documentation
none so far

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -2,13 +2,13 @@
...
 using System.Linq;
 using Unity.Entities;
-using UnityEngine;
 
 namespace Improbable.Gdk.Core.GameObjectRepresentation
 {
     /// <summary>
     ///     Gathers incoming dispatcher ops and invokes callbacks on relevant GameObjects.
     /// </summary>
+    [DisableAutoCreation]
```
jared-improbable: Can do we a test that iterates over all Assembly types and ensures that they have the `DisableAutoCreation` set? Perhaps with a whitelist of systems that don't need it (if they exist)?

workers/unity/Packages/com.improbable.gdk.core/Systems/CommandRequestTrackerSystem.cs
```diff
@@ -1,11 +1,12 @@
-﻿using System;
+using System;
```
jared-improbable: ??

workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerDisconnectCallbackSystem.cs
```diff
@@ -1,10 +1,11 @@
-﻿using System;
+using System;
```
jared-improbable: ??

workers/unity/Packages/com.improbable.gdk.core/CoreInitialization.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using Unity.Entities;
+
+namespace Improbable.Gdk.Core
+{
+    public static class CoreInitialization
```
samiwh: I'd call this something else. Maybe `WorldsInitializationHelper` or something as it doesn't initialise the core.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
jessicafalk: I guess we could have a test that iterates only over core & feature modules and checks all the systems in there. right now we don't have any systems in there that don't need it

workers/unity/Packages/com.improbable.gdk.core/Systems/CommandRequestTrackerSystem.cs
jessicafalk: no idea what happened there. will revert it

workers/unity/Assets/Playground/Scripts/OneTimeInitialisation.cs
```diff
@@ -18,38 +18,11 @@ private static void Init()
             }
 
             initialized = true;
-            SetupInjectionHooks();
-            PlayerLoopManager.RegisterDomainUnload(DomainUnloadShutdown, 1000);
+            CoreInitialization.SetupInjectionHooks();
+            PlayerLoopManager.RegisterDomainUnload(CoreInitialization.DomainUnloadShutdown, 1000);
```
firtina-improbable: I think this should be part of the new class as well, so the users would call only one method

firtina-improbable: "this" being `RegisterDomainUnload`

jessicafalk: I kept it here intentionally so that users can define their own `RegisterDomainUnload` if necessary. They might want to do more clean up than we provide

firtina-improbable: Agreed, sounds sensible

workers/unity/Packages/com.improbable.gdk.testutils/DisableAutoCreationSystemsTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            "Improbable.Gdk.Core",
+            "Improbable.Gdk.TransformSynchronization",
+            "Improbable.Gdk.PlayerLifecycle",
+        };
+
+        private static readonly string[] WhiteListedSystems = { };
+
+        [Test]
+        public void systems_should_have_DisableAutoCreation_tag()
```
jared-improbable: Nice!

PR #298, opened 2018-09-03
Disable QOS timeout

jessicafalk:

-------

#### Description
Disabling qos timeout for the default_launch config that should be used for `spatial local launch`.
That gets rid of the case where a worker disconnects when pausing the game and therefore causing a couple of issues.

#### Tests
tested it locally. Workers never disconnected and pausing/unpausing worked fine.
This should not be used for the cloud workflow

#### Documentation
no

#### Primary reviewers

### DIFF
--- a/default_launch.json
+++ b/default_launch.json
@@ -8,7 +8,13 @@
     "dimensions": {
       "xMeters": 5000,
       "zMeters": 5000
-    }
+    },
+    "legacy_flags": [
+      {
+        "name": "bridge_qos_max_timeout",
+        "value": "0"
+      }
+    ]
   },
   "workers": [
     {


default_launch.json
```diff
@@ -8,7 +8,13 @@
...
       "zMeters": 5000
-    }
+    },
+    "legacy_flags": [
+      {
+        "name": "bridge_qos_max_timeout",
+        "value": "0"
+      }
+    ]
   },
```
firtina-improbable: I see Unreal GDK also has max unacked pings rate set, do we know if we may need that too?

https://github.com/spatialos/UnrealGDKStarterProject/blob/master/spatial/default_launch.json#L19-L22

default_launch.json
```diff
@@ -8,7 +8,13 @@
     "dimensions": {
       "xMeters": 5000,
       "zMeters": 5000
-    }
+    },
+    "legacy_flags": [
+      {
+        "name": "bridge_qos_max_timeout",
+        "value": "0"
```
samiwh: I'm not sure about having this committed. It's more of a local debugging tool. If we need to add it in I'd suggest a different deployment config for it.

samiwh: Discussed offline. Putting this in the local launch config but not the cloud one is fine.

PR #178, opened 2018-08-03
Please take my contribution? It'll be worth it.

petemounce:
**Contributions**: We are not currently taking public contributions - see our [contributions](CONTRIBUTING.md) policy. However, we are accepting issues and we do want your [feedback](../README.md#give-us-feedback).

-------

Spike in BuildKite CI

I don't want to share the URL, but see BK.

BK user-docs.

If your change will take a long time to review, you can name at most two primary reviewers who are ultimately responsible for reviewing this request. @ mention them.

PR #299, opened 2018-09-03
Wire to BuildKite, now with branch prefix.

petemounce:
This is #178 but with a branch prefix so you can see it in other CI.


ci/build-test.cmd
```diff
@@ -0,0 +1,3 @@
+@echo off
+
+"%ProgramFiles%\git\bin\bash.exe" -c "./ci/build-test.sh"
```
gkassabli: Might be a n00b question, but is this supposed to work on non-Windows laptops?

jessicafalk: nope, buildkite currently only has windows agents

gkassabli: Got it. Do we intend to use buildkite for iOS/any other non-Windows builds at some point then?

jamiebrynes7: Yes - everything will be moved to Buildkite eventually

jessicafalk: yes, but it's not that easy, so focusing on windows first. but the plan is to completely move to buildkite at some point

gkassabli: Makes sense

jamiebrynes7:
Superseded by feature/bk-ci

PR #312, opened 2018-09-05
Cleanup Legacy Package

jamiebrynes7:
#### Description
We had a legacy package which only had the build system in it. We have recently chanaged the build system enough that it has diverged from the actual legacy code.

Moved it into `Improbable.Gdk.Core.BuildSystem` inside the Core package. Renamed all namespaces. Updated build scripts to invoke correct function.
#### Tests
Ran `ci/build-client.sh` and `ci/build-gamelogic.sh`.
#### Documentation
N/A - docs on build system are WIP
#### Primary reviewers
@jessicafalk @jared-improbable


zeroZshadow:
Please move the build system into its own package so it isn't manditory

PR #313, opened 2018-09-05
moving prefab preprocessor from playground to GameObjectRepresentation

jessicafalk:

-------

#### Description
Because if you create your new project and you don't have it, it will completely break.

#### Tests
tested in my own small test project

#### Documentation
none

#### Primary reviewers
@zeroZshadow


jessicafalk:
sorry about that, should be fixed now

zeroZshadow:
The namespace isn't fixed now

firtina-improbable:
LGTM for MVP, ideal solution is moving it into a module that adds on top of monobehaviour support.

jessicafalk:
@firtina-improbable  agree with that. definitely needs to be moved with the rest of the monobehaviour stuff!

I think I fixed everything now, please take another look @zeroZshadow

PR #314, opened 2018-09-05
Implement world command sending

firtina-improbable:

-------

#### Description
Example usage (to be fleshed out more in next PR):

```cs
    public class Terminator : MonoBehaviour
    {
        [Require] private WorldCommandsRequirables.WorldCommandRequestSender worldCommandRequestSender;
        [Require] private WorldCommandsRequirables.WorldCommandResponseHandler worldCommandResponseHandler;

        public void OnEnable()
        {
            worldCommandResponseHandler.OnDeleteEntityResponse += OnDeleteEntityResponse;
        }

        private void OnDeleteEntityResponse(WorldCommands.DeleteEntity.ReceivedResponse response)
        {
            if (response.Op.StatusCode != StatusCode.Success)
            {
                Debug.LogError("Could not terminate target. Possible reason: "+ response.Op.Message);
            }
        }

        public void Update()
        {
            if (Input.GetKeyDown(KeyCode.Backspace))
            {
                EntityId someEntityId = FindTargetToEliminate();
                worldCommandRequestSender.DeleteEntity(someEntityId);
            }
        }

        private EntityId FindTargetToEliminate()
        {
            // does not concern this PR
            throw new System.NotImplementedException();
        }
    }
```

#### Tests
To come with another PR.

#### Documentation
To be done in later phases of project

#### Primary reviewers
@samiwh @JonasImprobable

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
@@ -15,7 +15,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
     {
         private readonly Dictionary<Entity, MonoBehaviourActivationManager> entityToActivationManager =
             new Dictionary<Entity, MonoBehaviourActivationManager>();
-        private readonly Dictionary<Entity, InjectableStore> entityToReaderWriterStore =
+        internal readonly Dictionary<Entity, InjectableStore> entityToReaderWriterStore =
             new Dictionary<Entity, InjectableStore>();
 
         public readonly List<GameObjectComponentDispatcherBase> GameObjectComponentDispatchers =
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
@@ -1,0 +1,155 @@
+using System.Linq;
+using Unity.Collections;
+using Unity.Entities;
+
+#region Diagnostic control
+
+// Disable the "variable is never assigned" for injected fields.
+#pragma warning disable 649
+
+// ReSharper disable ClassNeverInstantiated.Global
+
+#endregion
+
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    public class GameObjectWorldCommandSystem : ComponentSystem
+    {
+        private struct ReserveEntityResponseData
+        {
+            [ReadOnly] public readonly int Length;
+            [ReadOnly] public EntityArray Entities;
+            [ReadOnly] public ComponentDataArray<Commands.WorldCommands.ReserveEntityIds.CommandResponses> Responses;
+        }
+
+        [Inject] private ReserveEntityResponseData reserveEntityResponseData;
+
+        private struct CreateEntityResponseData
+        {
+            [ReadOnly] public readonly int Length;
+            [ReadOnly] public EntityArray Entities;
+            [ReadOnly] public ComponentDataArray<Commands.WorldCommands.CreateEntity.CommandResponses> Responses;
+        }
+
+        [Inject] private CreateEntityResponseData createEntityResponseData;
+
+        private struct DeleteEntityResponseData
+        {
+            [ReadOnly] public readonly int Length;
+            [ReadOnly] public EntityArray Entities;
+            [ReadOnly] public ComponentDataArray<Commands.WorldCommands.DeleteEntity.CommandResponses> Responses;
+        }
+
+        [Inject] private DeleteEntityResponseData deleteEntityResponseData;
+
+        [Inject] private GameObjectDispatcherSystem GameObjectDispatcherSystem;
+
+        private static readonly InjectableId WorldCommandResponseHandlerInjectableId =
+            new InjectableId(InjectableType.WorldCommandResponseHandler, InjectableId.NullComponentId);
+
+        protected override void OnUpdate()
+        {
+            for (var i = 0; i < reserveEntityResponseData.Length; ++i)
+            {
+                if (reserveEntityResponseData.Responses[i].Responses.Count == 0)
+                {
+                    continue;
+                }
+
+                var entity = reserveEntityResponseData.Entities[i];
+
+                var worldCommandResponseHandlers = GetWorldCommandResponseHandlersForEntity(entity);
+
+                if (worldCommandResponseHandlers == null)
+                {
+                    continue;
+                }
+
+                foreach (var receivedResponse in reserveEntityResponseData.Responses[i].Responses)
+                {
+                    foreach (var worldCommandHandler in worldCommandResponseHandlers)
+                    {
+                        worldCommandHandler.OnReserveEntityIdsResponseInternal(receivedResponse);
+                    }
+                }
+            }
+
+            for (var i = 0; i < createEntityResponseData.Length; ++i)
+            {
+                if (createEntityResponseData.Responses[i].Responses.Count == 0)
+                {
+                    continue;
+                }
+
+                var entity = createEntityResponseData.Entities[i];
+
+                var worldCommandResponseHandlers =
+                    GetWorldCommandResponseHandlersForEntity(entity);
+
+                if (worldCommandResponseHandlers == null)
+                {
+                    continue;
+                }
+
+                foreach (var receivedResponse in createEntityResponseData.Responses[i].Responses)
+                {
+                    foreach (var worldCommandHandler in worldCommandResponseHandlers)
+                    {
+                        worldCommandHandler.OnCreateEntityResponseInternal(receivedResponse);
+                    }
+                }
+            }
+
+            for (var i = 0; i < deleteEntityResponseData.Length; ++i)
+            {
+                if (deleteEntityResponseData.Responses[i].Responses.Count == 0)
+                {
+                    continue;
+                }
+
+                var entity = deleteEntityResponseData.Entities[i];
+
+                var worldCommandResponseHandlers =
+                    GetWorldCommandResponseHandlersForEntity(entity);
+
+                if (worldCommandResponseHandlers == null)
+                {
+                    continue;
+                }
+
+                foreach (var receivedResponse in deleteEntityResponseData.Responses[i].Responses)
+                {
+                    foreach (var worldCommandHandler in worldCommandResponseHandlers)
+                    {
+                        worldCommandHandler.OnDeleteEntityResponseInternal(receivedResponse);
+                    }
+                }
+            }
+        }
+
+        private WorldCommandsRequirables.WorldCommandResponseHandler[] GetWorldCommandResponseHandlersForEntity(
+            Entity entity)
+        {
+            var entityToReaderWriterStore = GameObjectDispatcherSystem.entityToReaderWriterStore;
+
+            if (!entityToReaderWriterStore.TryGetValue(entity, out var injectableStore))
+            {
+                return null;
+            }
+
+            if (!injectableStore.TryGetInjectablesForComponent(WorldCommandResponseHandlerInjectableId,
+                out var injectables))
+            {
+                return null;
+            }
+
+            if (injectables.Count == 0)
+            {
+                return null;
+            }
+
+            return injectables.Cast<WorldCommandsRequirables.WorldCommandResponseHandler>().ToArray();
+        }
+    }
+}
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
@@ -21,7 +21,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
     {
         private struct ReserveEntityIdsResponseData
         {
-            [ReadOnly] public readonly int Length;
+            public readonly int Length;
             [ReadOnly] public EntityArray Entities;
             [ReadOnly] public ComponentDataArray<WorldCommands.ReserveEntityIds.CommandResponses> CommandResponses;
             [ReadOnly] public ComponentArray<GameObjectReference> HasGameObjectReference;
@@ -29,7 +29,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
 
         private struct CreateEntityResponseData
         {
-            [ReadOnly] public readonly int Length;
+            public readonly int Length;
             [ReadOnly] public EntityArray Entities;
             [ReadOnly] public ComponentDataArray<WorldCommands.CreateEntity.CommandResponses> CommandResponses;
             [ReadOnly] public ComponentArray<GameObjectReference> HasGameObjectReference;
@@ -37,7 +37,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
 
         private struct DeleteEntityResponseData
         {
-            [ReadOnly] public readonly int Length;
+            public readonly int Length;
             [ReadOnly] public EntityArray Entities;
             [ReadOnly] public ComponentDataArray<WorldCommands.DeleteEntity.CommandResponses> CommandResponses;
             [ReadOnly] public ComponentArray<GameObjectReference> HasGameObjectReference;
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Linq;
+using Improbable.Gdk.Core.Commands;
 using Unity.Collections;
 using Unity.Entities;
 
@@ -16,48 +18,43 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
 {
     public class GameObjectWorldCommandSystem : ComponentSystem
     {
-        private struct ReserveEntityResponseData
+        private struct ReserveEntityIdsResponseData
         {
             [ReadOnly] public readonly int Length;
             [ReadOnly] public EntityArray Entities;
-            [ReadOnly] public ComponentDataArray<Commands.WorldCommands.ReserveEntityIds.CommandResponses> Responses;
+            [ReadOnly] public ComponentDataArray<WorldCommands.ReserveEntityIds.CommandResponses> CommandResponses;
+            [ReadOnly] public ComponentArray<GameObjectReference> HasGameObjectReference;
         }
-
-        [Inject] private ReserveEntityResponseData reserveEntityResponseData;
 
         private struct CreateEntityResponseData
         {
             [ReadOnly] public readonly int Length;
             [ReadOnly] public EntityArray Entities;
-            [ReadOnly] public ComponentDataArray<Commands.WorldCommands.CreateEntity.CommandResponses> Responses;
+            [ReadOnly] public ComponentDataArray<WorldCommands.CreateEntity.CommandResponses> CommandResponses;
+            [ReadOnly] public ComponentArray<GameObjectReference> HasGameObjectReference;
         }
-
-        [Inject] private CreateEntityResponseData createEntityResponseData;
 
         private struct DeleteEntityResponseData
         {
             [ReadOnly] public readonly int Length;
             [ReadOnly] public EntityArray Entities;
-            [ReadOnly] public ComponentDataArray<Commands.WorldCommands.DeleteEntity.CommandResponses> Responses;
+            [ReadOnly] public ComponentDataArray<WorldCommands.DeleteEntity.CommandResponses> CommandResponses;
+            [ReadOnly] public ComponentArray<GameObjectReference> HasGameObjectReference;
         }
 
+        [Inject] private GameObjectDispatcherSystem gameObjectDispatcherSystem;
+        [Inject] private ReserveEntityIdsResponseData reserveEntityIdsResponseData;
+        [Inject] private CreateEntityResponseData createEntityResponseData;
         [Inject] private DeleteEntityResponseData deleteEntityResponseData;
-
-        [Inject] private GameObjectDispatcherSystem GameObjectDispatcherSystem;
 
         private static readonly InjectableId WorldCommandResponseHandlerInjectableId =
             new InjectableId(InjectableType.WorldCommandResponseHandler, InjectableId.NullComponentId);
 
         protected override void OnUpdate()
         {
-            for (var i = 0; i < reserveEntityResponseData.Length; ++i)
+            for (var i = 0; i < reserveEntityIdsResponseData.Length; ++i)
             {
-                if (reserveEntityResponseData.Responses[i].Responses.Count == 0)
-                {
-                    continue;
-                }
-
-                var entity = reserveEntityResponseData.Entities[i];
+                var entity = reserveEntityIdsResponseData.Entities[i];
                 var worldCommandResponseHandlers = GetWorldCommandResponseHandlersForEntity(entity);
 
                 if (worldCommandResponseHandlers == null)
@@ -65,7 +62,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
                     continue;
                 }
 
-                foreach (var receivedResponse in reserveEntityResponseData.Responses[i].Responses)
+                foreach (var receivedResponse in reserveEntityIdsResponseData.CommandResponses[i].Responses)
                 {
                     foreach (var worldCommandHandler in worldCommandResponseHandlers)
                     {
@@ -76,11 +73,6 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
 
             for (var i = 0; i < createEntityResponseData.Length; ++i)
             {
-                if (createEntityResponseData.Responses[i].Responses.Count == 0)
-                {
-                    continue;
-                }
-
                 var entity = createEntityResponseData.Entities[i];
                 var worldCommandResponseHandlers = GetWorldCommandResponseHandlersForEntity(entity);
 
@@ -89,7 +81,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
                     continue;
                 }
 
-                foreach (var receivedResponse in createEntityResponseData.Responses[i].Responses)
+                foreach (var receivedResponse in createEntityResponseData.CommandResponses[i].Responses)
                 {
                     foreach (var worldCommandHandler in worldCommandResponseHandlers)
                     {
@@ -100,11 +92,6 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
 
             for (var i = 0; i < deleteEntityResponseData.Length; ++i)
             {
-                if (deleteEntityResponseData.Responses[i].Responses.Count == 0)
-                {
-                    continue;
-                }
-
                 var entity = deleteEntityResponseData.Entities[i];
                 var worldCommandResponseHandlers = GetWorldCommandResponseHandlersForEntity(entity);
 
@@ -113,7 +100,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
                     continue;
                 }
 
-                foreach (var receivedResponse in deleteEntityResponseData.Responses[i].Responses)
+                foreach (var receivedResponse in deleteEntityResponseData.CommandResponses[i].Responses)
                 {
                     foreach (var worldCommandHandler in worldCommandResponseHandlers)
                     {
@@ -123,10 +110,10 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
             }
         }
 
-        private WorldCommandsRequirables.WorldCommandResponseHandler[] GetWorldCommandResponseHandlersForEntity(
+        private WorldCommands.Requirables.WorldCommandResponseHandler[] GetWorldCommandResponseHandlersForEntity(
             Entity entity)
         {
-            var entityToReaderWriterStore = GameObjectDispatcherSystem.entityToReaderWriterStore;
+            var entityToReaderWriterStore = gameObjectDispatcherSystem.entityToReaderWriterStore;
 
             if (!entityToReaderWriterStore.TryGetValue(entity, out var injectableStore))
             {
@@ -144,7 +131,8 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
                 return null;
             }
 
-            return injectables.Cast<WorldCommandsRequirables.WorldCommandResponseHandler>().ToArray();
+            return Array.ConvertAll(injectables.ToArray(),
+                injectable => (WorldCommands.Requirables.WorldCommandResponseHandler) injectable);
         }
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
@@ -58,7 +58,6 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
                 }
 
                 var entity = reserveEntityResponseData.Entities[i];
-
                 var worldCommandResponseHandlers = GetWorldCommandResponseHandlersForEntity(entity);
 
                 if (worldCommandResponseHandlers == null)
@@ -83,9 +82,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
                 }
 
                 var entity = createEntityResponseData.Entities[i];
-
-                var worldCommandResponseHandlers =
-                    GetWorldCommandResponseHandlersForEntity(entity);
+                var worldCommandResponseHandlers = GetWorldCommandResponseHandlersForEntity(entity);
 
                 if (worldCommandResponseHandlers == null)
                 {
@@ -109,9 +106,7 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
                 }
 
                 var entity = deleteEntityResponseData.Entities[i];
-
-                var worldCommandResponseHandlers =
-                    GetWorldCommandResponseHandlersForEntity(entity);
+                var worldCommandResponseHandlers = GetWorldCommandResponseHandlersForEntity(entity);
 
                 if (worldCommandResponseHandlers == null)
                 {
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
@@ -16,6 +16,7 @@ using Unity.Entities;
 
 namespace Improbable.Gdk.Core.GameObjectRepresentation
 {
+    [DisableAutoCreation]
     public class GameObjectWorldCommandSystem : ComponentSystem
     {
         private struct ReserveEntityIdsResponseData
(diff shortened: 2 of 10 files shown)


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                return null;
+            }
+
+            if (injectables.Count == 0)
+            {
+                return null;
+            }
+
+            return injectables.Cast<WorldCommandsRequirables.WorldCommandResponseHandler>().ToArray();
```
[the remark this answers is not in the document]
firtina-improbable: ??

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,155 @@
...
+            for (var i = 0; i < reserveEntityResponseData.Length; ++i)
+            {
+                if (reserveEntityResponseData.Responses[i].Responses.Count == 0)
+                {
+                    continue;
+                }
+
+                var entity = reserveEntityResponseData.Entities[i];
+
+                var worldCommandResponseHandlers = GetWorldCommandResponseHandlersForEntity(entity);
```
samiwh: This made more sense as a component to add onto an entity. This is fine if it's how everything else does it, but eventually we should replace all these with something that works with an ECS.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        [InjectableId(InjectableType.WorldCommandResponseHandler, InjectableId.NullComponentId)]
+        [InjectionCondition(InjectionCondition.RequireNothing)]
+        public class WorldCommandResponseHandler : IInjectable
+        {
+            private readonly ILogDispatcher logDispatcher;
+
+            private readonly List<Action<WorldCommands.ReserveEntityIds.ReceivedResponse>> reserveEntityIdsDelegates;
+            private readonly List<Action<WorldCommands.CreateEntity.ReceivedResponse>> createEntityDelegates;
+            private readonly List<Action<WorldCommands.DeleteEntity.ReceivedResponse>> deleteEntityDelegates;
```
samiwh: Just noticed this is missing entity queries.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
firtina-improbable: I agree, unfortunately need to jump through many hoops and create storage handlers and do a triple backflip to do that with the restrictions we have at the moment

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
firtina-improbable: That's another ticket

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,155 @@
...
+        [Inject] private GameObjectDispatcherSystem GameObjectDispatcherSystem;
+
+        private static readonly InjectableId WorldCommandResponseHandlerInjectableId =
+            new InjectableId(InjectableType.WorldCommandResponseHandler, InjectableId.NullComponentId);
+
+        protected override void OnUpdate()
+        {
+            for (var i = 0; i < reserveEntityResponseData.Length; ++i)
+            {
+                if (reserveEntityResponseData.Responses[i].Responses.Count == 0)
```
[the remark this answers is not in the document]
firtina-improbable: Unsure, shall I remove these checks

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,155 @@
...
+                {
+                    continue;
+                }
+
+                var entity = reserveEntityResponseData.Entities[i];
+
```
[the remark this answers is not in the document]
firtina-improbable: I'll clean them up in next commit

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+
+#region Diagnostic control
+
+// ReSharper disable UnusedMember.Local
+
+#endregion
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    public static partial class WorldCommandsRequirables
```
samiwh: Nit - I think Command rather than Commands reads better.
Non nit - why put Requirable in the name rather than as a nested class like everything else?

Also we should really change Requirables to Requirable although that can be done later.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
samiwh: Why is it separate?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
samiwh: It's an expensive way to do that. Anything that uses IEnumerable shouldn't be used in a hot path.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
samiwh: Which restrictions are that out of interest?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
samiwh: Will check but I think we guarantee if it's there it has something in it.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
firtina-improbable: nit: agreed
nit * -1: that doesn't sound too bad, will give it a try

/s/requirables/requirable/ : unsure, it's like this everywhere else so can be reconsidered later

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
samiwh: It can strictly happen if a system has deliberately removed all the responses. However in that case nothing would break here. So this is just an efficiency call. In which case it's probably more efficient not to make the check.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
firtina-improbable: Will go for
```cs
Array.ConvertAll(injectables.ToArray(),
                injectable => (WorldCommandRequirables.WorldCommandResponseHandler) injectable)
```

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            public WorldCommands.ReserveEntityIds.Request ReserveEntityIds(
+                uint numberOfEntityIds,
+                object context = null)
+            {
+                var reserveEntityIdCommandSender =
+                    entityManager.GetComponentData<WorldCommands.ReserveEntityIds.CommandSender>(entity);
+
+                var request = new WorldCommands.ReserveEntityIds.Request
```
samiwh: Prefer to use the CreateRequest functions I think.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
```diff
@@ -0,0 +1,98 @@
...
+            private readonly Entity entity;
+            private readonly EntityManager entityManager;
+
+            private WorldCommandRequestSender(Entity entity, EntityManager entityManager)
+            {
+                this.entity = entity;
+                this.entityManager = entityManager;
+            }
+
+            public WorldCommands.ReserveEntityIds.Request ReserveEntityIds(
```
samiwh: Why return the request? Would it make more sense to return the ID?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
firtina-improbable: These are the two main ones for me

- Cannot have lists in IComponentData
- Cannot have non-pure components on entities

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
```diff
@@ -0,0 +1,98 @@
...
+                uint numberOfEntityIds,
+                object context = null)
```
samiwh: We probably do also want the timeout to be exposed. Probably with a default value.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                var reserveEntityIdCommandSender =
```
samiwh: Can we move this to where it is used

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
firtina-improbable: I don't really know. I am guessing it should be simple to add once this one is in.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
```diff
@@ -0,0 +1,98 @@
...
+
+// ReSharper disable UnusedMember.Local
+
+#endregion
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    public static partial class WorldCommandsRequirables
+    {
+        [InjectableId(InjectableType.WorldCommandRequestSender, InjectableId.NullComponentId)]
```
samiwh: This one could be tied to the world command sender component. Not really much in it though.

samiwh: Misunderstood. Ignore this.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
firtina-improbable: Yes this is a bug, fixing

firtina-improbable: Also missing timeoutmillis

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
samiwh: Fair enough. Easier to review with less code :P.
In that case we can probably just immediately approve it once it's up.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
firtina-improbable: Agreed, will return the ID

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
firtina-improbable: I'll inline it

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
firtina-improbable: yes, will be added as part of createrequest fix

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
firtina-improbable: Additionally the needed data is already present somewhere and it would be duplicated. When in a polish pass we move them as part of entities, this hack will go away too.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
firtina-improbable: Roger, removing the check

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
samiwh: We have lots of lists on IComponentData. But there is also BufferArray for that. What is a non-pure component in this context?

samiwh: It is definitely possible to implement this. We don't have to do it now. But it can definitely be done. In fact you did it for world commands the previous PR.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
MPeti01: So that updated version will turn a HashSet into an array and then map it to another? Still could be lighter by just iterating over the HashSet and casting each element I think, I don't have a strong opinion about it though

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
MPeti01: Yeah I think that duplication of functionality is the biggest reason to just go like this for now, relying on the InjectableStore. We can later see about putting those on the entities more nicely

firtina-improbable: That's partially why it was a huge PR - the handlers/responders would need to reference other systems so they would need to either add more entities or register themselves by going through the world, then the system or handlers need to manage these components' storage and lifecycle,

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
samiwh: For removing the s I meant it should be done everywhere. So agreed it shouldn't be in this PR.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,155 @@
...
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    public class GameObjectWorldCommandSystem : ComponentSystem
+    {
+        private struct ReserveEntityResponseData
+        {
+            [ReadOnly] public readonly int Length;
+            [ReadOnly] public EntityArray Entities;
+            [ReadOnly] public ComponentDataArray<Commands.WorldCommands.ReserveEntityIds.CommandResponses> Responses;
```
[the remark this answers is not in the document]
firtina-improbable: Will say `CommandResponses` to match data

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
samiwh: Really the hashset probably shouldn't be a hashset. The majority of operations on it are iterating over it.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -14,7 +14,7 @@ internal class GameObjectDispatcherSystem : ComponentSystem
     {
         private readonly Dictionary<Entity, MonoBehaviourActivationManager> entityToActivationManager =
             new Dictionary<Entity, MonoBehaviourActivationManager>();
-        private readonly Dictionary<Entity, InjectableStore> entityToReaderWriterStore =
```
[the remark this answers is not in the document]
firtina-improbable: yes, that's ideal when the time comes.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
samiwh: The other option (which I think I prefer longer term) is to do this without the type erasure. However that is probably too large a change for now.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
samiwh: I think duplication is a lesser sin than making it harder to read and slower. But for now it doesn't matter too much.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
firtina-improbable: Creating a ticket for this, will link

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
firtina-improbable: UTY-1146

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
firtina-improbable: UTY-1146

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,155 @@
...
+// ReSharper disable ClassNeverInstantiated.Global
+
+#endregion
+
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    public class GameObjectWorldCommandSystem : ComponentSystem
+    {
+        private struct ReserveEntityResponseData
```
JonasImprobable: Nit: Rename struct to `ReserveEntityIdResponseData`. :)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
JonasImprobable: Yeah sorry, that was me arbitrarily splitting the work into two jira tickets.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                this.logDispatcher = logDispatcher;
+                reserveEntityIdsDelegates = new List<Action<WorldCommands.ReserveEntityIds.ReceivedResponse>>();
+                createEntityDelegates = new List<Action<WorldCommands.CreateEntity.ReceivedResponse>>();
+                deleteEntityDelegates = new List<Action<WorldCommands.DeleteEntity.ReceivedResponse>>();
+            }
+
+            public event Action<WorldCommands.ReserveEntityIds.ReceivedResponse> OnReserveEntityIdsResponse
+            {
+                add => reserveEntityIdsDelegates.Add(value);
```
JonasImprobable: Now that #288 has been merged, could you wrap this in
```
if (!VerifyNotDisposed())
{
  return;
}
``` 
blocks?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
firtina-improbable: o, no, I think it should filter on that!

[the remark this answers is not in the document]
firtina-improbable: fixing

firtina-improbable: agreed

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
JonasImprobable: Now that #288 has been merged, could you wrap this in
```
if (!VerifyNotDisposed())
{
  return;
}
``` 
blocks?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+
+#endregion
+
+namespace Improbable.Gdk.Core.GameObjectRepresentation
+{
+    public static partial class WorldCommandsRequirables
+    {
+        [InjectableId(InjectableType.WorldCommandResponseHandler, InjectableId.NullComponentId)]
+        [InjectionCondition(InjectionCondition.RequireNothing)]
+        public class WorldCommandResponseHandler : IInjectable
```
JonasImprobable: Now that #288 has been merged, can you make this inherit `RequirableBase`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
firtina-improbable: aye

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
firtina-improbable: Aye

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
jessicafalk: I really get the feeling that we should probably store that directly on the entity, if it is needed in multiple systems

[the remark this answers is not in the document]
jessicafalk: but leave it for now and we will refactor it later

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,138 @@
...
+        protected override void OnUpdate()
+        {
+            for (var i = 0; i < reserveEntityIdsResponseData.Length; ++i)
+            {
+                var entity = reserveEntityIdsResponseData.Entities[i];
+                var worldCommandResponseHandlers = GetWorldCommandResponseHandlersForEntity(entity);
+
+                if (worldCommandResponseHandlers == null)
+                {
+                    continue;
```
jessicafalk: can that actually happen? if it has a GameObjectReference, shouldn't it also be in the storage?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
```diff
@@ -0,0 +1,141 @@
...
+
+                internal void OnReserveEntityIdsResponseInternal(
+                    ReserveEntityIds.ReceivedResponse receivedResponse)
+                {
+                    GameObjectDelegates.DispatchWithErrorHandling(receivedResponse, reserveEntityIdsDelegates,
+                        logDispatcher);
+                }
+
+                internal void OnCreateEntityResponseInternal(
+                    CreateEntity.ReceivedResponse receivedResponse)
```
jessicafalk: should all these methods be on two lines?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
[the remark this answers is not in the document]
firtina-improbable: Aye, UTY-1145 :)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
firtina-improbable: If the entity has none injected I think it may be null... I'll do a little test though

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
firtina-improbable: After the namespace refactor, they now can fit into their lines, so tidying that up ✔️

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
firtina-improbable: aye, it is null (also we have a `count==0?null` in the method that returns it too

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
```diff
@@ -0,0 +1,102 @@
...
+                    entityManager.GetComponentData<CreateEntity.CommandSender>(entity)
+                        .RequestsToSend.Add(request);
+
+                    return request.RequestId;
+                }
+
+                public long DeleteEntity(
+                    EntityId entityId,
+                    uint? timeoutMillis = null,
+                    object context = null)
```
jessicafalk: same here with the lines ;)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
```diff
@@ -0,0 +1,138 @@
...
+                        logDispatcher);
+                }
+
+                [InjectableId(InjectableType.WorldCommandResponseHandler, InjectableId.NullComponentId)]
+                private class WorldCommandResponseHandlerCreator : IInjectableCreator
+                {
+                    public IInjectable CreateInjectable(
+                        Entity entity,
+                        EntityManager entityManager,
+                        ILogDispatcher logDispatcher)
```
jessicafalk: and here :P

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            [ReadOnly] public readonly int Length;
+            [ReadOnly] public EntityArray Entities;
+            [ReadOnly] public ComponentDataArray<WorldCommands.ReserveEntityIds.CommandResponses> CommandResponses;
+            [ReadOnly] public ComponentArray<GameObjectReference> HasGameObjectReference;
+        }
+
+        private struct CreateEntityResponseData
+        {
+            [ReadOnly] public readonly int Length;
```
jessicafalk: I'm pretty sure having `[ReadOnly]` here is not allowed. that only works on collections

firtina-improbable: It didn't complain but I am guessing it's a noop, will clean up

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandResponseHandler.cs
firtina-improbable: Well, almost. I'll still do what I can

firtina-improbable: Also tidying up on the responder too

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                return null;
+            }
+
+            if (injectables.Count == 0)
+            {
+                return null;
+            }
+
+            return Array.ConvertAll(injectables.ToArray(),
```
zeroZshadow: This is not oke in a hot loop

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
```diff
@@ -0,0 +1,139 @@
...
+                injectable => (WorldCommands.Requirables.WorldCommandResponseHandler) injectable);
```
samiwh: This still iterates over both a hashset and an array and allocates two new arrays in the process. And the result is only used to iterate over the data set once.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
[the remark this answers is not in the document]
MPeti01: It's a hashset at the moment because we remove arbitrary elements from it as monobehaviours are disabled. But thinking about it more, that's not a good reason if we iterate much more often than we insert and don't have loads of readers/etc of the same type for a single entity, ie it should just be a list.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
samiwh: If we are going to change the backing thing to a list then I don't mind too much what we do here now as it will be changed soon (and world commands aren't sent all that often at the moment)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
```diff
@@ -0,0 +1,93 @@
+﻿using Improbable.Gdk.Core.GameObjectRepresentation;
+using Improbable.Worker;
+using Unity.Entities;
+
+#region Diagnostic control
+
+// ReSharper disable MemberHidesStaticFromOuterClass
```
samiwh: Ooh I hadn't thought of this problem. I actually don't think we want to get rid of this warning is it really is us doing something wrong.

samiwh: I think we can get rid of the partial keyword from Reacquirables and it will do the same thing.

samiwh: At least we shouldn't disable resharper warnings for this. We can get rid of warnings that get passed on to users.

firtina-improbable: The culprits are the methods:

```
public long ReserveEntityIds(uint numberOfEntityIds, uint? timeoutMillis = null, object context = null)
```

`WorldCommands.Requirables.WorldCommandRequestSender#ReserveEntityIds` shadows `WorldCommands.ReserveEntityIds` class

samiwh: Actually are we? This will make the correct code but does it do it in the way we think it does. Will consult the standard.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectWorldCommandSystem.cs
firtina-improbable: Offline discussion: let's do the optimisation in next PR

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
samiwh: "Nested types may be declared in multiple parts by using the partial modifier. Typically, the containing type is declared using partial as well, and each part of the nested type is declared in a different part of the containing type." - C# language spec

We're all good. Feel free to get rid of the warning.

PR #318, opened 2018-09-05
Making set up of game object initialization simpler

jessicafalk:

-------

#### Description
in commit order
1. removing systems that were not needed and just made the overall module difficult to read
2. moved `EntityGameObjectCreator` from Playground into `Core.GameObjectRepresentation`
3. the changes necessary to hook everything up correctly again. Basically moved some component additions / removals into internal classes so that users don't have to touch them. Made the `GameObjectDispatcherSystem` responsible for taking care of its dictionaries
4. fixing playground & tests

the `GameObjectInitializationSystem` is still in Playground here. We might need to move this into core as well, if we don't want users to use systems at all. The overall idea would be that this system only calls the `EntityGameObjectCreator` and the `EntityGameObjectLinker` to set up and clean up everything.  The user can then implement the `EntityGameObjectCreator` which is made into an interface in Peter's PR.

#### Tests
tested in playground and my project

#### Documentation
none for now


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectCreator.cs
```diff
@@ -27,7 +26,8 @@ public EntityGameObjectCreator(World world, Dictionary<string, GameObject> initi
                 prefab = Resources.Load<GameObject>(prefabPath);
                 if (prefab == null)
                 {
-                    throw new PrefabNotFoundException($"Prefab for prefabPath {prefabPath} not found.");
+                    return null;
+                    //throw new PrefabNotFoundException($"Prefab for prefabPath {prefabPath} not found.");
```
JonasImprobable: Let's remove this commented-out line.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -9,31 +10,54 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
     ///     Gathers incoming dispatcher ops and invokes callbacks on relevant GameObjects.
     /// </summary>
     [DisableAutoCreation]
+    [AlwaysUpdateSystem]
```
JonasImprobable: Nice!

jessicafalk:
This PR will go in after https://github.com/spatialos/UnityGDK/pull/288 & https://github.com/spatialos/UnityGDK/pull/272, to make sure everything is in order. So please review those first :)

workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
```diff
@@ -12,25 +12,32 @@ namespace Playground
...
+            [ReadOnly] public SubtractiveComponent<EntityGameObjectHandle> NewlyCreatedEntities;
         }
 
         private struct RemovedEntitiesData
         {
             public readonly int Length;
             public EntityArray Entities;
-            [ReadOnly] public ComponentDataArray<GameObjectReferenceHandle> GameObjectReferenceHandles;
-            public SubtractiveComponent<GameObjectReference> NoGameObjectReference;
+            [ReadOnly] public SubtractiveComponent<Prefab.Component> PrefabNames;
```
zeroZshadow: Subtracting on SpatialEntityId should be enough?

workers/unity/Packages/com.improbable.gdk.core/Systems/UpdateGroups.cs
```diff
@@ -13,18 +13,6 @@ public class InternalSpatialOSReceiveGroup
 
         [UpdateInGroup(typeof(SpatialOSReceiveGroup))]
         [UpdateAfter(typeof(InternalSpatialOSReceiveGroup))]
-        public class EntityInitialisationGroup
```
JonasImprobable: What is the reason behind this change? Why not introduce a group rather than using
```
[UpdateBefore(typeof(SpatialOSReceiveGroup.GameObjectReceiveGroup))]
[UpdateAfter(typeof(SpatialOSReceiveGroup.InternalSpatialOSReceiveGroup))]
```
in `GameObjectInitializationSystem` et al?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -9,31 +10,54 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
...
     [UpdateInGroup(typeof(SpatialOSReceiveGroup.GameObjectReceiveGroup))]
     internal class GameObjectDispatcherSystem : ComponentSystem
     {
+        public struct GameObjectReferenceHandle : ISystemStateComponentData
```
zeroZshadow: This name no longer represents what this system tag is for

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectCreator.cs
```diff
@@ -27,7 +26,8 @@ public EntityGameObjectCreator(World world, Dictionary<string, GameObject> initi
                 prefab = Resources.Load<GameObject>(prefabPath);
                 if (prefab == null)
                 {
-                    throw new PrefabNotFoundException($"Prefab for prefabPath {prefabPath} not found.");
+                    return null;
```
JonasImprobable: What is the reasoning behind this change? We are not even checking for null in the return value in `GameObjectInitializationSystem`.

workers/unity/Packages/com.improbable.gdk.core/Systems/UpdateGroups.cs
jessicafalk: mainly because it's in playground and as we know: transitivity for unused groups doesn't work that well ;) we could introduce a group in the playground though. Can make that change

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -46,20 +70,40 @@ protected override void OnCreateManager(int capacity)
...
         }
 
+
+        protected override void OnUpdate()
+        {
+            for (var i = 0; i < addedEntitiesData.Length; i++)
+            {
+                var entity = addedEntitiesData.Entities[i];
+                CreateActivationManagerAndReaderWriterStore(entity);
+                PostUpdateCommands.AddComponent(entity, new GameObjectReferenceHandle());
```
zeroZshadow: Change to `PostUpdateCommands.AddComponent(entity, ComponentType.Create<GameObjectReferenceHandle>())`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectCreator.cs
firtina-improbable: I agree. This should remain an exception, especially if it's trying to load a prefab from a path that does not contain a prefab.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
firtina-improbable: Is this added because of RunDispatchers? I think it would be nicer to split them up to separate systems in that case... RunDispatchers looks out of place now.

Or, the other way around: the dispatcher system should not be dealing with added / removed entities.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -9,31 +10,54 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
...
+        public struct GameObjectReferenceHandle : ISystemStateComponentData
+        {
+        }
+
+        public struct AddedEntitiesData
+        {
+            public readonly int Length;
+            public EntityArray Entities;
+            [ReadOnly] public ComponentArray<GameObjectReference> GameObjectReferences;
+            [ReadOnly] public SubtractiveComponent<GameObjectReferenceHandle> GameObjectReferenceHandles;
```
firtina-improbable: Subtractive component field is missing the `No` prefix

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
jessicafalk: I do agree! However the dispatchers use dictionaries that are populated when entities get added. So we either have a static class with this dictioniaries (which I don't like) or all the setup and running logic in one system. Or: put it as a component on the entity that might also be possible, though not sure how much implementation changes you want to do before m1

firtina-improbable: I see, the last option is my favorite. Can you create a ticket with the requirements and the options you suggested for the M2 milestone please, since you have the fresh context on what may need to be / can be done?

jessicafalk: sure, will do that.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -9,31 +10,54 @@ namespace Improbable.Gdk.Core.GameObjectRepresentation
...
         private readonly Dictionary<Entity, MonoBehaviourActivationManager> entityToActivationManager =
             new Dictionary<Entity, MonoBehaviourActivationManager>();
+
         private readonly Dictionary<Entity, InjectableStore> entityToReaderWriterStore =
             new Dictionary<Entity, InjectableStore>();
 
-        public readonly List<GameObjectComponentDispatcherBase> GameObjectComponentDispatchers =
-            new List<GameObjectComponentDispatcherBase>();
+        [Inject] private AddedEntitiesData addedEntitiesData;
+        [Inject] private ExistingEntitiesData existingEntitiesData;
```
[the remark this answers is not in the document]
jessicafalk: nope, that was left over from when I tested stuff. will remove it

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
jessicafalk: UTY-1159

workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
```diff
@@ -12,25 +12,32 @@ namespace Playground
...
+
         private struct AddedEntitiesData
         {
             public readonly int Length;
             public EntityArray Entities;
             [ReadOnly] public ComponentDataArray<Prefab.Component> PrefabNames;
             [ReadOnly] public ComponentDataArray<Transform.Component> Transforms;
             [ReadOnly] public ComponentDataArray<SpatialEntityId> SpatialEntityIds;
-            [ReadOnly] public ComponentDataArray<NewlyAddedSpatialOSEntity> NewlyCreatedEntities;
+            [ReadOnly] public SubtractiveComponent<EntityGameObjectHandle> NewlyCreatedEntities;
```
jamiebrynes7: The name for this field is inaccurate

workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
```diff
@@ -12,25 +12,32 @@ namespace Playground
...
         private struct RemovedEntitiesData
         {
             public readonly int Length;
             public EntityArray Entities;
-            [ReadOnly] public ComponentDataArray<GameObjectReferenceHandle> GameObjectReferenceHandles;
-            public SubtractiveComponent<GameObjectReference> NoGameObjectReference;
+            [ReadOnly] public SubtractiveComponent<Prefab.Component> PrefabNames;
+            [ReadOnly] public SubtractiveComponent<Transform.Component> Transforms;
+            [ReadOnly] public SubtractiveComponent<SpatialEntityId> SpatialEntityIds;
+            [ReadOnly] public ComponentDataArray<EntityGameObjectHandle> NewlyCreatedEntities;
```
jamiebrynes7: General note - any `SubtractiveComponent<T>` should be called something like `No T Here` or similar.

workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
```diff
@@ -77,27 +84,15 @@ protected override void OnUpdate()
...
-
-                PostUpdateCommands.AddComponent(addedEntitiesData.Entities[i], gameObjectReferenceHandleComponent);
-
-                PostUpdateCommands.AddComponent(addedEntitiesData.Entities[i],
-                    requiresSpatialOSBehaviourManagerComponent);
-
-                viewCommandBuffer.AddComponent(entity, gameObjectReference);
-                entityGameObjectLinker.LinkGameObjectToEntity(gameObject, entity, spatialEntityId,
-                    viewCommandBuffer);
+                PostUpdateCommands.AddComponent(entity, new EntityGameObjectHandle());
```
jamiebrynes7: Is there not a `PostUpdateCommands.AddComponent<EntityGameObjectHandle>(entity)`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectCreator.cs
jamiebrynes7: Also commented code 😬

jessicafalk: ok, in that case I'll also implement that check inside the interface here, because there are definitely entities where we don't want to create a prefab

workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
jessicafalk: nope, there is not. it's very troll

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -12,38 +12,39 @@ namespace Improbable.Gdk.GameObjectCreation
...
+        }
 
         private struct AddedEntitiesData
         {
             public readonly int Length;
-            public EntityArray Entities;
+            [ReadOnly] public EntityArray Entities;
             [ReadOnly] public ComponentDataArray<SpatialEntityId> SpatialEntityIds;
-            [ReadOnly] public ComponentDataArray<NewlyAddedSpatialOSEntity> NewlyCreatedEntities;
+            [ReadOnly] public SubtractiveComponent<HasGameObjectSystemState> NoHasGameObjectSystemStates;
```
firtina-improbable: Minor: `NoHas` reads kind of strange, perhaps in this case `HasNo` or `DoesNotHave` or something may work better.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -12,38 +12,39 @@ namespace Improbable.Gdk.GameObjectCreation
     ///     and links it to the entity via the EntityGameObjectLinker. Also checks for entity removal and calls the
     ///     IEntityGameObjectCreator for cleanup.
     /// </summary>
-    [UpdateInGroup(typeof(SpatialOSReceiveGroup.EntityInitialisationGroup))]
     [DisableAutoCreation]
     internal class GameObjectInitializationSystem : ComponentSystem
     {
-        // TODO: Add our ISystemStateComponentData for all entities instead of ones with GameObjects, store spatial ID
-        // callbacks instead of in it, give that in removal callback instead of SpatialOSEntity struct.
+        public struct HasGameObjectSystemState : ISystemStateComponentData
```
firtina-improbable: Minor (may help with the other comment too) - if it has fields in it, it could be renamed to something else, e.g. `SpatialEntityIdData` or something... - can't think of good names here but I am guessing you get my point)

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -12,38 +12,39 @@ namespace Improbable.Gdk.GameObjectCreation
...
+        {
+            public SpatialEntityId EntityId;
```
firtina-improbable: Is it meant to contain a copy of another `IComponentData` here?

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -62,53 +63,46 @@ protected override void OnUpdate()
...
-                    gameObjectCreator.OnEntityRemoved(new SpatialOSEntity(entity, EntityManager), null);
-                }
-                else
-                {
-                    entityGameObjectCache.Remove(entity);
-                    LinkerSystem.Linker.UnlinkGameObjectFromEntity(gameObject, entity, true, PostUpdateCommands);
-                    gameObjectCreator.OnEntityRemoved(new SpatialOSEntity(entity, EntityManager), gameObject);
-                }
+                var spatialEntityId = EntityManager.GetComponentData<HasGameObjectSystemState>(entity).EntityId;
+                var gameObject = entityToGameObjects[entity];
```
firtina-improbable: I am guessing that there is no chance of `entityToGameObjects` to not contain a key for the `entity` here because `RemovedEntitiesData` filters via `HasGameObjectSystemState`, but this changes behaviour slightly, the users may still want the `onentityremoved` to be called for entities which do not have corresponding gameobjects.

We can handle that situation in another ticket though, as it is probably not the usual flow.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -62,53 +63,46 @@ protected override void OnUpdate()
...
 
                 if (gameObject == null)
                 {
                     continue;
                 }
 
-                LinkerSystem.Linker.LinkGameObjectToEntity(gameObject, entity, spatialEntityId, PostUpdateCommands, viewCommandBuffer);
-                entityGameObjectCache.Add(entity, gameObject);
+                linkerSystem.Linker.LinkGameObjectToEntity(gameObject, entity, spatialEntityId.EntityId, viewCommandBuffer);
+                PostUpdateCommands.AddComponent(entity, new HasGameObjectSystemState { EntityId = spatialEntityId });
```
firtina-improbable: Found one problem here actually... if the gameobject is null I am afraid `addedEntitiesData` will never be empty in following frames as long as entities are alive, probably because 
```
             [ReadOnly] public ComponentDataArray<NewlyAddedSpatialOSEntity> NewlyCreatedEntities; 
```

is no longer used

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
```diff
@@ -40,30 +41,17 @@ internal EntityGameObjectLinker(World world, ILogDispatcher logDispatcher)
...
-            viewCommandBuffer.AddComponent(entity, gameObjectReference);
-
-            var gameObjectReferenceHandleComponent = new GameObjectReferenceHandle();
-            entityCommandBuffer.AddComponent(entity, gameObjectReferenceHandleComponent);
-
-            gameObjectDispatcherSystem.CreateActivationManagerAndReaderWriterStore(entity, gameObject);
         }
 
-        public void UnlinkGameObjectFromEntity(GameObject gameObject, Entity entity, bool removeEcsComponents, EntityCommandBuffer entityCommandBuffer)
+        public void UnlinkGameObjectFromEntity(GameObject gameObject, ViewCommandBuffer viewCommandBuffer)
```
zeroZshadow: What is the viewCommandBuffer used for?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -143,46 +199,33 @@ protected override void OnUpdate()
...
-            }
-        }
-
-        internal void CreateActivationManagerAndReaderWriterStore(Entity entity, GameObject gameObject)
+        private void CreateActivationManagerAndReaderWriterStore(Entity entity)
         {
             if (entityToActivationManager.ContainsKey(entity))
             {
-                throw new ActivationManagerAlreadyExistsException($"MonoBehaviourActivationManager already exists for entity {entity.Index}.");
+                throw new SystemException($"MonoBehaviourActivationManager already exists for entity {entity.Index}.");
```
samiwh: I don't think this is the exception type you meant. You should never be throwing SystemException.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -138,8 +138,6 @@ private void AddCoreSystems()
             World.GetOrCreateManager<WorldCommandsCleanSystem>();
             World.GetOrCreateManager<WorldCommandsSendSystem>();
             World.GetOrCreateManager<CommandRequestTrackerSystem>();
-
-            // MonoBehaviour API
```
samiwh: Why remove this? I don't know when it was changed to say API (as that seems wrong), but it was meant to denote the stuff which isn't actually core but needs to be added here for now.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
samiwh: In this case the name should represent what the component denotes. The data is just what is needed to clean up. Putting this in an injection group is meant to denote that is has a game object associated with it.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
samiwh: `DenotesThereIsNoGameObject`

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -12,38 +12,39 @@ namespace Improbable.Gdk.GameObjectCreation
...
+            public SpatialEntityId EntityId;
+        }
 
         private struct AddedEntitiesData
         {
             public readonly int Length;
-            public EntityArray Entities;
+            [ReadOnly] public EntityArray Entities;
             [ReadOnly] public ComponentDataArray<SpatialEntityId> SpatialEntityIds;
-            [ReadOnly] public ComponentDataArray<NewlyAddedSpatialOSEntity> NewlyCreatedEntities;
```
samiwh: This one is still needed as we aren't putting the game object state component on unless the gameobject isn't null. Given that we wan't to call entity removed for all entities, whether or not they have a gameobject, I'd suggest renaming the game object component to something that states it's had its opportunity. Or we just make the Spatial ID component a system state data.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
samiwh: We do want to call it even when there is no gameobject, we will need to predicate this on something different.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/IEntityGameObjectCreator.cs
```diff
@@ -22,6 +23,6 @@ public interface IEntityGameObjectCreator
         /// <param name="linkedGameObject">
         ///     The GameObject linked to the entity, or null if there is none.
         /// </param>
-        void OnEntityRemoved(SpatialOSEntity entity, GameObject linkedGameObject);
+        void OnEntityRemoved(SpatialEntityId entityId, GameObject linkedGameObject);
```
samiwh: This can just be an EntityId. Inside the SpatialOSEntity it's just a regular EntityId.

samiwh: Also we have mismatched names here. Generally you shouldn't rename arguments between the interface and the implementation.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/SpatialOSEntity.cs
```diff
@@ -20,12 +21,12 @@ internal SpatialOSEntity(Entity entity, EntityManager entityManager)
             SpatialOSEntityId = entityManager.GetComponentData<SpatialEntityId>(entity).EntityId;
         }
 
-        public bool HasComponent<T>() where T : struct, ISpatialComponentData, IComponentData
```
samiwh: If we are removing that then we should also be giving access to other component types. Either we only guarantee that the Spatial components are there and we just give a user that, or we give them full access to the entity. This might come down to how much we want to be able to use this object elsewhere.

workers/unity/Assets/Playground/Scripts/EntityInitialization/ArchetypeInitializationSystem.cs
```diff
@@ -9,7 +9,7 @@ namespace Playground
     /// <summary>
     ///     Optionally adds a list of components to newly spawned entities according to an archetype definition.
     /// </summary>
-    [UpdateInGroup(typeof(SpatialOSReceiveGroup.EntityInitialisationGroup))]
+    [UpdateInGroup(typeof(SpatialOSReceiveGroup.GameObjectReceiveGroup))]
```
samiwh: These seems like the wrong name for a group this system would be in.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinkerSystem.cs
```diff
@@ -7,9 +7,11 @@ public class EntityGameObjectLinkerSystem : ComponentSystem
     {
         public EntityGameObjectLinker Linker;
 
+        [Inject] private WorkerSystem worker;
+
         protected override void OnCreateManager(int capacity)
         {
-            Linker = new EntityGameObjectLinker(World, World.GetExistingManager<WorkerSystem>().LogDispatcher);
+            Linker = new EntityGameObjectLinker(World, worker);
```
samiwh: I don't think `worker` exists at this point.

samiwh: Also why give it a worker and a world?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -103,9 +159,9 @@ private void GenerateComponentGroups()
             }
         }
 
-        protected override void OnUpdate()
+        private void RunDispatchers()
```
samiwh: I think rename this. It isn't just dispatching it's basically dealing with all the lifecycle steps.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreatorFromMetadata.cs
```diff
@@ -67,15 +67,15 @@ public GameObject OnEntityCreated(SpatialOSEntity entity)
             }
 
             var gameObject = Object.Instantiate(prefab, position, Quaternion.identity);
-            gameObject.name = $"{prefab.name}(SpatialOS: {entity.SpatialOSEntityId}, worker {workerType}";
+            gameObject.name = $"{prefab.name}(SpatialOS: {entity.SpatialOSEntityId}, Worker: {workerType}";
             return gameObject;
         }
 
-        public void OnEntityRemoved(SpatialOSEntity entity, GameObject linkedGameObject)
+        public void OnEntityRemoved(SpatialEntityId entityId, GameObject gameObject)
```
samiwh: Mentioned somewhere else but this should use the same arg names as the interface does. I don't really see anything wrong with linked being in the name, but I don't really mind one way or the other.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -12,38 +12,39 @@ namespace Improbable.Gdk.GameObjectCreation
...
         [Inject] private RemovedEntitiesData removedEntitiesData;
 
-        [Inject] private EntityGameObjectLinkerSystem LinkerSystem;
+        [Inject] private EntityGameObjectLinkerSystem linkerSystem;
         [Inject] private WorkerSystem worker;
 
         private ViewCommandBuffer viewCommandBuffer;
-        private readonly Dictionary<Entity, GameObject> entityGameObjectCache = new Dictionary<Entity, GameObject>();
         private readonly IEntityGameObjectCreator gameObjectCreator;
+        private readonly Dictionary<Entity, GameObject> entityToGameObjects = new Dictionary<Entity, GameObject>();
```
samiwh: nitpick - private readonly should go above non readonly. Injected is probably a special case that goes wherever it seems best.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/SpatialOSEntity.cs
samiwh: The purpose of this is to be the thing you use when you don't want to touch the ECS though, so restricting it might just make things easier.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/ToggleRotationCommandSender.cs
```diff
@@ -25,7 +25,7 @@ private void OnEnable()
         {
             if (obj.StatusCode != StatusCode.Success)
             {
-                GetComponent<SpatialOSComponent>().LogDispatcher.HandleLog(LogType.Error,
+                GetComponent<SpatialOSComponent>().Worker.LogDispatcher.HandleLog(LogType.Error,
```
jared-improbable: General comment: Logging is getting pretty verbose.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
jared-improbable: Interpolate in the `nameof` the class so it matches up when things are invariably refactored/renamed (there are a few instances of this)

workers/unity/Assets/Playground/Scripts/MonoBehaviours/ToggleRotationCommandSender.cs
firtina-improbable: I'm tempted to propose injecting the logger but I realise I am spending a lot of time with injectables so I am unsure if I'm making it a "when you have a hammer everything looks like a nail" moment.

How about adding logging methods to the spatialos component? Like, same signature as `Debug.Log` but `mySpatialOSComponent.Log` / `LogError`

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
jessicafalk: `Denotes`also fits with how we are doing it elsewhere, so that sounds and reads nicely. Not adding an underscore though :P

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
jessicafalk: should probably be directly the entity id. will change it

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
jessicafalk: Why would we want it to be called for entities that have no game objects, if that system is all about creating / deleting gameobjects?

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
jessicafalk: good spot! will add the `NewlyAddedSpatialOSEntity` back into the struct

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
jessicafalk: for nothing anymore :)

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
jessicafalk: in that case we should already provide a helper class for that as we do for the other modules. will add that

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/IEntityGameObjectCreator.cs
jessicafalk: yeah, that happened by accident. reverting it

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/SpatialOSEntity.cs
jessicafalk: discussed offline: reverting it for now as we don't have any use case where we would want to get a component that is not an `ISpatialComponentData`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinkerSystem.cs
jessicafalk: Tested it and the `worker` does exist

as to why both: we don't have access to the `World` from the `WorkerSystem` and we want to store the `World` inside the `SpatialOSComponent`

workers/unity/Assets/Playground/Scripts/MonoBehaviours/ToggleRotationCommandSender.cs
jessicafalk: the idea behind adding the logger to the worker was to be able to directly identify which worker is logging. Though there have already been discussions going on to refactor this

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
jessicafalk: actually, just saw that we already have it and we should have removed this line here completely

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinkerSystem.cs
firtina-improbable: the WorkerSystem could expose its own world (via base class) - or the linker could get the system using the world e.g. the deleted code can be moved within the `EntityGameObjectLinker`

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
jessicafalk: this would be interesting, but doesn't sound like it should be in this system. This would most likely be implemented in a separate one

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinkerSystem.cs
jessicafalk: the other question is: do we want to expose the world to the mono behaviours? we start mixing ECS and monobehaviours in a confusing way

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/GameObjectRepresentation/EntityGameObjectLinkerTests.cs
```diff
@@ -49,7 +50,7 @@ public void TearDown()
         public void LinkGameObjectToEntity_adds_SpatialOSComponent_component()
         {
             entityGameObjectLinker.LinkGameObjectToEntity(testGameObject, testEntity, testSpatialEntityId,
-                entityCommandBuffer, viewCommandBuffer);
```
JonasImprobable: We can delete the private `entityCommandBuffer` field now as it's not used anymore.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreatorFromMetadata.cs
```diff
@@ -67,11 +68,11 @@ public GameObject OnEntityCreated(SpatialOSEntity entity)
             }
 
             var gameObject = Object.Instantiate(prefab, position, Quaternion.identity);
-            gameObject.name = $"{prefab.name}(SpatialOS: {entity.SpatialOSEntityId}, worker {workerType}";
+            gameObject.name = $"{prefab.name}(SpatialOS: {entity.SpatialOSEntityId}, Worker: {workerType}";
```
JonasImprobable: Can we close the open bracket with a ")"? Can we add the entity index back again?

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -12,38 +14,41 @@ namespace Improbable.Gdk.GameObjectCreation
...
     ///     IEntityGameObjectCreator for cleanup.
     /// </summary>
-    [UpdateInGroup(typeof(SpatialOSReceiveGroup.EntityInitialisationGroup))]
     [DisableAutoCreation]
+    [UpdateInGroup(typeof(SpatialOSReceiveGroup.GameObjectReceiveGroup))]
     internal class GameObjectInitializationSystem : ComponentSystem
     {
-        // TODO: Add our ISystemStateComponentData for all entities instead of ones with GameObjects, store spatial ID
-        // callbacks instead of in it, give that in removal callback instead of SpatialOSEntity struct.
+        private struct HasGameObjectSystemState : ISystemStateComponentData
```
JonasImprobable: Nice!

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
JonasImprobable: Agreed. What they want is us exposing more of the dispatcher and giving users a general signal for entity added/removed. I'm ok with yoloing things together for now but we should revisit this.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreatorFromMetadata.cs
MPeti01: The entity index was removed because it really doesn't make sense if this API is supposed to be about abstracting away the ECS and apparently doesn't even identify the Entity uniquely in itself (which is a reason for either exposing nothing or exposing the entire Entity, and the former makes more sense in the context)

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
MPeti01: Matt also said it's reasonable to expect GameObject instantiation requiring cleanup to happen even in cases where they wouldn't want to link. Plus having it symmetric is less of a gotcha. I'm OK with it either way but when I previously asked around I did get the impression that's the functionality people would expect.

workers/unity/Packages/com.improbable.gdk.core/Systems/UpdateGroups.cs
JonasImprobable: That makes sense. Damn, this is really unfortunate.

workers/unity/Assets/Playground/Scripts/EntityInitialization/ArchetypeInitializationSystem.cs
```diff
@@ -9,7 +9,7 @@ namespace Playground
     /// <summary>
     ///     Optionally adds a list of components to newly spawned entities according to an archetype definition.
     /// </summary>
-    [UpdateInGroup(typeof(SpatialOSReceiveGroup.EntityInitialisationGroup))]
+    [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
```
JonasImprobable: Doesn't this technically need to run before GameObjects are created? Since this system could potentially add more components that users could rely on during the gameobject creation process?

MPeti01: How would they rely on those ECS components in the GameObject context? If they're writing a custom system to replace our GameObjectInitializationSystem they can 1) have that run after the Archetype stuff, 2) customize the Archetype stuff too. If they don't, they wouldn't have access to those components anyway right?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
```diff
@@ -40,30 +38,17 @@ internal EntityGameObjectLinker(World world, ILogDispatcher logDispatcher)
...
-            var gameObjectReference = new GameObjectReference { GameObject = gameObject };
-            viewCommandBuffer.AddComponent(entity, gameObjectReference);
-
-            var gameObjectReferenceHandleComponent = new GameObjectReferenceHandle();
-            entityCommandBuffer.AddComponent(entity, gameObjectReferenceHandleComponent);
-
-            gameObjectDispatcherSystem.CreateActivationManagerAndReaderWriterStore(entity, gameObject);
         }
 
-        public void UnlinkGameObjectFromEntity(GameObject gameObject, Entity entity, bool removeEcsComponents, EntityCommandBuffer entityCommandBuffer)
```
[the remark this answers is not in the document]
jessicafalk: that is a good point. In that case we need to add an additional check to see if the entity still exists. In the current use case, the entity is always destroyed / will get destroyed very soon

workers/unity/Assets/Playground/Scripts/EntityInitialization/ArchetypeInitializationSystem.cs
jessicafalk: this system is a playground system anyway. If users want to do that, they can create their own system and run it in the correct update loop. Right now, we are not depending on it so I don't see a reason to have it anything but the `SpatialOSUpdateGroup`

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreatorFromMetadata.cs
jessicafalk: will add the ")" back in though :)

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
samiwh: We definitely want it to be called for everything.
Also no user is going to get confused about it being called on something which has a no op. At worst they get one null ref and work out what is going on and that seems unlikely. The other way round you have a hard to debug issue where you can't work out why you aren't getting a remove entity for something you know was removed.

samiwh: Returning gameobject is a temporary hack to stop users tripping over a current limitation. Eventualyl we will just pass them the linker.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinkerSystem.cs
firtina-improbable: good question, the mixing does not sound nice, I'm OK with either way

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
jessicafalk: However in the cleanup, we only unlink the game object and destroy. If we never create it in the first place, I don't see a reason to have it clean up anything.  This could just cause more troubles...

we should create a separate system, if we want a general callback for added / removed entities

workers/unity/Assets/Playground/Scripts/MonoBehaviours/ToggleRotationCommandSender.cs
firtina-improbable: Yes having a simplified / not verbose logger API could be something for a next milestone

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
samiwh: That's just that one implementation of the creator. You could also want to return things to the pool, delete non linked entities. Then there is the issue above of how much of a gotcha is it.

samiwh: The alternative is that we make users write two implementions of two different interfaces that receive almost identical information and have to both deal with gameobject lifecycles.

samiwh: Once this is just in a generic gameobject feature module and the linker can support all the stuff it needs, it will just be a generic entity added and removed api that also passes you the worker information and the linker (or you already have the worker information and it just passes you the linker).

It's primary use will then be gamobject lifecycle stuff but it could be whatever. It's just a helper that relies on core and monobehaviours.

samiwh: Another idea was things you can inject that give you worker (rather than entity) callbacks and users bootstrap this somehow on the worker entity. But it is messier and has that bootstrapping problem.

samiwh: They'd have to deal with it in two places if they didn't link something explicitly but it does affect the scene.

samiwh: I don't follow the second paragraph.

jessicafalk: I'll just revert the change for now. We shouldn't discuss design decisions in a PR. Let's talk more next week when I'm back

samiwh: Eventually it will also be the case that you can link whenever. Or we make something that tell you when things are added and removed and it's up to you to grab a reference to the linker. However those are all variants on this that we can't do right now. But having two sets of static methods seems like it's worse.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinkerSystem.cs
samiwh: I think we want to give users the ability to use the ECS we just don't want them to have to use it. Have a reference to the world that they will not normally find is fine. Occasionally you need to get stuff for a hack.

samiwh: We don't want to mix things in places where users then have to use ECS objects in their monobehaviour workflow. If they grab that world object they have gone looking for it.

samiwh: So I think this code is fine.

samiwh: Also I'm not sure what other system you'd write without it that didn't just replace monobehaviours. Something needs to hold a reference to something that can see systems in order for that to work.

workers/unity/Assets/Playground/Scripts/EntityInitialization/ArchetypeInitializationSystem.cs
samiwh: Even in playground this is under constrained. Users need to be able able to rely on the new components being added while reading the initial component updates. This may eventually cause us odd bugs.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/ToggleRotationCommandSender.cs
samiwh: There are a couple of soft plans for the logger. Injecting it doesn't seem insane in monobehaviours. But yes, not for this PR.

workers/unity/Assets/Playground/Scripts/EntityInitialization/ArchetypeInitializationSystem.cs
jessicafalk: I guess we can just change it to being updates before `SpatialOSUpdateGroup`?

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -63,52 +68,51 @@ protected override void OnUpdate()
...
                 var entity = removedEntitiesData.Entities[i];
-
-                if (!entityGameObjectCache.TryGetValue(entity, out var gameObject))
+                var spatialEntityId = EntityManager.GetComponentData<InitializedEntitySystemState>(entity).EntityId;
+                var gameObject = entityToGameObjects[entity];
+                if (gameObject == null)
                 {
-                    // Entity without linked GameObject removed
-                    gameObjectCreator.OnEntityRemoved(new SpatialOSEntity(entity, EntityManager), null);
+                    gameObjectCreator.OnEntityRemoved(spatialEntityId, gameObject);
```
samiwh: Nit - can pull this out of the if

PR #320, opened 2018-09-05
Add tags properly

zeroZshadow:
#### Description
Starting entities preview 12, unity will be strict in disallowing empty `IComponentData` tags to be added through `EntityManager.AddComponentData`
This change fixes the these in advance.

The upgrade itself is blocked on a bug fix that is required for adding tags through the EntityCommandBuffers.


workers/unity/Assets/Playground/Scripts/EntityInitialization/GameObjectInitializationSystem.cs
```diff
@@ -1,10 +1,10 @@
 using System.Collections.Generic;
-using Generated.Playground;
```
jessicafalk: revert that change. prefabs are not in that system anymore and just makes merging harder :)

PR #321, opened 2018-09-05
Demonstrate example usage for world commands

firtina-improbable:

-------

#### Description
Adds example usage for entity reservation, creation and deletion

cube-spawning

#### Tests
These are the tests for playground

#### Primary reviewers

@MPeti01 + one non-monobehaviour team person


workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
```diff
@@ -0,0 +1,113 @@
...
+        private void OnEntityIdsReserved(WorldCommands.ReserveEntityIds.ReceivedResponse response)
+        {
+            if (!sentRequestIds.Remove(response.RequestId))
+            {
+                // This response was not for a command from this behaviour.
+                return;
+            }
+
+            var responseOp = response.Op;
+            // TODO make a ticket to make world command responses match with normal command responses
```
firtina-improbable: UTY-1162

workers/unity/Assets/Playground/Config/EntityTemplateUtils.cs
```diff
@@ -0,0 +1,14 @@
+using System.Collections.Generic;
+
+namespace Playground
+{
+    public static class EntityTemplateUtils
+    {
+        public static readonly List<string> AllWorkerAttributes =
```
samiwh: This would be better as part of worker utils. However it's also worth noting that this is the worker types and is only the attribute set by coincidence. The worker names and attributes don't have to be the same.

samiwh: This seems to be something we conflate in a lot of places though.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
```diff
@@ -0,0 +1,107 @@
...
+            cubeSpawnerCommandSender.SendSpawnCubeRequest(ownEntityId, new Empty());
+        }
+
+        private void SendDeleteCubeCommand()
+        {
+            var spawnedCubes = GetSpawnedCubes(cubeSpawnerReader.Data);
+
+            if (spawnedCubes.Count == 0)
+            {
+                logDispatcher.HandleLog(LogType.Warning, new LogEvent("No cubes left to delete."));
```
samiwh: This isn't really a warning; it is expected behaviour in a common non-fail-state.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                logDispatcher.HandleLog(LogType.Error,
+                    new LogEvent(string.Format(CouldNotDeleteEntityWithId,
+                        entityId, op.Message)));
+                return;
+            }
+
+            var spawnedCubes = CubeSpawnerInputBehaviour.GetSpawnedCubes(writer.Data);
+
+            if (!spawnedCubes.Remove(entityId))
```
samiwh: Is it valid to do this? Surely removing from that list leaves you in an inconsistent state.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                logDispatcher.HandleLog(LogType.Error,
+                    new LogEvent(string.Format(TheEntityHasBeenUnexpectedlyRemovedFromTheList, entityId)));
```
samiwh: Nit - if this is the only place this is used does it need to be pulled out into something else?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
firtina-improbable: I see, will make it a normal log

workers/unity/Assets/Playground/Config/CubeTemplate.cs
```diff
@@ -0,0 +1,44 @@
...
+            );
+
+            var cubeColor = CubeColor.Component.CreateSchemaComponentData();
+            var cubeTargetVelocity = CubeTargetVelocity.Component.CreateSchemaComponentData(new Vector3f { X = -2.0f });
+            var prefab = Prefab.Component.CreateSchemaComponentData(entityType);
+            var launchable = Launchable.Component.CreateSchemaComponentData(new EntityId(0));
+            var archetypeComponent = ArchetypeComponent.Component.CreateSchemaComponentData(entityType);
+
+            var entity = EntityBuilder.Begin()
+                .AddPosition(coords.X, 0, coords.Z, WorkerUtils.UnityGameLogic)
```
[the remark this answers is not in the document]
firtina-improbable: 👍

workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+        private void Update()
+        {
+            if (playerInputReader.Authority != Authority.Authoritative)
+            {
+                // Only send commands if we're the player with input
+                return;
+            }
```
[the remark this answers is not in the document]
firtina-improbable: Yes that sounds pretty bad, writer it is!

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
firtina-improbable: Could make an explicit copy of the list with this part missing

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
firtina-improbable: it was just a very long line

firtina-improbable: but funnily I have the basic string in the variable name so it's nonsense, inlining

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
firtina-improbable: will do that

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
```diff
@@ -0,0 +1,112 @@
...
+// ReSharper disable PossibleInvalidOperationException
+
+#endregion
+
+namespace Playground.MonoBehaviours
+{
+    public class SpawnCubeCommandReceiver : MonoBehaviour
+    {
+        [Require] private Transform.Requirables.Reader transformReader;
+        [Require] private CubeSpawner.Requirables.CommandRequestHandler commandRequestHandler;
```
samiwh: better to name the variable according to which command

workers/unity/Assets/Playground/Config/EntityTemplateUtils.cs
firtina-improbable: will move it into workerutils and rename to workertypes, if the user of the variable wants to use the worker types as attributes, it's on them (well us, but in a different context so I think the language works that way)

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
```diff
@@ -0,0 +1,112 @@
...
+        [Require] private CubeSpawner.Requirables.Writer writer;
```
samiwh: Unless there is only type in a behaviour it's better to name the writers according to which component they are on

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
firtina-improbable: I could go for cubeSpawnerCommandRequestHandler

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
firtina-improbable: +1

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
```diff
@@ -0,0 +1,112 @@
...
+            requestResponder.SendResponse(new Empty());
+
+            sentRequestIds.Add(worldCommandRequestSender.ReserveEntityIds(1));
+        }
+
+        private void OnEntityIdsReserved(WorldCommands.ReserveEntityIds.ReceivedResponse response)
+        {
+            if (!sentRequestIds.Remove(response.RequestId))
+            {
+                // This response was not for a command from this behaviour.
```
samiwh: Are we not filtering this at a layer above this? Or do we have multiple behavours per entity? Or is this trying to solve a race where the entity doing this is checked in then back out while a request was in progress (which I think core already handles gracefully)?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                return;
+            }
+
+            var location = transformReader.Data.Location;
+            var cubeEntityTemplate =
+                CubeTemplate.CreateCubeEntityTemplate(new Coordinates(location.X, location.Y + 2, location.Z));
+            var expectedEntityId = responseOp.FirstEntityId.Value;
+
+            sentRequestIds.Add(worldCommandRequestSender.CreateEntity(cubeEntityTemplate, expectedEntityId));
```
samiwh: I don't think request IDs are guaranteed to be unique across different worker types

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
firtina-improbable: It can potentially be picked up by this if this entity is used as a command sender by another behaviour, or even through ECS.

The filtering checks that the entity has the responders injected but not "what thing sent the request" - as potentially it could be something the users may want to handle like split sending / responding to different behaviours or even gradually move to/from ECS; I'm stretching my imagination quite far here though

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            if (!sentRequestIds.Remove(response.RequestId))
+            {
+                // This response was not for a command from this behaviour.
+                return;
+            }
+
+            var createEntityResponseOp = response.Op;
+
+            if (createEntityResponseOp.StatusCode != StatusCode.Success)
```
samiwh: Nitpick - we aren't dealing with the case where the entity was created but the response timed out. I don't think we have to deal with it but worth noting there is an unhandled failure case here.
Another failure case is you losing authority before the response comes back. Again I don't think we have to deal with it.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
```diff
@@ -0,0 +1,93 @@
...
+            sentRequestIds.Add(worldCommandRequestSender.DeleteEntity(entityId));
+        }
+
+
+        private void OnDeleteEntityResponse(WorldCommands.DeleteEntity.ReceivedResponse response)
+        {
+            if (!sentRequestIds.Remove(response.RequestId))
+            {
+                // This response was not for a command from this behaviour.
+                return;
```
zeroZshadow: The request Id has now leaked on the worker that did send. Maybe clean in OnDisable?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
```diff
@@ -0,0 +1,93 @@
...
+        private readonly HashSet<long> sentRequestIds = new HashSet<long>();
+
+        public void OnEnable()
+        {
+            logDispatcher = GetComponent<SpatialOSComponent>().LogDispatcher;
+            commandRequestHandler.OnDeleteSpawnedCubeRequest += OnDeleteSpawnedCubeRequest;
+            worldCommandResponseHandler.OnDeleteEntityResponse += OnDeleteEntityResponse;
+        }
+
+        private void OnDeleteSpawnedCubeRequest(CubeSpawner.DeleteSpawnedCube.RequestResponder requestResponder)
```
samiwh: I distinctly remember this was meant to be removed

workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            if (response.StatusCode != StatusCode.Success)
+            {
+                logDispatcher.HandleLog(LogType.Error, new LogEvent($"Delete error: {response.Message}"));
+            }
+        }
+
+        private void Update()
+        {
+            if (playerInputReader.Authority != Authority.Authoritative)
```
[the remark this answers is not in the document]
firtina-improbable: Yes, will go with the writer. I think it should be safe that a player would not lose authority over their input but if that's not the case I may need to rethink my assumptions

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
samiwh: I confused this with the pre-existing types when reading this. Which means it will also cause confusion for which you are meant to use elsewhere. Also I got very confused about the lack of something to send the response when everything else has one.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+        private void OnEnable()
+        {
+            var spatialOSComponent = GetComponent<SpatialOSComponent>();
+            logDispatcher = spatialOSComponent.LogDispatcher;
+            ownEntityId = spatialOSComponent.SpatialEntityId;
+
+            cubeSpawnerCommandResponseHandler.OnSpawnCubeResponse += OnSpawnCubeResponse;
```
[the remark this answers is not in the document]
firtina-improbable: Not necessary as the command response handlers get disposed of automatically.

I agree that this feels unintuitive, I will link to discussions about this offline.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
```diff
@@ -0,0 +1,93 @@
...
+            if (!spawnedCubes.Contains(entityId))
+            {
+                requestResponder.SendResponseFailure($"Requested entity id {entityId} not found in list.");
+            }
+            else
+            {
+                requestResponder.SendResponse(new Empty());
+            }
+
+            sentRequestIds.Add(worldCommandRequestSender.DeleteEntity(entityId));
```
samiwh: for low traffic things we could add the context. Alternatively we could split this into two lines. Having the main part of the line in the function arguments obfuscates the intent somewhat

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
firtina-improbable: Is that the `RequestResponder`? If that is the case it may be part of an open ticket

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
firtina-improbable: I'll go for the two lines

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
samiwh: Would it be better to see if the entity is there and let the component updates do the rest?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
samiwh: We won't have this playground by then :P

samiwh: Also we know how to solve it. The question is how correct or readable we want this code to be.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
samiwh: But do we guarantee that or is it an implementation detail?

workers/unity/Assets/Playground/Config/EntityTemplateUtils.cs
samiwh: In this case it makes sense to call it AllWorkerAttributes as that is how it's being used. But we should consider separating the type and layer into differnt variables.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
firtina-improbable: actually context will be better indeed, I can just pass `this`!

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
firtina-improbable: will use the context instead so won't need to worry about that

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
firtina-improbable: I am confident they are unique but I am aware it's not very obvious that they are, but I'll eliminate the problem here by using the context instead

workers/unity/Assets/Playground/Config/EntityTemplateUtils.cs
firtina-improbable: Ah I see

firtina-improbable: I find it quite convenient in one way to not have to represent a worker with different kinds of identifiers. Would it make sense to have a different number of attributes than worker types? I am imagining having different kinds of `UnityClient`s or even `UnityGameLogic`s with different purposes (e.g. mobile can see one component and standalone can see another / fire UnityGameLogic can write to one component / ice UnityGameLogic can write to another component)? Should we create a ticket to distinguish better?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
firtina-improbable: What do you mean by "the lack of something to send the response", and by "everything else"?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
samiwh: Ah nice idea.

firtina-improbable:
Running tests

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
```diff
@@ -0,0 +1,91 @@
...
+                logDispatcher.HandleLog(LogType.Error,
+                    new LogEvent(string.Format(CouldNotDeleteEntityWithId,
+                        entityId, op.Message)));
+                return;
+            }
+
+            var spawnedCubesCopy =
+                new List<EntityId>(CubeSpawnerInputBehaviour.GetSpawnedCubes(cubeSpawnerWriter.Data));
+
+            if (!spawnedCubesCopy.Remove(entityId))
```
samiwh: This isn't really a valid thing to do. If something goes wrong this will leave you with inconsistent data. Why not just use write?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/DeleteCubeCommandReceiver.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                logDispatcher.HandleLog(LogType.Error,
```
samiwh: I think this is the only place that removes from that list. So this error isn't needed if don't do this here.

PR #335, opened 2018-09-07
Fix a few problems I encountered during dogfooding

firtina-improbable:

-------

#### Description
Each commit to fix another problem
#### Tests
Testing with an external project
#### Documentation
Internal fixes

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/Common.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/Common.cs
@@ -121,7 +121,7 @@ namespace Improbable.Gdk.Tools
                 {
                     if (!splitPath.Contains(macPath))
                     {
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/com.improbable.gdk.tools/Common.cs
```diff
@@ -67,7 +67,14 @@ internal static string GetThisPackagePath()
             }
 
             path = path.Replace("file:", string.Empty);
-            path = $"{PackagesDir}/{path}";
+
+            if (Path.IsPathRooted(path))
+            {
+                // A "rooted path" is an absolute path, therefore it will should point directly at the package.
```
jared-improbable: Remove "should"

firtina-improbable: 👍

firtina-improbable:
doing a local sanity test with clean project just in case

firtina-improbable:
sanity test passed, things still work from clean state

PR #336, opened 2018-09-07
Add WorkType is a well known field for LogEvents

zeroZshadow:
#### Description
Logs are now much clearer when you have multiple workers in your editor

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
@@ -15,7 +15,7 @@ namespace Improbable.Gdk.Core
 
         private bool inHandleLog;
 
-        public Connection Connection { get; set; }
+        public Worker Worker { get; set; }
 
         private static readonly Dictionary<LogType, LogLevel> LogTypeMapping = new Dictionary<LogType, LogLevel>
         {
@@ -44,7 +44,7 @@ namespace Improbable.Gdk.Core
             // This is required to avoid duplicate forwarding caused by HandleLog also logging to console
             if (type == LogType.Exception)
             {
-                Connection?.SendLogMessage(LogLevel.Error, Connection.GetWorkerId(), $"{message}\n{stackTrace}");
+                Worker.Connection?.SendLogMessage(LogLevel.Error, Worker.Connection.GetWorkerId(), $"{message}\n{stackTrace}");
             }
         }
 
@@ -53,6 +53,11 @@ namespace Improbable.Gdk.Core
             inHandleLog = true;
             try
             {
+                if (!string.IsNullOrEmpty(Worker?.WorkerType))
+                {
+                    logEvent.WithField(LoggingUtils.WorkerType, Worker.WorkerType);
+                }
+
                 if (type == LogType.Exception)
                 {
                     // For exception types, Unity expects an exception object to be passed.
@@ -71,7 +76,7 @@ namespace Improbable.Gdk.Core
 
                 var logLevel = LogTypeMapping[type];
 
-                if (Connection == null || logLevel < minimumLogLevel)
+                if (Worker?.Connection == null || logLevel < minimumLogLevel)
                 {
                     return;
                 }
@@ -86,7 +91,7 @@ namespace Improbable.Gdk.Core
 
                 var entityId = LoggingUtils.ExtractEntityId(logEvent.Data);
 
-                Connection.SendLogMessage(logLevel, LoggingUtils.ExtractLoggerName(logEvent.Data), message, entityId);
+                Worker.Connection.SendLogMessage(logLevel, LoggingUtils.ExtractLoggerName(logEvent.Data), message, entityId);
             }
             finally
             {
@@ -96,7 +101,7 @@ namespace Improbable.Gdk.Core
 
         public void Dispose()
         {
-            Connection = null;
+            Worker = null;
             Application.logMessageReceived -= LogCallback;
         }
     }
--- a/workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
@@ -44,7 +44,7 @@ namespace Improbable.Gdk.Core
             // This is required to avoid duplicate forwarding caused by HandleLog also logging to console
             if (type == LogType.Exception)
             {
-                Worker.Connection?.SendLogMessage(LogLevel.Error, Worker.Connection.GetWorkerId(), $"{message}\n{stackTrace}");
+                Worker?.Connection?.SendLogMessage(LogLevel.Error, Worker.Connection.GetWorkerId(), $"{message}\n{stackTrace}");
             }
         }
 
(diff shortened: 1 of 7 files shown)


workers/unity/Packages/com.improbable.gdk.core/Logging/LogEvent.cs
```diff
@@ -53,19 +53,19 @@ public override string ToString()
...
             {
                 builder.AppendLine();
-
-                builder.AppendLine("Log event data:");
-
-                foreach (var entry in Data)
-                {
-                    builder.AppendLine($"'{entry.Key}': '{entry.Value}'");
-                }
+                builder.Append($"'{entry.Key}': '{entry.Value}'");
```
jared-improbable: Won't this in effect print WorkerType twice? Seems redundant.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -66,8 +66,8 @@ public bool TryGetEntity(EntityId entityId, out Entity entity)
 
                 var worker = new Worker(config.WorkerType, connection, logger, origin);
                 logger.HandleLog(LogType.Log, new LogEvent("Successfully created a worker")
-                    .WithField("WorkerType", worker.WorkerType)
-                    .WithField("WorkerId", worker.WorkerId));
+                    .WithField("WorkerId", worker.WorkerId)
+                    .WithField(LoggingUtils.WorkerType, worker.WorkerType));
```
jamiebrynes7: Why add this back in if its automatically added?

zeroZshadow: In this instance, Worker is still null. So I pass the name in manually

zeroZshadow: Oops this one doesn't need it.

workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
```diff
@@ -15,7 +15,7 @@ public class ForwardingDispatcher : ILogDispatcher
 
         private bool inHandleLog;
 
-        public Connection Connection { get; set; }
+        public Worker Worker { get; set; }
```
samiwh: Discussed offline that this needs to be WorkerSystem.

zeroZshadow: I really don't agree here however?
This entire thing has nothing todo with ecs, and the WorkerSystem is nothing but a helper to get to worker owned bits. So why not just link directly to the worker?
Originally the Worker constructor even set the connection on the logDispatched it was passed, now it sets the worker instead.

jessicafalk: agree with that. We might want to use the dispatcher in monobehaviours which should not be dependent on the `WorkerSystem`

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -98,7 +97,6 @@ public bool TryGetEntity(EntityId entityId, out Entity entity)
 
                     var worker = new Worker(config.WorkerType, connection, logger, origin);
                     logger.HandleLog(LogType.Log, new LogEvent("Successfully created a worker")
-                        .WithField("WorkerType", worker.WorkerType)
                         .WithField("WorkerId", worker.WorkerId));
```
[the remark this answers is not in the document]
zeroZshadow: I think at some point we need to refactor the logging again to deal with well known fields like this. Also others like `LoggerName`.
But for now, this is just ment to show clearly what workertype logged what, to make debugging in the editor easier.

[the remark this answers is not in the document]
jessicafalk: that's why I was wondering whether it should be the worker id instead of the worker type to make debugging even easier :P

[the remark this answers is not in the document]
zeroZshadow: The issue is that Worker ID's are massive, and the console only shows 2 rows of debug message.
Using the ID will most likely move the message itself to row 2

workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
jamiebrynes7: With the current impl - you wouldn't create a dispatcher in MB code, so whether it depends on `Worker` vs `WorkerSystem` doesn't really matter.

Having said that, I'm in the camp of `Worker` > `WorkerSystem`

workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
```diff
@@ -44,7 +44,7 @@ private void LogCallback(string message, string stackTrace, LogType type)
             // This is required to avoid duplicate forwarding caused by HandleLog also logging to console
             if (type == LogType.Exception)
             {
-                Connection?.SendLogMessage(LogLevel.Error, Connection.GetWorkerId(), $"{message}\n{stackTrace}");
+                Worker.Connection?.SendLogMessage(LogLevel.Error, Worker.Connection.GetWorkerId(), $"{message}\n{stackTrace}");
```
jamiebrynes7: Can `Worker` be null here?

jessicafalk: if you create a dispatcher and use it directly instead of giving it to the worker: yes

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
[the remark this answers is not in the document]
jessicafalk: fair enough, let's use the type for now

zeroZshadow:
Implemented feedback and updated to latest.
Logging tests don't seem to be working?

workers/unity/Packages/com.improbable.gdk.core/Logging/ForwardingDispatcher.cs
samiwh: WorkerSystem is not the helper. Worker is a set of helper methods for creating a worker. We just haven't got round to making the ecs based worker bootstrap.

samiwh: If we make it worker now we will just have to change it later.

jamiebrynes7: The class definition of `Worker` indicates it has the data definition of a `Worker` though - I don't see why this would be `WorkerSystem` as you'd just end up doing: `WorkerSystem.Worker`

jessicafalk: right now the data definition is in both `WorkerSystem` and `Worker`, which makes the overall definition of `WorkerSystem` and `Worker` more confusing. 
Even if you say that `Worker` is a set of helper methods, you could just as well use `Worker` instead of `WorkerSystem` everywhere outside of the ECS. 

For now, to keep it consistent with the MB workflow as well, I would use `WorkerSystem`, but we need to revisit on how we want to use the worker in an ECS flow vs a MB flow and remove any reference from `WorkerSystem` for anything that is not directly ECS related

PR #337, opened 2018-09-07
Rewrote parts of the heartbeat system to be more simplistic

zeroZshadow:
#### Description
Greatly simplified how heartbeats work.


callumlawson:
What was the actual issue with heartbeats that we are trying to address?
If this doesn't fix any known bug/correctness problem I am very wary of changing the implementation so close to the deadline!

zeroZshadow:
Its fine to put this on hold until we've tagged our release for the deadline

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
```diff
@@ -4,87 +4,78 @@
...
-                        if (EntityManager.HasComponent<HeartbeatData>(entity))
-                        {
-                            PostUpdateCommands.SetComponent(entity, failedHeartbeatsComponent);
-                        }
-                        else
-                        {
-                            PostUpdateCommands.AddComponent(entity, failedHeartbeatsComponent);
-                        }
+                        stillAlive = true;
+                        break;
```
jamiebrynes7: Is there a case where `break` drops information from a subsequent command responses?

zeroZshadow: Nop. Most of the time this system will run with just 1 response per entity, because it only sends them every X seconds. In the weird case more then 1 comes in, we only care about if all of the failed, to avoid false positives.

zeroZshadow:
It never waited X seconds after the response arrived. It send in intervals of X, but skipped entities that were still waiting on the previous heartbeat. This only happens if the interval is smaller then the command timeout, but then I see no harm in sending more.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
```diff
@@ -13,11 +13,10 @@ public class SendPlayerHeartbeatRequestSystem : ComponentSystem
         private struct Data
         {
             public readonly int Length;
-            [ReadOnly] public ComponentDataArray<PlayerHeartbeatClient.CommandSenders.PlayerHeartbeat> RequestSenders;
```
jessicafalk: why did you remove the `ReadOnly` tag?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
```diff
@@ -4,87 +4,78 @@
...
-                        if (EntityManager.HasComponent<HeartbeatData>(entity))
-                        {
-                            PostUpdateCommands.RemoveComponent<HeartbeatData>(entity);
-                        }
+                        var entityDeleteSender = heartbeatGroup.WorldCommandSenders[i];
+                        entityDeleteSender.RequestsToSend.Add(WorldCommands.DeleteEntity.CreateRequest
+                        (
+                            heartbeatGroup.SpatialEntityIds[i].EntityId
+                        ));
+                        heartbeatGroup.WorldCommandSenders[i] = entityDeleteSender;
```
jessicafalk: are line s68 and 73  needed? we only use `entityDeleteSender` once, so we could just use it directly instead of assigning it and then re-assigning it back to `heartbeatGroup.WorldCommandSenders[i]`

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
zeroZshadow: Because we alter the list inside this (even though under the hood this does not change the component)

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
zeroZshadow: Because assigning it directly does not write back to the ComponentDataArray

jessicafalk: but do we need to assign it back when adding a request to the list?

zeroZshadow: no, because the list Doesn't care. But this will allow us to use the change filters.

samiwh:
Haven't read through this yet but I wrote a ton of comments on the original one. Would be good to check that this doesn't reintroduce race conditions.

zeroZshadow:
rebased onto latest master, will now rewrite to use queries

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatRequestSystem.cs
```diff
@@ -7,34 +8,53 @@
...
+            group = GetComponentGroup(query);
         }
 
-        [Inject] private HeartbeatData heartbeatData;
-
         protected override void OnUpdate()
         {
-            for (var i = 0; i < heartbeatData.Length; i++)
+            var requestsType = GetArchetypeChunkComponentType<PlayerHeartbeatClient.CommandRequests.PlayerHeartbeat>(true);
+            var responderType = GetArchetypeChunkComponentType<PlayerHeartbeatClient.CommandResponders.PlayerHeartbeat>(true);
```
jamiebrynes7: This one isn't really `readonly`

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
```diff
@@ -23,11 +23,10 @@ public class SendPlayerHeartbeatRequestSystem : ComponentSystem
...
+            public ComponentDataArray<PlayerHeartbeatClient.CommandSenders.PlayerHeartbeat> RequestSenders;
```
jamiebrynes7: Missed a `[Inject]` API 😛

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
```diff
@@ -1,93 +1,93 @@
...
+                var deleteRequesters = chunk.GetNativeArray(entityDeleterType);
+                var spatialIds = chunk.GetNativeArray(spatialIdType);
 
-                foreach (var response in responses)
+                for (int i = 0; i < responses.Length; i++)
                 {
-                    if (response.StatusCode != StatusCode.Success)
+                    var heartbeatData = heartbeats[i];
+
+                    var stillAlive = false;
```
samiwh: I don't think this is the right name as it can still be alive if it didn't respond in time a couple of times.

samiwh: `responded` would work.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
```diff
@@ -1,93 +1,93 @@
...
                             ));
-                            data.WorldCommandSenders[i] = entityDeleteSender;
-                        }
-                    }
-                    else
-                    {
-                        if (EntityManager.HasComponent<HeartbeatData>(entity))
-                        {
-                            PostUpdateCommands.RemoveComponent<HeartbeatData>(entity);
+                            deleteRequesters[i] = entityDeleteSender;
```
samiwh: We don't need to write this back

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
```diff
@@ -1,93 +1,93 @@
...
+                    }
+                    else
+                    {
+                        heartbeatData.NumFailedHeartbeats += 1;
 
-                        if (failedHeartbeatsComponent.NumFailedHeartbeats >=
-                            PlayerLifecycleConfig.MaxNumFailedPlayerHeartbeats)
+                        if (heartbeatData.NumFailedHeartbeats >= PlayerLifecycleConfig.MaxNumFailedPlayerHeartbeats)
                         {
-                            Debug.LogFormat(Messages.DeletingPlayer, entityId);
```
samiwh: I think this should still log something.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
```diff
@@ -23,11 +23,10 @@ public class SendPlayerHeartbeatRequestSystem : ComponentSystem
...
             [ReadOnly] public ComponentDataArray<Authoritative<PlayerHeartbeatServer.Component>> AuthorityMarkers;
+            [ReadOnly] public ComponentDataArray<HeartbeatData> HasHeartbeatData;
             [ReadOnly] public ComponentDataArray<SpatialEntityId> SpatialEntityIds;
-            public EntityArray Entities;
-            public SubtractiveComponent<AwaitingHeartbeatResponseTag> NotAwaitingHeartbeatResponse;
```
samiwh: Doesn't removing this mean that it will send commands regardless of whether or not it has a response. Also that you could send two messages before receiving either response. Which in turn means you could miss the first heartbeat, but respond to the second, but end up still registering as missing a heartbeat.

samiwh: This is what adding extra components was meant to help. And generally how I would use an ECS in this case. If we want to avoid doing that right now we could add a bit to the heartbeat data.

zeroZshadow: I'm not sure what you mean with this? The change here means that the heartbeat will be send at a set interval, not caring about if it was responded to or not.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
zeroZshadow: Oops left over from ComponentDataArray, fixed

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
zeroZshadow: 👍

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
zeroZshadow: will do this one in the next commit after PR feedback commit

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatRequestSystem.cs
zeroZshadow: copy pasta error, fixed

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
samiwh: Which is the race that the awaiting response tag was meant to solve

samiwh: As we also aren't cross checking response orders against send orders.
We could also solve this race by having a permanent component store the data in state instead.
Then early out if we are waiting for a response.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
```diff
@@ -4,87 +4,78 @@
...
             }
         }
-
-        internal static class Messages
-        {
-            public const string FailedHeartbeat =
-                "Client of player entity {0} failed to respond to heartbeat. Number of consecutive failures: {1}/{2}.";
-
-            public const string DeletingPlayer =
-                "Client of player entity {0} failed to respond to heartbeat and exhausted all retries. Deleting player.";
```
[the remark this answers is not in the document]
samiwh: +1. This feature module should log what it is doing. Otherwise it will confuse new users. Eventually we can have an option to set a log level per feature module.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
zeroZshadow: I dont see how this is a race?

samiwh: I described the race in the first comment.
Server sends ping A
Server sends ping B
Client sends pong B
Client sends pong A too late (which arrived later)
Server receives pong B
Server receives pong B failure - updates heartbeats failed to 1

samiwh: Can also check against IDs. This is going to happen when overloaded although probably not much during normal operation (unless timeouts are too low). We could also not deal with this for now and fix it later but be aware this is semantically changing something that was done deliberately.
Arguably kicking clients more often when overloaded is fine, although that behaviour happens anyway. This just adds some oddity to it.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
```diff
@@ -20,19 +21,28 @@ namespace Improbable.Gdk.PlayerLifecycle
     [UpdateBefore(typeof(SpatialOSUpdateGroup))]
     public class SendPlayerHeartbeatRequestSystem : ComponentSystem
     {
-        private struct Data
+        private float timeOfNextHeartbeat = Time.time + PlayerLifecycleConfig.PlayerHeartbeatIntervalSeconds;
```
samiwh: Don't use Time.time it loses accuracy quite quickly. But it doesn't look like we are using this anywhere.

samiwh: Ah found it's use. Counting down might be slightly better until the timer module gets more love.

PR #342, opened 2018-09-10
UTY-1176: Code Generation Workflow Tweaks

jamiebrynes7:
#### Description
Added a marker file to `Temp` to indicate auto codegen has run already. This should stop the constant popups every recompile.

This gets removed when Unity closes so it will always generate when first started.
#### Tests
Tested workflow locally - worked nicely.
#### Documentation
N/A 
#### Primary reviewers
@jared-improbable

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
@@ -20,18 +20,31 @@ namespace Improbable.Gdk.Tools
         private static readonly string SchemaCompilerRelativePath =
             $"../build/CoreSdk/{Common.CoreSdkVersion}/schema_compiler/schema_compiler";
 
+        private static readonly string StartupCodegenMarkerFile =
+            Path.GetFullPath("Temp/ImprobableCodegen.marker");
+
         /// <summary>
         ///     Ensure that code is generated on editor startup.
         /// </summary>
         static GenerateCode()
         {
-            if (EditorApplication.isPlayingOrWillChangePlaymode)
+            if (!CanGenerateOnLoad())
             {
-                // Don't generate code when entering PlayMode.
                 return;
             }
 
             EditorApplication.delayCall += Generate;
+        }
+
+        private static bool CanGenerateOnLoad()
+        {
+            if (EditorApplication.isPlayingOrWillChangePlaymode)
+            {
+                // Don't generate code when entering PlayMode.
+                return false;
+            }
+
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -20,20 +20,33 @@ internal static class GenerateCode
         private static readonly string SchemaCompilerRelativePath =
             $"../build/CoreSdk/{Common.CoreSdkVersion}/schema_compiler/schema_compiler";
 
+        private static readonly string StartupCodegenMarkerFile =
+            Path.GetFullPath("Temp/ImprobableCodegen.marker");
```
firtina-improbable: `/`, may be better to use `Path.Join`

firtina-improbable: I think it's safe to use `"Temp"` here but may be even safer to do e.g. `Path.Join(Application.dataPath, "..", "Temp", "filename.ext")`

callumlawson: if this is the right folder, we should use this:
https://docs.unity3d.com/ScriptReference/Application-temporaryCachePath.html

firtina-improbable: This one could potentially clash with other projects:

`%USERPROFILE%/AppData/Local/Temp/{company name}/{project name}`

Where {project name} is likely `unity` and {company name} is likely `DefaultCompany`

zeroZshadow: That shouldn't be the windows temp, but the project temp

jamiebrynes7: There's an easy way to test this 😉

jamiebrynes7: It is indeed - `%USERPROFILE%/AppData/Local/Temp/{company name}/{project name}`

jamiebrynes7: It also doesn't get cleaned when Unity closes or opens. Am going to revert back to the original

firtina-improbable: re: the original: reminder about my comments about Path.Join

PR #352, opened 2018-09-10
UTY-1125: Renamed title from Logging and accessing the worker to SpatialOSComponent utility API

JonasImprobable:

-------

#### Description
Renamed the title to avoid confusions as discussed with @martinzlocha 
The previous title suggested that this article might be about logging which is explained properly in a different article. This article should just link to the proper logging article.
#### Tests
#### Documentation
#### Primary reviewers


docs/README.md
```diff
@@ -50,7 +50,7 @@ Once you are comfortable with the GDK and want to start a separate project, look
 * [Sending and receiving events](content/gameobject/sending-receiving-commands.md)
 * [Sending and receiving commands](content/gameobject/sending-receiving-events.md)
 * [World commands](content/gameobject/world-commands.md)
-* [Logging and accessing the worker](content/gameobject/logging-and-accessing-the-worker.md)
+* [SpatialOSComponent utility API](content/gameobject/spatialoscomponent-utility-api.md)
```
callumlawson: I'd say this is a worse title as I, as a user, have no idea what the SpatialOSComponent utility API could do for me. 
Logging and accessing the worker is helpful as I know why I might want to read the linked doc.

callumlawson: I see the point about logging, but you should find a name more in line with "Sending and reciving events" centric to the task the user will have in mind.

MPeti01: The main problem is with the design of the component itself in the first place imo. It's just a place where a bunch of unrelated things ended up, the previous title was not really describing the contents. For now I think it's OK like this and when we hopefully break it up we can make the docs make more sense too.

JonasImprobable: Agreed. Once we can properly inject your entityid or the logger, this component will have a clearer role or will be removed altogether. The current state is more of a stop gap.

JonasImprobable: Renamed to `Retrieving your EntityId and accessing the worker` for now.

PR #359, opened 2018-09-10
Prevent invalid game objects from breaking injection

firtina-improbable:

-------

#### Description
Unity makes invalid components `null` but somehow still lets them be returned by component list queries.

When we're iterating over these any method calls throw exceptions and prevent further iterations.

#### Tests
Tested with a project that has bad gameobjects, it does not break any more, other game objects with monobehaviours still work well

#### Documentation
Internal


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
```diff
@@ -3,6 +3,7 @@
 using Improbable.Worker;
 using Unity.Entities;
 using UnityEngine;
+using Object = System.Object;
```
samiwh: why not just use `object`

firtina-improbable: Ah, this using is no longer necessary, I was using `Object.ReferenceEquals` but I converted them to `ReferenceEquals` instead

PR #362, opened 2018-09-11
Instantiating a prefab with broken/missing components fixed

zeroZshadow:
Instantiating a prefab with broken/missing components will no longer break the GDK


zeroZshadow:
Seems we already have this and it was just... sitting there https://github.com/spatialos/UnityGDK/pull/359

PR #365, opened 2018-09-11
Some code hygiene

jared-improbable:


workers/unity/Assets/Playground/Scripts/MonoBehaviours/UpdateSpinnerColor.cs
```diff
@@ -4,22 +4,29 @@
...
+#pragma warning disable 649
+
+#endregion
+
 namespace Playground.MonoBehaviours
 {
     public class UpdateSpinnerColor : MonoBehaviour
     {
-        [Require] private SpinnerColor.Requirable.Writer writer;
+        [Require] private SpinnerColor.Requirable.Writer spinnerColorWriter;
```
samiwh: I'm not sure I agree that we need to have longer names in unambiguous contexts. Although I don't really mind one way or the other.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -35,11 +35,11 @@ private void OnDestroy()
             Dispose();
         }
 
-        public async Task Connect(string workerType, ILogDispatcher logger)
+        protected async Task Connect(string workerType, ILogDispatcher logger)
```
samiwh: This is meant to be public. We want other things to be able to call connect on the derived class. We currently don't because we do things in start but that won't work in all cases.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -201,12 +201,17 @@ public void Dispose()
         {
             Worker?.Dispose();
             Worker = null;
-            // Check needed for the case that play mode is exited before the connection can complete
+            // Handle the case that play mode is exited before the connection can complete.
```
samiwh: This is actually the case where everything behaves well. The case described here is the do nothing one.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -201,12 +201,17 @@ public void Dispose()
...
             if (Application.isPlaying)
             {
                 Destroy(levelInstance);
                 Destroy(this);
             }
         }
+
+        private static string CreateNewWorkerId(string workerType)
```
samiwh: Can this go above the dispose function

workers/unity/Packages/com.improbable.gdk.core/Commands/WorldCommands.cs
```diff
@@ -14,34 +14,34 @@ internal static void AddWorldCommandRequesters(World world, EntityManager manage
         {
             var createEntitySender = new CreateEntity.CommandSender
             {
-                Handle = CreateEntity.RequestsProvider.Allocate(world)
+                Handle = CreateEntity.RequestsProvider.Allocate(world),
+                RequestsToSend = new List<CreateEntity.Request>()
```
samiwh: Is this right? The handle needs to point to this list so I'm sure this is correct.

workers/unity/Packages/com.improbable.gdk.core/Logging/LoggingUtils.cs
```diff
@@ -1,28 +1,33 @@
...
-                }
+                return null;
+            }
 
-                if (dataEntityId is long)
-                {
+            switch (dataEntityId)
+            {
+                case EntityId _:
+                    return (EntityId) dataEntityId;
```
samiwh: can avoid the second cast by giving a name instead of the `_`. Not sure if that's better though.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/UpdateSpinnerColor.cs
jared-improbable: I think we need to set a good example in our code, and `writer` isn't doing that, IMO.

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -22,13 +22,7 @@ public static T GetCommandLineValue<T>(IList<string> arguments, string configKey
...
         }
 
-        public static bool TryGetCommandLineValue<T>(IList<string> arguments, string configKey, out T configValue)
-        {
-            var dict = ParseCommandLineArgs(arguments);
-            return TryGetConfigValue(dict, configKey, out configValue);
-        }
-
-        public static bool TryGetConfigValue<T>(Dictionary<string, string> dictionary, string configName, out T value)
+        private static bool TryGetConfigValue<T>(Dictionary<string, string> dictionary, string configName, out T value)
```
samiwh: Nice. Been meaning to clean this up for ages.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/TakeRawTransformUpdateSystem.cs
```diff
@@ -24,7 +24,7 @@ private struct Data
 
         protected override void OnUpdate()
         {
-            for (int i = 0; i < data.Length; ++i)
+            for (var i = 0; i < data.Length; ++i)
```
samiwh: I don't actually like var in the transform stuff as a lot of these are over uints rather than int. Although I don't care that much.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
samiwh: Actually we haven't made this virtual yet. And this meets are use cases for now.
In that case I retract the above. Unless we want to rejig all of this protected is fine.

workers/unity/Packages/com.improbable.gdk.core/Commands/WorldCommands.cs
jared-improbable: I don't understand - I can't see how `Handle` is related to `RequestsToSend` et al, at all.
If they are somehow, we need to fix our API, because that is guaranteed to be misused.

workers/unity/Packages/com.improbable.gdk.core/Logging/LoggingUtils.cs
jared-improbable: Oh duh, of course.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/TakeRawTransformUpdateSystem.cs
jared-improbable: I don't really care much either, though our current code settings do prefer it

workers/unity/Packages/com.improbable.gdk.core/Commands/WorldCommands.cs
jamiebrynes7: This is a workaround for blittable (and lack there of). To enable any lists in an `IComponentData`, we have a handle that points to it somewhere else. `RequestToSend` is a property that fetches that list using the Handle.

This API needs some tweaking for safety - but something similar to it is unavoidable until we have a replacement for lists and maps and strings.

jared-improbable: Wow, that's really fragile.
Let me see what I can do.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/RequiredFieldInjector.cs
```diff
@@ -89,7 +94,7 @@ public void DisposeAllRequiredFields(MonoBehaviour behaviour)
                             .WithField("Field", field.Name));
                     }
 
-                    requirableToBeDisposed.Dispose();
+                    requirableToBeDisposed?.Dispose();
```
firtina-improbable: I think in this case it may be better if the `== null` branch causes a `continue` instead

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -142,19 +142,19 @@ protected virtual LocatorConfig GetLocatorConfig(string workerType)
...
+                        new LogEvent($"Failed to create worker")
                             .WithField("WorkerType", workerType)
                             .WithField("Message", e.Message));
                     attempts--;
                 }
             }
 
             throw new ConnectionFailedException(
-                $"Exceeded maximum connection attempts ({attempts})",
+                $"Tried to connect {attempts} times - giving up.",
```
firtina-improbable: Driveby: What will `attempts` be in this case? It feels like it will be 0.

This method should not be modifying the input parameter - can/should we lint for this (when we reenable linting...)? It's usually a code smell.

I think what'd be best is if it:
- rename the input parameter to `maximumAttempts` (or similar)
- copied the input parameter to be `remainingAttempts`

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -142,19 +142,19 @@ protected virtual LocatorConfig GetLocatorConfig(string workerType)
                 catch (ConnectionFailedException e)
                 {
                     logger.HandleLog(LogType.Error,
-                        new LogEvent($"Failed attempt to create worker")
+                        new LogEvent($"Failed to create worker")
```
firtina-improbable: Still feels like "an attempt" failed. 

Spinoffs below:

Should that become a warning instead of an error?

Should it print out how many attempts are left / which attempt it was?

If there are 3 failed attempts the user will see:

- "{error} Failed to create worker"
- "{error} Failed to create worker"
- "{error} Failed to create worker"
- "{exception} Tried to connect 3 times - giving up"

Doesn't look too bad actually, but still something to consider.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
jared-improbable: Ah yes, I missed we were modifying the original input variable. Will clean up...

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
samiwh: The main reason I had attempt in here before is because the Task<Worker> hasn't failed. It just hasn't finished yet. The latter message that said it failed to create a worker meant that the potential worker really was dead and should be removed. I agree that a warning might be better here, which would itself make it clear that it might still succeed.

workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
samiwh: Actually thinking about this again, does it make any sense for this to be protected and the class not be abstract. This makes it mostly useless if you don't inherit from it.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnectorBase.cs
```diff
@@ -146,19 +148,19 @@ protected virtual void HandleWorkerConnectionFailure()
...
             }
 
             throw new ConnectionFailedException(
-                $"Exceeded maximum connection attempts ({attempts})",
+                $"Tried to connect {maxAttempts} times - giving up.",
                 ConnectionErrorReason.ExceededMaximumRetries);
         }
 
-        private bool ShouldUseLocator()
+        private static bool ShouldUseLocator()
```
samiwh: We will likely have to make this virtual in the near future. Not really a big deal and we can do that when we need to though.

PR #371, opened 2018-09-12
UTY-1160: IEntityGameObjectCreator.OnEntityAdded now returns a bool

JonasImprobable:

-------

#### Description
Describe your changes here.
#### Tests
#### Documentation
#### Primary reviewers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.gameobjectcreation/IEntityGameObjectCreator.cs
+++ b/workers/unity/Packages/com.improbable.gdk.gameobjectcreation/IEntityGameObjectCreator.cs
@@ -1,28 +1,32 @@
-using Improbable.Gdk.Core;
 using Improbable.Worker;
 using UnityEngine;
 
 namespace Improbable.Gdk.GameObjectCreation
 {
     /// <summary>
-    ///     Interface for listening for SpatialOS Entity creation to be used for binding GameObjects.
+    ///     Interface for listening for SpatialOS Entity check-out and check-in events for defining custom companion GameObjects creation/deletion logic.
     ///     Implementing classes can be passed to GameObjectCreationSystemHelper in order to be called.
     /// </summary>
     public interface IEntityGameObjectCreator
     {
         /// <summary>
         ///     Called when a new SpatialOS Entity is checked out by the worker.
+        ///     Implement this method to define custom logic for creating a GameObject to represent an entity.
         /// </summary>
+        /// <param name="gameObject">
+        ///     The GameObject created to represent the entity.
+        /// </param>
         /// <returns>
-        ///     A GameObject to be linked to the entity, or null if no GameObject should be linked.
+        ///     true, if a GameObject should be created to represent the entity and if said GameObject was successfully created.
         /// </returns>
-        GameObject OnEntityCreated(SpatialOSEntity entity);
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/com.improbable.gdk.gameobjectcreation/IEntityGameObjectCreator.cs
```diff
@@ -1,28 +1,32 @@
...
         /// <summary>
         ///     Called when a new SpatialOS Entity is checked out by the worker.
+        ///     Implement this method to define custom logic for creating a GameObject to represent an entity.
         /// </summary>
+        /// <param name="gameObject">
+        ///     The GameObject created to represent the entity.
+        /// </param>
         /// <returns>
-        ///     A GameObject to be linked to the entity, or null if no GameObject should be linked.
+        ///     true, if a GameObject should be created to represent the entity and if said GameObject was successfully created.
```
jessicafalk: this seems like we are overloading it a bit. would it be better to have a separate method: `ShouldCreateGameObject` 

and `OnEntityAdded` only returns `true`, if a `GameObject` was created

firtina-improbable:
What functionality does this enable?

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/IEntityGameObjectCreator.cs
JonasImprobable: I initially implemented a `ShouldCreateGameObject` method but the code felt awkward. It felt like you forcefully separated code that usually would be together into two locations and sometimes need to duplicate code. Example:
You want to implement an `IEntityGameObjectCreator` which can differentiate between two types of creatures, foxes and rabbits. In that case a hypothetical `ShouldCreateGameObject` would need to check whether a newly added entity is either a fox or a rabbit based on its attached components. The `OnEntityAdded` method would then again independently check whether the entity needs a fox gameobject or a rabbit gameobject. This might however also be a symptom of something different being wrong with the API (e.g. what if we could register two Entitycreators, one for foxes and one for rabbits). Happy to discuss this in person.

JonasImprobable: Even that point aside,it felt weird to split "does my entity have certain components" checks and "get something from an entity without safety checks just because the ShouldCreateGameObject passed" logic.

JonasImprobable:
After looking over this one more time, I'm not sure if my proposed API change is really an improvement to the API. Will park this PR for now and potentially close it. The problem is that I'm not happy with the alternative of providing a `ShouldCreateGameObject` method either. Something just feels fundamentally off with the current state of the API. Happy to discuss this in person.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/IEntityGameObjectCreator.cs
JonasImprobable: Looking over the changes, I have to admit though that my proposed solution isn't great either. I agree with you on the overload point.

shir0i84:
This PR is to address UTY-1160, and has been addressed offline.

PR #376, opened 2018-09-13
[WIP]Feature/uty 1080 fix mobile connection

gkassabli:

-------

#### Description
This PR adds changes to make mobile versions work again after latest master changes
#### Tests
The feature is only tested on iOS.
Run local deployment -> Build for iOS in Unity -> Run iOS build on simulator -> ensure the connection is established as expected
#### Documentation
Will be documented later with the whole process
#### Primary reviewers
@jessicafalk

### DIFF
--- a/workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
+++ b/workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
@@ -2,6 +2,9 @@ using System;
 using System.Collections;
 using System.Linq;
 using System.Threading.Tasks;
+#if UNITY_ANDROID
+using Improbable.Gdk.Android;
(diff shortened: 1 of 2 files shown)


workers/unity/Assets/Playground/Scripts/Worker/WorkerConnectorBase.cs
```diff
@@ -123,11 +126,19 @@ protected virtual ReceptionistConfig GetReceptionistConfig(string workerType)
             }
             else
             {
+#if UNITY_ANDROID
```
samiwh: The function that gets the config is virtual to deal with cases like this. Please override rather than add conditional compilation to the base class.

jessicafalk:
closing this as obsolete

PR #330, opened 2018-09-06
UTY-1098: Unit Tests for Receive System

jamiebrynes7:
#### Description
Added a ton of tests for the `SpatialOSReceiveSystem`. Also added some utility for testing:
* `WorkerOpFactory` and `WrappedOp<T>` for safely creating and using worker ops during testing
* `TestLogDispatcher` for a finer grained assertion experience vs `LogAssert` 

Was going to put things in commit order but the rebase was being lame. 
#### Tests
Lotta tests here
#### Documentation
N/A
#### Primary reviewers
If your change will take a long time to review, you can name at most two primary reviewers who are ultimately responsible for reviewing this request. @ mention them.
@firtina-improbable @samiwh

PR #384, opened 2018-09-15
Update Generated Code Logs and Exceptions

jamiebrynes7:
#### Description
A couple changes:
* To maintain consistency with the changes in the `SpatialOSReceiveSystem` made in #330 - receiving an OP and not having the EntityId in the Worker throws instead of logs.
* A command response that is received after the entity has migrated/was deleted now just logs instead of errors (as this is a very much expected state).
* Driveby - make attribute for testing internal instead of public.

**Last commit is generated code**
#### Tests
Ran all tests + ran playground.
#### Documentation
N/A
#### Primary reviewers
@jessicafalk

### DIFF
--- a/test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingularTranslation.cs
+++ b/test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingularTranslation.cs
@@ -41,10 +41,7 @@ namespace Improbable.Gdk.Tests
 
             public override void OnAddComponent(AddComponentOp op)
             {
-                if (!IsValidEntityId(op.EntityId, "AddComponentOp", out var entity))
-                {
-                    return;
-                }
+                var entity = TryGetEntityFromEntityId(op.EntityId);
 
                 var data = Improbable.Gdk.Tests.ExhaustiveBlittableSingular.Serialization.Deserialize(op.Data.SchemaData.Value.GetFields(), World);
                 data.DirtyBit = false;
@@ -103,10 +100,7 @@ namespace Improbable.Gdk.Tests
 
             public override void OnRemoveComponent(RemoveComponentOp op)
             {
-                if (!IsValidEntityId(op.EntityId, "RemoveComponentOp", out var entity))
-                {
-                    return;
-                }
+                var entity = TryGetEntityFromEntityId(op.EntityId);
 
                 entityManager.RemoveComponent<Improbable.Gdk.Tests.ExhaustiveBlittableSingular.Component>(entity);
 
@@ -130,10 +124,7 @@ namespace Improbable.Gdk.Tests
 
             public override void OnComponentUpdate(ComponentUpdateOp op)
             {
-                if (!IsValidEntityId(op.EntityId, "OnComponentUpdate", out var entity))
-                {
-                    return;
-                }
+                var entity = TryGetEntityFromEntityId(op.EntityId);
 
                 if (entityManager.HasComponent<NotAuthoritative<Improbable.Gdk.Tests.ExhaustiveBlittableSingular.Component>>(entity))
                 {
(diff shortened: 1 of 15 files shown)


workers/unity/Packages/com.improbable.gdk.core/CodegenAdapters/ComponentDispatcherHandler.cs
```diff
@@ -41,11 +42,30 @@ protected ComponentDispatcherHandler(WorkerSystem worker, World world)
             Worker = worker;
         }
 
+        /// <summary>
+        ///     A helper method that wraps TryGetEntity on the Worker. This throws if the entity does not exist.
+        ///     The reason is throws is that this method is called on receiving ops and not having that Entity Id is
```
jessicafalk: `the reason is throws is`-> `the reason it throws is`

test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveMapValueTranslation.cs
```diff
@@ -206,21 +197,12 @@ public override void OnComponentUpdate(ComponentUpdateOp op)
...
-                {
-                    return;
-                }
-
+                var entity = TryGetEntityFromEntityId(op.EntityId);
                 ApplyAuthorityChange(entity, op.Authority, op.EntityId);
             }
 
             public override void OnCommandRequest(CommandRequestOp op)
             {
```
jessicafalk: line 228: (for some reason I can' t add a comment on that line) - do we want to change the LogError to an exception / log?

jamiebrynes7: Ah yes probably - as its a similar case to an `UnknownComponentId`

jessicafalk: lgtm :)

workers/unity/Packages/com.improbable.gdk.core/CodegenAdapters/ComponentDispatcherHandler.cs
samiwh: Also "*if* this method".
Can actually remove the first part of the sentence. "This throws if the entity does not exist; calling this method on receive ops ..."

jamiebrynes7: Sure will readjust the wording :)

workers/unity/Packages/com.improbable.gdk.core/CodegenAdapters/ComponentDispatcherHandler.cs
```diff
@@ -60,5 +59,19 @@ protected Entity TryGetEntityFromEntityId(EntityId entityId)
...
+        /// <summary>
+        ///     An exception for when the ComponentDispatcherHandler receive a CommandRequest or CommandResponse
+        ///     with an index that it does not know.
+        /// </summary>
+        protected class UnknownCommandIndexException : Exception
+        {
+            /// <param name="commandIndex">The unknown command index.</param>
+            /// <param name="componentName">The name of the component that the command was received on.</param>
+            public UnknownCommandIndexException(uint commandIndex, string componentName)
+                : base($"Unknown command index: {commandIndex} received for {componentName}")
```
samiwh: Is there any way a user can cause this that doesn't involve magic? If not then we might need to be clearer in the error message.

samiwh: Actually this index is generated by the connection, not by the bridge, so we know any commands we receive responses for must have been sent by this worker.

test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingularTranslation.cs
```diff
@@ -203,13 +179,7 @@ public override void OnCommandResponse(CommandResponseOp op)
...
                 {
                     default:
-                        LogDispatcher.HandleLog(LogType.Error, new LogEvent(CommandIndexNotFound)
-                            .WithField(LoggingUtils.LoggerName, LoggerName)
-                            .WithField(LoggingUtils.EntityId, op.EntityId.Id)
-                            .WithField("CommandIndex", commandIndex)
-                            .WithField("Component", "Generated.Improbable.Gdk.Tests.ExhaustiveBlittableSingular")
-                        );
-                        break;
+                        throw new UnknownCommandIndexException(commandIndex, "ExhaustiveBlittableSingular");
```
samiwh: why are we doing a switch and only catching the default case?

samiwh: Ah never mind

PR #385, opened 2018-09-15
UTY-1099: Add Unit Tests to SpatialOS Send System

jamiebrynes7:
#### Description
A few unit tests for the `SpatialOSSendSystem` - mostly focused around registering custom replicators.
#### Tests
These are tests
#### Documentation
N/A
#### Primary reviewers
@zeroZshadow

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Attributes/DisableAutoRegisterAttribute.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Attributes/DisableAutoRegisterAttribute.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    ///     This attribute denotes that a CodegenAdapter class should not be automatically registered with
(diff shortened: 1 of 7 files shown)


workers/unity/Packages/com.improbable.gdk.core/Attributes/DisableAutoRegisterAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    ///     This attribute denotes that a CodegenAdapter class should not be automatically registered with
+    ///     that which consumes them. The intended use for this attribute is testing.
+    /// </summary>
+    internal class DisableAutoRegisterAttribute : Attribute
```
jessicafalk: don't we already have this attribute?

jessicafalk: ah nvm, just saw that you moved it from one class to another

PR #390, opened 2018-09-17
small fix for UnlinkGameObjectFromEntity

jessicafalk:

-------

#### Description
we forgot to remove the `GameObjectReference` from the entity and to actually flush the `viewCommandBuffer` on `OnDestroyManager`.

#### Tests
tested it locally.

#### Documentation
none needed

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -107,6 +107,7 @@ protected override void OnDestroyManager()
                 gameObjectCreator.OnEntityRemoved(spatialEntityId, gameObject);
             }
 
+            viewCommandBuffer.FlushBuffer();
```
firtina-improbable: Is the flush buffer here necessary if there are objects to remove in the above loop? It feels not transparent at all that the `viewCommandBuffer` may or may not be dirtied depending on multiple conditions... The code around here smells pretty nasty... can we check if `entityToGameObjects` here is empty and put all of them here so it will be easier to understand?

jessicafalk:
so it did work before in the playground. however I did some changes in a private repo, where I realized that unlinking the gameobject from the entity actually didn't unlink everything correctly. i.e. unlinking the game object, but keeping the entity alive => the entity still had a `GameObjectReference`

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
jessicafalk: you mean inside the if-check? we could do it, but we do it that way everywhere else as well. Iterate through everything and then flush no matter whether we added anything to the buffer or not

firtina-improbable: I mean it feels odd / unusual to do this in OnDestroyManager and also that the System hands the command buffer to something else to populate.

If someone does a quick glance at this method; at first sight it feels like it's doing two different conceptually not necessarily related things: 1: iterating over some stuff, 2: flushing the view command buffer.

If we wrap all of this in `if(not empty){loop; flush;}` it will be faster to understand.

jessicafalk: sure, will wrap it in there

PR #391, opened 2018-09-17
UTY-1166: Menu item for additional Unity clients

Zomtorg:

-------

#### Description
Add a menu item in the SpatialOS dropdown for launching an additional local client. Beware - code duplication is involved, could be refactored if required.
#### Tests
Manual testing.
#### Documentation
None
#### Primary reviewers
@jessicafalk @jamiebrynes7


workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -29,7 +29,14 @@ private static void BuildConfigMenu()
         private static void LaunchMenu()
         {
             Debug.Log("Launching SpatialOS locally...");
-            EditorApplication.delayCall += Launch;
+            EditorApplication.delayCall += LaunchSpatialLocally;
+        }
+
+        [MenuItem("SpatialOS/Launch additional client")]
```
jamiebrynes7: Nit: `Launch standalone client` - the user may not have a client running yet

jamiebrynes7: Similar changes to logs/function names throughout

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -29,7 +29,14 @@ private static void BuildConfigMenu()
         private static void LaunchMenu()
         {
             Debug.Log("Launching SpatialOS locally...");
-            EditorApplication.delayCall += Launch;
+            EditorApplication.delayCall += LaunchSpatialLocally;
```
jamiebrynes7: Nit: If we are going to rename this - can we call it `LaunchLocalDeployment` ?

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -109,6 +169,25 @@ public static void Launch()
             };
         }
 
+        private static string GetCommand()
+        {
+            if (Application.platform == RuntimePlatform.OSXEditor)
+            {
+                return "osascript";
```
jessicafalk: is this needed?

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,19 +49,72 @@ public static void BuildConfig()
...
+            process.Exited += (sender, args) =>
+            {
+                // N.B. This callback is run on a different thread.
+                if (process.ExitCode == 0)
+                {
+                    return;
+                }
+
+                var logPath = Path.Combine(SpatialProjectRootDir, "logs");
+                var latestLogFile = Directory.GetFiles(logPath, "*unityclient.log")
```
jamiebrynes7: As far as I'm aware the log will always be called `external-default-unityclient.log`

jamiebrynes7: For a standalone client

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
jessicafalk: ah just saw that this was used before as well. as it works on mac, I guess we do need it

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,19 +49,72 @@ public static void BuildConfig()
...
+                var message = $"Unity Client local launch logfile: {latestLogFile.FullName}";
+
+                if (WasProcessKilled(process))
+                {
+                    Debug.Log(message);
+                }
+                else
+                {
+                    var content = File.ReadAllText(latestLogFile.FullName);
+                    message = $"{message}\n{content}";
```
jamiebrynes7: Don't rebind the variable - just create a new one 😄 

Makes the code easier to follow for someone else. You can also inline this in the `Debug` call.

`Debug.LogError($"{message}\n{content}");`

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
jessicafalk: if you specify it to be called that way in the json file

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: Was a leftover from the previous code - perhaps some golden variable for debugging.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: Definitely makes more sense as `standalone`.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
jamiebrynes7: Yes but a user could change it from `*unityclient.log` as well. This may want to read the worker json files

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,19 +49,70 @@ public static void BuildConfig()
             }
         }
 
-        public static void Launch()
+        public static void LaunchClient()
```
jessicafalk: after looking at the two methods it looks like the code is nearly completely the same for `LaunchClient` and `LaunchLocalDeployment`. Try to factor it out in a generic `Launch` method in which you can specify which spatial command you are running.
The only difference I see right now are the logfile name and the logs. The logs can be made a bit more generic, while the logfile name can be passed as a parameter

Zomtorg: Definitely, will do.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -99,16 +108,33 @@ public static void Launch()
...
-                    Debug.LogError(message);
+                    Debug.LogError($"{message}\n{content}");
                 }
 
                 process.Dispose();
                 process = null;
             };
         }
 
+        private static string GetCommand()
```
jessicafalk: these two methods can be inlined again.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,20 +49,22 @@ public static void BuildConfig()
             }
         }
 
-        public static void Launch()
+        public static void LaunchClient()
         {
-            BuildConfig();
+            Launch("Launch standalone client", "local worker launch UnityClient default", "spatial_*.log");
```
jessicafalk: think you got the log file names wrong. need to be the other way around `spatial_*.log` for launching a deployment and vice versa

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,20 +49,22 @@ public static void BuildConfig()
...
-                command = "osascript";
-                commandArgs = $"-e 'tell application \"Terminal\"\nactivate\ndo script \"cd {SpatialProjectRootDir} && {Common.SpatialBinary} local launch\"\nend tell'";
-            }
+        public static void LaunchLocalDeployment()
+        {
+            Launch("Launch SpatialOS locally", "local launch", "*unityclient.log");
+        }
 
-            var processInfo = new ProcessStartInfo(command, commandArgs)
+        public static void Launch(string commandName, string commandArgs, string logFileName)
```
jamiebrynes7: Instead of passing a string - what about defining an enum?

So the signature would look something like: 

`public static void Launch(string commandName, LaunchType launchType, string logFileName, param string[] additionalArgs)`

Then its invoked like: `Launch("Launch standalone client", LaunchType.CLIENT, "logFile", "UnityClient", "default")` ? 

You could go further and infer the `commandName` and `logFile` from the `enum` as well.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -99,16 +108,33 @@ public static void Launch()
...
+            }
+            return Common.SpatialBinary;
+        }
+
+        private static string GenerateCommandArgs(string command)
+        {
+            string generatedCommandArgs = command;
+            if (Application.platform == RuntimePlatform.OSXEditor)
+            {
+                generatedCommandArgs = $"-e 'tell application \"Terminal\"\nactivate\ndo script \"cd {SpatialProjectRootDir} && {Common.SpatialBinary} {command}\"\nend tell'";
```
firtina-improbable: Driveby opportunity 😈 : while we're touching this can we convert this to a verbatim interpolated string please? 

```cs
activate
do script ""cd {SpatialProjectRootDir} && {Common.SpatialBinary} local launch""
end tell'";
```

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: Oops, fixed them 👍

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -29,7 +31,14 @@ private static void BuildConfigMenu()
         private static void LaunchMenu()
         {
             Debug.Log("Launching SpatialOS locally...");
-            EditorApplication.delayCall += Launch;
+            EditorApplication.delayCall += LaunchLocalDeployment;
+        }
+
+        [MenuItem("SpatialOS/Launch standalone client")]
+        private static void LaunchAdditionalClientMenu()
```
samiwh: Name this according to the menu item

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -29,7 +31,14 @@ private static void BuildConfigMenu()
...
+        {
+            Debug.Log("Launching a standalone client locally...");
```
samiwh: Remove the ...

samiwh: And the locally.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: Sure thing 🔫

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
samiwh: These definitely shouldn't be two functions; they both rely on specialised output of the other. This also needs a more descriptive name.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,20 +51,35 @@ public static void BuildConfig()
             }
         }
 
-        public static void Launch()
+        public static void LaunchClient()
+        {
+            GetClientLogFilename();
+            Launch("Launch standalone client", "local worker launch UnityClient default", "*unityclient.log");
```
samiwh: This is very hard coded to using `UnityClient` and makes an assumption that they have a default config? They all use the same log file too by the look of it.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,20 +51,35 @@ public static void BuildConfig()
...
+
+            var logConfigPath = Path.Combine(SpatialProjectRootDir, "workers", "unity", clientConfigFilename);
+            var configFileJson = File.ReadAllText(logConfigPath);
+            var dict = Json.Deserialize(configFileJson) as Dictionary<string, object>;
 
-            var processInfo = new ProcessStartInfo(command, commandArgs)
+            return "WIP";
+        }
+
+        public static void Launch(string commandName, string commandArgs, string logFileName)
```
samiwh: Passing strings around everywhere for launching different processes makes me feel uneasy. I'd much rather we just make these separate functions.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,20 +51,35 @@ public static void BuildConfig()
...
+        private static string GetClientLogFilename()
+        {
+            string clientConfigFilename = "spatialos.UnityClient.worker.json";
+
+            var logConfigPath = Path.Combine(SpatialProjectRootDir, "workers", "unity", clientConfigFilename);
+            var configFileJson = File.ReadAllText(logConfigPath);
+            var dict = Json.Deserialize(configFileJson) as Dictionary<string, object>;
 
-            var processInfo = new ProcessStartInfo(command, commandArgs)
+            return "WIP";
```
samiwh: If this is still wip put that in the PR name

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,20 +51,35 @@ public static void BuildConfig()
...
+        public static void LaunchClient()
+        {
+            GetClientLogFilename();
+            Launch("Launch standalone client", "local worker launch UnityClient default", "*unityclient.log");
+        }
+
+        public static void LaunchLocalDeployment()
         {
             BuildConfig();
+            Launch("Launch SpatialOS locally", "local launch", "spatial_*.log");
```
samiwh: Do we not already have tools for this?

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
samiwh: Probably fine for anything we do now. But something to note.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: Removed 🙏

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: Migrated to enums as recommended by @jamiebrynes7 and some dicts to mappings.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: No idea - where can I take a look at the tools?

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: Merged them now 👍 - still not sure if the names are descriptive though.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,17 +49,86 @@ public static void BuildConfig()
...
+            process.Exited += (sender, args) =>
+            {
+                // N.B. This callback is run on a different thread.
+                if (process.ExitCode == 0)
+                {
+                    return;
+                }
+
+                var logPath = Path.Combine(SpatialProjectRootDir, "logs");
+                var latestLogFile = Directory.GetFiles(logPath, "external-default-unityclient.log")
```
jessicafalk: is it possible to extract the logpath directly from the `spatialos.UnityClient.worker.json` file?

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,17 +54,156 @@ public static void BuildConfig()
...
+                }
+
+                process.Dispose();
+                process = null;
+            };
+        }
+
+        private static string GetClientLogFilename()
+        {
+
```
jessicafalk: remove new line

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,17 +54,156 @@ public static void BuildConfig()
...
+            var logConfigPath = Path.Combine(SpatialProjectRootDir, "workers", "unity");
+            var configFileJson = File.ReadAllText(Path.Combine(logConfigPath, ClientConfigFilename));
+            var dict = Json.Deserialize(configFileJson);
+            var defaultLogFileName = "unitygamelogic.log";
```
jessicafalk: this is the default name for the game logic worker. We want the default name for the client log file :)

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,17 +54,156 @@ public static void BuildConfig()
...
+            int argumentIdx = 0;
+            for (argumentIdx = 0; argumentIdx < arguments.Count; argumentIdx++)
+            {
+                // The next argument would be the name of the log file
+                if (arguments[argumentIdx].Equals("-logfile"))
+                {
+                    argumentIdx++;
+                    break;
+                }
+
```
jessicafalk: new line

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,17 +54,154 @@ public static void BuildConfig()
...
+            // Logger file not found - using default one
+            if (argumentIdx >= arguments.Count)
+            {
+                Debug.LogError($"Config file {ClientConfigFilename} doesn't contain '-logfile' argument within 'external' -> 'default' -> '{currentOS}' -> arguments.");
+                return defaultLogFileName;
+            }
+
+            // Strip any relative pathing
+            var logFileName = ((string)arguments[argumentIdx]).Substring(((string) arguments[argumentIdx]).LastIndexOf('/'));
```
jamiebrynes7: What happens here if the user gives a log file path like: 

`../../my-special-logs/client.log`? 

Won't this just return `client.log`?

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,17 +54,154 @@ public static void BuildConfig()
...
+            var configFileJson = File.ReadAllText(Path.Combine(logConfigPath, ClientConfigFilename));
+            var dict = Json.Deserialize(configFileJson);
+            var defaultLogFileName = "*unityclient.log";
+            Dictionary<string, object> tempDict;
+
+            var currentOS = "windows";
+            if (Application.platform == RuntimePlatform.OSXEditor)
+            {
+                currentOS = "macos";
+            }
```
jamiebrynes7: Use a ternary:

`var currentOs = Application.platform == RuntimePlatform.OSXEditor ? "macos" : "windows";`

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,17 +54,154 @@ public static void BuildConfig()
...
+                    break;
+                }
+            }
+
+            // Logger file not found - using default one
+            if (argumentIdx >= arguments.Count)
+            {
+                Debug.LogError($"Config file {ClientConfigFilename} doesn't contain '-logfile' argument within 'external' -> 'default' -> '{currentOS}' -> arguments.");
+                return defaultLogFileName;
+            }
```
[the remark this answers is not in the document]
firtina-improbable: Do you need the `.SkipWhile` if you have the `.FirstOrDefault`?

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
firtina-improbable: Can use `Path.GetFullPath` if appropriate

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,17 +54,154 @@ public static void BuildConfig()
...
+            };
+        }
+
+        private static string GetClientLogFilename()
+        {
+            var logConfigPath = Path.Combine(SpatialProjectRootDir, "workers", "unity");
+            var configFileJson = File.ReadAllText(Path.Combine(logConfigPath, ClientConfigFilename));
+            var dict = Json.Deserialize(configFileJson);
+            var defaultLogFileName = "*unityclient.log";
+            Dictionary<string, object> tempDict;
```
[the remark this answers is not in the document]
Zomtorg: It's instantiated in the middle of two if statements and in my opinion having it here is clearer from a readability standpoint.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: That's correct - I've based the assumption of retrieving the filename only rather than combining the paths based on the current way the spatial logs are operated on:

    var logPath = Path.Combine(SpatialProjectRootDir, "logs");
     var latestLogFile = Directory.GetFiles(logPath, "spatial_*.log")
                    .Select(f => new FileInfo(f))
                    .OrderBy(f => f.LastWriteTimeUtc).LastOrDefault();

jamiebrynes7: Nothing requires the user to place the log in `logs` however though - we should be able to resolve the path regardless of where it points (as its relative to the json file)

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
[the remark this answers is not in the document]
jamiebrynes7: It needs a more descriptive name then - as it gives little clue as to what this variable is for or why its needed

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -42,17 +54,154 @@ public static void BuildConfig()
...
+                return defaultLogFileName;
+            }
+
+            tempDict = defaultValue as Dictionary<string, object>;
+
+            if (!tempDict.TryGetValue(currentOS, out var currentOSValue))
+            {
+                Debug.LogError($"Config file {ClientConfigFilename} doesn't contain key '{currentOS}' within 'external' -> 'default'.");
+                return defaultLogFileName;
+            }
```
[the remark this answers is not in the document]
Zomtorg: Good call - added to the code.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
Zomtorg: Fixed now.

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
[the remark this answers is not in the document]
Zomtorg: Awesome, changed it.

jessicafalk:
overall looks good, but could we add a check whether the client was build out?
right now, if you don't have the client built and try to run it, you get the following  error:
`Could not find a standalone client log file C:\Users\jessica\Projects\improbable\UnityGDK\logs\external-default-unityclient.log.`, which is a symptom of not having it built out, but not the cause

it would also be nice, if we could move the two menu items into their own category, like we have done for example with the "Build Unityclient/UnityGamelogic ... " menu items

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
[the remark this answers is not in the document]
firtina-improbable: > 
> 
> Yes - otherwise you will return the first argument in the list that is not `-logfile`. So if the args were:
> 
> ```
>           "+workerType",
>           "UnityClient",
>           "+infraServicesUrl",
>           "http://127.0.0.1:21000",
>           "+projectName",
>           "${IMPROBABLE_PROJECT_NAME}",
>           "-logfile",
>           "../../logs/external-default-unityclient.log"
> ```
> 
> You will get `+workerType`

oooh... good one

Zomtorg:
Added the grouping items and the check if the Unity Client has been built

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -98,10 +251,7 @@ public static void Launch()
                 }
                 else
                 {
-                    var content = File.ReadAllText(latestLogFile.FullName);
-                    message = $"{message}\n{content}";
-
-                    Debug.LogError(message);
+                    Debug.LogError($"{message}");
```
firtina-improbable: Looks like string interpolation is no longer necessary :)

firtina-improbable: Or the `if/else` block itself

Zomtorg: Fixed it now so it makes more sense.

PR #392, opened 2018-09-17
Move the Configuring your project section higher.

martinzlocha:

-------

#### Description
Slightly reordered doc order after discussion with @ElleEss.
#### Tests
#### Documentation
#### Primary reviewers


MPeti01:
I'm pretty sure it was like that but then @callumlawson moved it down? It also kind of makes sense to have it low because it's advanced project setup stuff and the signposting directs you to it only after looking at the stuff above. I don't have a strong opinion though, feel free to merge if it's been well discussed

martinzlocha:
@MPeti01 Personally I think that the "Starting workers" part is very essential and you will have to look at that section before writing any code.

jamiebrynes7:
It depends what we expect consumers of these docs will more immediately want for M2.

Its likely `Starting workers` is very important, `Setting up a new project` less so

MPeti01:
@martinzlocha Tbh I'd kind of agree *if* that workers doc had more information about where you might want to edit the WorkerConnectors, ie AddWorkerSystems(), which it really should. What's in there right now is not that useful for someone starting out imo

jessicafalk:
it is important as you need to know how to actually add workers to the scene

however, I'm also in the process of adding more information about the connection flow

PR #396, opened 2018-09-18
UTY-1234: helper methods to determine if entity is on worker

paulbalaji:

-------

#### Description
Add IsEntityOnThisWorker helper method to SpatialOsComponent, decided not to include a version that takes an Entity, to avoid iterating through a whole Dictionary
#### Tests
none
#### Documentation
none
#### Primary reviewers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSComponent.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSComponent.cs
@@ -11,5 +11,10 @@ namespace Improbable.Gdk.GameObjectRepresentation
         public Entity Entity;
         public World World;
         public WorkerSystem Worker;
+
+        public bool IsEntityOnThisWorker(EntityId entityId)
+        {
+            return Worker.EntityIdToEntity.ContainsKey(entityId);
+        }
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSComponent.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSComponent.cs
@@ -16,10 +16,5 @@ namespace Improbable.Gdk.GameObjectRepresentation
         {
             return Worker.EntityIdToEntity.ContainsKey(entityId);
         }
-
-        public bool IsEntityOnThisWorker(Entity entity)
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/SpatialOSComponent.cs
```diff
@@ -11,5 +11,15 @@ public class SpatialOSComponent : MonoBehaviour
...
         public WorkerSystem Worker;
+
+        public bool IsEntityOnThisWorker(EntityId entityId)
+        {
+            return Worker.EntityIdToEntity.ContainsKey(entityId);
+        }
+
+        public bool IsEntityOnThisWorker(Entity entity)
+        {
+            return Worker.EntityIdToEntity.ContainsValue(entity);
```
jamiebrynes7: This will be super heavy

jamiebrynes7: I'm not sure the utility of an `Entity` overload for this anyway

jessicafalk: agreed, do we need this is or is the entity id version enough for now?

paulbalaji: +1 on it being heavy

I think EntityId version should be enough for now, but I left in the Entity version in case some user wants to use the heavy method - but do we want to let users even do that?

paulbalaji: I'll remove the heavy Entity version, to discourage users from doing bad things

jamiebrynes7: Yup - also not sure of the use case where you have the `Entity` but not the EntityId (or have a quick way of accessing it)

MPeti01:
Wouldn't you want to check for GameObjects rather than Entities? That's what the ticket says and would typically be more useful. Ie take a GameObject, dig through its SpatialOSComponent to find the EntityId and check that?

jamiebrynes7:
@MPeti01 This PR is orthogonal - there's another PR for that

PR #400, opened 2018-09-18
[WIP]Feature/uty 1063 shared common schema types

Zomtorg:

-------

#### Description
Create a schema of commonly used types and wrappers. Feel free to suggest other types.
#### Tests
Manually testing if code generation works - appears as a generated .cs file.
#### Documentation
Not documented.


jared-improbable:
Can you provide a strong justification for these pre-made types? Is it an inordinate burden for users to write a new type?

jessicafalk:
I think it makes sense for the empty type, but maybe less for the others? as we would lose the names in e.g. command responses / requests

workers/unity/Packages/com.improbable.gdk.core/Schema/common.schema
```diff
@@ -0,0 +1,40 @@
+package improbable.core;
+import "improbable/vector3.schema";
+
+type EmptyMessage {}
+
+type BoolMessage {
+    bool value = 1;
+}
+
+type IntMessage {
```
JonasImprobable: How about renaming this to `Int32Message`?

samiwh: There's actually 8 different types of int available. I'm not sure that it adds much to make each primitive type it's own message unless we are using them a lot.
Empty is the message that is really useful here.

Zomtorg:
I agree with @jessicafalk about the semantics of using types, which wrap simple values for the sake of readability, such as:
`type CreatePlayerRequestType
{
    improbable.Vector3f position = 1;
}`
I'm also not sure whether wrapping all different types would provide much value.

jessicafalk:
I would say: keep the `EmptyMessage` and ensure that we don't redefine it in any of the other schemas and remove the rest. If we discover that we do actually need them, we can still add them later

JonasImprobable:
Since we seem to have different opinions on this matter, how about we only provide the empty type for now?

Zomtorg:
Refactored the code to use the new Empty type.

PR #403, opened 2018-09-19
UTY-1102: Added way to link GameObject to the Worker entity

MPeti01:

-------

#### Description
Added an optional parameter when adding the GameObject creation feature module to have a GameObject representing the worker entity. This lets users implement per-worker singleton behaviours that can also [Require] things that don't need authority, i.e. command senders/receivers.

Also driveby fixed gameobject creation assembly name.
#### Tests
Unit tests included!
#### Documentation
Will be looking at where this best fits in the MonoBehaviour docs right away, this seems like a good point to review though.
#### Primary reviewers
@samiwh @JonasImprobable


workers/unity/Packages/com.improbable.gdk.gameobjectcreation/Tests/WorkerEntityLinkingTests.cs
```diff
@@ -0,0 +1,82 @@
+using System.Runtime.CompilerServices;
```
samiwh: What are we using this for?

MPeti01: Forgetting to clean up weird things added by the IDE while autocompleting as usual

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/WorkerEntityGameObjectLinkerSystem.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            for (int i = 0; i < connectedData.Length; i++)
+            {
+                linkerSystem.Linker.LinkGameObjectToEntity(gameObjectToLink, connectedData.Entities[i],
+                    new EntityId(0), viewCommandBuffer);
+            }
+
+            for (int i = 0; i < disconnectedData.Length; i++)
+            {
+                linkerSystem.Linker.UnlinkGameObjectFromEntity(gameObjectToLink, disconnectedData.Entities[i],
```
samiwh: Note for docs - we don't delete the gameobject. I think that is fine. But it should be noted and probably added to the connector stuff.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
```diff
@@ -22,7 +22,7 @@ namespace Improbable.Gdk.GameObjectRepresentation
     [DisableAutoCreation]
     [AlwaysUpdateSystem]
     [UpdateInGroup(typeof(SpatialOSReceiveGroup.GameObjectReceiveGroup))]
-    internal class GameObjectDispatcherSystem : ComponentSystem
+    public class GameObjectDispatcherSystem : ComponentSystem
```
samiwh: If this is only used in tests that I'd rather make it visible to the test dll than make it public.

samiwh: Alternatively we should use the helper methods to add these systems to test worlds.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/WorkerEntityGameObjectLinkerSystem.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        [Inject] private WorkerConnectedData connectedData;
+        [Inject] private WorkerDisconnectedData disconnectedData;
+
+        [Inject] private EntityGameObjectLinkerSystem linkerSystem;
+        [Inject] private WorkerSystem workerSystem;
+
+        private ViewCommandBuffer viewCommandBuffer;
+        private GameObject gameObjectToLink;
```
JonasImprobable: How about renaming this variable to `workerGameObject`?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/GameObjectDispatcherSystem.cs
MPeti01: Will do

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
```diff
@@ -1,20 +1,26 @@
...
 using Unity.Entities;
+using UnityEngine;
 
 namespace Improbable.Gdk.GameObjectCreation
 {
     public static class GameObjectCreationSystemHelper
     {
-        public static void EnableStandardGameObjectCreation(World world)
+        public static void EnableStandardGameObjectCreation(World world, IEntityGameObjectCreator creator = null,
+            GameObject workerGameObject = null)
```
JonasImprobable: It looks like users need to right now determine themselves, when to create/clean up their worker entity gameobject. Would it make more sense to extend the `IEntityGameObjectCreator` with OnConnected/OnDisconnected callbacks to give users a more consistent way to set up/clean up their worker entity gameobject?

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/WorkerEntityGameObjectLinkerSystem.cs
MPeti01: Yes I was considering which way is better and decided to go with this one in the end, since it's created by the user in the first place.

MPeti01: For the connectors, I didn't add it yet because I couldn't think of anything sensible to add to the playground and I'm confident it works because of the unit tests. Should I just add a blank GameObject to the workers in Playground, or should I try to come up with something good to add there?

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
MPeti01: I did think of that, but decided against it because 1) it feels weird to have callback-style methods that would be called just once over the lifetime of the object, 2) that would enforce that the users must override all methods even if they don't need, which is annoying boilerplate. Not 100% convinced, but this seemed like the better option.

MPeti01: An example in the WorkerConnectors would certainly help a bit with that

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/WorkerEntityGameObjectLinkerSystem.cs
samiwh: I think what you have here is fine. We should think about adding something in a follow up though. We could rewrite a player lifecycle in it if we wanted to.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
JonasImprobable: As long as everyone agrees that this is a stop gap, I'm fine with what this PR does for now. Longer term, I'd like to see the GameObjectCreationSystemHelper + IEntityGameObjectCreator replaced with something that looks more like a dispatcher.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
```diff
@@ -1,20 +1,26 @@
 using Improbable.Gdk.Core;
 using Unity.Entities;
+using UnityEngine;
 
 namespace Improbable.Gdk.GameObjectCreation
 {
     public static class GameObjectCreationSystemHelper
     {
-        public static void EnableStandardGameObjectCreation(World world)
+        public static void EnableStandardGameObjectCreation(World world, IEntityGameObjectCreator creator = null,
```
JonasImprobable: We still need to update the `WorkerUtils` file as part of this PR right? Otherwise we are not using any `IEntityGameObjectCreator` anymore by default.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
```diff
@@ -1,20 +1,26 @@
...
+                creator = new GameObjectCreatorFromMetadata(workerSystem.WorkerType, workerSystem.Origin,
+                    workerSystem.LogDispatcher);
+            }
 
-        public static void EnableStandardGameObjectCreation(World world, IEntityGameObjectCreator creator)
-        {
             world.CreateManager<GameObjectInitializationSystem>(creator);
+            if (workerGameObject != null)
+            {
+                world.CreateManager<WorkerEntityGameObjectLinkerSystem>(workerGameObject);
```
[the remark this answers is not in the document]
samiwh: World gives you an error. I can't remember what but it won't add two systems.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
samiwh: I think it adds one in the function but I agree this looks odd. Slightly better to keep this as two overloads I think.

MPeti01: If you read below, I'm using null as a special value to add the default parametrized by the stuff from the World. It's confusing, so I'll revert back to 2 methods instead

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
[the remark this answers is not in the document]
JonasImprobable: I guess what I tried to say is that it is not apparent right now when you should be allowed to call this method.
Call this method after you connected, and the worker gameobject is not set up correctly since you don't have the `OnConnected` tag anymore. Call this method a second time and it sometimes errors and sometimes doesn't depending on whether `workerGameObject` is null. It is also not apparent whether we allow the worker gameobject can be re-set to a different gameobject mid lifecycle.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
JonasImprobable: Ah sorry, didn't read until the end.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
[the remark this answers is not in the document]
JonasImprobable: I assume that right now, the idea is to always call this method at most once ever and do that before we connect. If that's the case, we should intercept the cases where the method is called when it shouldn't be called and log a descriptive error message.

PR #405, opened 2018-09-19
UTY-1215: entity builder helpers in feature modules

paulbalaji:

-------

#### Description
Added helper methods to add feature module-specific components to an entity builder - so users do not have to worry about what components each feature module may use. Will also check if the snapshot is valid and if playground isn't broken. Also combined SystemHelper classes with EntityBuilderHelper, unifying helper methods into one class per feature module. 
#### Tests
- check snapshot sanity
- check worker/client in playground
#### Documentation
none
#### Primary reviewers


workers/unity/Assets/Playground/Config/CubeTemplate.cs
```diff
@@ -29,13 +22,14 @@ public static Entity CreateCubeEntityTemplate(Coordinates coords)
...
                 .SetPersistence(true)
                 .SetReadAcl(WorkerUtils.AllWorkerAttributes)
-                .AddComponent(transform, WorkerUtils.UnityGameLogic)
                 .AddComponent(cubeColor, WorkerUtils.UnityGameLogic)
                 .AddComponent(cubeTargetVelocity, WorkerUtils.UnityGameLogic)
-                .AddComponent(launchable, WorkerUtils.UnityGameLogic)
-                .Build();
+                .AddComponent(launchable, WorkerUtils.UnityGameLogic);
+
+            TransformSynchronizationEntityBuilderHelper.AddComponents(entity, WorkerUtils.UnityGameLogic,
```
firtina-improbable: This class name feels... long.

/me braces for big discussion

How about `TransformSynchronizationModule.AddEntityComponents`?

firtina-improbable: Or extension methods? (😅)

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleEntityBuilderHelper.cs
```diff
@@ -0,0 +1,16 @@
+using Improbable.Gdk.Core;
+using Improbable.PlayerLifecycle;
+
+namespace Improbable.Gdk.PlayerLifecycle
+{
+    public class PlayerLifecycleEntityBuilderHelper
+    {
+        public static void AddComponents(EntityBuilder entityBuilder, string clientAccess, string serverAccess)
```
firtina-improbable: Should they return the `entityBuilder` they received? It fits with the builder pattern better.

e.g. 
```
return entityBuilder.AddComponent(xyz)
    .AddComponent(abc);
```

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -40,14 +36,14 @@ public static class PlayerTemplate
...
                 .AddComponent(launcher, WorkerUtils.UnityGameLogic)
-                .AddComponent(clientHeartbeat, clientAttribute)
-                .AddComponent(serverHeartbeat, WorkerUtils.UnityGameLogic)
                 .AddComponent(score, WorkerUtils.UnityGameLogic)
                 .AddComponent(cubeSpawner, WorkerUtils.UnityGameLogic);
 
+            TransformSynchronizationEntityBuilderHelper.AddComponents(entityBuilder, clientAttribute);
+            PlayerLifecycleEntityBuilderHelper.AddComponents(entityBuilder, clientAttribute, WorkerUtils.UnityGameLogic);
+
             return entityBuilder.Build();
```
firtina-improbable: If the `entityBuilder` is somehow made immutable (because builder patterns make that trivial) then the helper.AddComponents could potentially be no-op at this point. See my comment on them returning the entity builder.

That way:

```c#
entityBuilder = TransformSynchronizationEntityBuilderHelper.AddComponents(entityBuilder, clientAttribute);
entityBuilder = PlayerLifecycleEntityBuilderHelper.AddComponents(entityBuilder, clientAttribute, WorkerUtils.UnityGameLogic);

return entityBuilder.Build();
```

workers/unity/Assets/Playground/Config/CubeTemplate.cs
paulbalaji: I like this^

we could probably then merge {module}SystemHelper and {module}EntityBuilderHelper classes into the same class

workers/unity/Assets/Playground/Config/CubeTemplate.cs
```diff
@@ -29,13 +22,14 @@ public static Entity CreateCubeEntityTemplate(Coordinates coords)
...
                 .SetPersistence(true)
                 .SetReadAcl(WorkerUtils.AllWorkerAttributes)
-                .AddComponent(transform, WorkerUtils.UnityGameLogic)
                 .AddComponent(cubeColor, WorkerUtils.UnityGameLogic)
                 .AddComponent(cubeTargetVelocity, WorkerUtils.UnityGameLogic)
-                .AddComponent(launchable, WorkerUtils.UnityGameLogic)
-                .Build();
+                .AddComponent(launchable, WorkerUtils.UnityGameLogic);
+
+            entity = TransformSynchronizationEntityBuilderHelper.AddComponents(entity, WorkerUtils.UnityGameLogic,
```
firtina-improbable: nit: variable name `entity` => `entityBuilder`

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformSynchronizationEntityBuilderHelper.cs
```diff
@@ -0,0 +1,26 @@
+using Improbable.Gdk.Core;
+using Improbable.Transform;
+using Unity.Entities;
+
+namespace Improbable.Gdk.TransformSynchronization
+{
+    public static class TransformSynchronizationEntityBuilderHelper
```
firtina-improbable: **\*Performs comment necromancy\***

image

workers/unity/Assets/Playground/Config/CubeTemplate.cs
paulbalaji: I also considered this change, left out in case there was any particular reason the cube template used a different name to the player template

workers/unity/Assets/Playground/Config/CubeTemplate.cs
```diff
@@ -12,30 +13,23 @@ public static Entity CreateCubeEntityTemplate(Coordinates coords)
...
                 .SetPersistence(true)
                 .SetReadAcl(WorkerUtils.AllWorkerAttributes)
-                .AddComponent(transform, WorkerUtils.UnityGameLogic)
                 .AddComponent(cubeColor, WorkerUtils.UnityGameLogic)
                 .AddComponent(cubeTargetVelocity, WorkerUtils.UnityGameLogic)
-                .AddComponent(launchable, WorkerUtils.UnityGameLogic)
-                .Build();
+                .AddComponent(launchable, WorkerUtils.UnityGameLogic);
+
+            TransformSynchronizationHelper.AddComponents(ref entityBuilder, WorkerUtils.UnityGameLogic,
```
jamiebrynes7: Is there a way to do with without breaking the fluent API?

firtina-improbable: /whispers "extension methods"

jamiebrynes7: It was a leading question 😉

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformSynchronizationHelper.cs
```diff
@@ -1,9 +1,43 @@
+using Improbable.Gdk.Core;
+using Improbable.Transform;
 using Unity.Entities;
 
 namespace Improbable.Gdk.TransformSynchronization
 {
-    public static class TransformSynchronizationSystemHelper
+    public static class TransformSynchronizationHelper
     {
+        public static EntityBuilder AddTransformSynchronizationComponents(this EntityBuilder entityBuilder, string writeAccess,
```
jamiebrynes7: Instead of these overloads, can you just inline the default quaternion?

paulbalaji: I can not because the default Quaternion appears to be (0,0,0,0), not (1,0,0,0) - and can't do `new Quaternion(1f, 0f, 0f, 0f)` because default parameter needs to be a compile-time constant

jared-improbable: Use `Quaternion.identity` - also, the identity Quaternion is `0,0,0,1`...

jamiebrynes7:
@jared-improbable For Unity Quaternion it is as its defined as (x,y,z,w), but our Transform Quaternion is defined as (w,x,y,z) 😬

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformSynchronizationHelper.cs
paulbalaji: as Jamie said, our Quaternions are wxyz but Unity uses xyzw

though I could just make the function take in Unity Quaternions and abstract away our Quaternions in the AddComponent method

samiwh: The arguments to this should just use the default Unity types, Vector3 and Quaternion. We can convert those inside the feature modules using the extensions methods.

samiwh: The physics tick and ticks per second shouldn't be set as arguments either. They should be set to 0 and the 1/fixedDeltaTime.

JonasImprobable:
Please update `UnityGDK/docs/content/entity-templates.md` accordingly or create a ticket for tracking the update work. :)

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformSynchronizationHelper.cs
paulbalaji: discussed offline:

Quaternion.identity is not a compile-time constant :/ - so we need separate methods

docs/content/entity-templates.md
```diff
@@ -53,6 +53,56 @@ public static class CreatureTemplate
 
 You can use this entity template to spawn a Creature entity. This can be done using the [MonoBehaviour](gameobject/world-commands.md) or [ECS](ecs/world-commands.md) workflow.
 
+### Feature Module Components
+
+To take advantage of some feature modules, you need to add the components that the modules will make use of.  This is done by calling the appropriate helper method for each desired feature module.
```
JonasImprobable: Nit: double " " between the two sentences. Also, maybe change "add the components" to "add SpatialOS components".

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -22,29 +22,26 @@ public static class PlayerTemplate
                     $"Expected an attribute that is not \"{WorkerUtils.UnityClient}\" but none was found.");
             }
 
-            var transform =
-                TransformInternal.Component.CreateSchemaComponentData(new Location(),
-                    new Quaternion { W = 1, X = 0, Y = 0, Z = 0 }, new Velocity(0.0f, 0.0f, 0.0f), 0, 0.0f);
+            const string CharacterType = "Character";
```
jared-improbable: I don't see a point in having a string constant if it's just used once...

paulbalaji: thanks for the spot - seems to have carried forward from a merge, will remove

docs/content/entity-templates.md
```diff
@@ -53,6 +53,54 @@ public static class CreatureTemplate
...
+            .AddPosition(position.X, position.Y, position.Z, WorkerUtils.UnityGameLogic)
+            ...
+            ...
+            .AddPlayerLifecycleComponents(clientAttribute, WorkerUtils.UnityGameLogic);
+
+        return entityBuilder.Build();
+    }
+}
+```
+
```
ElleEss: The whole doc needs a once-over as it's got a lot of unclear passive voice and prophecies about the future (will) rather than statements of fact - I know you didn't write that, though.
There's quite a few bits and bobs in this diff, so I am going to go through the whole thing here - I am adding my edits to the text, below in here and added questions in "<Editor questions:....>".

Notes to explain the edits:
We use newspaper titles (sentences without full stops) for our title case.
"Feature Module(s)" and "Core Module" are proper nouns (blame Callum!).
We avoid "will" wherever possible because it reads like a prophecy for the future (and implies instability).
We avoid passive voice as it doesn't make it clear who does the action - you can tell passive voice by adding "by zombies" to a sentence. If the sentence makes sense with "by zombies" attached to it, it's in passive voice (IE "This is done by zombies calling the appropriate helper method".
We only use the bold (**) for titles, to call out "Note:" as well as for anything on a UI.

-----

### Feature Module components
 To take advantage of some Feature Modules, you need to add the SpatialOS components that these modules make use of. You do this by calling the appropriate helper method for each feature module. This document lists the helper methods and how to use them.

_<Editor questions: why "some" feature modules?>_
_<Editor questions: add the components to what or where?>_
_<Editor questions: calling the appropriate from where?>_
_<Editor questions: "This document lists the helper methods and how to use them." Is this right?>_

 #### Transform synchronization module
When you want to give a [worker](./worker.md) authority [authority](LINK!TBD) over the transform synchronisation of a SpatialOS component, you use the _<Editor questions: name?>_ helper method. This Feature Module requires the attribute of the worker you are giving authority to.
_<Editor questions: where do we find out about attributes?>_
You have the option to set a starting `location`, `velocity` and `rotation`. If you do not provide these, location and velocity are `default(Vector3)`, and rotation is `Quaternion.identity`.

Example showing _<Editor questions: what are we showing here?>_

```c sharp
public static class CubeTemplate
{
    public static Entity CreateCubeEntityTemplate(Coordinates coords)
    {
        var entityBuilder = EntityBuilder.Begin()
            .AddPosition(coords.X, coords.Y, coords.Z, WorkerUtils.UnityGameLogic)
            ...
            ...
            .AddTransformSynchronizationComponents(WorkerUtils.UnityGameLogic,
				location: coords.NarrowToUnityVector());
         return entityBuilder.Build();
    }
}
```
 #### Player lifecycle
 The helper method _<Editor questions: what's it called?>_ for this Feature Module _<Editor questions: does what?>_ . It requires a `clientAttribute` and a `serverAttribute`.
 
Example showing _<Editor questions: what are we showing here?>_
```c sharp
public static class PlayerTemplate
{
    public static Entity CreatePlayerEntityTemplate(List<string> clientAttributeSet, Improbable.Vector3f position)
    {
        // Obtain unique client attribute
        var clientAttribute = clientAttributeSet.First(attribute => attribute != WorkerUtils.UnityClient);
         var entityBuilder = EntityBuilder.Begin()
            .AddPosition(position.X, position.Y, position.Z, WorkerUtils.UnityGameLogic)
            ...
            ...
            .AddPlayerLifecycleComponents(clientAttribute, WorkerUtils.UnityGameLogic);
         return entityBuilder.Build();
    }
}
```

paulbalaji:
@ElleEss 

Took on your feedback, updated the doc and answering a couple of unanswered editor questions:

why "some" feature modules?
- We might release a feature module that may not require users to add extra SpatialOS components (for example, snapshot tooling might not need a special component for the tool to work)

where do we find out about attributes?
- I'm not even sure if it's explained at the moment

docs/content/entity-templates.md
```diff
@@ -53,6 +53,60 @@ public static class CreatureTemplate
 
 You can use this entity template to spawn a Creature entity. This can be done using the [MonoBehaviour](gameobject/world-commands.md) or [ECS](ecs/world-commands.md) workflow.
 
+### Feature Module components
```
jessicafalk: another question to this doc, unrelated to the actual changes (sorry), but we never specify what kind of entity we mean here. SpatialOS entity or Unity ECS entity, @ElleEss  what would be the best way to do this?

ElleEss: Yeah, my first pass had a comment at the top that the existing text on here needs work - it's not so clear - loads of passive voice, not clear what type of entity it is....the first half needs a review....

ElleEss: The disadvantage of reviewing docs in PRs is that you can't see them in context or make suggestions about unedited content which might need updating. Docs exist very much in context so the human-brain compiler can make sure it all makes sense.

paulbalaji: creating new ticket for Entity Template doc review (UTY-1252) + moving discussion to gdoc 

going to merge these changes in to unblock

PR #366, opened 2018-09-11
Reimplement `WorkerType` attribute from SDK

firtina-improbable:
**Contributions**: We are not currently taking public contributions - see our [contributions](CONTRIBUTING.md) policy. However, we are accepting issues and we do want your [feedback](../README.md#give-us-feedback).

-------

#### Description
This is how it could look like if we want to consider re-adding the `WorkerType` attribute for MonoBehaviours.

Example usage:

```cs
namespace Assets.Gamelogic.Shaders
{
    // Enable this MonoBehaviour on client workers only
    [WorkerType(WorkerUtils.UnityClient)]
    public class ReapplyShaders : MonoBehaviour
    {
        private Renderer[] renderers;

        private void Awake()
        {
            renderers = GetComponentsInChildren<Renderer>();
        }

        private void Start()
        {
            // ...
        }
    }
}
```

#### Tests
Tested locally on a port of Pirates.

`MonoBehaviours` do get enabled/disabled depending on which worker type they are on even if they inject nothing. The prefab preprocessor correctly disables them along with other behaviours even if they inject nothing.

Added automated testing

#### Documentation
Needs documentation

PR #408, opened 2018-09-20
Replace `world.GetOrCreateManager<WorkerSystem>()` by `world.GetExistingManager<WorkerSystem>()`

firtina-improbable:

-------

#### Description
This is because `WorkerSystem` requires constructor arguments.
If WorkerSystem is not on the world, that's an error, so now we throw exceptions.
#### Tests
Playground works as expected, tests green
#### Documentation
Internal


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -80,11 +82,15 @@ private static bool IsBehaviourEnabledInEditor(Object obj)
 
         private static bool DoesBehaviourNeedConditionalEnabling(Type targetType)
         {
-            return
-                targetType.GetCustomAttribute<WorkerTypeAttribute>() != null ||
-                targetType
-                    .GetFields(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public)
-                    .Any(field => Attribute.IsDefined(field, typeof(RequireAttribute), false));
+            if (!ConditionalEnableCache.ContainsKey(targetType))
```
jared-improbable: How about `TryGet(targetType, out bool result)`?

firtina-improbable: Yes, looks a bit nicer

workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerSystem.cs
```diff
@@ -27,6 +27,12 @@ public WorkerSystem(Connection connection, ILogDispatcher logDispatcher, string
             Origin = origin;
         }
 
+        public static WorkerSystem Instantiate(World world,
```
samiwh: Why add this?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Editor/Prefabs/PrefabPreprocessor.cs
```diff
@@ -80,11 +82,16 @@ private static bool IsBehaviourEnabledInEditor(Object obj)
 
         private static bool DoesBehaviourNeedConditionalEnabling(Type targetType)
         {
-            return
-                targetType.GetCustomAttribute<WorkerTypeAttribute>() != null ||
-                targetType
-                    .GetFields(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public)
-                    .Any(field => Attribute.IsDefined(field, typeof(RequireAttribute), false));
+            if (!ConditionalEnableCache.TryGetValue(targetType, out var needsConditionalEnabling))
```
samiwh: Was the lack of this causing a performance problem?

workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerSystem.cs
firtina-improbable: The main benefit of using named parameters instead of relying on the `params object[]` via `CreateManager` is type checking at compile time rather than runtime.

This way, if we want to add another parameter or adjust the constructor in another way, there would be runtime problems whose discovery would be more difficult when compared to this.

zeroZshadow: You still have this issue now? Since you would add the parameter to the constructor AND the `Instantiate` function.
Just moving the problem a layer deeper. I'm rather against this change.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Editor/Prefabs/PrefabPreprocessor.cs
firtina-improbable: This is a continuation of the previous PR #366, where opportunities to improve the neighborhood of the affected code have been discovered.

workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerSystem.cs
firtina-improbable: Let's say you update the constructor without this change; you need to manually find all places where you have called `CreateManager<WorkerSystem>(...)` and ensure you have modified them appropriately. If you miss a few, you won't get compile problems, but it may get triggered when the piece of code gets executed.

With this change, you modify the constructor and `Instantiate`, and all the places that use it have to be fixed until the user code can compile again.

However, since this code is used in the unconditionally for all execution paths that use `WorkerSystem`, the errors should be caught immediately when you run anything of that sort.

Therefore I'm happy to revert this.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Editor/Prefabs/PrefabPreprocessor.cs
firtina-improbable: In short: no real problem, it's a driveby.

samiwh: But what is the drive-by fixing? It makes the code harder to read and adds state to and editor only script, which previously had none. So if this isn't something that causes an issue it probably shouldn't be added.

firtina-improbable: Not a fix, an optimization. 

How much faster? I'm unsure, I don't think it's worth a benchmark.

Making the code harder to read: I sort of see what you mean.

I think it's worth shaving off some time from large projects, but can be done at a later phase instead. I'll revert it until it becomes an issue.

firtina-improbable:
Hey, my first XS pr!

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectCreationSystemHelper.cs
```diff
@@ -8,7 +8,7 @@ public static class GameObjectCreationSystemHelper
     {
         public static void EnableStandardGameObjectCreation(World world, GameObject workerGameObject = null)
         {
-            var workerSystem = world.GetOrCreateManager<WorkerSystem>();
+            var workerSystem = world.GetExistingManager<WorkerSystem>();
```
jessicafalk: should we throw an exception if this returns null?

firtina-improbable: this should already do that :)

jessicafalk: `GetExistingManager` won't throw an exception, it will return null

firtina-improbable:
The PR is now something completely different and no longer XS

PR #410, opened 2018-09-20
Removed an unneeded class from the public API

jared-improbable:


JonasImprobable:
For the record, I'm personally on board with removing the wrapping entity. I know that we have different opinions on this. While I'm ok with leaving this in for now, I'd appreciate if we could sit together and reach a decisive consensus on this at some point.

PR #413, opened 2018-09-20
Report correct load metric if target frame rate is -1.

martinzlocha:

-------

#### Description
If the frame rate is unbounded (-1) the calculation for the load is incorrect and shows very high numbers. It would be better to use a default value instead.

#### Tests
Ran locally, worked.

#### Documentation

#### Primary reviewers


workers/unity/Assets/Playground/Scripts/Metrics/MetricSendSystem.cs
```diff
@@ -54,6 +55,12 @@ protected override void OnUpdate()
         private static float DefaultLoadCalculation(float fps)
         {
             float targetFps = Application.targetFrameRate;
+
+            if (targetFps == -1)
+            {
+                targetFps = DefaultTargetFrameRate;
```
samiwh: This doesn't seem like it would give you the correct load.

martinzlocha: Why not? You can try calculating the load with fps = 30 and 60 and then at 30 load is 1 and at 60 its 0.

PR #418, opened 2018-09-24
Unify Test Namespaces

jamiebrynes7:
#### Description
Small cosmetic changes - make all the namespaces in the `test-project` unified for easier viewing in the TestRunner. Driveby on the gitignore
#### Tests
Yes
#### Documentation
N/A
#### Primary reviewers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,5 +61,6 @@ test-project/Assets/Plugins/Improbable.meta
 test-project/ProjectSettings/UnityConnectSettings.asset
 test-project/Assets/Generated/Source.meta
 test-project/Assets/Generated/Source/**/*.*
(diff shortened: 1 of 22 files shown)


samiwh:
Why have editmode test in the namespace before the thing they are tests for? Wouldn't that make more sense the other way around?
Also what is the motivation for doing this now?

jamiebrynes7:
@firtina-improbable Its then consistent with the assembly.

@samiwh Why `editmode` in the namespace - same answer as above. I'm not married to it.

And why not - I'm already waiting for review on 3 docs, pushing more docs doesn't feel productive (was also a driveby from over the weekend)

PR #420, opened 2018-09-24
UTY-356 Add Steam token config option

Zomtorg:

-------

#### Description
Implement the option for setting Steam credentials.
#### Tests
Automated unit tests for missing credentials, deployment tag adhering to a regex, and the expected behavior have been implemented.
#### Documentation
None
#### Primary reviewers
@jessicafalk @oblm

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Config/LocatorConfig.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Config/LocatorConfig.cs
@@ -53,7 +53,8 @@ namespace Improbable.Gdk.Core
 
         public void SetLoginToken(string loginToken)
         {
-            if (loginToken.Equals(string.Empty) && !LocatorParameters.CredentialsType.Equals(null))
+            // Steam token already provided, no need to initialise default values
+            if (loginToken.Equals(string.Empty) && LocatorParameters.CredentialsType == LocatorCredentialsType.Steam)
             {
                 return;
             }
@@ -91,8 +92,8 @@ namespace Improbable.Gdk.Core
                 parsedArgs, RuntimeConfigNames.SteamDeploymentTag, string.Empty);
             var steamToken = CommandLineUtility.GetCommandLineValue(
                 parsedArgs, RuntimeConfigNames.SteamToken, string.Empty);
-            config.SetLoginToken(loginToken);
             config.SetSteamCredentials(steamToken, steamDeploymentTag);
+            config.SetLoginToken(loginToken);
             config.SetProjectName(projectName);
             config.LocatorHost = CommandLineUtility.GetCommandLineValue(
                 parsedArgs, RuntimeConfigNames.LocatorHost, RuntimeConfigDefaults.LocatorHost);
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/com.improbable.gdk.core/Config/LocatorConfig.cs
```diff
@@ -16,10 +19,30 @@ public LocatorConfig()
...
+                    ValidateConfig(LocatorParameters.Steam.DeploymentTag, RuntimeConfigNames.SteamDeploymentTag);
+                    ValidateSteamDeploymentTag(LocatorParameters.Steam.DeploymentTag);
+                    break;
+            }
+        }
+
+        private void ValidateSteamDeploymentTag(string deploymentTag)
+        {
+            const string regex = @"^[A-Za-z0-9][A-Za-z0-9_]*$";
+            var match = Regex.Match(deploymentTag, regex);
```
jessicafalk: I think that it actually needs to be one specific tag:  `steam_deployment`
https://docs.improbable.io/reference/12.0/unitysdk/customize/steam#distributing-with-steam

workers/unity/Packages/com.improbable.gdk.core/Config/LocatorConfig.cs
```diff
@@ -30,20 +53,48 @@ public void SetProjectName(string projectName)
...
                 parsedArgs, RuntimeConfigNames.LoginToken, string.Empty);
             var projectName = CommandLineUtility.GetCommandLineValue(
                 parsedArgs, RuntimeConfigNames.ProjectName, string.Empty);
+            var steamDeploymentTag = CommandLineUtility.GetCommandLineValue(
+                parsedArgs, RuntimeConfigNames.SteamDeploymentTag, string.Empty);
+            var steamToken = CommandLineUtility.GetCommandLineValue(
+                parsedArgs, RuntimeConfigNames.SteamToken, string.Empty);
+
+            // Assign Steam Credentials first so that loginToken can fall back to default values if not provided
+            config.SetSteamCredentials(steamToken, steamDeploymentTag);
```
jessicafalk: a login token will fail as well, if it uses the default values. We literally need to provide a login token or steam credentials to go successfully through the locator flow.  So we should have a check here: do we have a steamToken? yes, go set it, no, go set the logintoken

Zomtorg: Having a second look at the implementation I have now baked the logic in the functions to cover the cases you mentioned.

workers/unity/Packages/com.improbable.gdk.core/Config/LocatorConfig.cs
Zomtorg: Here's where I derived the regex from: https://github.com/improbable/everything/blob/3da51c9907e18f0b52172cdb6dfb2c6bef04ce7e/worker_sdk/csharp/sdk/worker/Parameters.cs#L212

We can also hardcode the deployment tag to be "steam_deployment" if that's the case?

jamiebrynes7:
Closing this for the time being - can reopen when it becomes a priority again

PR #421, opened 2018-09-24
Cleanup of WorkerConnectorBase

samiwh:
#### Description
Made the worker connector a little less specific to playground
#### Tests
Ran it
#### Documentation
Docs currently still very WIP


workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -59,7 +71,14 @@ await Worker.CreateWorkerAsync(GetReceptionistConfig(workerType), logger, origin
                 }
 
                 var worker = await ConnectWithRetries(connectionDelegate, MaxConnectionAttempts, logger, workerType);
-                InitializeWorker(worker);
+
+                Worker = worker;
```
jamiebrynes7: I think the `worker` variable is not needed here

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -1,26 +1,38 @@
...
-        public GameObject LevelPrefab;
         public int MaxConnectionAttempts = 3;
         public bool UseExternalIp;
 
         public Worker Worker;
 
-        private GameObject levelInstance;
+        private List<Action<Worker>> workerConnectedCallbacks = new List<Action<Worker>>();
+
+        public event Action<Worker> OnWorkerCreationFinished
```
jamiebrynes7: This event doesn't appear to be used anywhere - is this in preparation for something else? Or was this an FR?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
samiwh: Yep. Will remove.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
samiwh: Yes this was an FR. We want to use it to tell things like UI elements that stuff has happened.

samiwh:
Merging this to unblock people. Can deal with nitpicks later.

PR #425, opened 2018-09-25
WIP: Small docs fixes.

oblm:
#### Description
WIP: Small docs fixes.
#### Tests
None
#### Documentation
This is documentation.

### DIFF
--- a/docs/content/build.md
+++ b/docs/content/build.md
@@ -16,13 +16,13 @@ To build the bridge and launch configurations for each of your server-workers an
 * From the Unity Editor menu, select **SpatialOS** > **Build worker configs**.
 
     or
-* In a terminal window, from the root of your game project, run:  `spatial build build-config` in the [`spatial` CLI](https://docs.improbable.io/reference/latest/shared/glossary#the-spatial-command-line-tool-cli).
+* In a terminal window, from the root of your game project, run: `spatial build build-config` via the [`spatial` CLI](https://docs.improbable.io/reference/latest/shared/glossary#the-spatial-command-line-tool-cli).
 
 ### Preparing the build configuration of your workers
 To prepare the build of your server-workers and client-workers you can create a new [scriptable object](https://docs.unity3d.com/ScriptReference/ScriptableObject.html) (Unity documentation) containing the information needed to build your workers correctly.
 
 To do this:
-1. In your Unity Editor, go to the Project window and select  **Assets** > **Create** > **SpatialOS** > **Build Configuration**. This creates a Unity Asset called `SpatialOS Build Configuration`.
+1. In your Unity Editor, go to the Project window and select **Assets** > **Create** > **SpatialOS** > **Build Configuration**. This creates a Unity Asset called `SpatialOS Build Configuration`.
 1. Click on this Asset to view it in the Unity Editor Inspector. Here you can configure each server-worker and client-worker for both local and cloud deployments. You can configure:
     * Which Scenes each worker should contain in its build.
     * Which platforms you want to build it for.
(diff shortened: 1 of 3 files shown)


docs/content/build.md
```diff
@@ -1,46 +1,55 @@
...
-## Building your workers
+---
 
-Before running a [deployment of your game](deploy.md), you need to:
-1. Build the bridge and launch configuration of your [workers](workers.md)
-2. Prepare the build configuration of your workers
-3. Build your workers
+## How to build your game.
 
-To build your workers, you don't need to know what the bridge and launch configurations are, but if you do want to know more about them, see the SpatialOS documentation on [bridge configuration](https://docs.improbable.io/reference/latest/shared/worker-configuration/bridge-config#bridge-configuration) and [launch configuration](https://docs.improbable.io/reference/latest/shared/worker-configuration/launch-configuration#worker-launch-configuration).
```
jessicafalk: this is something that we did delibaretly as Laura found that there were way too many links in point 2. please check with her first

oblm: will check this with @ElleEss,thanks for spotting.

oblm: signed off with @ElleEss

docs/content/build.md
```diff
@@ -1,46 +1,54 @@
...
+1. Build your workers.
+2. Build the bridge configuration and launch configuration of your [workers](workers.md).
+3. Prepare the build configuration of your workers.
+
+### Building your workers
+
+To build your workers:
+
+* With your project open in the Unity Editor, select **SpatialOS** > **Build `WorkerName` for local** or **Build `WorkerName` for cloud** where `WorkerName` is the name of any worker you want to build.
+
```
ElleEss: We would usually use "<WorkerName>" in this context. I think it was a mash-up of two styles in the previous version of this text.

docs/content/build.md
```diff
@@ -1,46 +1,54 @@
...
-2. Prepare the build configuration of your workers
-3. Build your workers
+## How to build your game
 
-To build your workers, you don't need to know what the bridge and launch configurations are, but if you do want to know more about them, see the SpatialOS documentation on [bridge configuration](https://docs.improbable.io/reference/latest/shared/worker-configuration/bridge-config#bridge-configuration) and [launch configuration](https://docs.improbable.io/reference/latest/shared/worker-configuration/launch-configuration#worker-launch-configuration).
+Before running a [deployment of your game](deploy.md) locally or in the cloud, you need to:
+
+1. Build your workers.
+2. Build the bridge configuration and launch configuration of your [workers](workers.md).
+3. Prepare the build configuration of your workers.
```
ElleEss: <3 the match between this and the titles.

docs/content/build.md
```diff
@@ -1,46 +1,54 @@
...
+
+* With your project open in the Unity Editor, select **SpatialOS** > **Build `WorkerName` for local** or **Build `WorkerName` for cloud** where `WorkerName` is the name of any worker you want to build.
+
+This starts the build process for the worker you selected and deployment type (cloud or local).
+
+> **It’s finished building when:** there is a message in the Unity Editor console.
+
+You can now find the built-out workers in `build/assembly/worker` in the root of the SpatialOS GDK for Unity project.
 
 ### Building the bridge and launch configuration of your workers
```
ElleEss: configurations

There are two right? This reads as if it's one.

docs/content/build.md
oblm: Do you mean that it should be:

>"WorkerName"

instead of
>`WorkerName`
?

docs/content/build.md
oblm: good spot, will fix

docs/content/build.md
oblm: thanks, I did it myself

docs/content/build.md
```diff
@@ -1,46 +1,54 @@
...
-To build the bridge and launch configurations for each of your server-workers and client-workers, you can either:
 
-* From the Unity Editor menu, select **SpatialOS** > **Build worker configs**.
+The [bridge](https://docs.improbable.io/reference/latest/shared/worker-configuration/launch-configuration#worker-bridge-configuration) and [launch](https://docs.improbable.io/reference/latest/shared/worker-configuration/launch-configuration#worker-launch-configuration) confgurations are pre-configued for you. You only need to build them. You do this for each of your [server-workers](https://docs.improbable.io/reference/latest/shared/glossary#server-worker) and [client-workers](https://docs.improbable.io/reference/latest/shared/glossary#client-worker). To do this you can either:
+
```
ElleEss: Typo for "confgurations"  and "pre-configued" in this para.

Good effort from us (fist bump) for finding text for bit of a tricky thing to express (for the reasons we discussed). 

Looking at it, I would prefer it if we did say something about not needing to find out about the bridge and launcher because we want to try to keep using the GDK as simple as possible. I would suggest something along the lines of:

"The [bridge](https://docs.improbable.io/reference/latest/shared/worker-configuration/launch-configuration#worker-bridge-configuration) and [launch](https://docs.improbable.io/reference/latest/shared/worker-configuration/launch-configuration#worker-launch-configuration) configurations are pre-configured for you. You don't need to learn about them, you only need to build them. To do this..."

docs/content/build.md
ElleEss: Couldn't do this "<" and ">".

docs/content/build.md
ElleEss: You are brilliant.

docs/content/build.md
ElleEss: "find out about them"?

docs/content/build.md
ElleEss: I can't show you because GitHub formats it and makes it disapper...<like this>.

ElleEss: It just disappeared.

ElleEss: This:
`<WorkerName>`
But without the backticks.

ElleEss: It's really frustrating not to be able to just write it.

oblm: I've removed the backticks to illustrate that this stops `<WorkerName>` rendering at all:

https://github.com/spatialos/UnityGDK/blob/399b41b24e3f18ac5ba34cdacc51b7c6b15b70a0/docs/content/build.md

oblm: I'm unsure what presentation you want here, please clarify.

PR #430, opened 2018-09-26
Mac Osascript Regression Fix

jamiebrynes7:
#### Description
Actually run the command given rather than `spatial osascript`
#### Tests
Manually
#### Documentation
N/A
#### Primary reviewers
@jared-improbable

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
@@ -204,7 +204,7 @@ namespace Improbable.Gdk.Tools
                 command = "osascript";
                 commandArgs = $@"-e 'tell application ""Terminal""
                                      activate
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -204,7 +204,7 @@ public static void LaunchLocalDeployment()
                 command = "osascript";
                 commandArgs = $@"-e 'tell application ""Terminal""
                                      activate
-                                     do script ""cd {SpatialProjectRootDir} && {Common.SpatialBinary} {command}""
+                                     do script ""cd {SpatialProjectRootDir} && {Common.SpatialBinary} local launch""
```
mvriel: Could do the same thing here as in the other location? To re-include the prior commandArgs definition?

gkassabli: Should we use commandArgs here?

PR #439, opened 2018-10-01
WIP UTY-347 Helpers for creating attributes from a workerId

tenevdev:
#### Description
Worker attributes for a specific worker id have the format `workerId:<id>`. These helpers encapsulate this format when creating attributes from id's.

#### Tests
Played the playground project.

#### Documentation
Not documented.

### DIFF
--- a/workers/unity/Assets/Playground/Config/WorkerUtils.cs
+++ b/workers/unity/Assets/Playground/Config/WorkerUtils.cs
@@ -16,8 +16,35 @@ namespace Playground
             new List<string>
             {
                 UnityGameLogic,
-                UnityClient
+                UnityClient,
             };
+
+        /// <summary>
+        /// Useful for specifying write access for a single client.
+        /// For example, an input component that drives your character.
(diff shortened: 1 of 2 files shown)


samiwh:
I'm currently changing this for the sake of an internal FR. This feature module requires lots of other large changes so I wouldn't bother with these changes right now.

workers/unity/Assets/Playground/Config/WorkerUtils.cs
```diff
@@ -16,9 +16,36 @@ public static class WorkerUtils
...
 
+        /// <summary>
+        /// Useful for specifying write access for a single client.
+        /// For example, an input component that drives your character.
+        /// </summary>
+        /// <param name="workerId"></param>
+        /// <returns></returns>
+        public static string SpecificClient(string workerId)
+        {
+            return "workerId:" + workerId;
```
samiwh: This is actually more brittle than what is currently there.

PR #440, opened 2018-10-01
Added an OwningWorker component and helpers

samiwh:
#### Description
Added an a component to be used when a player is first checked out
#### Tests
Ran it
#### Documentation
Pending

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
+++ b/workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
@@ -1,3 +1,4 @@
+using System;
 using Improbable.Gdk.Core;
 using Improbable.PlayerLifecycle;
 using Unity.Entities;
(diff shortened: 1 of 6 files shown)


workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -7,14 +8,44 @@ namespace Improbable.Gdk.PlayerLifecycle
...
+            var worker = workerWorld.GetExistingManager<WorkerSystem>();
+
+            if (worker == null)
+            {
+                throw new InvalidOperationException("Provided World does not have an associated worker");
+            }
+
+            if (!worker.TryGetEntity(entityId.EntityId, out var entity))
+            {
+                throw new InvalidOperationException(
```
firtina-improbable: Should this return false instead?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -7,14 +8,44 @@ namespace Improbable.Gdk.PlayerLifecycle
...
+                    $"Entity with SpatialOS Entity ID {entityId.EntityId.Id} is not in this worker's view");
+            }
+
+            if (entityManager.HasComponent<OwningWorker.Component>(entity))
```
firtina-improbable: Do you mean **not** has?

samiwh: Yep

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
samiwh: I don't think so. This could only ever happen through gross misuse of the API

samiwh: Also we don't know that the answer is false. This is being asked for information you can't know. So saying no is strictly wrong.

firtina-improbable: I see, I agree

PR #441, opened 2018-10-01
UTY-347 Add helper for creating an attribute from worker id

tenevdev:
#### Description
A helper to protect us from breaking changes to the attribute string format.

#### Tests
None

#### Documentation
In comments.


tenevdev:
I've linked this PR and the slack thread to the ticket on JIRA. Feel free to close it as it's not a satisfactory solution for me as well...

tenevdev:
@samiwh nothing for now because we accept strings in the entity builder. Hopefully, we could move to a type which prevents people from using the `"workerId:%s"` format in their code, so that we can change it at some point if we need to without breaking user code. Then class methods here will return said type as well instead of a string...

samiwh:
Closing this as it is now orphaned

PR #445, opened 2018-10-02
UTY-1181 Adding plugin directories compatibility logic

gkassabli:
This PR adds a logic that allows us to mark plugins compatibility platform after download

-------

#### Description
We use different plugins for different platforms and we need to mark their compatibility in order to avoid conflicts in Unity (i.e. iOS Worker DLL and non-iOS one). This PR adds a logic that marks plugins compatibility based on our plugin structure.
#### Tests
Run Download plugins in Unity and ensure that plugins are marked as expected.
#### Documentation
What would be a good place to document this?

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
@@ -27,10 +27,10 @@ namespace Improbable.Gdk.Tools
 
         private static readonly List<PluginDirectoryCompatibility> PluginsCompatibilityList = new List<PluginDirectoryCompatibility>
         {
-            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/OSX", false, new List<BuildTarget> { BuildTarget.StandaloneOSX }, null, true),
-            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/Linux", false, new List<BuildTarget> { BuildTarget.StandaloneLinuxUniversal }, null, true),
-            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/Windows", false, new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, null, true),
-            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Sdk/Common", true, null, null, true),
+            PluginDirectoryCompatibility.PluginDirectoryIncludePlatforms("Assets/Plugins/Improbable/Core/OSX", new List<BuildTarget> { BuildTarget.StandaloneOSX }, true),
+            PluginDirectoryCompatibility.PluginDirectoryIncludePlatforms("Assets/Plugins/Improbable/Core/Linux", new List<BuildTarget> { BuildTarget.StandaloneLinuxUniversal }, true),
+            PluginDirectoryCompatibility.PluginDirectoryIncludePlatforms("Assets/Plugins/Improbable/Core/Windows", new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, true),
+            PluginDirectoryCompatibility.PluginDirectoryExcludePlatforms("Assets/Plugins/Improbable/Sdk/Common", null, true),
         };
 
         [MenuItem(DownloadForceMenuItem, false, DownloadForcePriority)]
@@ -147,6 +147,9 @@ namespace Improbable.Gdk.Tools
             return exitCode == 0 ? DownloadResult.Success : DownloadResult.Error;
         }
 
+        /// <summary>
+        ///     Sets plugin platform compatibility based on directory structure
+        /// </summary>
         private static void SetPluginsCompatibility()
         {
             foreach (var pluginDirectoryCompatibility in PluginsCompatibilityList)
@@ -184,7 +187,21 @@ namespace Improbable.Gdk.Tools
 
         private class PluginDirectoryCompatibility
         {
-            public PluginDirectoryCompatibility(string path,
+            public static PluginDirectoryCompatibility PluginDirectoryIncludePlatforms(string path,
+                List<BuildTarget> compatiblePlatforms,
+                bool editorCompatible)
+            {
+                return new PluginDirectoryCompatibility(path, false, compatiblePlatforms, null, editorCompatible);
+            }
+
+            public static PluginDirectoryCompatibility PluginDirectoryExcludePlatforms(string path,
+                List<BuildTarget> incompatiblePlatforms,
+                bool editorCompatible)
+            {
+                return new PluginDirectoryCompatibility(path, true, null, incompatiblePlatforms, editorCompatible);
+            }
+
+            private PluginDirectoryCompatibility(string path,
                 bool anyPlatformCompatible,
                 List<BuildTarget> compatiblePlatforms,
                 List<BuildTarget> incompatiblePlatforms,
@@ -192,16 +209,6 @@ namespace Improbable.Gdk.Tools
             {
                 Path = path;
                 AnyPlatformCompatible = anyPlatformCompatible;
-                if (anyPlatformCompatible && compatiblePlatforms != null)
-                {
-                    Debug.LogError($"Bad plugin compatibility configuration for  {path}. Compatible with any shouldn't be used with a list of compatible platforms");
-                }
-
-                if (!anyPlatformCompatible && incompatiblePlatforms != null)
-                {
-                    Debug.LogError($"Bad plugin compatibility configuration for  {path}. Not compatible with any shouldn't be used with a list of incompatible platforms");
-                }
-
                 CompatiblePlatforms = compatiblePlatforms ?? new List<BuildTarget>();
                 IncompatiblePlatforms = incompatiblePlatforms ?? new List<BuildTarget>();
                 EditorCompatible = editorCompatible;
--- a/workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
@@ -30,7 +30,7 @@ namespace Improbable.Gdk.Tools
             new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/OSX", false, new List<BuildTarget> { BuildTarget.StandaloneOSX }, null, true),
             new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/Linux", false, new List<BuildTarget> { BuildTarget.StandaloneLinuxUniversal }, null, true),
             new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/Windows", false, new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, null, true),
-            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Sdk/OSX", true, null, null, true),
+            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Sdk/Common", true, null, null, true),
         };
 
         [MenuItem(DownloadForceMenuItem, false, DownloadForcePriority)]
--- a/workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
@@ -27,10 +27,10 @@ namespace Improbable.Gdk.Tools
 
         private static readonly List<PluginDirectoryCompatibility> PluginsCompatibilityList = new List<PluginDirectoryCompatibility>
         {
-            PluginDirectoryCompatibility.PluginDirectoryIncludePlatforms("Assets/Plugins/Improbable/Core/OSX", new List<BuildTarget> { BuildTarget.StandaloneOSX }, true),
-            PluginDirectoryCompatibility.PluginDirectoryIncludePlatforms("Assets/Plugins/Improbable/Core/Linux", new List<BuildTarget> { BuildTarget.StandaloneLinuxUniversal }, true),
-            PluginDirectoryCompatibility.PluginDirectoryIncludePlatforms("Assets/Plugins/Improbable/Core/Windows", new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, true),
-            PluginDirectoryCompatibility.PluginDirectoryExcludePlatforms("Assets/Plugins/Improbable/Sdk/Common", null, true),
+            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/OSX", new List<BuildTarget> { BuildTarget.StandaloneOSX }, true),
+            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Linux", new List<BuildTarget> { BuildTarget.StandaloneLinuxUniversal }, true),
+            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Windows", new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, true),
+            PluginDirectoryCompatibility.CreateAllCompatible("Assets/Plugins/Improbable/Sdk/Common"),
         };
 
         [MenuItem(DownloadForceMenuItem, false, DownloadForcePriority)]
@@ -45,8 +45,6 @@ namespace Improbable.Gdk.Tools
             }
 
             Download();
-            AssetDatabase.Refresh();
-            SetPluginsCompatibility();
         }
 
         private static void RemoveMarkerFile()
@@ -109,9 +107,7 @@ namespace Improbable.Gdk.Tools
                 return DownloadResult.AlreadyInstalled;
             }
 
-            DownloadResult result = Download();
-            SetPluginsCompatibility();
-            return result;
+            return Download();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -136,6 +150,81 @@ private static DownloadResult Download()
...
 
+        private static void SetPluginsCompatibility()
+        {
+            foreach (var pluginDirectoryCompatibility in pluginsCompatibilityList)
+            {
+                var pluginPaths = AssetDatabase.FindAssets("", new[] { pluginDirectoryCompatibility.Path });
+                foreach (var pluginPath in pluginPaths)
+                {
+                    var plugin = AssetImporter.GetAtPath(AssetDatabase.GUIDToAssetPath(pluginPath)) as PluginImporter;
+                    if (!plugin)
```
jamiebrynes7: Do an explicit null check. `plugin == null`

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -136,6 +150,81 @@ private static DownloadResult Download()
...
+                List<BuildTarget> compatiblePlatforms,
+                List<BuildTarget> incompatiblePlatforms,
+                bool editorCompatible)
+            {
+                Path = path;
+                AnyPlatformCompatible = anyPlatformCompatible;
+                CompatiblePlatforms = compatiblePlatforms;
+                IncompatiblePlatforms = incompatiblePlatforms;
+                EditorCompatible = editorCompatible;
+            }
```
jamiebrynes7: If you do validation of the arguments here - you can pull a lot of the nested logic out from above into here. Logically - the validation should happen here as well imo

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -25,6 +25,17 @@ internal static class DownloadCoreSdk
...
         private const string DownloadForceMenuItem = "SpatialOS/Download CoreSdk (force)";
 
+        private static readonly List<PluginDirectoryCompatibility> pluginsCompatibilityList = new List<PluginDirectoryCompatibility>
+        {
+            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/Android", false, new List<BuildTarget> { BuildTarget.Android }, null, false),
+            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/iOS", false, new List<BuildTarget> { BuildTarget.iOS }, null, false),
+            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/OSX", false, new List<BuildTarget> { BuildTarget.StandaloneOSX }, null, true),
+            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/Linux", false, new List<BuildTarget> { BuildTarget.StandaloneLinuxUniversal }, null, true),
+            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/Windows", false, new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, null, true),
+            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Sdk/OSX", true, null, new List<BuildTarget> { BuildTarget.iOS }, true),
```
DanailPenev: This would require changes to the download program itself to not name the folder OSX as it is every other platform apart from iOS.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -25,6 +25,17 @@ internal static class DownloadCoreSdk
 
         private const string DownloadForceMenuItem = "SpatialOS/Download CoreSdk (force)";
 
+        private static readonly List<PluginDirectoryCompatibility> pluginsCompatibilityList = new List<PluginDirectoryCompatibility>
```
DanailPenev: It would be great if there was a way to link this to the paths in `.DownloadCoreSdk.cs`. Otherwise if we change the folder hierarchy we have to change hardcoded paths in 2 files.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -25,6 +25,17 @@ internal static class DownloadCoreSdk
...
+        {
+            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/Android", false, new List<BuildTarget> { BuildTarget.Android }, null, false),
```
DanailPenev: Either leave the mobile folders out (to add them together with the rest of the mobile stuff) or add them to the Download program. Will be more consistent.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: I would love that as well. The problem is that we deliberately wanted to keep DownloadCoreSDK separated from Unity logic so that it can be run separately (in case Unity bugs with menu item). Considering that, I couldn't come up with a lightweight solution to share the paths (one solution that came to my mind is shared config file, but that one seemed a bit too heavy to just avoid duplicating strings in 2 places).
Happy to hear the suggestions!

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: Good point, didn't clean this up when rebasing

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: Yeah, the name is confusing. Shall we rename it to something like "Common" for instance to avoid the confusion?

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: (y)

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: Good point, let me rework that

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -25,6 +25,14 @@ internal static class DownloadCoreSdk
 
         private const string DownloadForceMenuItem = "SpatialOS/Download CoreSdk (force)";
 
+        private static readonly List<PluginDirectoryCompatibility> PluginsCompatibilityList = new List<PluginDirectoryCompatibility>
+        {
+            new PluginDirectoryCompatibility("Assets/Plugins/Improbable/Core/OSX", false, new List<BuildTarget> { BuildTarget.StandaloneOSX }, null, true),
```
gkassabli: > Copying the lost comment

It would be great if there was a way to link this to the paths in .DownloadCoreSdk.cs. Otherwise if we change the folder hierarchy we have to change hardcoded paths in 2 files.
The problem is that we deliberately wanted to keep DownloadCoreSDK separated from Unity logic so that it can be run separately (in case Unity bugs with menu item). Considering that, I couldn't come up with a lightweight solution to share the paths (one solution that came to my mind is shared config file, but that one seemed a bit too heavy to just avoid duplicating strings in 2 places).
Happy to hear the suggestions!

jared-improbable: It's in a separate process because for some reason Unity project file generation doesn't include references to the assembly that allows access to the Zipfile functionality, so we can't unzip in a platform-agnostic way. There's no concern with "glitches in the menu or anything - just an annoying Unity quirk.

Regardless, we don't change these paths often, so I don't think it's a huge concern.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -136,6 +147,73 @@ private static DownloadResult Download()
...
+        }
+
+        private class PluginDirectoryCompatibility
+        {
+            public PluginDirectoryCompatibility(string path,
+                bool anyPlatformCompatible,
+                List<BuildTarget> compatiblePlatforms,
+                List<BuildTarget> incompatiblePlatforms,
+                bool editorCompatible)
+            {
```
jamiebrynes7: Actually on second thought - if combinations of arguments can produce an invalid state - why not provide multiple constructors/static methods that restrict the argument combinations?

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: That makes sense, thanks for the context!

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -136,6 +147,73 @@ private static DownloadResult Download()
             return exitCode == 0 ? DownloadResult.Success : DownloadResult.Error;
         }
 
+        private static void SetPluginsCompatibility()
+        {
+            foreach (var pluginDirectoryCompatibility in PluginsCompatibilityList)
+            {
+                var pluginPaths = AssetDatabase.FindAssets("", new[] { pluginDirectoryCompatibility.Path });
```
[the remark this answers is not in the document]
gkassabli: It is used for setting plugin compatibility below. AssetDatabase is only used to enumerate over the content of directory and create PluginImporter instance.
Is there any API call that would be more convenient to use here that I didn't notice?

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: Fair enough, I'll add that

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -100,7 +109,9 @@ internal static DownloadResult TryDownload()
                 return DownloadResult.AlreadyInstalled;
             }
 
-            return Download();
+            DownloadResult result = Download();
+            SetPluginsCompatibility();
```
jamiebrynes7: Given this is called after `Download` in both cases - can this call be moved into the `Download()` function?

gkassabli: We have to refresh Asset Database before we can set plugins compatibility. This part actually doesn't work well with generation, as refresh doesn't happen.
Let me move that into Download along with AssetDatabase.Refresh() call

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -25,6 +25,14 @@ internal static class DownloadCoreSdk
 
         private const string DownloadForceMenuItem = "SpatialOS/Download CoreSdk (force)";
 
+        private static readonly List<PluginDirectoryCompatibility> PluginsCompatibilityList = new List<PluginDirectoryCompatibility>
+        {
+            PluginDirectoryCompatibility.PluginDirectoryIncludePlatforms("Assets/Plugins/Improbable/Core/OSX", new List<BuildTarget> { BuildTarget.StandaloneOSX }, true),
+            PluginDirectoryCompatibility.PluginDirectoryIncludePlatforms("Assets/Plugins/Improbable/Core/Linux", new List<BuildTarget> { BuildTarget.StandaloneLinuxUniversal }, true),
+            PluginDirectoryCompatibility.PluginDirectoryIncludePlatforms("Assets/Plugins/Improbable/Core/Windows", new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, true),
+            PluginDirectoryCompatibility.PluginDirectoryExcludePlatforms("Assets/Plugins/Improbable/Sdk/Common", null, true),
```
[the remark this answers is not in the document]
gkassabli: (y)

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -133,9 +140,89 @@ private static DownloadResult Download()
                 EditorApplication.UnlockReloadAssemblies();
             }
 
+            AssetDatabase.Refresh();
+            SetPluginsCompatibility();
```
jamiebrynes7: Small optimisation - these can be done conditionally on the result of the download

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -25,6 +25,14 @@ internal static class DownloadCoreSdk
 
         private const string DownloadForceMenuItem = "SpatialOS/Download CoreSdk (force)";
 
+        private static readonly List<PluginDirectoryCompatibility> PluginsCompatibilityList = new List<PluginDirectoryCompatibility>
+        {
+            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/OSX", new List<BuildTarget> { BuildTarget.StandaloneOSX }, true),
+            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Linux", new List<BuildTarget> { BuildTarget.StandaloneLinuxUniversal }, true),
+            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Windows", new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, true),
```
jessicafalk: only StandaloneWindows64 if the dll is in the `x86_64` folder. if it's in `x86`, it needs to be marked as Windows32

gkassabli: Good point, let me specify that

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: (y)

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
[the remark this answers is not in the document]
jared-improbable: https://docs.unity3d.com/ScriptReference/PluginImporter.GetImporters.html or https://docs.unity3d.com/ScriptReference/PluginImporter.GetAllImporters.html have been used in the past.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -133,7 +141,92 @@ private static DownloadResult Download()
...
+
+        /// <summary>
+        ///     Sets plugin platform compatibility based on directory structure
+        /// </summary>
+        private static void SetPluginsCompatibility()
+        {
+            foreach (var pluginDirectoryCompatibility in PluginsCompatibilityList)
+            {
+                var pluginPaths = AssetDatabase.FindAssets("", new[] { pluginDirectoryCompatibility.Path });
+                foreach (var pluginPath in pluginPaths)
```
jared-improbable: How about `foreach (var pluginPath in pluginPaths.OfType<PluginImporter>())`?
Then you can drop the first chunk of the loop.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -25,6 +25,15 @@ internal static class DownloadCoreSdk
 
         private const string DownloadForceMenuItem = "SpatialOS/Download CoreSdk (force)";
 
+        private static readonly List<PluginDirectoryCompatibility> PluginsCompatibilityList = new List<PluginDirectoryCompatibility>
+        {
+            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/OSX", new List<BuildTarget> { BuildTarget.StandaloneOSX }, true),
```
jared-improbable: A named parameter will make the meaning of all the `true/false` arguments clearer.

gkassabli: Good point, I am still getting used to C# conventions.
Would it be a better idea to only name the last parameter, assuming that path and platform list are implied from the method call, or is it better to explicitly state all parameters?

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: (y)

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
jared-improbable: Yeah, sorry, that's what I meant. Just the last parameter needs to be named; I think the context is obvious for the other parameters.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
gkassabli: Just realised that this won't work, because we only have asset GUID, not asset itself from AssetDatabase - we can't distinguish between plugin and non-plugin until we get actual instance

PR #448, opened 2018-10-03
Allow getting of EntityIds from EntityQuery results

firtina-improbable:

-------

#### Description
This has some uses (e.g. if users want to send commands to these entities) and it's still safe to access the entity ids.

#### Tests
Adjusted relevant tests, also tested locally, this code:
```csharp
        public void OnEnable()
        {
            worldCommandResponseHandler.OnEntityQueryResponse += OnEntityQueryResponse;
        }

        private void OnEntityQueryResponse(WorldCommands.EntityQuery.ReceivedResponse response)
        {
            Debug.Log($"Entity query response {response.StatusCode}");

            if (response.StatusCode == StatusCode.Success)
            {
                Debug.Log($"count: {response.ResultCount}");

                foreach (var responseEntityId in response.EntityIds)
                {
                    Debug.Log($"Result entity id: {responseEntityId.Id}");
                }
            }
        }
```

Produces these logs:
image

#### Documentation
Will update relevant docs if pr goes in:
- known issues?
- worldcommands docs in gdocs

#### Primary reviewers
@paulbalaji

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Commands/WorldCommands.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Commands/WorldCommands.cs
@@ -700,11 +700,6 @@ namespace Improbable.Gdk.Core.Commands
 
             public static Request CreateRequest(Improbable.Worker.Query.EntityQuery entityQuery, uint? timeoutMillis = null, Object context = null)
             {
-                if (entityQuery.ResultType is SnapshotResultType)
-                {
-                    Debug.LogWarning("Cannot safely access component data from entity query - this is a known issue. To protect its integrity, the worker will drop the request before sending.");
-                }
-
                 return new Request
                 {
                     EntityQuery = entityQuery,
@@ -718,7 +713,7 @@ namespace Improbable.Gdk.Core.Commands
             {
                 public StatusCode StatusCode { get; }
                 public string Message { get; }
-                public Dictionary<EntityId, Entity> Result { get; }
+                public EntityId[] EntityIds { get; }
                 public int ResultCount { get; }
                 public Request RequestPayload { get; }
                 public object Context { get; }
@@ -728,7 +723,7 @@ namespace Improbable.Gdk.Core.Commands
                 {
                     StatusCode = op.StatusCode;
                     Message = op.Message;
-                    Result = op.Result;
+                    EntityIds = op.Result?.Keys.ToArray();
                     ResultCount = op.ResultCount;
                     RequestPayload = req;
                     Context = context;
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/com.improbable.gdk.core/Commands/WorldCommands.cs
```diff
@@ -700,11 +700,6 @@ public struct Request
 
             public static Request CreateRequest(Improbable.Worker.Query.EntityQuery entityQuery, uint? timeoutMillis = null, Object context = null)
             {
-                if (entityQuery.ResultType is SnapshotResultType)
-                {
-                    Debug.LogWarning("Cannot safely access component data from entity query - this is a known issue. To protect its integrity, the worker will drop the request before sending.");
```
paulbalaji: perhaps we could leave the warning in, but let the user know that only the entity ids will be returned and not the component data

firtina-improbable: It'd be too spammy imo.

I would make the warning optional somehow... like a `bool iKnowWhatIAmDoing`, or 

`#if DISABLE_WARNING_ENTITY_QUERIES_CAN_ONLY_RETURN_ENTITY_IDS`

Or a `CreateRequestForEntityIds` next to it that will omit the warning, or something like that?

samiwh: Why remove this. This seems correct.

workers/unity/Packages/com.improbable.gdk.core/Systems/WorldCommandsSendSystem.cs
```diff
@@ -121,11 +121,6 @@ protected override void OnUpdate()
                 var entity = entityQuerySenderData.Entities[i];
                 foreach (var req in sender.RequestsToSend)
                 {
-                    if (req.EntityQuery.ResultType is SnapshotResultType)
-                    {
-                        continue;
```
samiwh: This is also correct code.

workers/unity/Packages/com.improbable.gdk.core/Commands/WorldCommands.cs
```diff
@@ -718,7 +713,7 @@ public struct ReceivedResponse
             {
                 public StatusCode StatusCode { get; }
                 public string Message { get; }
-                public Dictionary<EntityId, Entity> Result { get; }
+                public EntityId[] EntityIds { get; }
```
samiwh: I don't want to change the API in advance of us fixing it.

PR #451, opened 2018-10-05
Fix initial compatibility update

gkassabli:
Move initial compatibility update to the time where assets are imported + add lazy update to avoid reimport loop

-------

#### Description
Plugin compatibility settings were not applied correctly at the initial launch. This was caused by the AssetDatabase not importing plugins right after the download.
To address the issue plugins compatibility update was moved to asset postprocessing update. Additionally plugin compatibility is now only updated when it is different from expected. This is done to avoid import cycle.
#### Tests
Do a clean checkout -> ensure that Plugin compatibility is set correctly when Unity is loaded for the first time
#### Documentation
Do we need to document this? If so, what would be a good place for that?

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
@@ -13,6 +13,26 @@ namespace Improbable.Gdk.Tools
         Error
     }
 
+    internal class PluginPostprocessor : AssetPostprocessor
+    {
+        static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths)
+        {
+            var processPlugins = false;
+            foreach (var asset in importedAssets)
+            {
+                if (DownloadCoreSdk.IsImprobablePlugin(asset))
+                {
+                    processPlugins = true;
+                }
+            }
+
+            if (processPlugins)
+            {
+                DownloadCoreSdk.SetPluginsCompatibility();
+            }
+        }
+    }
+
     internal static class DownloadCoreSdk
     {
         private const int DownloadForcePriority = 50;
@@ -30,7 +50,7 @@ namespace Improbable.Gdk.Tools
             PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/OSX", new List<BuildTarget> { BuildTarget.StandaloneOSX }, editorCompatible: true),
             PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Linux", new List<BuildTarget> { BuildTarget.StandaloneLinux64 }, editorCompatible: true),
             PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Windows/x86_64", new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, editorCompatible: true),
-            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Windows/x86", new List<BuildTarget> { BuildTarget.StandaloneWindows }, false),
+            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Windows/x86", new List<BuildTarget> { BuildTarget.StandaloneWindows }, editorCompatible: true),
             PluginDirectoryCompatibility.CreateAllCompatible("Assets/Plugins/Improbable/Sdk/Common"),
         };
 
@@ -46,6 +66,8 @@ namespace Improbable.Gdk.Tools
             }
 
             Download();
+            SetPluginsCompatibility();
+			AssetDatabase.Refresh();
         }
 
         private static void RemoveMarkerFile()
@@ -88,9 +110,14 @@ namespace Improbable.Gdk.Tools
             return !failedPlugins.Any();
         }
 
+        internal static bool IsImprobablePlugin(string assetPath)
+        {
+            return assetPath.StartsWith("Assets/Plugins/Improbable");
+        }
+
         private static bool ShouldCheckPluginForLock(PluginImporter p)
         {
-            if (!p.assetPath.StartsWith("Assets/Plugins/Improbable"))
+            if (!IsImprobablePlugin(p.assetPath))
             {
                 return false;
             }
@@ -146,18 +173,21 @@ namespace Improbable.Gdk.Tools
                 return DownloadResult.Error;
             }
 
-            AssetDatabase.Refresh();
-            SetPluginsCompatibility();
             return DownloadResult.Success;
         }
 
         /// <summary>
         ///     Sets plugin platform compatibility based on directory structure
         /// </summary>
-        private static void SetPluginsCompatibility()
+        internal static void SetPluginsCompatibility()
         {
+	        AssetDatabase.StartAssetEditing();
             foreach (var pluginDirectoryCompatibility in PluginsCompatibilityList)
             {
+                if (!Directory.Exists(pluginDirectoryCompatibility.Path))
+                {
+                    continue;
+                }
                 var pluginPaths = AssetDatabase.FindAssets("", new[] { pluginDirectoryCompatibility.Path });
                 foreach (var pluginPath in pluginPaths)
                 {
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -146,15 +168,13 @@ private static DownloadResult Download()
                 return DownloadResult.Error;
             }
 
-            AssetDatabase.Refresh();
```
firtina-improbable: If you don't do a `Refresh` here it's likely that the DLLs may not be imported at all until something else triggers a refresh somewhere else.

For an alternate approach for the fix, it may help to take advantage of `AssetDatabase.StartAssetEditing` and `try{...}finally{AssetDatabase.StopAssetEditing()}` - this way all of the dlls will be grouped together in one import.

I was playing around with this script for an example:

```cs
using System.IO;
using UnityEditor;
using UnityEngine;
using Directory = UnityEngine.Windows.Directory;

public class TestEditorScript : AssetPostprocessor
{
    [MenuItem("Test/Imports")]
    public static void TestBatchImports()
    {
        var testFilesPath = Path.GetFullPath(Path.Combine(Application.dataPath, "test-files"));

//        AssetDatabase.StartAssetEditing();

        Directory.CreateDirectory(testFilesPath);

        try
        {
            for (var i = 0; i < 3; ++i)
            {
                var fileName = $"file-{i}.txt";
                var fullFilePath = Path.Combine(testFilesPath, fileName);
                File.WriteAllText(fullFilePath, $"{Random.Range(0, 100)}-{Random.Range(0, 100)}");

                Debug.Log($"<color=green>Triggering an import: {fileName}</color>");

                var assetPathInAssets = Path.Combine("Assets", "test-files", fileName);
                var importer = AssetImporter.GetAtPath(assetPathInAssets);
                Debug.Log($"<color=green>Importer for {assetPathInAssets} is null before first import: {(importer == null)}</color>");

                AssetDatabase.ImportAsset(assetPathInAssets);
                Debug.Log($"<color=green>Import triggered: {fileName}</color>");

                importer = AssetImporter.GetAtPath(assetPathInAssets);
                Debug.Log($"<color=green>Importer for {assetPathInAssets} is null after first import: {(importer == null)}</color>");

                Debug.Log($"<color=green>Triggering a second import: {fileName}</color>");
                AssetDatabase.ImportAsset(assetPathInAssets);
                Debug.Log($"<color=green>Second import triggered: {fileName}</color>");
            }
        }
        finally
        {
//            AssetDatabase.StopAssetEditing();
        }
    }

    static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets,
        string[] movedFromAssetPaths)
    {
        if (importedAssets.Length > 0)
        {
            Debug.Log($"<color=red>Imported assets: {string.Join(",", importedAssets)}</color>");
        }

        if (deletedAssets.Length > 0)
        {
            Debug.Log($"<color=red>Deleted assets: {string.Join(",", deletedAssets)}</color>");
        }
    }
}
```

In the default configuration the script creates a few files and imports them one by one (without using `AssetDatabase.StartAssetEditing()`).

If you copy in this script to a Unity project and select "Test/Imports", you get this sequence of logs:
image

This highlights a few things (some of these things we may know already but may help to highlight):
- the asset importer is null before the first import for the asset happens (i.e. before the AssetPostprocessor gets executed)
- the imports are synchronous, so one import will finish before `ImportAsset` returns
- after the import is completed, the asset importer is not null any more, so we can now modify import settings

For the next step:
- delete the `test-files` folder, 
- uncomment `StartAssetEditing` and `StopAssetEditing` to enable batch imports 
- select **Test** / **Imports**

image

This run highlights:
- First of all everything is much faster! (can see this from the timestamps too)
- However, asset importers are null even after the first "import" call is finished (probably because the "import" call now only flags for `StopAssetEditing` to do the importing) - so guessing they will not be null after `StopAssetEditing`

To test the last guess:
```cs

        finally
        {
            AssetDatabase.StopAssetEditing();
            
            for (var i = 0; i < 3; ++i)
            {
                var fileName = $"file-{i}.txt";

                var assetPathInAssets = Path.Combine("Assets", "test-files", fileName);
                var importer = AssetImporter.GetAtPath(assetPathInAssets);
                Debug.Log($"<color=green>Importer for {assetPathInAssets} is null after StopAssetEditing: {(importer == null)}</color>");
            }
        }
```

image

--------------

## Recommendation

Based on these findings, my recommendation would be to do this:

- `private static DownloadResult Download()`
  - StartAssetEditing()
  - Download all dlls
  - Trigger `AssetDatabase.ImportAsset` on all written dll paths within assets
  - StopAssetEditing() (let dlls get imported first time, likely with wrong settings!)
  - SetPluginsCompatibility()
- `private static void SetPluginsCompatibility()`
  - StartAssetEditing
  - Iterate over all written dll paths within assets:
    - Adjust importer settings if necessary (we know _(†)_ importers have to exist at this point)
    - Trigger a reimport if necessary
  - StopAssetEditing (let dlls get imported with correct importer settings)

I think it would not need the post processor in this case, unless we want to handle the situation when the user (or something else) replaces the plugin dlls outside these methods.

-----------

## † Scary story time

Note: **I'm not sure** what happens during `StopAssetEditing` if one import fails (for example, if two dlls have same name and same platform). If the importing process for DLLs does not continue for all DLLs in this case, at worst case multiple passes of SetPluginsCompatibility may need to happen, until AssetImporters exist and have been configured correctly for all DLLs.

This is probably unlikely but still may be worth a test.

Could the next few paragraphs happen for example?

In near worst case, we have let's say 5 very unfortunate dlls.
- first one gets imported with default settings (wrong)
- second one tries to get imported with default settings (wrong), fails, process stops here?
- third one does not get imported
- fourth one does not get imported
- fifth one does not get imported

In this case the first and second ones will have asset importers, but not the third and the fourth and the fifth. So the plugin compatibility method would be able to fix the first two, then, in near worst case (again)
- first one gets imported fine
- second one gets imported fine
- third one gets imported with default settings (wrong) but does not fail because it may only clash with the fourth (so evil)
- fourth one tries to get imported with default settings (wrong), fails
- fifth one does not get imported

Now first, second are all good, third and fourth can be fixed because there will be importers, but fifth is still in trouble and without an importer to fix.

Plugin compatibility needs to run again, new result:
- first one gets imported fine
- second one gets imported fine
- third one gets imported fine
- fourth one gets imported fine
- fifth one gets imported with default settings (wrong)

Plugin compatibility needs to run again, it sees that all plugins do have the right import settings, so it can emit a happy beep after the final reimport.
- first one gets imported fine
- second one gets imported fine
- third one gets imported fine
- fourth one gets imported fine
- fifth one gets imported fine

Let's hope this is not necessary...

firtina-improbable: Also have we played with different parameters to `Refresh`, e.g. `AssetDatabase.Refresh(ImportAssetOptions.ForceSynchronousImport | ImportAssetOptions.ForceUpdate | ImportAssetOptions.ImportRecursive)` ? It's possible that the default options may not have created the correct settings in time for the call to the compatibility fixer

gkassabli: Thanks for the detailed research, @firtina-improbable ! Let me try and refactor the diff using your suggestion.
Re AssetDatabase.Refresh - yes, this was the first approach I've tried, unfortunately even with all options you've mentioned, AssetDatabase didn't seem to be in the complete state after the import, possibly because of the issue you've described

gkassabli: Chatted offline about this - for the system to work properly we'll need to set up communication between downloading module and plugin processor. While this is something we'll need to add eventually if we are to download plugins depending on a platform, it is probably overkill for a simple issue like this one.

workers/unity/Packages/com.improbable.gdk.tools/DownloadCoreSdk.cs
```diff
@@ -29,8 +49,7 @@ internal static class DownloadCoreSdk
         {
             PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/OSX", new List<BuildTarget> { BuildTarget.StandaloneOSX }, editorCompatible: true),
             PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Linux", new List<BuildTarget> { BuildTarget.StandaloneLinux64 }, editorCompatible: true),
-            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Windows/x86_64", new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, editorCompatible: true),
-            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Windows/x86", new List<BuildTarget> { BuildTarget.StandaloneWindows }, false),
+            PluginDirectoryCompatibility.CreateWithCompatiblePlatforms("Assets/Plugins/Improbable/Core/Windows", new List<BuildTarget> { BuildTarget.StandaloneWindows64 }, editorCompatible: true),
```
gkassabli: x86 directory was throwing warnings, was not consistent with Linux one and we don't actually have any Windows x86 plugins

jessicafalk: that is not correct. we are simply not downloading it right now, however, we do have x86 plugins that could be used

I fixed the linux one last night. please run cloud deployments and test everything before merging in PRs

gkassabli: Yeah, saw that, thanks, Jess!
Do we have a list of things worth checking before merging PR? Alternatively is there an extended CI run we can run?

In regards to x86 - we don't have it in the list of downloads for now, is there any other way we can extract it?
Alternatively, if we want to cover the architectures that we might potentially download are there any other architectures that we should add?

jessicafalk: no, unfortunately we haven' set up CI for that yet. You will have to manually start a local and a cloud deployment and test all platforms that way

x86: you can only download the via spatial package retrieve, I believe. I think these are all the architectures that we need to test. We only support linux for x64, windows is the only one with x86 and 64

PR #453, opened 2018-10-07
using our special singleton for build configurations

jessicafalk:

-------

#### Description
Using Unity's `ScriptableSingleton` was not a good idea. lazy-loading the asset sometimes doesn't work.

#### Tests
tested locally and in cloud. 
tested building a worker without clicking on an asset before

#### Documentation
none needed

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.buildsystem/Configuration/SingletonScriptableObject.cs
```diff
@@ -0,0 +1,119 @@
...
+
+namespace Improbable.Gdk.BuildSystem.Configuration
+{
+    public abstract class SingletonScriptableObject<TSelf> : ScriptableObject
+        where TSelf : SingletonScriptableObject<TSelf>
+    {
+        private static readonly List<TSelf> Instances =
+            new List<TSelf>();
+        
+        private static readonly HashSet<Type> LoadingInstances = new HashSet<Type>();
```
jessicafalk: @firtina-improbable  moved this from the loader class into here. seems to work just fine. was there a reason to have it in a separate class?

firtina-improbable: I think previously the reason was this:

Each "non-generic implementation" of this generic class would have its own copy of `LoadingInstances` rather than "one hashset to contain them all". For example `SingletonScriptableObject<Foo>.LoadingInstances != SingletonScriptableObject<Bar>.LoadingInstances`.

If it is declared within the implementation, it'd effectively be the same thing as `static bool IsLoading` - since `LoadingInstances` in each version will have only one key now: `typeof(TSelf)`. See experiment:  https://ideone.com/kXVsAi

But I think the `IsLoading` was a bit scary to use because it wasn't very apparent that it'd be defined "per non-generic implementation". e.g. https://stackoverflow.com/questions/9647641/resharper-warns-static-field-in-generic-type

workers/unity/Packages/com.improbable.gdk.buildsystem/Configuration/SpatialOSBuildConfiguration.cs
```diff
@@ -43,7 +43,7 @@ internal void UpdateEditorScenesForBuild()
                     .ToArray();
         }
 
-        private void OnEnable()
+        public override void OnEnable()
         {
```
firtina-improbable: Call base class here

workers/unity/Packages/com.improbable.gdk.buildsystem/Configuration/SingletonScriptableObject.cs
jessicafalk: aahhh, that makes sense. ok, will add the other class back in.

PR #455, opened 2018-10-08
UTY-1306: Game Client Crash Fix

jamiebrynes7:
#### Description
If a client quits and Unity starts destroying objects - there is no guarantee the order in which this happens.

So you can get into the situation where the GameObject has been destroyed but the linking system has not, meaning when OnDestroyManager occurs - it will throw null refs.

Misc changes due to code cleanup
#### Tests
Can test when the `pinned/starter-project` branch is ff'ed.
#### Documentation
N/A
#### Primary reviewers
@mattyoung-improbable

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
@@ -30,11 +30,11 @@ namespace Improbable.Gdk.GameObjectRepresentation
         ///     Links a GameObject to an ECS Entity.
         /// </summary>
         /// <remarks>
-        ///     All <see cref="UnityEngine.Component"/>s on the GameObject will be inserted onto the ECS Entity. Note
+        ///     All <see cref="UnityEngine.Component" />s on the GameObject will be inserted onto the ECS Entity. Note
         ///     that children and parent components are not added.
         /// </remarks>
         /// <remarks>
-        ///     A <see cref="SpatialOSComponent"/> will be added to the GameObject.
+        ///     A <see cref="SpatialOSComponent" /> will be added to the GameObject.
         /// </remarks>
         /// <remarks>
         ///     If a Monobehaviour exists multiple times on the GameObject, only the first occurence is added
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/EntityGameObjectLinker.cs
```diff
@@ -102,16 +102,22 @@ public void LinkGameObjectToEntity(GameObject gameObject, Entity entity, ViewCom
...
         ///     An instance of the ViewCommandBuffer. Must be flushed to apply the changes to the ECS Entity.
         /// </param>
-        public void UnlinkGameObjectFromEntity(GameObject gameObject, Entity entity, ViewCommandBuffer viewCommandBuffer)
+        public void UnlinkGameObjectFromEntity(GameObject gameObject, Entity entity,
+            ViewCommandBuffer viewCommandBuffer)
         {
-            if (entityManager.Exists(entity))
+            if (!entityManager.Exists(entity))
+            {
+                return;
```
JonasImprobable: The approach in this PR feels more like a hack. This if check here for instance feels like a place where we'd throw errors if we wanted to do things "properly". I'm ok with merging this PR if the idea is to quickly patch things up for the short term. Ideally, we should ensure that all gameobjects are cleaned up before unity's automatic scene cleanup behaviour kicks in. This was achieved in the SDK by cleaning up all gameobjects in an OnApplicationQuit method attached to the bootstrap gameobject representing the worker connection. That method was fired before unity's own scene cleanup.

jamiebrynes7: Yeah this is a shorter term fix for a crash on close in the ECS-FPS client

JonasImprobable: Ok, that makes sense. Can you create a ticket for cleaning this up properly?

jamiebrynes7: Yes will do

PR #472, opened 2018-10-09
Fix improbadocs issues for first part of ecs content

firtina-improbable:


docs/content/ecs/accessing-worker-info.md
```diff
@@ -2,7 +2,7 @@
```
jamiebrynes7: We have two headings in a row - we can probably drop the top one and promote the others

PR #507, opened 2018-10-10
Docs: Add ToCs

tenevdev:
#### Description
ToCs would help navigating the glossary a lot. They should be useful for the tutorial and getting started, too.

#### Tests
Local preview

Health pickup tutorial looks like this:
image

Glossary looks like this:
image

### DIFF
--- a/docs/content/glossary.md
+++ b/docs/content/glossary.md
@@ -4,8 +4,6 @@
 
 # Glossary
 
-<%(TOC)%>
-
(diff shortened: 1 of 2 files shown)


jamiebrynes7:
The TOC on the glossary looks a little funky

jamiebrynes7:
image

tenevdev:
I think it will look nicer when we add some structure to it. I discussed having alphabetical and semantic orderings with @ElleEss. For reference check out https://docs.improbable.io/reference/13.3/shared/glossary

jamiebrynes7:
Yeah if we do that sort of thing it'll be much nicer 😄

firtina-improbable:
OK, I'll rebase and get fresh screenshots

firtina-improbable:
Fresh screenshots are now in the PR description

For glossary, happy to join categorization discussions, waiting for responses from @ernopp and @callumlawson

ernopp:
Sorry for the slow response. I think that since the health pick up tutorial is so long it is worth having a ToC for it so people can more easily navigate it and know where they are with it

PR #509, opened 2018-10-10
Replace usages of LogEvent default constructor

tenevdev:
#### Description
The default constructor doesn't initialise the Data field because this is a struct. This may cause NPEs. Calling the non-default constructor is a temporary solution.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
@@ -271,7 +271,7 @@ namespace Improbable.Gdk.GameObjectRepresentation
             catch (Exception e)
             {
                 // Log the exception but do not rethrow it.
-                logger.HandleLog(LogType.Exception, new LogEvent("Caught exception").WithException(e));
+                logger.HandleLog(LogType.Exception, new LogEvent("Caught exception in MonoBehaviour").WithException(e));
             }
         }
 
--- a/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
@@ -271,7 +271,7 @@ namespace Improbable.Gdk.GameObjectRepresentation
             catch (Exception e)
             {
                 // Log the exception but do not rethrow it.
-                logger.HandleLog(LogType.Exception, new LogEvent().WithException(e));
+                logger.HandleLog(LogType.Exception, new LogEvent("Caught exception").WithException(e));
             }
         }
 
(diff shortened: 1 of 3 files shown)


tenevdev:
As stated in the description - this is a temporary solution. I'd like to raise another PR which makes it impossible to make this mistake again. However, this will require structural changes to how we log events. The performance gained having log event be a struct is unclear. Let me know if anyone has data supporting this design decision.

jamiebrynes7:
We are planning a refactor to logging soon ™️ so we can add this onto the list of things to revisit

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -271,7 +271,7 @@ private void RunWithExceptionHandling(Action action)
             catch (Exception e)
             {
                 // Log the exception but do not rethrow it.
-                logger.HandleLog(LogType.Exception, new LogEvent().WithException(e));
+                logger.HandleLog(LogType.Exception, new LogEvent("Caught exception").WithException(e));
```
jamiebrynes7: `Caught exception while executing an action` 

If we know these are only used in callbacks in the Monobehaviour workflow we could even be more explicit?

tenevdev: I just didn't want to make any assumptions about what the caller is doing. Also "while executing action" seemed redundant, the user will see a stack trace anyway.

jamiebrynes7: The content of the action could throw, but these actions are all callbacks registered against a Reader/Writer right now I think

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/MonoBehaviourActivationManager.cs
```diff
@@ -271,7 +271,7 @@ private void RunWithExceptionHandling(Action action)
             catch (Exception e)
             {
                 // Log the exception but do not rethrow it.
-                logger.HandleLog(LogType.Exception, new LogEvent("Caught exception").WithException(e));
+                logger.HandleLog(LogType.Exception, new LogEvent("Caught exception in MonoBehaviour").WithException(e));
```
jamiebrynes7: `in a MonoBehaviour`

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/ReadersWriters/GameObjectDelegates.cs
```diff
@@ -35,7 +35,7 @@ public static class GameObjectDelegates
                 catch (Exception e)
                 {
                     // Log the exception but do not rethrow it, as other delegates should still get called
-                    logDispatcher.HandleLog(LogType.Exception, new LogEvent().WithException(e));
+                    logDispatcher.HandleLog(LogType.Exception, new LogEvent("Caught exception in a dispatcher callback").WithException(e));
```
samiwh: dispatcher callback doesn't tell anything to a user

PR #520, opened 2018-10-11
Mobile local dev: Adding mobile assemblies and DeviceInfo logic

gkassabli:

-------

#### Description
This PR adds Mobile assemblies Improbable.Gdk.Mobile, Improbable.Gdk.Mobile.Android and Improbable.Gdk.Mobile.iOS as well as logic to detect Android and iOS simulator
#### Tests
Tested along with other PRs on iOS and Android devices and simulators
#### Primary reviewers
cc @DanailPenev


workers/unity/Packages/com.improbable.gdk.mobile/iOS/Utility/DeviceInfo.cs
```diff
@@ -0,0 +1,12 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile.Ios
+{
+    public class DeviceInfo
+    {
+        public static bool IsIosSimulator()
+        {
+            return SystemInfo.deviceModel.Equals("x86_64");
```
jessicafalk: is this enough to reliably detect an ios emulator?

gkassabli: Yes, I think this is taken from UIDevice model value, which is always x86_84 for simulators. All actual devices are named "iPhone.." etc.

jessicafalk: @DanailPenev are we sure that this won't overlap with the android emulator values by accident?

jessicafalk: just want to check whether we need additional checks somewhere to clearly identify which x86_64 it is

DanailPenev: I mean it is code that should only ever be run on iOS devices.

gkassabli: Oh, sorry, I misunderstood the question, was only thinking that we shouldn't mix it with iPhone device.
We can add `#if UNITY_IOS` to be extra safe

jessicafalk: never assume anything :P there will be people trying to do it

jessicafalk: though I guess that this code will be in a package that is only available in editor and ios,right?

gkassabli: Thinking about it, someone could have `isiOSSimulator` and `isAndroidSimulator` in the same method, so I think it would be better to be safer to have a directive to compile this out

DanailPenev: The code won't compile for Android because the iOS assembly is only available for iOS and Editor.

jessicafalk: yeah, so it should be fine

gkassabli: Fair enough, let's keep it as is in that case

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/DeviceInfo.cs
```diff
@@ -0,0 +1,30 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile.Android
+{
+    public class DeviceInfo
```
jamiebrynes7: This can be `static`

workers/unity/Packages/com.improbable.gdk.mobile/iOS/Utility/DeviceInfo.cs
```diff
@@ -0,0 +1,12 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile.Ios
+{
+    public class DeviceInfo
+    {
+        public static bool IsIosSimulator()
```
firtina-improbable: Wondering if it may be better to have a common abstract base class that can be extended per different platforms? It's not really clear how the `Is*Emulator/Simulator` API will be used though, so this could probably be refactored in beta phase.

gkassabli: Agree. It is hard to say what would be preferable for developers, but the current structure of having core Mobile assembly + platform specific iOS and Android assemblies should allow us to refactor classes the way we need when we understand the requirements better

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/DeviceInfo.cs
samiwh: Would also mildly rather change the name of the type than just the namespace. It's not immediately obvious in code which `DeviceInfo` you are looking at.

workers/unity/Packages/com.improbable.gdk.mobile/iOS/Utility/DeviceInfo.cs
samiwh: That's only fine if it doesn't compile with a comprehensible error message. Otherwise someone will spend ages banging their head against the wall of why them accessing the wrong identically named object doesn't compile.

gkassabli: Good point, let me rename those to AndroidDeviceInfo and iOSDeviceInfo to address that

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
```diff
@@ -0,0 +1,30 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile.Android
+{
+    public static class AndroidDeviceInfo
+    {
+        public const string AndroidStudioEmulatorDefaultCallbackIp = "10.0.2.2";
+
+        private static readonly AndroidJavaObject Build = new AndroidJavaObject("android.os.Build");
```
firtina-improbable: I have a few issues with these:
- should this object be disposed?
- it feels a bit dangerous to have these be statically evaluated (it'd be evaluated probably when you call IsAndroidStudioEmulator the first time)

If the values will never change and will never be used in other methods, I recommend to:
- remove the GetValue method
- move these into the method, which would:
    - get the java object
    - get the values
    - do the check
    - save into a bool
    - dispose of the java object
    - then reuse the bool when it may be queried again

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+        private static readonly string Fingerprint = GetValue("FINGERPRINT");
+        private static readonly string Model = GetValue("MODEL");
+        private static readonly string Brand = GetValue("BRAND");
+        private static readonly string Device = GetValue("DEVICE");
+        private static readonly string Product = GetValue("PRODUCT");
+
+        public static bool IsAndroidStudioEmulator()
+        {
+            return Fingerprint.StartsWith("generic") || Fingerprint.StartsWith("unknown") ||
```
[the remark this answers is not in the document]
gkassabli: Are stackOverflow links allowed? :)

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
gkassabli: This looks like a reasonable change to me in any case. I'll add that.

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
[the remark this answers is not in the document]
firtina-improbable: good question - cc @jared-improbable thoughts?

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
```diff
@@ -0,0 +1,39 @@
...
+        public const string AndroidStudioEmulatorDefaultCallbackIp = "10.0.2.2";
+
+        private static bool isInitialised = false;
+        private static bool isEmulator = false;
+
+        public static bool IsAndroidStudioEmulator()
+        {
+            if (!isInitialised)
+            {
+                AndroidJavaObject build = new AndroidJavaObject("android.os.Build");
```
firtina-improbable: Is it safe to dispose of this object? It's a disposable so I'm guessing it should be disposed

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
```diff
@@ -0,0 +1,39 @@
...
+                    product.Equals("google_sdk");
+                isInitialised = true;
+            }
+
+            return isEmulator;
+        }
+
+        private static string GetValue(AndroidJavaObject build, string parameter)
+        {
+            return build.GetStatic<string>(parameter);
```
jessicafalk: seeing as we only use the method here, I guess we can inline this? there is no real benefit of having this method

workers/unity/Packages/com.improbable.gdk.mobile/iOS/Utility/iOSDeviceInfo.cs
```diff
@@ -0,0 +1,12 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile.Ios
+{
+    public static class iOSDeviceInfo
+    {
+        public static bool IsiOSSimulator()
```
jessicafalk: `isSimulator` is probably fine? has the class is already called `iOSDeviceInfo` and it makes the method easier to read

workers/unity/Packages/com.improbable.gdk.mobile/iOS/Utility/iOSDeviceInfo.cs
```diff
@@ -0,0 +1,12 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile.Ios
```
jessicafalk: still has the old naming

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
jessicafalk: could just use `using` to automatically dispose of it?

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
```diff
@@ -0,0 +1,39 @@
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile.Android
+{
+    public static class AndroidDeviceInfo
+    {
+        public const string AndroidStudioEmulatorDefaultCallbackIp = "10.0.2.2";
+
+        private static bool isInitialised = false;
+        private static bool isEmulator = false;
```
jessicafalk: could be changed to an enum:
```
enum DeviceType = {
  UNKNOWN,
  PHONE,
  EMULATOR
}
```

that makes it easier to later add detection whether it's a phone or a tablet. Depending on that you often want to change how to display things.

jessicafalk: actually, I really want this to be changed to an enum :)

PR #521, opened 2018-10-11
Mobile local dev: Adding mobile connection workers

gkassabli:

-------

#### Description
This PR adds MobileWorkerConnector to be used as a base for mobile Connectors. It also integrates it into Playground along with necessary logic
#### Tests
Tested along with other PRs in `mobile_merge` on iOS and Android devices and simulators
#### Primary reviewers
cc @DanailPenev


workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
+﻿using Improbable.Gdk.Core;
+using Improbable.Gdk.Mobile;
+using Playground;
+using UnityEngine;
+using UnityEngine.UI;
+
+public abstract class AbstractMobileClientWorker : MobileWorkerConnector
```
jamiebrynes7: Nit: Would prefer `Base` to `Abstract` as a prefix

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    public GameObject ConnectionPanel;
```
jamiebrynes7: I think our style is `[SerializeField] private` for these sorts of things

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+    public GameObject Level;
+
+    private GameObject levelInstance;
+    private bool connected;
+    private InputField input;
+    private Button button;
```
jamiebrynes7: A slightly more descriptive names for these could be helpful - e.g `connectButton`, `ipAddrInput`

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+    private Text errorMessage;
+
+    public void Awake()
+    {
+        input = ConnectionPanel.transform.Find("ConnectInput").GetComponent<InputField>();
+        button = ConnectionPanel.transform.Find("ConnectButton").GetComponent<Button>();
+        errorMessage = ConnectionPanel.transform.Find("ConnectionError").GetComponent<Text>();
+
+        input.text = PlayerPrefs.GetString("cachedIp");
```
jamiebrynes7: Why is this not done in the `ConnectionPanel` game object? (which owns this input field)

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+            return;
+        }
+
+        levelInstance = Instantiate(Level, transform);
+        levelInstance.transform.SetParent(null);
+
+        connected = true;
+        ConnectionPanel.SetActive(false);
+
+        PlayerPrefs.SetString("cachedIp", input.text);
```
jamiebrynes7: Magic string (`cachedIp`)

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+        PlayerPrefs.Save();
+    }
+
+    protected override void HandleWorkerConnectionFailure()
+    {
+        errorMessage.text = "Connection failure. Please check the IP address";
+    }
+
+    protected string GetIpFromField()
```
jamiebrynes7: This can be made a property:


```
protected string IpAddress => input?.text;
```

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
```diff
@@ -0,0 +1,26 @@
...
+#if UNITY_ANDROID
+            var hostIp = GetIpFromField();
+            if (Application.isMobilePlatform && DeviceInfo.IsAndroidStudioEmulator() && hostIp.Equals(string.Empty))
+            {
+                return DeviceInfo.AndroidStudioEmulatorDefaultCallbackIp;
+            }
+
+            return hostIp;
+#else
+            throw new NotImplementedException("Incompatible platform: Please use Android");
```
jamiebrynes7: I'm not sure `NotImplementedException` is the correct one here

jamiebrynes7: Similar comment in `IosClientWorkerConnector`

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
```diff
@@ -0,0 +1,20 @@
+using Improbable.Gdk.Core;
+
+namespace Improbable.Gdk.Mobile
+{
+    public abstract class MobileWorkerConnector : WorkerConnector
```
jamiebrynes7: Wait why do we have a class tree like:

`AbstractMobileWorkerConnector` > `MobileWorkerConnector` > `WorkerConnector` ? 

Can't the mobile ones just be wrapped together into one?

DanailPenev: The idea is that `MobileWorkerConnector` is the core mobile class, part of the mobile package while `AbstractMobileWorkerConnector` is our Playground implementation and it exists because we have overlapping logic for Android and iOS.

jamiebrynes7: Hmmmm I see - I would suggest rethinking the name of the one in Playground as its super confusing at the moment

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
samiwh: Why? Abstract is better imo

jessicafalk: `Base` should be at the end of the classname imo :P

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+    private bool connected;
+    private InputField input;
+    private Button button;
+    private Text errorMessage;
+
+    public void Awake()
+    {
+        input = ConnectionPanel.transform.Find("ConnectInput").GetComponent<InputField>();
+        button = ConnectionPanel.transform.Find("ConnectButton").GetComponent<Button>();
+        errorMessage = ConnectionPanel.transform.Find("ConnectionError").GetComponent<Text>();
```
samiwh: Can't these be linked in the editor?

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    public void Start()
+    {
+        button.onClick.AddListener(Connect);
+    }
+
+    public async void Connect()
+    {
+        errorMessage.text = "";
+        await Connect(WorkerUtils.UnityClient, new ForwardingDispatcher()).ConfigureAwait(false);
```
samiwh: Please make a specific worker type

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
samiwh: +1

workers/unity/Assets/Playground/Scripts/Worker/IosClientWorkerConnector.cs
```diff
@@ -0,0 +1,27 @@
...
+#endif
+using Improbable.Gdk.Core;
+
+namespace Playground.Worker
+{
+    public class IosClientWorkerConnector : AbstractMobileClientWorker
+    {
+        protected override string GetHostIp()
+        {
+#if UNITY_IOS
```
samiwh: Does this need to be conditionally compiled away? Can it not be checked at runtime?

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
samiwh: Don't name the more derived one to sound like it's less derived.

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+    public override void Dispose()
+    {
+        if (levelInstance != null)
+        {
+            Destroy(levelInstance);
+        }
+
+        if (connected)
+        {
+            Destroy(this);
```
jessicafalk: this looks like we call destroy to early.

remove all of that except for the levelInstance part and just call `base.Dispose();`, see https://github.com/spatialos/gdk-for-unity/blob/a8e7bb33e8042062bae3cbe9c16cfea5472fd788/workers/unity/Assets/Playground/Scripts/Worker/ClientWorkerConnector.cs how it should look like

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
jessicafalk: yeah, that looks like the wrong exception

jessicafalk: and it also doesn't give much information. how do I use android?

workers/unity/Assets/Playground/Scripts/Worker/IosClientWorkerConnector.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using UnityEngine;
+#if UNITY_IOS
+using Improbable.Gdk.Mobile.Ios;
+#endif
+using Improbable.Gdk.Core;
```
jessicafalk: order them alphabetically

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    public async void Connect()
+    {
+        errorMessage.text = "";
+        await Connect(WorkerUtils.UnityClient, new ForwardingDispatcher()).ConfigureAwait(false);
+    }
+
+    protected override void HandleWorkerConnectionEstablished()
+    {
+        WorkerUtils.AddClientSystems(Worker.World);
```
samiwh: Are we sure these should be the same systems?

workers/unity/Assets/Playground/Scripts/Worker/IosClientWorkerConnector.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Playground.Worker
```
jessicafalk: just use `Playground`

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
samiwh: I thought the controls one at least was different.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -262,7 +262,7 @@ protected virtual void HandleWorkerConnectionFailure()
                 ConnectionErrorReason.ExceededMaximumRetries);
         }
 
-        private static string CreateNewWorkerId(string workerType)
+        protected static string CreateNewWorkerId(string workerType)
```
jessicafalk: why make it protected?

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
samiwh: Destroy is a deferred call. A gameobject doesn't get destroyed until the end of the tick.

workers/unity/Assets/Playground/Scripts/Worker/IosClientWorkerConnector.cs
```diff
@@ -0,0 +1,27 @@
...
+#if UNITY_IOS
+            var hostIp = GetIpFromField();
+            if ((Application.isEditor || DeviceInfo.IsIosSimulator()) && hostIp.Equals(string.Empty))
+            {
+                return RuntimeConfigDefaults.ReceptionistHost;
+            }
+
+            return hostIp;
+#else
+            throw new NotImplementedException("Incompatible platform: Please use iOS");
```
samiwh: Same as the above.

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
DanailPenev: Controls are not in these PRs. It's an MVP of connecting and being able to dev locally; controls implementation seemed too convoluted and time consuming for the time being.

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
```diff
@@ -0,0 +1,81 @@
...
+        input = ConnectionPanel.transform.Find("ConnectInput").GetComponent<InputField>();
+        button = ConnectionPanel.transform.Find("ConnectButton").GetComponent<Button>();
+        errorMessage = ConnectionPanel.transform.Find("ConnectionError").GetComponent<Text>();
+
+        input.text = PlayerPrefs.GetString("cachedIp");
+    }
+
+    public void Start()
+    {
+        button.onClick.AddListener(Connect);
```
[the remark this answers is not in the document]
DanailPenev: Good question, will move it.

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
DanailPenev: They can. Is it a better practice to link both the parent object and its children as opposed to just linking the parent and getting the children from there? I was not aware.

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
jamiebrynes7: Link to Amazon page to buy an Android phone :trollface:

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
DanailPenev: It can't be accessed from the Mobile package or the Playground when `private`.

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
jessicafalk: yeah, this doesn't make too much sense... 
as people might only use the iOS worker or the android worker instead of both, I would suggest just duplicating the code to make it easier to understand.

Additionally, all the splash screen stuff can be moved to it's own game object

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
jessicafalk: better to link as it means that Unity doesn't need to search for stuff during runtime

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
jessicafalk: but still think we should call the base for this unless there is a reason for doing it like this

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
jessicafalk: do we need to?

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
DanailPenev: The reason for doing it like this is allowing connection retry with the same GameObject i.e. to allow fixing a wrong IP, for example.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
DanailPenev: We are overriding the `GetReceptionistConfig()` where we need to give a `workerId` in the `ReceptionistConfig` object

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
samiwh: More explicit / less brittle

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
samiwh: Re-read this. Very against not calling base.Dispose(). EWe could easily change the base behaviour to do more and not know to change this. Why not make the thing that gets and IP something else and have that create a gamobject whenever it tries to make a connection?

workers/unity/Assets/Playground/Scripts/Worker/IosClientWorkerConnector.cs
jessicafalk: nope, because we allow the mobile package only in editor, ios or android builds. therefore, we need to use these preprocessors to get a succeeding build

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
jessicafalk: yeah, that's what should happen. create the worker gameobject only when necessary. reusing the current one could potentially cause issues

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
jessicafalk: that tells me how to buy an android phone, not how to use it :trollface:

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
jessicafalk: this can be removed now :P

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
jessicafalk: this also doesn't make any sense regarding the worker type defined in the worker.json. Not sure if that could cause problems

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
firtina-improbable: aye, should be a const string in some static class ideally

workers/unity/Assets/Playground/Scripts/Worker/AbstractMobileClientWorker.cs
firtina-improbable: +1, in a dispose if you don't call base.dispose if the base changes in the future it will be the result of very funky bugs that may be a pain to diagnose, if we fix the reason that makes it inconvenient to dispose the parent, it will be better

DanailPenev:
This PR is now updated with all the feedback received. We now have 2 separate worker types for Android and iOS. The weird class tree was simplified by having the same code in the `AndroidClientWorkerConnector` and the `iOSClientWorkerConnector`. Exceptions were updated to be more accurate. Finally, the splash screen logic was moved into a separate prefab that spawns the relevant worker type and disposes of it if it fails.

workers/unity/Assets/Playground/Improbable.Playground.asmdef
```diff
@@ -9,10 +9,13 @@
...
+        "Improbable.Gdk.Mobile",
+        "Improbable.Gdk.Android",
+        "Improbable.Gdk.Ios"
     ],
     "optionalUnityReferences": [],
     "includePlatforms": [],
     "excludePlatforms": [],
     "allowUnsafeCode": false
-}
+}
```
jamiebrynes7: Nit: Newline

workers/unity/Assets/Playground/Improbable.Playground.asmdef
```diff
@@ -9,10 +9,13 @@
         "Unity.Entities.Hybrid",
         "Unity.Mathematics",
         "Improbable.Gdk.Generated",
-        "Improbable.Gdk.StandardLibraryUtils"
+        "Improbable.Gdk.StandardLibraryUtils",
+        "Improbable.Gdk.Mobile",
+        "Improbable.Gdk.Android",
+        "Improbable.Gdk.Ios"
```
jamiebrynes7: I thought we were going with `iOS`?

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        private void TryConnect()
+        {
+            worker = Instantiate(clientWorkerPrefab);
+            var workerConnector = worker.GetComponent<IMobileClient>();
+
+            if (workerConnector == null)
+            {
+                return;
```
jamiebrynes7: Is this an error that should be logged/thrown?

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -0,0 +1,57 @@
...
+            }
+
+            workerConnector.IpAddress = IpAddress;
+            workerConnector.ConnectionScreenController = this;
+            workerConnector.TryConnect();
+        }
+
+        public void OnSuccess()
```
jamiebrynes7: `OnSuccess` of what?

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        [SerializeField] private GameObject connectionPanel;
+        [SerializeField] private InputField ipAddressInput;
+        [SerializeField] private Button connectButton;
+        [SerializeField] private Text errorMessage;
+        [SerializeField] private GameObject clientWorkerPrefab;
+
+        private GameObject worker;
+
+        private string IpAddress => ipAddressInput != null ? ipAddressInput.text : null;
```
firtina-improbable: `ipAddressInput?.text`

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            await Connect(WorkerUtils.AndroidClient, new ForwardingDispatcher()).ConfigureAwait(false);
+        }
+
+        protected override void HandleWorkerConnectionEstablished()
+        {
+            ConnectionScreenController.OnSuccess();
+
+            WorkerUtils.AddClientSystems(Worker.World);
+            if (level == null)
```
jamiebrynes7: Nit: Newline in between if statement and previous statement

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
```diff
@@ -0,0 +1,60 @@
...
+#if UNITY_ANDROID
+            if (Application.isMobilePlatform && DeviceInfo.IsAndroidStudioEmulator() && IpAddress.Equals(string.Empty))
+            {
+                return DeviceInfo.AndroidStudioEmulatorDefaultCallbackIp;
+            }
+
+            return IpAddress;
+#else
+            throw new PlatformNotSupportedException(
+                "This method is only defined for the Android platform. Please check your build settings.");
```
jamiebrynes7: Its not just this method - its the whole class really, right?

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Improbable.Gdk.Core;
+using Improbable.Gdk.Mobile;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace Playground
+{
+    public class ConnectionScreenController : MonoBehaviour
+    {
+        private const string CachedIp = "cachedIp";
```
firtina-improbable: This string `cachedIp` for playerprefs needs to be more specific to improbable/GDK for unity

Also the variable name should reflect that it's used for player prefs

workers/unity/Assets/Playground/Scripts/Worker/IosClientWorkerConnector.cs
```diff
@@ -0,0 +1,59 @@
...
+        public async void TryConnect()
+        {
+            await Connect(WorkerUtils.iOSClient, new ForwardingDispatcher()).ConfigureAwait(false);
+        }
+
+        protected override void HandleWorkerConnectionEstablished()
+        {
+            ConnectionScreenController.OnSuccess();
+
+            WorkerUtils.AddClientSystems(Worker.World);
```
jamiebrynes7: Nit: Whitespace

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
jamiebrynes7: Same comment in the iOS one

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -0,0 +1,57 @@
...
+        [SerializeField] private Text errorMessage;
+        [SerializeField] private GameObject clientWorkerPrefab;
+
+        private GameObject worker;
+
+        private string IpAddress => ipAddressInput != null ? ipAddressInput.text : null;
+
+        public void Awake()
+        {
+            ipAddressInput.text = PlayerPrefs.GetString(CachedIp);
```
firtina-improbable: optional: use the second parameter in GetString to return something explicitly invalid

workers/unity/Assets/Playground/Improbable.Playground.asmdef
DanailPenev: Naming will be fixed before merging in the big master PR as github is pretty shit with renaming these and reverts them back all the time. We are going with `iOS`.

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
DanailPenev: This is literally only here, because the compiler is complaining. Being able to use this method on a different platform should be impossible, because of the assembly/platform limitations. An alternative is to wrap the entire class in preprocessor directives but I think it is much uglier+ it brings other problems of the class not being detected and what-not else.

firtina-improbable: This error may be confusing to users and may not really help them, I'd try to see what's the last user action that may trigger this, e.g. "You can use an AndroidWorkerConnector only for the Android Platform".

Additionally, perhaps it'd be better to throw something within OnEnable (or probably better, TryConnect?) if this behaviour is added when not UNITY_ANDROID?

workers/unity/Assets/Playground/Scripts/Worker/IosClientWorkerConnector.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+#if UNITY_IOS
+            if ((Application.isEditor || DeviceInfo.IsIosSimulator()) && IpAddress.Equals(string.Empty))
+            {
+                return RuntimeConfigDefaults.ReceptionistHost;
+            }
+
+            return IpAddress;
+#else
+            throw new PlatformNotSupportedException(
```
firtina-improbable: See note in Android version

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
firtina-improbable: Guessing this conversation can probably be resolved

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
jamiebrynes7: This was more a nit about the comment in the exception 😛

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: Aye, or show something in the UI

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
DanailPenev: ^ This doesn't work because of Unity complaining about their underlying null checks or something like that.
```'?.' on a type deriving from 'UnityEngine.Object' bypasses the lifetime check on the underlying Unity engine object```

Comment (firtina-improbable) [reply workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs]: Ah, yes :(

File: workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs (hunk 22 above)
Comment (jessicafalk) [reply workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs]: should mirror the `OnConnectionFailed` method name. so something like `OnConnectionSucceeded`

File: workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs [hunk 30]
```diff
@@ -0,0 +1,60 @@
...
+        private GameObject levelInstance;
+
+        public async void TryConnect()
+        {
+            await Connect(WorkerUtils.AndroidClient, new ForwardingDispatcher()).ConfigureAwait(false);
+        }
+
+        protected override void HandleWorkerConnectionEstablished()
+        {
+            ConnectionScreenController.OnSuccess();
```
jessicafalk: would move that to the end of the method so that the level and systems are already loaded / ready before the screen gets disabled.

workers/unity/Assets/Playground/Scripts/Worker/IMobileClient.cs
```diff
@@ -0,0 +1,11 @@
+namespace Playground
+{
+    public interface IMobileClient
```
jessicafalk: would change that name to probably something like `MobileConnectionController`?
Not sure what would be a good name, but `IMobileClient` seems to broad

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -32,6 +32,7 @@ private void TryConnect()
 
             if (workerConnector == null)
             {
+                errorMessage.text = "Failed to create a mobile worker connector.";
```
firtina-improbable: Can improve this message: the actual problem is that the worker connector behaviour could not be found on the client worker prefab

DanailPenev: I am not sure if we should dive into that much detail in the error message, considering it's shown in the UI.

firtina-improbable: Anything that will help the developer resolve the problem. From this message by itself, what can they do to fix it? An alternative is to also log an error so they can use the stack trace from the editor.

DanailPenev: Should it be throwing an exception, rather than simply logging?

firtina-improbable: Exception sounds better.

Also destroy the worker before throwing the exception :)

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        protected override void HandleWorkerConnectionFailure()
+        {
+            ConnectionScreenController.OnConnectionFailed();
+        }
+
+        protected override string GetHostIp()
+        {
+#if UNITY_ANDROID
+            if (Application.isMobilePlatform && DeviceInfo.IsAndroidStudioEmulator() && IpAddress.Equals(string.Empty))
```
jessicafalk: `IpAddress.Equals(string.Empty)` should be `string.Empty.Equals(IpAddress)` as `IpAddress` can be `null`

workers/unity/Assets/Playground/Scripts/Worker/IosClientWorkerConnector.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace Playground
+{
+    public class iOSClientWorkerConnector : MobileWorkerConnector, IMobileConnectionController
+    {
+        public string IpAddress { get; set; }
+        public ConnectionScreenController ConnectionScreenController { get; set; }
+
+        [SerializeField] private GameObject level;
+
+        private GameObject levelInstance;
```
samiwh: Does this not need to be removed on dispose or destroy? (Or disable)

workers/unity/Assets/Playground/Scripts/Worker/IMobileConnectionController.cs
```diff
@@ -0,0 +1,11 @@
+namespace Playground
+{
+    public interface IMobileConnectionController
```
samiwh: Can this not be rolled into the mobile connector?

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace Playground
+{
+    public class AndroidClientWorkerConnector : MobileWorkerConnector, IMobileConnectionController
+    {
+        public string IpAddress { get; set; }
+        public ConnectionScreenController ConnectionScreenController { get; set; }
+
+        [SerializeField] private GameObject level;
+
+        private GameObject levelInstance;
```
samiwh: Is the platform specific worker connectors meant to be just for the playground workers or used more generally? If used more generally then non platform stuff like this can ve moved into the mobile worker connector. However then I'd possibly recommend having platform specific worker connectors (that just do the platform specific stuff) then having the specific playground version doing stuff like the level instances, as that is a specific opinion that the main connector has gone away from.

jessicafalk: not sure I'm understanding this

workers/unity/Assets/Playground/Scripts/Worker/IMobileConnectionController.cs
jessicafalk: the main issue here is that this needs to be used by both worker connectors. However, as it is playground specific we don't to move it into the `MobileWorkerConnector`.

We have the following options:
- expose this interface
- move it into the specific types and do a type check in the `ScreenController` to identify the exact type
- have a screen controller for each platform which contain the exact some logic, but would get rid of the interface and the type checks

We decided to go with the interface for this one as it made sense for this specific case

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
DanailPenev: These platform specific worker connectors are meant to be just for the playground workers.

workers/unity/Assets/Playground/Scripts/Worker/IMobileConnectionController.cs
firtina-improbable: Would it help to make the name highlight it's only for the playground? May be overkill though

jessicafalk: `IMobileConnectionControllerPlaygroundOnly` 😂

firtina-improbable: 💯

workers/unity/Assets/Playground/Scripts/Worker/IosClientWorkerConnector.cs
jessicafalk: yep, seems like that go removed by accident. will add it back in

jessicafalk: added it in.

PR #523, opened 2018-10-11
Mobile local dev: Unity project changes + IL2CPP correction

gkassabli:

-------

#### Description
This PR adds Unity project settings changes to support iOS and Android. It also corrects IL2CPP assembly linking that is used in iOS.
#### Tests
Tested along with other PRs in `mobile_merge` on iOS and Android devices and simulators
#### Primary reviewers
cc @DanailPenev

### DIFF
--- a/workers/unity/ProjectSettings/ProjectSettings.asset
+++ b/workers/unity/ProjectSettings/ProjectSettings.asset
@@ -151,7 +151,10 @@ PlayerSettings:
   androidSupportedAspectRatio: 1
   androidMaxAspectRatio: 2.1
   applicationIdentifier:
-    Android: io.improbable.demo
+    Android: com.improbable.gdk
+    Standalone: com.improbable.gdk
+    iOS: com.improbable.gdk
+    tvOS: com.improbable.gdk
   buildNumber: {}
   AndroidBundleVersionCode: 1
   AndroidMinSdkVersion: 16
--- a/workers/unity/ProjectSettings/ProjectSettings.asset
+++ b/workers/unity/ProjectSettings/ProjectSettings.asset
@@ -47,7 +47,7 @@ PlayerSettings:
   defaultScreenWidthWeb: 960
   defaultScreenHeightWeb: 600
   m_StereoRenderingPath: 0
-  m_ActiveColorSpace: 1
+  m_ActiveColorSpace: 0
   m_MTRendering: 1
   m_StackTraceTypes: 010000000100000001000000010000000100000001000000
   iosShowActivityIndicatorOnLoading: -1
@@ -150,7 +150,8 @@ PlayerSettings:
   resolutionScalingMode: 0
   androidSupportedAspectRatio: 1
   androidMaxAspectRatio: 2.1
-  applicationIdentifier: {}
+  applicationIdentifier:
+    Android: io.improbable.demo
   buildNumber: {}
   AndroidBundleVersionCode: 1
   AndroidMinSdkVersion: 16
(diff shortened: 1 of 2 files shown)


workers/unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -546,8 +639,8 @@ PlayerSettings:
   webGLLinkerTarget: 0
   scriptingDefineSymbols:
     1: UNITY_POST_PROCESSING_STACK_V2;UNITY_DISABLE_AUTOMATIC_SYSTEM_BOOTSTRAP
-    4: UNITY_POST_PROCESSING_STACK_V2
-    7: UNITY_POST_PROCESSING_STACK_V2
+    4: UNITY_POST_PROCESSING_STACK_V2;UNITY_DISABLE_AUTOMATIC_SYSTEM_BOOTSTRAP
+    7: UNITY_POST_PROCESSING_STACK_V2;UNITY_DISABLE_AUTOMATIC_SYSTEM_BOOTSTRAP
```
samiwh: Why is this added back in

gkassabli: I think this was accidental, I'll revert this

workers/unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -546,8 +639,8 @@ PlayerSettings:
   webGLLinkerTarget: 0
   scriptingDefineSymbols:
     1: UNITY_POST_PROCESSING_STACK_V2;UNITY_DISABLE_AUTOMATIC_SYSTEM_BOOTSTRAP
-    4: UNITY_POST_PROCESSING_STACK_V2
-    7: UNITY_POST_PROCESSING_STACK_V2
+    4: UNITY_POST_PROCESSING_STACK_V2;
```
jessicafalk: this looks like it needs the `UNITY_DISABLE_AUTOMATIC_SYSTEM_BOOTSTRAP` symbol

workers/unity/ProjectSettings/ProjectSettings.asset
DanailPenev: @samiwh what do you mean back in? It's still present in master for PC. This just adds the symbol for Android and iOS.

jessicafalk: why did this got removed?

jessicafalk: agreeing with @DanailPenev  here, we need these symbols for mobile support.

samiwh: Out of interest what breaks then if you don't enable the automatic bootstrap on mobile?

jessicafalk: none of the systems in the playground are annotated with `DisableAutoCreation`. You might remember the discussion we had back then when deciding whether to have a default world or not. Those will be added to the default world. There are a few systems in the playground that also use the WorkerSystem. Those will throw errors without that symbol

jessicafalk: in general: Either we have the automatic bootstrap enabled for all platforms or for none. Having it only for a few just seems wrong and could introduce issues without us realizing it

workers/unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -47,7 +47,7 @@ PlayerSettings:
   defaultScreenWidthWeb: 960
   defaultScreenHeightWeb: 600
   m_StereoRenderingPath: 0
-  m_ActiveColorSpace: 1
```
firtina-improbable: Is this intentional?

gkassabli: Yes, without this change either iOS Sim or iOS device (don't remember which exactly) doesn't work correctly.

firtina-improbable: 👍

workers/unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -150,7 +150,8 @@ PlayerSettings:
   resolutionScalingMode: 0
   androidSupportedAspectRatio: 1
   androidMaxAspectRatio: 2.1
-  applicationIdentifier: {}
+  applicationIdentifier:
+    Android: com.improbable.gdk
```
jessicafalk: while you are on that. could you also fix that for the other platforms? I believe the standalone and ios player settings use the default one `com.Company.ProductName`....

should really change that 😂

gkassabli: (y)

PR #527, opened 2018-10-11
Migrate CI to the Shared Repo

jamiebrynes7:
#### Description
Migrated CI to use the shared repo. 
#### Tests
Tested locally, the CI pass will reveal whether it really worked.
#### Documentation
N/A
#### Primary reviewers

### Note that the empty ci/prepare-unity.sh is because the build step runs this explicitly. Once this migration is merged - I can remove the build step and this empty script


ci/bootstrap.sh
```diff
@@ -1,4 +1,15 @@
 #!/usr/bin/env bash
```
jared-improbable: Make sure to `chmod +x` on all of these shell scripts

jessicafalk: yeah, please do that

PR #529, opened 2018-10-11
WIP remove component data arrays in the clean system

samiwh:
Read in commit order

Need to sort out the casing on generated variable names

#### Description
Need to kill them data arrays
Also hoping to see this scale better with archetype count increasing. Next we get anything out of jobifying it.
#### Tests
Tests and playground all still run
#### Documentation
Not user facing


zeroZshadow:
I would invert how the check for components are currently done.

Instead of looping for all entities once, and checking per iteration what to remove, check once per chunk, and iterate per component found.

PR #533, opened 2018-10-12
Removed the component data arrays from the world command send system

samiwh:
#### Description
Archetype chunk queries in the world command send system
Also made it one query rather than 4
#### Tests
All tests and playground still pass


workers/unity/Packages/com.improbable.gdk.core/Systems/WorldCommandsSendSystem.cs
```diff
@@ -75,67 +57,81 @@ protected override void OnUpdate()
...
-                foreach (var req in sender.RequestsToSend)
+                var entityArray = chunk.GetNativeArray(entityType);
+
+                var createEntitySenders = chunk.GetNativeArray(createEntityType);
+                for (int i = 0; i < createEntitySenders.Length; ++i)
                 {
-                    var reqId = connection.SendDeleteEntityRequest(req.EntityId, req.TimeoutMillis);
-                    deleteEntityStorage.CommandRequestsInFlight.Add(reqId.Id,
-                        new CommandRequestStore<WorldCommands.DeleteEntity.Request>(entity, req, req.Context, req.RequestId));
+                    var sender = createEntitySenders[i];
```
zeroZshadow: since the sender isnt used directly, it would be better to make this `RequestsToSend` directly, to avoid the second lookup for the Clear

workers/unity/Packages/com.improbable.gdk.core/Systems/WorldCommandsSendSystem.cs
```diff
@@ -75,67 +57,81 @@ protected override void OnUpdate()
...
+                    var entity = entityArray[i];
+                    foreach (var req in sender.RequestsToSend)
+                    {
+                        var reqId = connection.SendCreateEntityRequest(req.Entity, req.EntityId, req.TimeoutMillis);
+                        createEntityStorage.CommandRequestsInFlight.Add(reqId.Id,
+                            new CommandRequestStore<WorldCommands.CreateEntity.Request>(entity, req, req.Context, req.RequestId));
+                    }
+
+                    sender.RequestsToSend.Clear();
```
zeroZshadow: I'm not sure what the overhead is for calling `Clear` on an already empty list

workers/unity/Packages/com.improbable.gdk.core/Systems/WorldCommandsSendSystem.cs
```diff
@@ -75,67 +57,81 @@ protected override void OnUpdate()
...
 
-                sender.RequestsToSend.Clear();
-            }
+            var createEntityType = GetArchetypeChunkComponentType<WorldCommands.CreateEntity.CommandSender>();
+            var deleteEntityType = GetArchetypeChunkComponentType<WorldCommands.DeleteEntity.CommandSender>();
+            var reserveEntityIdsType = GetArchetypeChunkComponentType<WorldCommands.ReserveEntityIds.CommandSender>();
+            var entityQueryType = GetArchetypeChunkComponentType<WorldCommands.EntityQuery.CommandSender>();
 
-            for (var i = 0; i < deleteEntitySenderData.Length; i++)
+            var chunkArray = group.CreateArchetypeChunkArray(Allocator.Temp);
```
zeroZshadow: Would a using block be safer here? so `Dispose` happens even if something breaks? (To avoid crashing)

samiwh: IMO it would be less safe. Using will create a copy, so with unsafe code that you didn't write it's something to avoid.

workers/unity/Packages/com.improbable.gdk.core/Systems/WorldCommandsSendSystem.cs
samiwh: This isn't a change from the previous system. So no worse than it previously was. It also doesn't appear in the profiler so I'm going to leave this for now.

workers/unity/Packages/com.improbable.gdk.core/Systems/WorldCommandsSendSystem.cs
samiwh: Yep, will change.

workers/unity/Packages/com.improbable.gdk.core/package.json
```diff
@@ -7,7 +7,7 @@
     "description": "SpatialOS GDK Core Module. Made by Improbable.",
     "dependencies": {
         "com.improbable.gdk.tools": "0.1.0",
-        "com.unity.entities": "0.0.12-preview.13",
+        "com.unity.entities": "0.0.12-preview.17",
```
jamiebrynes7: Lil' driveby here

workers/unity/Packages/com.improbable.gdk.core/Systems/WorldCommandsSendSystem.cs
zeroZshadow: This wasnt changed because it was forgotten. This change was made in the normal command senders however
https://github.com/spatialos/gdk-for-unity/blob/master/workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentConversionGenerator.tt#L619

PR #543, opened 2018-10-15
Force Position to Replicate Last

jamiebrynes7:
#### Description
This PR forces the position component to be replicated last. Otherwise, the position change could trigger a load balancing operation that would cause valid update ops to be dropped.
#### Tests
Verified that position was hit last in the loop.
#### Documentation
N/A - Should this be documented? 
#### Primary reviewers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSSendSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSSendSystem.cs
@@ -17,6 +17,9 @@ namespace Improbable.Gdk.Core
     [UpdateInGroup(typeof(SpatialOSSendGroup.InternalSpatialOSSendGroup))]
     public class SpatialOSSendSystem : ComponentSystem
     {
+        // Can't access the generated component ID in Core code.
+        private const uint PositionComponentId = 54;
+
         private Connection connection;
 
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSSendSystem.cs
```diff
@@ -89,6 +92,16 @@ private void PopulateDefaultComponentReplicators()
 
                 AddComponentReplicator(componentReplicationHandler);
             }
+
+            // Force the position component to be replicated last to delay authority changes until this frame's updates
```
samiwh: I think this sentence can just say what it is doing: moving position to last.

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSSendSystem.cs
```diff
@@ -89,6 +92,16 @@ private void PopulateDefaultComponentReplicators()
...
+            // are applied as expected. If the position update is applied earlier and triggers an authority change -
```
samiwh: doesn't need a dash. Maybe "A position update can trigger an authority change, which could cause subsequent updates to be dropped".

PR #544, opened 2018-10-15
Adding scripts for iOS CI

gkassabli:

-------

#### Description
This PR adds scripts that are going to be ran for building iOS client on CI agent
#### Tests
Run the script locally > ensure build succeeds.
Run the script on CI > ensure it succeeds and is stable
#### Primary reviewers
cc @DanailPenev


ci/build-test-ios.sh
```diff
@@ -0,0 +1,13 @@
...
+set -e -u -o -x pipefail
+
+cd "$(dirname "$0")/../"
+
+# Get shared CI and prepare unity
+ci/bootstrap.sh
+.shared-ci/scripts/prepare-unity.sh
+
+ci/test.sh
+ci/build-ios_client.sh
```
gkassabli: Not sure if copying the script is best solution. Considering that we will have Android client build as well, converting this to parameterised script looks slightly better idea to me.
What do people think?

jamiebrynes7: I'd like to add a generic one to https://improbableio.atlassian.net/browse/UTY-986 but haven't got around to it 🤷‍♂️

ci/build-test-ios.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -e -u -o -x pipefail
+
+cd "$(dirname "$0")/../"
+
+# Get shared CI and prepare unity
+ci/bootstrap.sh
+.shared-ci/scripts/prepare-unity.sh
+
+ci/test.sh
```
jamiebrynes7: You don't need to call test. This script can just be `build-ios.sh`

ci/build-test-ios.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
jessicafalk: don't think we need this script. we can just call the build ios client script directly in TeamCity.

ci/build-ios_client.sh
```diff
@@ -0,0 +1,33 @@
...
+spatial build build-config IosClient
+
+pushd "${UNITY_PROJECT_DIR}"
+    dotnet run -p ../../.shared-ci/tools/RunUnity/RunUnity.csproj -- \
+        -projectPath "${UNITY_PROJECT_DIR}" \
+        -batchmode \
+        -quit \
+        -logfile "$(pwd)/../../logs/iOSClientBuild.log" \
+        -executeMethod "Improbable.Gdk.BuildSystem.WorkerBuilder.Build" \
+        +buildWorkerTypes "IosClient" \
```
jessicafalk: change the name from "IosClient" to "iOSClient"

ci/build-ios_client.sh
```diff
@@ -0,0 +1,33 @@
...
+UNITY_PROJECT_DIR="$(pwd)/workers/unity"
+
+markStartOfBlock "$0"
+
+markStartOfBlock "Building iOSClient"
+
+TARGET=${1:-local}
+
+# Back-compat: we need to run this to generate the bridge configurations.
+spatial build build-config IosClient
```
jessicafalk: do we need to run this for the build step? this is only needed when running a deployment

jessicafalk: ah just saw, that we also did this in the other scripts. not sure why though

ci/build-ios_client.sh
```diff
@@ -0,0 +1,33 @@
...
+
+pushd "${UNITY_PROJECT_DIR}"
+    dotnet run -p ../../.shared-ci/tools/RunUnity/RunUnity.csproj -- \
```
jessicafalk: just out of curiosity: what does the output of building it via the build system look like?

ci/build-ios_client.sh
jamiebrynes7: I think this was before we built the worker configs during the build (actually does this happen from the CLI? If not we can add it there and get rid of these) 

The idea being the build script would build everything required to run a worker

ci/build-ios_client.sh
jamiebrynes7: Its piped into the log file - but it looks like a normal Unity build tbh

ci/build-test-ios.sh
jamiebrynes7: See spatialos/gdk-for-unity-shared-ci#2 for this generic build script.

Also no idea why that JIRA link was copy pasted 😂

ci/build-test-ios.sh
jessicafalk: yes, that script is mainly for ci afaik. If you want to build your client, you would usually do it from the unity menu. 
so, just adding an additional line to the build-test script to build an ios client if on a mac agent should be enough

ci/build-test.sh
```diff
@@ -11,3 +11,8 @@ ci/bootstrap.sh
 ci/test.sh
 .shared-ci/scripts/build.sh "workers/unity" UnityClient local "$(pwd)/logs/UnityClientBuild.log"
 .shared-ci/scripts/build.sh "workers/unity" UnityGameLogic cloud "$(pwd)/logs/UnityGameLogicBuild.log"
+
+# Build iOS client during Mac OS CI build
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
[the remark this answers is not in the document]
gkassabli: 👍

ci/build-test.sh
```diff
@@ -8,6 +8,13 @@ cd "$(dirname "$0")/../"
 ci/bootstrap.sh
 .shared-ci/scripts/prepare-unity.sh
 
+source ".shared-ci/scripts/pinned-tools.sh"
+
 ci/test.sh
 .shared-ci/scripts/build.sh "workers/unity" UnityClient local "$(pwd)/logs/UnityClientBuild.log"
 .shared-ci/scripts/build.sh "workers/unity" UnityGameLogic cloud "$(pwd)/logs/UnityGameLogicBuild.log"
+
+# Build iOS client during Mac OS CI build
```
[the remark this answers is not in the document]
gkassabli: yeah, was left from to clarify old condition. Code is quite self-explanatory right now, I'll remove it

PR #549, opened 2018-10-17
Added Dynamic in for generic component functionality

samiwh:
#### Description
Basically the same as the C# Worker SDK Dyanmic.ForEachComponent and Dynamic.ForComponent but with a slightly different interface.

Here's a silly example of printing all transform updates on a given entity
```
using System;
using Improbable.Gdk.Core;
using Improbable.Transform;
using Improbable.Worker;
using Improbable.Worker.Core;
using Unity.Entities;
using UnityEngine;

[WorkerType("UnityClient")]
public class ForEachComponentTest : MonoBehaviour
{
    private struct Handler : Dynamic.IHandler
    {
        public World World;
        public Dispatcher Dispatcher;
        public EntityId EntityId;

        public void Accept<TData, TUpdate>(uint componentId,
            ComponentDeserializer<TData> deserializeComponentAdded,
            UpdateDeserializer<TUpdate> deserializeComponentUpdate) where TData : ISpatialComponentData
            where TUpdate : ISpatialComponentUpdate
        {
            if (componentId != TransformInternal.ComponentId)
            {
                return;
            }

            var capturedHandler = this;
            Dispatcher.OnComponentUpdate(op =>
            {
                if (op.EntityId != capturedHandler.EntityId)
                {
                    return;
                }

                var transformUpdate = deserializeComponentUpdate(op.Update, capturedHandler.World);
                PrintTransformUpdate(transformUpdate);
            });
        }
    }

    private void OnEnable()
    {
        var spatialOSComponent = GetComponent<SpatialOSComponent>();
        Dynamic.ForEachComponent(new Handler
        {
            World = spatialOSComponent.World,
            Dispatcher = spatialOSComponent.World.GetExistingManager<SpatialOSReceiveSystem>().Dispatcher,
            EntityId = spatialOSComponent.SpatialEntityId
        });
    }

    private static void PrintTransformUpdate(object update)
    {
        if (!(update is TransformInternal.Update))
        {
            return;
        }

        var c = (TransformInternal.Update) update;
        if (c.Location.HasValue)
        {
            Debug.Log(c.Location.Value.X);
        }
    }
}

```

### DIFF
--- a/test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingular.cs
+++ b/test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingular.cs
@@ -2,14 +2,18 @@
 // DO NOT EDIT - this file is automatically regenerated.
 // ===========
 
-using Unity.Entities;
 using Improbable.Gdk.Core;
+using Improbable.Worker.Core;
+using System;
 using System.Collections.Generic;
+using Unity.Entities;
 
 namespace Improbable.Gdk.Tests
 {
     public partial class ExhaustiveBlittableSingular
     {
+        public const uint ComponentId = 197720;
+
         public struct Component : IComponentData, ISpatialComponentData
         {
             public uint ComponentId => 197720;
@@ -651,6 +655,38 @@ namespace Improbable.Gdk.Tests
             public global::System.Collections.Generic.List<Update> Updates
             {
                 get => Improbable.Gdk.Tests.ExhaustiveBlittableSingular.ReferenceTypeProviders.UpdatesProvider.Get(handle);
+            }
+        }
+
+        internal class ExhaustiveBlittableSingularDynamic : IDynamicInvokable
+        {
+            public uint ComponentId => ExhaustiveBlittableSingular.ComponentId;
+
+            private static Component DeserializeData(ComponentData data, World world)
+            {
+                var schemaDataOpt = data.SchemaData;
+                if (!schemaDataOpt.HasValue)
+                {
+                    throw new ArgumentException($"Can not deserialize an empty {nameof(ComponentData)}");
+                }
+
+                return Serialization.Deserialize(schemaDataOpt.Value.GetFields(), world);
+            }
+
+            private static Update DeserializeUpdate(ComponentUpdate update, World world)
+            {
+                var schemaDataOpt = update.SchemaData;
+                if (!schemaDataOpt.HasValue)
+                {
+                    throw new ArgumentException($"Can not deserialize an empty {nameof(ComponentUpdate)}");
+                }
+
+                return Serialization.DeserializeUpdate(schemaDataOpt.Value);
+            }
+
+            public void InvokeHandler(Dynamic.IHandler handler)
+            {
+                handler.Accept<Component, Update>(ExhaustiveBlittableSingular.ComponentId, DeserializeData, DeserializeUpdate);
             }
         }
     }
--- a/test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveMapKey.cs
+++ b/test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveMapKey.cs
@@ -2,14 +2,18 @@
 // DO NOT EDIT - this file is automatically regenerated.
 // ===========
 
-using Unity.Entities;
 using Improbable.Gdk.Core;
+using Improbable.Worker.Core;
+using System;
 using System.Collections.Generic;
+using Unity.Entities;
 
 namespace Improbable.Gdk.Tests
 {
     public partial class ExhaustiveMapKey
     {
+        public const uint ComponentId = 197719;
+
         public struct Component : IComponentData, ISpatialComponentData
         {
             public uint ComponentId => 197719;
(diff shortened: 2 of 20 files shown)


workers/unity/Packages/com.improbable.gdk.core/Dynamic/Dynamic.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            if (!idsToComponents.TryGetValue(componentId, out var component))
+            {
+                throw new ArgumentException($"Unknown component ID {componentId}.");
+            }
+
+            component.DynamicInvoker(handler);
+        }
+
+        public static uint GetComponentId<T>() where T : ISpatialComponentData
```
firtina-improbable: I see this is not being used anywhere, is it preparation for the future?

firtina-improbable: nvm, I missed the point of the PR

jessicafalk:
What is the slightly different interface / what was the reason for implementing this?

samiwh:
@jessicafalk Because we are using a very different set of types and have no need or desire to introduce the framework that the normal Worker SDK has to. We could do the whole metaclass thing and we could use this foreach to create another dispatcher that uses generics but right now that seems like overkill.

workers/unity/Packages/com.improbable.gdk.core/Dynamic/Dynamic.cs
```diff
@@ -0,0 +1,69 @@
...
+                        var instance = (ISpatialComponentData) Activator.CreateInstance(type);
+                        componentsToIds.Add(type, instance.ComponentId);
+                    }
+                }
+            }
+        }
+
+        public static void ForEachComponent(IHandler handler)
+        {
+            foreach (var component in idsToComponents.Values)
```
zeroZshadow: This allocates I think, would iterating over the map directly not be better?

samiwh: Forgot this was a map; it was a list at first. Will change.

test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveMapKey.cs
```diff
@@ -1738,5 +1741,14 @@ public struct ReceivedUpdates : IComponentData
                 get => Improbable.Gdk.Tests.ExhaustiveMapKey.ReferenceTypeProviders.UpdatesProvider.Get(handle);
             }
         }
+
+        internal class ExhaustiveMapKeyDynamic : IDynamicInvokable
+        {
+            public uint ComponentId => ExhaustiveMapKey.ComponentId;
+            public Action<Dynamic.IHandler> DynamicInvoker => (handler =>
+            {
+                handler.Accept(ExhaustiveMapKey.ComponentId, Serialization.Deserialize);
```
jamiebrynes7: If we expect users to deserialise `ComponentUpdateOp`s with this, giving them just the `Deserialize` function may not be correct. You will need to give both as you can have a partial `ComponentUpdate` but not a partial `ComponentData`

samiwh: Ah yep. Forgot to go back and add that in. Will do that now.

samiwh: Fixed now.

workers/unity/Packages/com.improbable.gdk.core/Dynamic/Dynamic.cs
samiwh: Actually I'm not sure this is worth changing as this is one time init code and I don't think there is a substantially better way of doing this now. You can avoid some minor things by getting an enumerator directly.

samiwh: I think they both call the direct non allocating GetEnumerator and the only difference is not calling dispose in the finally block.

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentDataGenerator.tt
```diff
@@ -152,5 +156,37 @@ namespace <#= qualifiedNamespace #>
...
+        {
+            public uint ComponentId => <#= componentDetails.ComponentName  #>.ComponentId;
+
+            private static Component DeserializeData(ComponentData data, World world)
+            {
+                var schemaDataOpt = data.SchemaData;
+                if (!schemaDataOpt.HasValue)
+                {
+                    throw new ArgumentException($"Can not deserialize an empty {nameof(ComponentData)}");
+                }
```
jamiebrynes7: I don't _think_ these checks are required. I think the only reason its nullable is that if you pass in an actual vtable you will get the result of that instead of the raw schema data

jamiebrynes7: Discussed offline - protection against users passing in their own potentially malformed ops and reduces boilerplate when using the deserialize functions

samiwh: Yeah, this is really a marginal check. Arguably this is no better than the error you'd get if this just fell through though.

samiwh:
@jamiebrynes7 Yep. Will redo codegen again shortly. Currently discussing another slight change which would mean it needed to be pushed again.

test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveBlittableSingular.cs
```diff
@@ -653,5 +657,37 @@ public struct ReceivedUpdates : IComponentData
...
+        internal class ExhaustiveBlittableSingularDynamic : IDynamicInvokable
+        {
+            public uint ComponentId => ExhaustiveBlittableSingular.ComponentId;
+
+            private static Component DeserializeData(ComponentData data, World world)
+            {
+                var schemaDataOpt = data.SchemaData;
+                if (!schemaDataOpt.HasValue)
+                {
+                    throw new ArgumentException($"Can not deserialize an empty {nameof(ComponentData)}");
```
jared-improbable: Mega-nit: Prefer "cannot" as it's more common.

jared-improbable: Mega-nit: Prefer "cannot" as it's more common.

workers/unity/Packages/com.improbable.gdk.core/Dynamic/Dynamic.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        private static readonly Dictionary<uint, IDynamicInvokable> idsToComponents =
+            new Dictionary<uint, IDynamicInvokable>();
+
+        private static readonly Dictionary<Type, uint> componentsToIds = new Dictionary<Type, uint>();
+
+        static Dynamic()
+        {
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
```
jared-improbable: Drop some `foreach` nesting and tests: `AppDomain.CurrentDomain.GetAssemblies().SelectMany(assembly => assembly.GetTypes()).Where(type => !type.IsAbstract)`

PR #551, opened 2018-10-17
WIP Entity queries - now with snapshots

samiwh:
Wip because docs etc will need changing but this works completely and could be merged in

#### Description
Entity query results now have an entity snapshot that lives as long as it needs to, instead of the raw entity map
#### Tests
pending
#### Documentation
pending
#### Primary reviewers
I'll probably find someone


workers/unity/Packages/com.improbable.gdk.core/Commands/WorldCommands.cs
```diff
@@ -865,7 +865,7 @@ public static class EntityQuery
             /// <remarks>
             ///     Do not use the default constructor. Use <see cref="EntityQuery.CreateRequest"/> instead.
             ///     Using <see cref="EntityQuery.CreateRequest"/> will ensure a correctly formed structure.
-            /// </remarks>
+            /// </remarks
```
firtina-improbable: looks like an accident

samiwh: It is indeed. Will fix.

spatialos.json
```diff
@@ -1,8 +1,8 @@
 {
     "name": "unity_gdk",
     "project_version": "0.0.1",
-    "sdk_version": "14.0-b6143-48ac8-WORKER-SNAPSHOT",
+    "sdk_version": "14.0-b6757-b205a-WORKER-SNAPSHOT",
     "dependencies": [
         {"name": "standard_library", "version": "14.0-b6143-48ac8-WORKER-SNAPSHOT"}
```
jamiebrynes7: Change this one for consistency - or remove it? Not sure what purpose this serves given we manually download the packages

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentDataGenerator.tt
```diff
@@ -76,6 +76,20 @@ namespace <#= qualifiedNamespace #>
 <# } #>
         }
 
+        public struct Snapshot : ISpatialComponentSnapshot
+        {
+            public uint ComponentId => <#= componentDetails.ComponentId #>;
+
+<# foreach(var fieldDetails in fieldDetailsList) { #>
+<# if (fieldDetails.IsBlittable) { #>
```
jamiebrynes7: This if-else doesn't appear to be necessary

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentDataGenerator.tt
```diff
@@ -123,6 +137,22 @@ namespace <#= qualifiedNamespace #>
...
+            public static <#= componentNamespace #>.Snapshot DeserializeSnapshot(global::Improbable.Worker.Core.SchemaObject obj, global::Unity.Entities.World world)
+            {
+                var component = new <#= componentNamespace #>.Snapshot();
+
+<# foreach (var fieldDetails in fieldDetailsList) { #>
+<# if (!fieldDetails.IsSingular) { #>
+                component.<#= fieldDetails.PascalCaseName#> = new <#= fieldDetails.Type #>();
+<# } #>
+                {
+                    <#= fieldDetails.GetDeserializeString("component." + fieldDetails.PascalCaseName, "obj", 5) #>
```
jamiebrynes7: Instead of the if + the GetDeserialiseString - why not add a `GetDeserialiseSnapshotString` which handles both?

jamiebrynes7: Actually you don't even need the if part. The DeserialiseString includes instantiating the field. Look at some of the non-blittable generated code for examples

spatialos.json
samiwh: It's needed for a few things. Will discuss offline.

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentDataGenerator.tt
samiwh: Yep. Originally had them different. Didn't notice they are now the same.

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentDataGenerator.tt
samiwh: You're right. I'm always caught out by the "Allocate" function not allocating anything. It makes it look like it's doing something very different.

Also that string looks like it's doing something weird for non-blittable types. I'm seeing multiple `=` on the same line. Will refactor that.

spatialos.json
samiwh: Good point. The starter project needs to be changed too for the next release.
If we need to change the snapshot again we should just be able to take whichever is latest.

jessicafalk: but why is the standard library not changed?

samiwh: Oh that's a mistake

PR #552, opened 2018-10-17
Bugfix/build support on mac

firtina-improbable:

-------

#### Description
Bug found: Mac editor cannot build for OSX

#### Tests
- [x] test on a Mac

#### Documentation
Also improved the error message a bit

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.buildsystem/Util/BuildSupportChecker.cs
+++ b/workers/unity/Packages/com.improbable.gdk.buildsystem/Util/BuildSupportChecker.cs
@@ -30,6 +30,12 @@ public static class BuildSupportChecker
         return buildTargets
             .Where(target =>
             {
+                if (target == BuildTarget.StandaloneOSX && Application.platform == RuntimePlatform.OSXEditor)
+                {
+                    // OSXEditor will always have StandaloneOSX support
+                    return false;
+                }
+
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.buildsystem/Util/BuildSupportChecker.cs
```diff
@@ -30,6 +30,12 @@ public static BuildTarget[] GetBuildTargetsMissingBuildSupport(params BuildTarge
         return buildTargets
             .Where(target =>
             {
+                if (target == BuildTarget.StandaloneOSX && Application.platform == RuntimePlatform.OSXEditor)
```
jessicafalk: lol, that is so troll. However are we sure that works for all other platforms?  didn't take a look at the original PR, but checking a directory name seems a bit brittle.

jessicafalk: took a look at the code. Why are we not changing the value inside the `BuildTargetSupportDirectoryNames` dictionary? ah, is it because the playbackengine for mac on mac is directly integrated and doesn't show up as a directory?

firtina-improbable: For Windows it works for all values supplied
For Mac it's exactly as you said, it works for all values supplied except standalone osx support because it is directly integrated and doesn't show up as a directory (it's in another directory bundled within the app)

PR #554, opened 2018-10-18
Correcting Mobile assemblies naming + reworking Android device info

gkassabli:

-------

#### Description
Assembly renaming to have uniform names + rework AndroidDeviceInfo to just enum publicly
#### Tests
Ensure Android compiles correctly
#### Primary reviewers
cc @DanailPenev


workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
```diff
@@ -2,22 +2,22 @@
 
 namespace Improbable.Gdk.Mobile.Android
 {
+    public enum AndroidDeviceType
```
firtina-improbable: Does the iOS version also return a device type?

Wondering if we can have a MobileDeviceType {Unknown, Physical, Virtual} in that case?

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
```diff
@@ -2,22 +2,22 @@
...
-            Unknown,
-            Physical,
-            Emulator
-        }
-
-        private static DeviceType activeDeviceType = DeviceType.Unknown;
+        private static AndroidDeviceType activeDeviceType = AndroidDeviceType.Unknown;
 
-        public static bool IsEmulator()
+        public static AndroidDeviceType ActiveDeviceType()
```
DanailPenev: We should maybe wait for the other PR to go in first so we can address these API changes in the AndroidClientWorkerConnector here.

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
DanailPenev: Do we need the Android prefix, considering the class is already called `AndroidDeviceInfo`?

jessicafalk: nope

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
jessicafalk: agree with that

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
```diff
@@ -2,22 +2,22 @@
...
+    public enum AndroidDeviceType
+    {
+        Unknown,
+        Physical,
+        Emulator
+    }
+
     public static class AndroidDeviceInfo
     {
         public const string AndroidStudioEmulatorDefaultCallbackIp = "10.0.2.2";
```
DanailPenev: Also somewhat unrelated but if we are not calling the emulator `AndroidStudioEmulator` nowhere else, we can probably drop AndroidStudio from here. Considering it is the default emulator of choice I think we can go with that change.

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
jessicafalk: yes, that would be perfect :)

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
jessicafalk: we could also make this a property instead of a method

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
DanailPenev: That would be nice and would enforce consistency between the two classes (Android and iOS) at compile time.

gkassabli: Good point, let me add that

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
gkassabli: 👍

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
gkassabli: I already have the commit with needed changes to other PR for compatibility. Will add it if this PR goes in first. Does that work @DanailPenev ?

gkassabli: 👍

jessicafalk: it would look something like that:
```
private static AndroidDeviceType activeDeviceType = AndroidDeviceType.Unknown;
public static AndroidDeviceType ActiveDeviceType
{
    get
    {
        if (activeDeviceType == AndroidDeviceType.Unknown) 
        {
            // do the initialization
        }
        return activeDeviceType;
    }
}

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
gkassabli: Actually, after trying that - I am not sure that is a great idea: Android and iOS device types are different. Android uses emulator, while iOS uses simulator. Later on we might want to add more concrete device types, for instance separate them by generation or by different emulator types and having them all in one enum wouldn't make a lot of sense.
I am also not sure if the consistency here would be that beneficial.
What do people think?

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
jessicafalk: Would prefer to have it in one PR as it is an API change. having those two things in two PRs is a bit confusing. And it also shouldn't take that much more time to wait for the other PR to go in

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
gkassabli: The enum was taken out of AndroidDeviceInfo to allow using it directly (just AndroidDeviceType instead of AndroidDeviceInfo.DeviceType).

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
gkassabli: Sure, we can wait with this one

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/AndroidDeviceInfo.cs
jessicafalk: I would still vote for Firtina's suggestion for now: `MobileDeviceType {Unknown, Physical, Virtual} `

I agree that this might change later. and we will most likely want to differentiate between phone and tablet very soon.
However, for now I would keep it simple and only split it out when it becomes necessary

PR #562, opened 2018-10-22
Added Android dependencies and building to ci script

DanailPenev:

-------

#### Description
Adding build steps for `AndroidClient`, as well as a reference to the preparation script from `shared-ci`.

#### Tests
How to test the preparation script from `shared-ci` repo:
- Checkout this branch
- Run `ci/bootstrap.sh`
- Go in `.shared-ci` and checkout the `feature/add-android-dependencies-unity` branch
- Comment out the call to bootstrap in `<ROOT_DIR>/ci/build-test.sh` (this would reset back to master)
- Run `ci/build-test.sh`
- ???
- Profit

#### Documentation
#### Primary reviewers

### DIFF
--- a/ci/build-test.sh
+++ b/ci/build-test.sh
@@ -7,7 +7,10 @@ cd "$(dirname "$0")/../"
 # Get shared CI and prepare unity
 ci/bootstrap.sh
 .shared-ci/scripts/prepare-unity.sh
-.shared-ci/scripts/prepare-unity-mobile.sh "$(pwd)/logs/PrepareUnityMobile.log"
+
+if isWindows; then
+    .shared-ci/scripts/prepare-unity-mobile.sh "$(pwd)/logs/PrepareUnityMobile.log"
+fi
 
 source ".shared-ci/scripts/pinned-tools.sh"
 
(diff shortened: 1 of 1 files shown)


ci/build-test.sh
```diff
@@ -7,7 +7,9 @@ cd "$(dirname "$0")/../"
 # Get shared CI and prepare unity
 ci/bootstrap.sh
 .shared-ci/scripts/prepare-unity.sh
+.shared-ci/scripts/prepare-unity-mobile.sh "$(pwd)/logs/PrepareUnityMobile.log"
 
 ci/test.sh
 .shared-ci/scripts/build.sh "workers/unity" UnityClient local "$(pwd)/logs/UnityClientBuild.log"
 .shared-ci/scripts/build.sh "workers/unity" UnityGameLogic cloud "$(pwd)/logs/UnityGameLogicBuild.log"
+.shared-ci/scripts/build.sh "workers/unity" AndroidClient local "$(pwd)/logs/AndroidClientBuild.log"
```
jessicafalk: need to either
1. add a check if thsi is running on a windows agent
2. make sure that we can also run it in ci for mac.

would prefer 2, but that could be enabled later

ci/build-test.sh
```diff
@@ -7,13 +7,18 @@ cd "$(dirname "$0")/../"
 # Get shared CI and prepare unity
 ci/bootstrap.sh
 .shared-ci/scripts/prepare-unity.sh
+.shared-ci/scripts/prepare-unity-mobile.sh "$(pwd)/logs/PrepareUnityMobile.log"
```
gkassabli: Should we execute this only for Windows?

DanailPenev: Yup, good point.

jessicafalk: we probably do want to run it for both mac and windows at some point. but seeing as we are very limited with the number of mac agents we have and mac is kind of rekt anyway. we can isolate it to windows for now

jessicafalk: decided to execute it for both as it is possible for only one platform to fail

PR #563, opened 2018-10-22
entity queries now work

samiwh:
#### Description
Docks to go with the entity query pr
#### Documentation
This is it


docs/known-issues.md
```diff
@@ -9,7 +9,6 @@ Known issue = any major user-facing bug or lack of user-facing feature that:
...
 | `bool` is not a [blittable type](https://docs.microsoft.com/en-us/dotnet/framework/interop/blittable-and-non-blittable-types), so will cause an error if used to store component data.                                                                                                                                       | 2018/09/17 | None         | Use `BlittableBool` instead.                                          | No     |
-| Accessing component data from an entity query will segfault, therefore requests where `ResultType` is `SnapshotResultType` will be dropped before sending.                                                                                                                                                                   | 2018/09/19 | None         | Don't send entity queries where `ResultType` is `SnapshotResultType`. | No     |
```
jessicafalk: do we remove the known issues or mark them as resolved? the second one issue is resolved and marked as such, but not removed from the table.

would personally prefer to remove

firtina-improbable: +1 to remove

samiwh: Wait we have a fixed column?

jessicafalk: it is weird. as said offline, let's just remove it

PR #564, opened 2018-10-22
UTY-1363: Use SchemaObject.IndexObject instead of GetObject to remove duplicate events

JonasImprobable:

-------

#### Description
Use `SchemaObject.IndexObject` to retrieve all events if multiple events are received as part of the same component update. `SchemaObject.GetObject` only gives you the last element in a repeated proto object (according to Sami and David) leading to us retrieving the last event multiple times.
#### Tests
Manual tests. I'm not aware of a way to test this nicely at a unit test level. Help appreciated.
#### Documentation
As far as I understand we don't do explicit release/bugfix notes atm.
#### Primary reviewers
@samiwh @mattyoung-improbable


JonasImprobable:
Investigations to see whether this bug is repeated elsewhere are pending.

PR #567, opened 2018-10-22
Don't modify build settings scene configuration

firtina-improbable:

-------

#### Description
This is no longer necessary, it only causes annoying side effects now.

#### Tests
- [x] Build all local workers and test that all still work well
- [x] Build all cloud workers and test that all still work well

#### Documentation
Internal


firtina-improbable:
Tests confirm both local + cloud deployments work with workers built with these configs and no scenes selected in the BuildSettings interface

PR #568, opened 2018-10-22
Workaround for OS UI input crash

firtina-improbable:

-------

Looks like this in editor:
image

In phone:
image

#### Tests
- [x] test in ios simulator, can input, can still connect
- [x] test in physical iPhone, can input, can still connect
- [x] test in android emulator, can input, can still connect
- [x] test in physical android, can input, can still connect

#### Documentation
Need to update the ios/android docs a bit, needs new screenshots


workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -1,32 +1,86 @@
 ﻿using Improbable.Gdk.Core;
-using Improbable.Gdk.Mobile;
 using UnityEngine;
-using UnityEngine.UI;
 
 namespace Playground
 {
+    [ExecuteInEditMode]
```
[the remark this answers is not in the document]
firtina-improbable: so that you can preview it in the editor

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -1,32 +1,86 @@
...
+        public void OnGUI()
+        {
+            if (!showConnectionGui)
+            {
+                return;
+            }
+
+            using (new GUILayout.AreaScope(new Rect(
+                GuiPadding,
+                Screen.height * 0.4f,
```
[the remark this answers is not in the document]
firtina-improbable: Aye

workers/unity/Assets/Playground/Scripts/UI/ResizedGui.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public static string TextArea(string text)
+        {
+            TempGuiContent.text = text;
+            return GUILayout.TextArea(text, 
+                GUILayout.Height(GUI.skin.textArea.CalcHeight(TempGuiContent, Screen.width)));
+        }
+    }
+}
```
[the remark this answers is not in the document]
firtina-improbable: Ow!

workers/unity/ProjectSettings/EditorBuildSettings.asset
```diff
@@ -5,13 +5,13 @@ EditorBuildSettings:
   m_ObjectHideFlags: 0
   serializedVersion: 2
   m_Scenes:
-  - enabled: 1
+  - enabled: 0
```
gkassabli: We probably don't want to commit this part, right?

firtina-improbable: reverting

firtina-improbable:
Tested on all four virtual/physical combos, all of them work ✅

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -1,32 +1,87 @@
...
         public void Awake()
         {
-            ipAddressInput.text = PlayerPrefs.GetString(HostIpPlayerPrefsKey);
-            connectButton.onClick.AddListener(TryConnect);
+            ipAddressText = PlayerPrefs.GetString(HostIpPlayerPrefsKey);
+        }
+
+        public void OnGUI()
+        {
+            if (!showConnectionGui)
```
jessicafalk: how about just disabling the gameobject if we don't want to show it

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -1,32 +1,87 @@
...
-        [SerializeField] private Button connectButton;
-        [SerializeField] private Text errorMessage;
         [SerializeField] private GameObject clientWorkerPrefab;
+        [SerializeField] private Font font;
+        [SerializeField] private float screenWidthFontRatio = 20;
 
-        private GameObject worker;
+        private bool showConnectionGui = true;
+        private bool isConnecting = false;
+        private string ipAddressText = null;
```
jessicafalk: the above two are assigned default values. no need to assign ;)

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -1,32 +1,87 @@
...
-using UnityEngine.UI;
 
 namespace Playground
 {
+    [ExecuteInEditMode]
     public class ConnectionScreenController : MonoBehaviour
     {
         private const string HostIpPlayerPrefsKey = "SpatialOSHostIp";
+        private const int GuiPadding = 10;
+        private const float VerticalPositionRatio = 0.4f;
```
jessicafalk: nit: should we group `SerializeField` fields and non-serializefield together?

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -1,32 +1,87 @@
...
+                        GUI.skin.button,
+                        GUI.skin.textArea
+                    ))
+                    {
+                        ResizedGui.Label("Enter IP address:");
+                        ipAddressText = ResizedGui.TextField(ipAddressText);
+
+                        if (ResizedGui.Button("Connect"))
+                        {
+                            if (Application.isPlaying)
```
jessicafalk: nit: can combine these two ifs :)

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -1,32 +1,87 @@
...
+                            {
+                                TryConnect();
+                            }
+                        }
+
+                        if (!string.IsNullOrEmpty(errorMessage))
+                        {
+                            ResizedGui.Label("Error:");
+
```
jessicafalk: nit: whiteline :D

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
jessicafalk: that way we could also get rid of the try-finally part

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
jessicafalk: ah forget it. completely skipped that they are all const

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -1,32 +1,87 @@
...
+                Screen.height * VerticalPositionRatio,
+                Screen.width - GuiPadding * 2,
+                Screen.height - GuiPadding * 2)))
+            {
+                var guiEnabled = GUI.enabled;
+
+                GUI.enabled = !isConnecting;
+                try
+                {
+                    using (new ResizedGui(font, screenWidthFontRatio,
```
jessicafalk: this is probably due to me not knowing UI stuff in Unity too well, but why do we instantiate the `ResizedGUI`and then later we only use static methods. do we need this?

workers/unity/Assets/Playground/Scripts/UI/ResizedGui.cs
```diff
@@ -0,0 +1,69 @@
...
+                guiStyle.fontSize = fontSizeCache[i];
+            }
+        }
+
+        private static readonly GUIContent TempGuiContent = new GUIContent();
+
+        public static void Label(string labelText)
+        {
+            TempGuiContent.text = labelText;
+            GUILayout.Label(labelText, GUILayout.Height(GUI.skin.label.CalcSize(TempGuiContent).y));
```
jessicafalk: should we return here as well_ just to stay consistent

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
jessicafalk: is it to initialize the font and fontsize?

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: Yes, will clean up

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: Agreed, we can even destroy it 😁

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: You're right about the font and fontsize initialization, but it's also for restoration.

The struct is doing a trick here similar to [EditorGUILayout.HorizontalScope](https://docs.unity3d.com/ScriptReference/EditorGUILayout.HorizontalScope.html) - within the context of this using scope, all of these `GUI.skin` objects will be modified to have different font size configurations. When the `using` block is over, the skin objects will be restored to original state.

The alternative would be to do something like this:

```c#
var oldLabelFontSize = GUI.skin.label.fontSize;
var oldTextFieldFontSize = GUI.skin.textField.fontSize;
var oldButtonFontSize = GUI.button.fontSize;
var oldTextAreaFontSize = GUI.textArea.fontSize;

GUI.skin.label.fontSize = wantedFontSize;
GUI.skin.textField.fontSize = wantedFontSize;
GUI.skin.button.fontSize = wantedFontSize;
GUI.skin.textArea.fontSize = wantedFontSize;

// do some GUI actions here with different font sizes

GUI.skin.label.fontSize = oldLabelFontSize;
GUI.skin.textField.fontSize = oldTextFieldFontSize;
GUI.skin.button.fontSize = oldButtonFontSize;
GUI.skin.textArea.fontSize = oldTextAreaFontSize;
```

The static methods are helpers to ensure the vertical sizing will be fixed - a side effect of wanting to do non-traditional font sizes, because Unity assumes every line will be 16 pixels high. Can probably find an alternative to the static methods - just saw "line height" configuration potential ( note to self: try https://docs.unity3d.com/ScriptReference/EditorGUIUtility-singleLineHeight.html  )

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: will do

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: it haunts me

workers/unity/Assets/Playground/Scripts/UI/ResizedGui.cs
firtina-improbable: Can't, labels are void too

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: Checked, it's read only :(

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: Done

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: Eradicated!

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: Cleaned

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
firtina-improbable: Removed bool, it now destroys the gameobject on success

firtina-improbable: Tested this and it doesn't cause any problems as expected

workers/unity/Assets/Playground/Scripts/UI/ResizedGui.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using UnityEditor;
```
jessicafalk: will this work in a built-out player?

firtina-improbable: good catch!

firtina-improbable: fixed

gkassabli:
Just tested this on device and it looks like if we press connect without dismissing keyboard, keyboard stays there.
Can we fix that issue?

firtina-improbable:
Experimented a bit with different solutions, each worse than the other unfortunately. It seems to be a problem in Unity Input that is well known in the industry.

firtina-improbable:
Sanity test passed!

PR #574, opened 2018-10-23
Command closures in monobehaviours

samiwh:
As usual, it's in commit order

#### Description
Who loves mad hacks. I love mad hacks.
Hack to add closures ahead of planned work to do this properly.

Also a driveby to change `VerifyNotDisposed` to `IsValid`

### DIFF
--- a/test-project/Assets/Generated/Source/improbable/gdk/tests/blittabletypes/BlittableComponentGameObjectComponentDispatcher.cs
+++ b/test-project/Assets/Generated/Source/improbable/gdk/tests/blittabletypes/BlittableComponentGameObjectComponentDispatcher.cs
@@ -2,6 +2,7 @@
 // DO NOT EDIT - this file is automatically regenerated.
 // ===========
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Unity.Entities;
@@ -300,18 +301,23 @@ namespace Improbable.Gdk.Tests.BlittableTypes
                     for (var i = 0; i < entities.Length; i++)
                     {
                         var injectableStore = entityToInjectableStore[entities[i]];
-                        if (!injectableStore.TryGetInjectablesForComponent(commandResponseHandlerInjectableId, out var commandResponseHandlers))
-                        {
-                            continue;
-                        }
+                        injectableStore.TryGetInjectablesForComponent(commandResponseHandlerInjectableId,
+                            out var commandResponseHandlers);
 
                         var commandResponseList = commandResponseLists[i];
-                        foreach (Requirable.CommandResponseHandler commandResponseHandler in commandResponseHandlers)
+                        foreach (var commandResponse in commandResponseList.Responses)
                         {
-                            foreach (var commandResponse in commandResponseList.Responses)
+                            if (commandResponseHandlers != null)
                             {
-                                commandResponseHandler.OnFirstCommandResponseInternal(commandResponse);
+                                foreach (Requirable.CommandResponseHandler commandResponseHandler in
+                                    commandResponseHandlers)
+                                {
+                                    commandResponseHandler.OnFirstCommandResponseInternal(commandResponse);
+                                }
                             }
+
+                            var callback = commandResponse.Context as Action<FirstCommand.ReceivedResponse>;
+                            callback?.Invoke(commandResponse);
                         }
                     }
                 }
@@ -323,21 +329,27 @@ namespace Improbable.Gdk.Tests.BlittableTypes
                     for (var i = 0; i < entities.Length; i++)
                     {
                         var injectableStore = entityToInjectableStore[entities[i]];
-                        if (!injectableStore.TryGetInjectablesForComponent(commandResponseHandlerInjectableId, out var commandResponseHandlers))
-                        {
-                            continue;
-                        }
+                        injectableStore.TryGetInjectablesForComponent(commandResponseHandlerInjectableId,
+                            out var commandResponseHandlers);
 
                         var commandResponseList = commandResponseLists[i];
-                        foreach (Requirable.CommandResponseHandler commandResponseHandler in commandResponseHandlers)
+                        foreach (var commandResponse in commandResponseList.Responses)
                         {
-                            foreach (var commandResponse in commandResponseList.Responses)
+                            if (commandResponseHandlers != null)
                             {
-                                commandResponseHandler.OnSecondCommandResponseInternal(commandResponse);
+                                foreach (Requirable.CommandResponseHandler commandResponseHandler in
+                                    commandResponseHandlers)
+                                {
+                                    commandResponseHandler.OnSecondCommandResponseInternal(commandResponse);
+                                }
                             }
+
+                            var callback = commandResponse.Context as Action<SecondCommand.ReceivedResponse>;
+                            callback?.Invoke(commandResponse);
                         }
                     }
                 }
+
                 Profiler.EndSample();
             }
 
--- a/test-project/Assets/Generated/Source/improbable/gdk/tests/blittabletypes/BlittableComponentMonoBehaviourCommandHandlers.cs
+++ b/test-project/Assets/Generated/Source/improbable/gdk/tests/blittabletypes/BlittableComponentMonoBehaviourCommandHandlers.cs
@@ -103,7 +103,7 @@ namespace Improbable.Gdk.Tests.BlittableTypes
                 public long SendFirstCommandRequest(EntityId entityId, global::Improbable.Gdk.Tests.BlittableTypes.FirstCommandRequest payload,
                     uint? timeoutMillis = null, bool allowShortCircuiting = false, object context = null)
                 {
-                    if (!VerifyNotDisposed())
+                    if (!IsValid())
                     {
                         return -1;
                     }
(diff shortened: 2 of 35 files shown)


workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
```diff
@@ -94,6 +123,71 @@ public long ReserveEntityIds(uint numberOfEntityIds, uint? timeoutMillis = null,
...
+                ///     The EntityTemplate object that defines the SpatialOS components on the to-be-created entity.
+                /// </param>
+                /// <param name="callback">
+                ///     A callback that will be called with the command response.
+                /// </param>
+                /// <param name="timeoutMillis">
+                ///     (Optional) The command timeout in milliseconds. If not specified, will default to 5 seconds.
+                /// </param>
+                /// <returns>The request ID of the command request.</returns>
+                public long CreateEntity(EntityTemplate entityTemplate, Action<CreateEntity.ReceivedResponse> callback,
```
jessicafalk: can we just merge the two methods and use the following method signature:
`public long CreateEntity(EntityTemplate entityTemplate,  Action<CreateEntity.ReceivedResponse> callback,EntityId entityId = null, uint? timeoutMillis = null)`

samiwh: We can but I split them out because the entity ID is configuration, it's part of the payload. We do the odd things of having world command payloads as individual parameters and component commands as a struct.

So yes, but it means you are putting a callback inside the payload parameters.

jessicafalk: not sure I understand what exactly the difference would be?

also, I'm not sure what to think about putting the callback into the context object...
What if you would want to depend on the context when the callback gets triggered. Or have extra information while still triggering the callback

samiwh: You want to keep parameters grouped based on what they are to make it clear what they are for. If you mix configuration and data it makes the api a lot more confusing. At a glance, if I didn’t know how the commands all work, I’d have thought an optional entity I’d parameter, after the callback, would be the entity ID that wants the response or something. Arguably that means those parameters should be a struct, but I didn’t have time to change things like that as a drive by.

A better callback api would give you access to a context object too, however you can also just capture that in your callback, which is what expects you to do. I also doubt there are many use cases for mixing closures and handlers that need context beyond their request ID.

But ultimately this is a hack being made for a specific purpose, before work is done that makes most of it moot. Which is why this most just needs to exist rather than be perfect.

test-project/Assets/Generated/Source/improbable/gdk/tests/blittabletypes/BlittableComponentGameObjectComponentDispatcher.cs
```diff
@@ -300,18 +301,24 @@ public override void InvokeOnCommandResponseCallbacks(Dictionary<Unity.Entities.
...
+                        bool thereIsResponseHandler =
+                            injectableStore.TryGetInjectablesForComponent(commandResponseHandlerInjectableId,
+                                out var commandResponseHandlers);
 
                         var commandResponseList = commandResponseLists[i];
-                        foreach (Requirable.CommandResponseHandler commandResponseHandler in commandResponseHandlers)
+                        foreach (var commandResponse in commandResponseList.Responses)
                         {
-                            foreach (var commandResponse in commandResponseList.Responses)
+                            if (thereIsResponseHandler)
```
jared-improbable: Just check if `commandResponseHandlers` is `null` and get rid of the bool?

test-project/Assets/Generated/Source/improbable/gdk/tests/blittabletypes/BlittableComponentMonoBehaviourCommandHandlers.cs
```diff
@@ -114,6 +114,20 @@ public CommandRequestSender(Entity entity, EntityManager entityManager, ILogDisp
                     return request.RequestId;
                 }
 
+                public long SendFirstCommandRequest(EntityId entityId, global::Improbable.Gdk.Tests.BlittableTypes.FirstCommandRequest payload,
+                    Action<FirstCommand.ReceivedResponse> callback, uint? timeoutMillis = null, bool allowShortCircuiting = false)
+                {
+                    if (!VerifyNotDisposed())
```
jared-improbable: Rename the function to get rid of this double-negative. e.g. `IsValid`

test-project/Assets/Generated/Source/improbable/gdk/tests/blittabletypes/BlittableComponentGameObjectComponentDispatcher.cs
samiwh: sure

test-project/Assets/Generated/Source/improbable/gdk/tests/blittabletypes/BlittableComponentMonoBehaviourCommandHandlers.cs
samiwh: This isn't my function and that drive by will take longer. Can do that in another PR.

samiwh: (actually I look into how many places this function is. It might just be a string replace)

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
jessicafalk: hm... can't say I like that we are now hacking this together instead of properly designing it. It looks like we should rethink the internal implementation instead of adding hacks on top of it.

though I get  why we are doing it and the hack is not too big, so I guess it's fine for now...

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
```diff
@@ -60,6 +61,34 @@ public long ReserveEntityIds(uint numberOfEntityIds, uint? timeoutMillis = null,
...
+                /// </summary>
+                /// <param name="numberOfEntityIds">The number of entity IDs to reserve.</param>
+                /// <param name="callback">
+                ///     A callback that will be called with the command response.
+                /// </param>
+                /// <param name="timeoutMillis">
+                ///     (Optional) The command timeout in milliseconds. If not specified, will default to 5 seconds.
+                /// </param>
+                /// <returns>The request ID of the command request.</returns>
+                public long ReserveEntityIds(uint numberOfEntityIds, Action<ReserveEntityIds.ReceivedResponse> callback,
```
firtina-improbable: It looks like if the user calls the above method with a `context = Action<WorldCommands.ReserveEntityIds.ReceivedResponse>` that will effectively do the same thing as this method.

Why not instruct the users in the above method's documentation (and other places), to pass in a `Action<WorldCommands.ReserveEntityIds.ReceivedResponse>` if they would like that callback to be called?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
firtina-improbable: What is currently preventing us from using a struct for world command payloads instead of parameters to these methods?

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
firtina-improbable: I understand.

workers/unity/Packages/com.improbable.gdk.core/GameObjectRepresentation/Injection/WorldCommandRequestSender.cs
firtina-improbable: OK, it will need to be considered in future polish passes it seems.

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityGameObjectCommandHandlersGenerator.tt
```diff
@@ -97,6 +100,20 @@ namespace <#= qualifiedNamespace #>
                     return request.RequestId;
                 }
 
+                public long Send<#= commandDetails.CommandName #>Request(EntityId entityId, <#= commandDetails.FqnRequestType #> payload,
+                    Action<<#= receivedResponseType #>> callback, uint? timeoutMillis = null, bool allowShortCircuiting = false)
+                {
+                    if (!IsValid())
+                    {
+                        return -1;
```
jamiebrynes7: Quick refresher - why do return an invalid request id instead of throwing here?

samiwh: I'm not entirely sure. The effect is that it will send the command but not give you and ID which seems wrong. I don't think I changed the behaviour here though. So this can be fixed as part of the subscriptions work.

jamiebrynes7: Yeah this is the same behaviour - just not sure why we are doing C style errors

PR #578, opened 2018-10-26
UTY-356: Allow SteamDeploymentTag and SteamTicket to be set independently

JonasImprobable:

-------

#### Description
It turns out that my previous solution (parse steam ticket from command lines) is not sufficient because the steam ticket is usually not known at that time. Allow SteamDeploymentTag and SteamTicket to be set independently as a prerequisite for retrieving the steam ticket at runtime.

#### Tests
Manual experiments with the Steam API.

#### Documentation
In a different PR.

#### Primary reviewers
@firtina-improbable @jessicafalk


workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <param name="errorMessage">Steam auth session token in case true is returned.</param>
+        /// <param name="errorMessage">Reason for failing the retrieval in case false is returned.</param>
+        /// <returns>True, if a Steam auth session token was successfully retrieved.</returns>
+        private bool GetSteamAuthSessionTicket(out string steamTicket, out string errorMessage)
+        {
+            steamTicket = string.Empty;
+            errorMessage = "Method not implemented.";
+            return false;
+
+            // Uncomment this code and remove the preceding method stub after installing Steamworks.NET. (https://steamworks.github.io)
```
jessicafalk: could we use preprocessor macros for these?

firtina-improbable: Nice idea, can have this kind of workflow:
- install steamworks.net
- change project config to define preprocessor token

workers/unity/Packages/com.improbable.gdk.core/Config/LocatorConfig.cs
```diff
@@ -179,11 +200,16 @@ public static LocatorConfig CreateConnectionConfigFromCommandLine(Dictionary<str
...
             var steamDeploymentTag = CommandLineUtility.GetCommandLineValue(
                 parsedArgs, RuntimeConfigNames.SteamDeploymentTag, string.Empty);
+            if (!string.IsNullOrEmpty(steamDeploymentTag))
+            {
+                config.SetSteamDeploymentTag(steamDeploymentTag);
+            }
+
             var steamTicket = CommandLineUtility.GetCommandLineValue(
                 parsedArgs, RuntimeConfigNames.SteamTicket, string.Empty);
             if (!string.IsNullOrEmpty(steamDeploymentTag) && !string.IsNullOrEmpty(steamTicket))
```
firtina-improbable: Looks like we may need to remove the check for `steamDeploymentTag` here

firtina-improbable: Needs a test for when there's only steam ticket and no deployment tag

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// </summary>
+        /// <remarks>
+        ///     All other configuration parameters are populated according to the logic of the parent class implementation.
+        ///     (Based on the command line arguments)
+        /// </remarks>
+        /// <param name="workerType">The type of the worker to create.</param>
+        /// <returns>The Locator connection configuration</returns>
+        protected override LocatorConfig GetLocatorConfig(string workerType)
+        {
+            var config = base.GetLocatorConfig(workerType);
```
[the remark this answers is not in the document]
JonasImprobable: It's possible to pass in a steam ticket through command line arguments and the parent class method would successfully parse and use it in those cases. This class here exists because you Usually don't know the ticket in advance to pass it in as a command line argument. I could slightly change the logic so that we only retrieve a ticket if the command line argument was not specified, i.e. the command line argumet takes priority.

workers/unity/Packages/com.improbable.gdk.core/Config/LocatorConfig.cs
JonasImprobable: Oh shoot. Thanks for spotting.

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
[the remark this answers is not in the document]
firtina-improbable: I see, that would be nice indeed!

But my comment was about potential error cases, if the command line makes the base class try using a `LoginToken` we'll get an exception saying "can't use both logintoken and steam", but if the base class won't try this it should be fine

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <returns>True, if a Steam auth session token was successfully retrieved.</returns>
+        private bool GetSteamAuthSessionTicketFacepunch(out string steamTicket, out string errorMessage)
+        {
+            steamTicket = string.Empty;
+            errorMessage = "Method not implemented.";
+            return false;
+
+            // Uncomment this code and remove the preceding method stub after installing Facepunch.Steamworks. (https://github.com/Facepunch/Facepunch.Steamworks)
+            /*
+            if (Facepunch.Steamworks.Client.Instance == null)
```
JonasImprobable: This is WIP and isn't really working yet as we depend on the Start() function for initialising the steam client being called before the Start() function of our WorkerConnector which attempts to connect. Just wanted to save my progress over the weekend.

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
JonasImprobable: I'm not fully sure here. The preprocessor tag feels like magic. We are already struggling with getting everyone to install the correct .NET SDK or Unity version and this looks like something that people might get wrong and that is hard to debug. Uncommenting code feels more "idiot proof" to me here and since this is supposed to just serve as an example (almost like a tutorial but we've written the code for you), maybe commented code is fine?

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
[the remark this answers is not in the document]
JonasImprobable: Oh dear, that some serious OOP spaghetti here. I would propose adding another command line argument called something like `connection_method` with `receptionist, locator_logintoken, locator_steam` being possible values. If we had that we could just ignore superfluous args rather than determining the connection method implicitly. The problem here is that the launcher is still starting game clients with hardcoded commandline args so this change is hard to make. My suggestion would be: let's temporarily accept the current imperfect situation and do this properly once launcher command line args are configurable.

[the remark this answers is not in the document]
JonasImprobable: Fixed!

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -93,15 +93,31 @@ public async Task Connect(string workerType, ILogDispatcher logger)
                 ConnectionDelegate connectionDelegate;
                 if (ShouldUseLocator())
                 {
+                    var config = GetLocatorConfig(workerType);
+                    if (!config.Validate(out var errorMessage))
```
JonasImprobable: This is here to handle an edge case where we neither use the login token flow nor the steam flow because neither command line arguments were specified. I wanted to handle the problem on this level because 
- we are not using the Validate() function outside of tests for some reason
- that way this issue doesn't need to be addressed at the `SteamClientWorkerConnector` level.

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
```diff
@@ -0,0 +1,138 @@
...
+        protected override LocatorConfig GetLocatorConfig(string workerType)
+        {
+            var config = base.GetLocatorConfig(workerType);
+            if (config.LocatorParameters.CredentialsType == LocatorCredentialsType.LoginToken)
+            {
+                // Game client is started via the Launcher. Do not retrieve a Steam auth session ticket and connect normally.
+                return config;
+            }
+
+            if (!string.IsNullOrEmpty(config.LocatorParameters.Steam.Ticket))
```
JonasImprobable: We are still relying on the steam deployment tag to be passed through the command line not be empty. This is now handled by calling config.Validate() in WorkerConnector.cs.

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
```diff
@@ -0,0 +1,83 @@
...
+            errorMessage = string.Empty;
+            if (!SteamManager.Initialized)
+            {
+                errorMessage = "SteamManager is not initialized. Ensure that your Steam client is running and that you are logged in. " +
+                    "Also ensure that this code is not called before SteamManager.Awake().";
+                return false;
+            }
+
+            var steamTicketData = new byte[1024];
+            var hAuthTicket = SteamUser.GetAuthSessionTicket(steamTicketData, 1024, out var steamTicketLength);
```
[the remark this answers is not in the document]
JonasImprobable: I've wrapped the code in a try catch block to catch potential exceptions.

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
jamiebrynes7: How do we expect users to find this? Are there going to be docs signposting this?

jamiebrynes7:
Needs an entry in `CHANGELOG.md` 😉

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
JonasImprobable: There will be a docs page for this. The docs are currently being stalled due to legal concerns. ;)

jared-improbable: How do we test this commented code?

JonasImprobable:
@jared-improbable: 
Re: SteamWorks/Facepunch connection feature module: The SteamClientWorkerConnector class is part of the Playground project and derives from the ClientWorkerConnector class which is also part of Playground. The reason this is not a feature module (yet) is because
- This feature is small in scope, it would be a single file containing a single library method (GetSteamAuthTicket()) which itself depends on third party code. Not sure if this is a valid reason but a feature module felt like overkill.
- Users might want a bespoke login flow or use a different Steam library at which point they will need to write their own WorkerConnector anyway. The SteamClientWorkerConnector serves as an example, almost like a tutorial.

Re: Testing code: This is a very good question... Testing this end-to-end would require a test machine have the third party libraries installed and have a steam client running where we are logged in. I also don't see a way to mock this.

I guess if the maintainability concerns are too high, there is an argument to not include this into our codebase altogether. The burden would then be shifted on the tutorials docs page which would then instruct users to write the code we didn't provide to them. Because of that, my slightly shaky idea was to include this file to complement the docs page.

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
JonasImprobable: Sorry, the way how github presents these comments confused me. I replied on the "main thread" of this PR.

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
```diff
@@ -0,0 +1,181 @@
...
+    ///     The SteamClientWorkerConnector class provides a sample solution for connecting a game client published
+    ///     on Steam to a SpatialOS deployment. It provides example code for retrieving a
+    ///     Steam auth session token at runtime using either Steamworks.NET (https://steamworks.github.io) or
+    ///     Facepunch.Steamworks (https://github.com/Facepunch/Facepunch.Steamworks).
+    ///     You can also use the ClientWorkerConnector parent class to parse the Steam auth session token
+    ///     from the commandline arguments in case the Steam auth session token is known in advance.
+    ///     You will need to install either Steamworks.NET (https://steamworks.github.io) or
+    ///     Facepunch.Steamworks (https://github.com/Facepunch/Facepunch.Steamworks) in order to use this code.
+    /// </summary>
+    public class SteamClientWorkerConnector : ClientWorkerConnector
```
jessicafalk: should this inherit from `ClientWorkerConnector`? wouldn't it be better to inherit directly from `WorkerConnector`?

workers/unity/Assets/Playground/Scripts/Worker/SteamClientWorkerConnector.cs
jessicafalk: don't like the idea of having an example that is completely commented out. This also means we can not really test it..
 Would really prefer to have some kind of switch...

How generic is this example implementation? can it be reused?

jessicafalk: e.g. SteamWorksNet and Facepunch are two different libraries, right? we could split it out into two different worker connectors that serve as examples or a default implementation for anyone who wants to use steam

JonasImprobable:
Closing this PR for now. I will create a separate repo containing a full feature module which contains all dependencies.

JonasImprobable:
Nevermind, I'm dumb. We still need to partial changes in this PR as a prerequisite for the feature module.

jamiebrynes7:
Closing this as the work to re-implement the feature on top of changes since this PR is less than the work required to rebase this PR.

PR #588, opened 2018-10-30
splitting worker connector into standalone and mobile

jessicafalk:

-------

#### Description
`UseExternalIp` needs to be set to true for mobile no matter what and we can't use command line arguments in mobile either. So after a discussion with Sami, we decided to split it out into a PC and a mobile worker connector that both inherit from the `WorkerConnector`

#### Tests
Tested locally.
tested on mobile. 
both worked

#### Documentation
need to check if we need to change worker docs

#### Primary reviewers
@samiwh

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Worker/StandaloneWorkerConnector.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Worker/StandaloneWorkerConnector.cs
@@ -3,10 +3,7 @@ using UnityEngine;
 
 namespace Improbable.Gdk.Core
 {
-    /// <summary>
-    ///     Connect workers via Monobehaviours.
-    /// </summary>
-    public class StandaloneWorkerConnector : WorkerConnector
+    public abstract class StandaloneWorkerConnector : WorkerConnector
     {
         /// <summary>
         ///     Denotes whether to connect using an external IP address.
--- a/workers/unity/Packages/com.improbable.gdk.core/Worker/StandaloneWorkerConnector.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Worker/StandaloneWorkerConnector.cs
@@ -1,0 +1,74 @@
+using System;
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    ///     Connect workers via Monobehaviours.
+    /// </summary>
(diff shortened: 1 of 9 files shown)


workers/unity/Packages/com.improbable.gdk.core/Worker/StandaloneWorkerConnector.cs
```diff
@@ -0,0 +1,71 @@
+using System;
+using UnityEngine;
+
+namespace Improbable.Gdk.Core
+{
+    public abstract class StandaloneWorkerConnector : WorkerConnector
```
samiwh: Is it called standalone? What kind of worker isn't standalone?

jessicafalk: That's Unity terminology for building for OSX/Linux/Windows. Happy to rename it to something different, if you got a suggestion

jessicafalk: we could also call it `DefaultWorkerConnector`?

DanailPenev: I prefer `StandaloneWorkerConnector` to `DefaultWorkerConnector` as default can mean a lot of things. Also adhering to Unity's terminology makes sense in general to me.

samiwh: I think standalone refers to it being an executable. So for a build target that makes sense. For something connecting the, now running, worker it makes less sense.

samiwh: PCWorkerConnector or something maybe. It's also specifically for windows, mac, and linux too, not for the other standalone stuff like ps4

samiwh: Standalone also makes it sounds like an adjective on the worker connector. Basically I just don't like "standalone" as a word here.

samiwh: I'd also prefer Default to Standalone as that is actually more of what it is. Standalone is unity terminology but for something else really.

PR #591, opened 2018-11-01
[WIP] shooting & respawning for mobile

jessicafalk:

-------

#### Description
simple shooting & respawning implementation. Currently only tested in the Editor.
No real validation on server side is done, it's more about demonstrating the feature

important to note:
- I changed the authority of the transform stuff to the gamelogic worker to make it work without having to add even more scripts and commands
- currently it includes the `UnityClient` for testing purposes, will remove these references before merging it in

#### Tests
tested in the editor

#### Documentation

#### Primary reviewers
@samiwh @jamiebrynes7

### DIFF
--- a/schema/playground/health.schema
+++ b/schema/playground/health.schema
@@ -2,9 +2,8 @@ package playground;
 
 import "from_gdk_packages/com.improbable.gdk.core/common.schema";
 
-type TakeDamageRequest
-{
-	int32 damage = 1;
+type TakeDamageRequest {
+    int32 damage = 1;
 }
 
 component Health {
--- a/schema/playground/health.schema
+++ b/schema/playground/health.schema
@@ -1,0 +1,14 @@
+package playground;
+
+import "from_gdk_packages/com.improbable.gdk.core/common.schema";
+
+type TakeDamageRequest
+{
+	int32 damage = 1;
+}
+
+component Health {
+    id = 12021;
+    int32 health = 1;
+    command improbable.common.Empty take_damage(TakeDamageRequest);
+}
--- a/schema/playground/shot.schema
+++ b/schema/playground/shot.schema
@@ -1,0 +1,16 @@
+package playground;
+
+import "improbable/vector3.schema";
+import "from_gdk_packages/com.improbable.gdk.core/common.schema";
+
+type ShootRequest {
+    int32 damage = 1;
+    improbable.Vector3f start_point = 2;
+    improbable.Vector3f direction = 3;
+}
+
+component ShootBullet
+{
+    id = 12020;
+	command improbable.common.Empty shoot(ShootRequest);
+}
--- a/schema/playground/shot.schema
+++ b/schema/playground/shot.schema
@@ -9,8 +9,7 @@ type ShootRequest {
     improbable.Vector3f direction = 3;
 }
 
-component ShootBullet
-{
+component Shooting {
     id = 12020;
-	command improbable.common.Empty shoot(ShootRequest);
+    command improbable.common.Empty shoot(ShootRequest);
 }
--- a/workers/unity/Assets/Playground/Config/PlayerTemplate.cs
+++ b/workers/unity/Assets/Playground/Config/PlayerTemplate.cs
@@ -21,6 +21,9 @@ namespace Playground
             var score = Score.Component.CreateSchemaComponentData(0);
             var cubeSpawner = CubeSpawner.Component.CreateSchemaComponentData(new List<EntityId>());
 
+            var health = Health.Component.CreateSchemaComponentData(100);
+            var shootBullet = ShootBullet.Component.CreateSchemaComponentData();
+
             var entityBuilder = EntityBuilder.Begin()
                 .AddPosition(0, 0, 0, clientAttribute)
                 .AddMetadata("Character", WorkerUtils.UnityGameLogic)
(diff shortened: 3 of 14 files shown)


schema/playground/shot.schema
```diff
@@ -0,0 +1,16 @@
...
+
+type ShootRequest {
+    int32 damage = 1;
+    improbable.Vector3f start_point = 2;
+    improbable.Vector3f direction = 3;
+}
+
+component ShootBullet
+{
+    id = 12020;
```
samiwh: Inconsistent formatting in this file

schema/playground/shot.schema
```diff
@@ -0,0 +1,16 @@
...
+import "improbable/vector3.schema";
+import "from_gdk_packages/com.improbable.gdk.core/common.schema";
+
+type ShootRequest {
+    int32 damage = 1;
+    improbable.Vector3f start_point = 2;
+    improbable.Vector3f direction = 3;
+}
+
+component ShootBullet
```
samiwh: Don't name the component for the specific command. Shooting component or something would be better.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/TriggerShotBehaviour.cs
```diff
@@ -0,0 +1,38 @@
+using Improbable;
+using Improbable.Gdk.GameObjectRepresentation;
+using Improbable.Worker;
+using UnityEngine;
+
+namespace Playground
+{
+    [WorkerType(WorkerUtils.UnityClient, WorkerUtils.AndroidClient, WorkerUtils.iOSClient)]
```
samiwh: This is client side authoritative then?

schema/playground/shot.schema
```diff
@@ -0,0 +1,16 @@
+package playground;
+
+import "improbable/vector3.schema";
+import "from_gdk_packages/com.improbable.gdk.core/common.schema";
+
+type ShootRequest {
+    int32 damage = 1;
+    improbable.Vector3f start_point = 2;
```
samiwh: Why have both of these? Wouldn't a hit point make more sense?

schema/playground/health.schema
```diff
@@ -0,0 +1,14 @@
+package playground;
+
+import "from_gdk_packages/com.improbable.gdk.core/common.schema";
+
+type TakeDamageRequest
+{
```
samiwh: Inconsistent brackets

schema/playground/health.schema
```diff
@@ -0,0 +1,14 @@
+package playground;
```
samiwh: Why is this being added to playground? Is it needed for testing?

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -31,7 +34,9 @@ public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improba
                 .AddComponent(launcher, WorkerUtils.UnityGameLogic)
                 .AddComponent(score, WorkerUtils.UnityGameLogic)
                 .AddComponent(cubeSpawner, WorkerUtils.UnityGameLogic)
-                .AddTransformSynchronizationComponents(clientAttribute)
+                .AddComponent(health, WorkerUtils.UnityGameLogic)
+                .AddComponent(shootBullet, WorkerUtils.UnityGameLogic)
+                .AddTransformSynchronizationComponents(WorkerUtils.UnityGameLogic)
```
samiwh: Why exactly do we need this change. This was put here to make playground feel slightly nicer until we get rid of most of it. It will work fine with server side authority but without any csp it will feel irritating to use.

jessicafalk: just to make the respawning a bit simpler. otherwise we would need another script, command and component that the client is  authoritative over to make it works

schema/playground/health.schema
jessicafalk: testing controls and features on mobile. if it will ever be merged into master is another question. but as long as we don't have a proper starter project for mobile, playground it is

schema/playground/shot.schema
jessicafalk: hit point and then check whether a game object is at that position? basically the idea is that the client only informs the server of the position of the entity and the direction of the shot and the server then checks what got hit

workers/unity/Assets/Playground/Scripts/MonoBehaviours/TriggerShotBehaviour.cs
jamiebrynes7: Not with that attitude you can't!

jessicafalk: lol

samiwh: Server side authority isn't listening for input on the server, its not sending "trusted" commands from the client.

samiwh: Server side movement doesn't mean listening for keyboard presses.

schema/playground/shot.schema
samiwh: That position and direction might not be consistent with what the other client sees though.
But the point of my question was why have both? What is the extra information used for?
If you had a hit point that would be used for playing animations etc.

schema/playground/health.schema
samiwh: Why not make a starter project repo for mobile then?
I'd rather not have a grab bag of features in playground. It will make cleaning it up much harder if nothing else.

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
samiwh: This just seems like another reason this shouldn't be in playground. When this eventually becomes a proper gym you don't want to have to change behaviour to make an unrelated starter project easier.

schema/playground/shot.schema
jessicafalk: the information is there to do a raycast to identify which object was hit. 
or do you mean that we should change it to just sending which entity got hit directly and not do that on server-side?

schema/playground/health.schema
jessicafalk: it's now gonna be moved to a different repo. We haven't really decided on a starter porject yet, so gonna create a proper repo later

jessicafalk:
closing this for now

PR #600, opened 2018-11-02
Add known issue that Bossa reported.

oblm:
#### Description
Add known issue that Bossa reported.

#### Tests
None

#### Documentation
This is documentation.


oblm:
@jamiebrynes7 it's unclear from the defect report but it looks like it's an incompatibility cause by having either no _Visual Studio Tools for Unity_ installed, or an old incompatible version.

jamiebrynes7:
In which case I would argue this is not a known issue with the GDK

docs/known-issues.md
```diff
@@ -9,3 +9,4 @@ Known issue = any major user-facing bug or lack of user-facing feature that:
...
 | The Unity Editor doesn't correctly import navmeshes.                                                                                                                                                                                                                                                                         | 2018/10/12 | None                                                                                                   | Manually regenerate the navmesh after opening the project.            |
+| Error: Project file has no matching compilation unit                                                                                                                                                                                                                                                                         | 2018/11/02 | [1059037](https://issuetracker.unity3d.com/issues/project-file-has-no-matching-compilation-unit)                                                                                                   | Install [these upgrades](https://issuetracker.unity3d.com/issues/project-file-has-no-matching-compilation-unit) from Microsoft.            |
```
jessicafalk: this sounds like it would be better in the troubleshooting section as it means that people installed a too old VS version (might also be good to check the setup guide whether we specify the VS version)

jamiebrynes7: We have a warning in the setup guide about older versions of VS.

We did have a specific version that was verified but we removed it for some reason

oblm:
Abandoning this PR and instead placing the the troubleshooting section.

PR #611, opened 2018-11-12
Sidegrade to the 13.4 snapshot

jamiebrynes7:
#### Description
In preparation for being on a stable Worker SDK version - eating the breaking changes for this now. Changes include:
- `RequestId` is now just a `uint`
- `Dispatcher` is now user code
- `EntityId` is now a long in the Worker SDK. Re-introduced it to reduce breaking changes.

Commits are roughly ordered.

#### Tests
Ran playground locally. Ran test project tests.

#### Documentation
How is this documented (for example: release note, upgrade guide, feature page, in-code documentation)?
Its not. May require an upgrade hint in the release notes.

PR #618, opened 2018-11-14
Sidegrade to 13.4 snapshot 

jamiebrynes7:
#### Description
The saga continues (from #611)! This PR comes with:
- A new snapshot!
- Improved commit ordering
- RequestIds as longs instead of uints
- Lots of `CInterop`y-ness
#### Tests
The standard stuff really.

#### Documentation
Changelog.


workers/unity/Packages/com.improbable.gdk.core/Dispatcher.cs
```diff
@@ -0,0 +1,494 @@
...
+                ExitUpdateGuard();
+            }
+        }
+
+        public void InvokeAllReverse(T op)
+        {
+            try
+            {
+                EnterUpdateGuard();
+                foreach (var pair in map)
```
samiwh: We can save the performance things for later but this does need to be reversed.

workers/unity/Packages/com.improbable.gdk.core/SchemaObjectExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Improbable.WorkerCore;
+
+namespace Improbable.Gdk.Core
+{
+    public static class SchemaObjectExtensions
+    {
+        public static void AddEntityId(this SchemaObject obj, uint fieldId, EntityId entityId)
```
samiwh: Is it not going to get confusing to have these for schema objects but not for the connection?

workers/unity/Packages/com.improbable.gdk.core/Dispatcher.cs
davedissian: This should be `mapReversed`.

workers/unity/Packages/com.improbable.gdk.core/SchemaObjectExtensions.cs
samiwh: I'd rather not add the extension methods unless it's there's something that makes the conversion particularly irritating.

jamiebrynes7: What APIs do we expose directly on the connection where a user hits an `long` instead of an `EntityId`? 

The alternative is to special case EntityId **everywhere** in the codegen

jamiebrynes7: In a normal workflow

workers/unity/Packages/com.improbable.gdk.core/Dispatcher.cs
jamiebrynes7: Can fix 👍

workers/unity/Packages/com.improbable.gdk.core/SchemaObjectExtensions.cs
jamiebrynes7: Discussed offline - this is okay for now, we can revisit in the future when we do passes over the code generator

PR #629, opened 2018-11-22
Entity Template v2

jamiebrynes7:
#### Description
Following on from various discussions about the entity builder and its limitations we introduce a more flexible, less opinionated `EntityTemplate` which is actually reusable.

Featuring:
- Add/Get/Remove/Replace components from the `EntityTemplate` class.
- Automatic ACL creation (carried on from the Builder)
- Overwrite component write access. 
- Tests!

Replaced all user code in Playground with the template version.

Note that a few things were removed in the process:
- `EntityBuilder`
- `CreateSchemaComponentData` generated code - as its no longer used
- `NonBlittableReader/WriterTests` - as these used `CreateSchemaComponentData` and were a relic of when blittable/non-blittable components had completely different code paths. Since these do not exercise different code paths from the reader tests, they can be safely removed.

#### Tests
Regenerated snapshot, ran around in playground, spawned cubes. All worked fine.
Ran all tests, they were 💚 

#### Documentation
Changelog entry. Will require a corresponding docs PR.

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentDataGenerator.tt
```diff
@@ -213,6 +193,16 @@ var fieldDetails = fieldDetailsList[i]; #>
 <# } #>
             }
 
+            public static void SerializeSnapshot(<#= componentNamespace #>.Snapshot snapshot, global::Improbable.Worker.CInterop.SchemaObject obj)
+            {
+<# for (var i = 0; i < fieldDetailsList.Count; i++) {
+var fieldDetails = fieldDetailsList[i]; #>
+                {
```
zeroZshadow: scoping isn't needed here

jamiebrynes7: It is as the serialisation string may be multi-line with local variables

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
```diff
@@ -63,7 +64,16 @@ private void OnEntityIdsReserved(WorldCommands.ReserveEntityIds.ReceivedResponse
 
             var location = transformReader.Data.Location;
             var cubeEntityTemplate =
-                CubeTemplate.CreateCubeEntityTemplate(new Coordinates(location.X, location.Y + 2, location.Z));
+                CubeTemplate.CreateCubeEntityTemplate();
```
paulbalaji: can probably put 66/67 on one line

workers/unity/Packages/com.improbable.gdk.core/Utility/EntityTemplate.cs
```diff
@@ -1,42 +1,272 @@
...
 
-            hasBeenUsed = true;
+            entityData[snapshot.ComponentId] = snapshot;
+        }
+
+        /// <summary>
+        ///     Removes a component snapshot from the EntityTemplate, if it exists.
+        /// </summary>
+        /// <typeparam name="TSnapshot">The type of the component snapshot.</typeparam>
+        public void RemoveComponent<TSnapshot>() where TSnapshot : struct, ISpatialComponentSnapshot
```
samiwh: Would it be preferable to have this is a `SetComponent` function rather than separate add and replace?
Also in our current use cases would it be better to have this return a copy but with the replacement?

workers/unity/Packages/com.improbable.gdk.core/Utility/EntityTemplate.cs
```diff
@@ -1,42 +1,272 @@
...
+
+        /// <summary>
+        ///     Creates an <see cref="Entity"/> instance from this template.
+        /// </summary>
+        /// <remarks>
+        ///     This function allocates native memory. The <see cref="Entity"/> returned from this function should
+        ///     be handed to a GDK API, which will take ownership, or otherwise must be disposed of manually.
+        /// </remarks>
+        /// <returns>The Entity object.</returns>
+        public Entity GetEntity()
```
samiwh: Rather than replace things in the template, we could make this a builder that adds the replacements in at this point as overrides.

workers/unity/Packages/com.improbable.gdk.core/Utility/EntityTemplate.cs
```diff
@@ -1,42 +1,272 @@
...
-        public void Dispose()
+        /// <summary>
+        ///     Attempts to overwrite a component snapshot in the EntityTemplate with another one of the same type.
+        /// </summary>
+        /// <param name="snapshot">The component snapshot that will be inserted into the EntityTemplate.</param>
+        /// <typeparam name="TSnapshot">The type of the component snapshot.</typeparam>
+        /// <exception cref="InvalidOperationException">
+        ///     Thrown if there is no component snapshot of type <see cref="TSnapshot"/> to replace.
+        /// </exception>
+        public void ReplaceComponent<TSnapshot>(TSnapshot snapshot) where TSnapshot : struct, ISpatialComponentSnapshot
```
samiwh: Discussed offline - worth considering having a set function instead of replace

workers/unity/Packages/com.improbable.gdk.core/Utility/EntityTemplate.cs
samiwh: This was meant to be on replace rather than remove

workers/unity/Packages/com.improbable.gdk.core/Utility/EntityTemplate.cs
jamiebrynes7: Discussed offline - this can be implemented on top of the existing functionality when we have the use case for it.

workers/unity/Packages/com.improbable.gdk.core/Utility/EntityTemplate.cs
jamiebrynes7: Renamed `ReplaceComponent` -> `SetComponent` and removed the check as discussed offline.

PR #630, opened 2018-11-23
Kport updates week of 19/11/2018

firtina-improbable:

See individual commits for descriptions

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/Config/LocatorConfigTests.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/Config/LocatorConfigTests.cs
@@ -15,6 +15,20 @@ namespace Improbable.Gdk.Core.EditmodeTests
         }
 
         [Test]
+        public void ConnectionTimeoutMillis_should_be_60000_by_default()
+        {
+            var config = GetDefaultWorkingConfigWithLoginToken();
+            Assert.AreEqual(60000, config.ConnectionTimeoutMillis);
+        }
+
+        [Test]
+        public void DefaultCommandTimeoutMillis_should_be_5000_by_default()
+        {
+            var config = GetDefaultWorkingConfigWithLoginToken();
+            Assert.AreEqual(5000, config.DefaultCommandTimeoutMillis);
+        }
+
+        [Test]
         public void Validate_should_return_true_when_LocatorConfig_is_set_to_default_value()
         {
             var config = GetDefaultWorkingConfigWithLoginToken();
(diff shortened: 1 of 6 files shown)


workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/Config/LocatorConfigTests.cs
```diff
@@ -14,6 +14,20 @@ public void UseExternalIp_should_be_true_by_default()
             Assert.IsTrue(config.UseExternalIp);
         }
 
+        [Test]
+        public void ConnectionTimeoutMillis_should_be_60000_by_default()
```
samiwh: This is an oddly specific test. Is this something we want?

workers/unity/Packages/com.improbable.gdk.core/Utility/EntityBuilder.cs
```diff
@@ -243,6 +248,16 @@ public void SetComponentWriteAccess(uint componentId, string attribute)
                 writePermissions[componentId] = attribute;
             }
 
+            public string GetComponentWriteAccess(uint componentId)
```
samiwh: There is nothing from the name of these functions that tells me what the difference is.

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/Config/LocatorConfigTests.cs
firtina-improbable: The main benefit of it was to prevent Rider from complaining "this isn't used anywhere"

workers/unity/Packages/com.improbable.gdk.core/Utility/EntityBuilder.cs
firtina-improbable: I based them to be similar to the other methods near them

jamiebrynes7: See #629 as this is changed. Do we need a `Get` for this?

firtina-improbable: The get was necessary to prevent unneeded modification of this property

firtina-improbable: A user would not want to check / keep track of "did I write to this thing before?"

jamiebrynes7: So the logic was: if the write access is set, don't modify it, else set it?

firtina-improbable: yes

PR #633, opened 2018-11-26
Release 0.1.3 Prep

jamiebrynes7:
Release prep - increment versions and update changelog.

Steam stuff purposefully omitted as we don't have a worked example of how to use this.


jamiebrynes7:
Manual validation passed ✔️

PR #635, opened 2018-11-26
CopySchema now scans for nested package dependencies

paulbalaji:

-------

#### Description
* `CopySchema()` now scans for indirect dependencies (dependencies of dependencies, and so on)
* some tidying up of code
* adding some test dependencies

#### Tests
* created `dependencytest`, `dependencytestchild` and `dependencytestgrandchild` packages with slightly different schema files
* made `test-project` depend on `dependencytest`
* made `dependencytest` depend on `dependencytestchild`
* made `dependencytestchild` depend on `dependencytestgrandchild`
* the child/grandchild schema were copied into the correct directory

#### Documentation
n/a

#### Primary reviewers


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Added
 
 - Added support for the Steam authentication flow.
+- `CopySchema` now scans for nested package dependencies.
```
jamiebrynes7: This is probably in `Fixed` - as this is really a bug (imo)

workers/unity/Packages/com.improbable.gdk.tools/Common.cs
```diff
@@ -83,17 +83,38 @@ internal static string GetThisPackagePath()
...
-                    kv => (string) kv.Value);
+                return ParseDependencies($"{PackagesDir}/manifest.json");
             }
             catch (Exception e)
             {
                 throw new ArgumentException($"Failed to parse manifest file: {e.Message}");
             }
         }
 
+        internal static Dictionary<string, string> GetPackageDependencies(string packagePath)
```
jamiebrynes7: Since this is just a wrapper around `ParseDependencies`, why not just merge the two?

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -215,6 +231,13 @@ private static void CopySchemaFiles(string schemaRoot, KeyValuePair<string, stri
             }
         }
 
+        private static Dictionary<string, string> GetPackagePaths(this Dictionary<string, string> packages)
```
jamiebrynes7: Maybe `GetLocalPathsInPackage` ? To make it explicit this looks for local dependencies and resolves the path? 

Also don't like that this is an extension method

CHANGELOG.md
jamiebrynes7: Also this message isn't very useful to a user:

```
Fixed a bug where Code Generation would not capture schema in a nested package dependency
```

or something similar, (and better pls)

workers/unity/Packages/com.improbable.gdk.tools/Common.cs
paulbalaji: since `GetManifestDependencies` uses `ParseDependencies`, I thought to keep things separate so that if an exception throws we can tell the difference between a top-level manifest error or a package dependency error

jamiebrynes7: Hmm does that matter that we can distinguish between them in code? Surely the file path is sufficient information?

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
jamiebrynes7: Extension methods = magic and should be used carefully 😉

workers/unity/Packages/com.improbable.gdk.tools/Common.cs
paulbalaji: this is true, unifying them then streamlines the `CopySchema` function too

workers/unity/Packages/com.improbable.gdk.tools/Common.cs
```diff
@@ -79,18 +81,22 @@ internal static string GetThisPackagePath()
...
-                    Json.Deserialize(File.ReadAllText($"{PackagesDir}/manifest.json", Encoding.UTF8));
-                return ((Dictionary<string, object>) manifest["dependencies"]).ToDictionary(kv => kv.Key,
+                var package = Json.Deserialize(File.ReadAllText(filePath, Encoding.UTF8));
+                return ((Dictionary<string, object>) package["dependencies"]).ToDictionary(kv => kv.Key,
                     kv => (string) kv.Value);
             }
             catch (Exception e)
             {
-                throw new ArgumentException($"Failed to parse manifest file: {e.Message}");
+                throw new ArgumentException($"Failed to parse dependencies: {e.Message}");
```
jamiebrynes7: Probably want to print the `e.Message` to give a hint of why it failed

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Added support for the Steam authentication flow.
 
+### Fixed
+
+- Fixed a bug where nested package dependencies were not captured during Code Generation.
```
jamiebrynes7: I'd double check with @ElleEss on this messaging

workers/unity/Packages/com.improbable.gdk.tools/Common.cs
paulbalaji: good shout

PR #638, opened 2018-11-26
Add Explicit Restart Step

jamiebrynes7:
Due to feedback, reintroducing the restart step as it has hit some people.

### DIFF
--- a/docs/content/get-started/set-up.md
+++ b/docs/content/get-started/set-up.md
@@ -36,8 +36,6 @@ Unity gives you Windows build support by default.
 
 - Verified with versions `2.1.3xx` and `2.1.4xx`
 
-> **Note:** After installing the .NET Core SDK, you should restart any Unity and Unity Hub processes. This will prevent errors where Unity cannot find the `dotnet` executable.
-
 **Step 3.** Run the **<a href="https://console.improbable.io/installer/download/stable/latest/win" data-track-link="SpatialOS Installer Downloaded|product=Docs|platform=Win|label=Win" target="_blank">SpatialOS Installer</a>**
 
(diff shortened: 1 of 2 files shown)


docs/content/get-started/set-up.md
```diff
@@ -67,6 +65,10 @@ Once you have navigated to the **Workloads** tab:
 
 Once you have installed [Rider](https://www.jetbrains.com/rider/), install the [**Unity Support** plugin](https://github.com/JetBrains/resharper-unity) for a better experience.
 
+**Step 5.** Restart your machine.
+
+> If you do not restart your machine, you may experience errors when opening a GDK project.
+
```
ElleEss: Text is tip-top.

Looking at the formatting of the page, we are gonna lose this step under the VS and Rider bits of additional info, especially in the Windows instructions  - I think we need to either:
(1) Move the info about VS and Rider to the right as a Note (like the notes above) - not sure that would work for Windows as loads of text and an image.
(2) Move the info about VS and Rider to below step 5, with a note on step 4 to look at the info below.
(3) Something else.

docs/content/get-started/set-up.md
```diff
@@ -105,6 +105,10 @@ Once you have installed [Visual Studio](https://www.visualstudio.com/downloads/)
 
 Once you have installed [Rider](https://www.jetbrains.com/rider/), install the [**Unity Support** plugin](https://github.com/JetBrains/resharper-unity) for a better experience.
 
+**Step 5.** Restart your machine.
```
ElleEss: Same comment about the formatting.
Text is tip-top.

jamiebrynes7:
This PR isn't cool enough for @ElleEss

PR #639, opened 2018-11-26
"Clean all workers" now cleans worker configs

paulbalaji:

-------

#### Description
- updated `Clean` function to clean the entire `build/assembly` directory, instead of just `build/assembly/worker`

#### Tests
tried out locally, renamed stuff to reproduce error and "clean all workers" button now fixes the issue reported in the ticket

#### Documentation
n/a

#### Primary reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Added support for the Steam authentication flow.
 
+### Changed
+
+- `Clean all workers` now cleans worker configs in addition to built-out workers.
+
 ## `0.1.3` - 2018-11-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Added support for the Steam authentication flow.
 
-### Changed
+### Fixed
 
 - `Clean all workers` now cleans worker configs in addition to built-out workers.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Added support for the Steam authentication flow.
 
+### Changed
```
jamiebrynes7: Arguably, this could be `fixed` as well 🤷‍♂️. Not sure which is more correct though

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -118,7 +121,7 @@ public static void BuildWorkerForEnvironment(string workerType, BuildEnvironment
 
         public static void Clean()
         {
-            Directory.Delete(PlayerBuildDirectory, true);
+            Directory.Delete(AssetDatabaseDirectory, true);
```
jamiebrynes7: Do we want to not clean the `PlayerBuildDirectory`, if not, you can remove the static string above?

paulbalaji: `PlayerBuildDirectory` is used in other parts of the class, so we can't remove the static string

We can do the hot swap on this line because `AssetDatabaseDirectory` is the parent of `PlayerBuildDirectory`

jamiebrynes7: Ah nice 👍

PR #644, opened 2018-12-05
Fixing launch configuration for OSX

jessicafalk:
#### Description
Describe your changes here.

#### Tests
Ran it locally with `spatial local launch`, `spatial local worker launch UnityGameLogic default` and `spatial local worker launch UnityClient default` to start up multiple workers. Worked fine.

#### Documentation
release notes

#### Primary reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - `Clean all workers` now cleans worker configs in addition to built-out workers.
+- Fixed a bug where you could only start each  built-out worker once on OSX.
 
 ## `0.1.3` - 2018-11-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Fixed
 
 - `Clean all workers` now cleans worker configs in addition to built-out workers.
-- Fixed a bug where you could only start each built-out worker once on OSX.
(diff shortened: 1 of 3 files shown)


workers/unity/spatialos.UnityGameLogic.worker.json
```diff
@@ -52,6 +52,7 @@
       "macos": {
         "command": "open",
         "arguments": [
+          "-n",
           "./build/worker/UnityGameLogic@Mac/UnityGameLogic@Mac.app",
```
firtina-improbable: Does this not need to change to be similar to the managed launch? What happens if the user tries to launch two instances of the game logic externally?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Fixed
 
 - `Clean all workers` now cleans worker configs in addition to built-out workers.
+- Fixed a bug where you could only start each built-out worker once on OSX.
```
firtina-improbable:
```suggestion
- Fixed a bug where you could start each built-out worker only once on OSX.
```

workers/unity/spatialos.UnityGameLogic.worker.json
jessicafalk: nope, the difference is that with external you try to start a worker and with managed the runtime tries to start a worker. therefore, open won't work for managed launch, but works fine for external

PR #645, opened 2018-12-05
Adding support for launching an android client over ADB from the editor

zeroZshadow:
#### Description
This PR contains some additional changes that were related or required:

* Updating the local deployment launcher to use a runtime ip if provided.
* Fix AndroidClient build options.
* Mobile connection screen now skips input if IP is provided through launch argument.

#### Tests
I've build and run for the phone a few times with different settings.

### DIFF
--- a/workers/unity/Assets/Playground/Config/BuildConfiguration.asset
+++ b/workers/unity/Assets/Playground/Config/BuildConfiguration.asset
@@ -35,10 +35,10 @@ MonoBehaviour:
     - {fileID: 102900000, guid: 82ccd0fd96a2ba24f88d0c7d01592f2e, type: 3}
     LocalBuildConfig:
       BuildPlatforms: 32
-      BuildOptions: 0
+      BuildOptions: 1
     CloudBuildConfig:
       BuildPlatforms: 0
-      BuildOptions: 16384
+      BuildOptions: 0
   - WorkerType: iOSClient
     ScenesForWorker:
     - {fileID: 102900000, guid: 424c4b2c090874512badc5e7ee757d5f, type: 3}
--- a/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
+++ b/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
@@ -23,6 +23,14 @@ namespace Playground
 
         public void Awake()
         {
+            var hostIp = GetIPFromExtras();
+            if (!string.IsNullOrEmpty(hostIp))
+            {
+                ipAddressInput.text = hostIp;
+                TryConnect();
+                return;
+            }
+
             ipAddressInput.text = PlayerPrefs.GetString(HostIpPlayerPrefsKey);
             connectButton.onClick.AddListener(TryConnect);
         }
@@ -63,6 +71,13 @@ namespace Playground
             workerConnector.IpAddress = IpAddress;
             workerConnector.ConnectionScreenController = this;
             workerConnector.TryConnect();
+        }
+
+        private string GetIPFromExtras()
+        {
+            var arguments = CommandLineUtility.GetArguments();
+            var hostIp = CommandLineUtility.GetCommandLineValue(arguments, RuntimeConfigNames.ReceptionistHost, string.Empty);
+            return hostIp;
         }
     }
 }
--- a/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
+++ b/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
@@ -23,7 +23,7 @@ namespace Playground
 
         public void Awake()
         {
-            var hostIp = GetIPFromExtras();
+            var hostIp = GetReceptionistHostFromArguments();
             if (!string.IsNullOrEmpty(hostIp))
             {
                 ipAddressInput.text = hostIp;
@@ -73,7 +73,7 @@ namespace Playground
             workerConnector.TryConnect();
         }
 
-        private string GetIPFromExtras()
+        private string GetReceptionistHostFromArguments()
         {
             var arguments = CommandLineUtility.GetArguments();
             var hostIp = CommandLineUtility.GetCommandLineValue(arguments, RuntimeConfigNames.ReceptionistHost, string.Empty);
--- a/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
+++ b/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
@@ -1,4 +1,5 @@
 using Improbable.Gdk.Core;
+using Improbable.Gdk.Mobile.Android;
 using UnityEngine;
 using UnityEngine.UI;
 
@@ -75,8 +76,9 @@ namespace Playground
 
         private string GetReceptionistHostFromArguments()
         {
-            var arguments = CommandLineUtility.GetArguments();
-            var hostIp = CommandLineUtility.GetCommandLineValue(arguments, RuntimeConfigNames.ReceptionistHost, string.Empty);
+            var arguments = LaunchArguments.GetArguments();
+            var hostIp =
+                CommandLineUtility.GetCommandLineValue(arguments, RuntimeConfigNames.ReceptionistHost, string.Empty);
             return hostIp;
         }
     }
--- a/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
+++ b/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
@@ -1,5 +1,4 @@
 using Improbable.Gdk.Core;
-using Improbable.Gdk.Mobile.Android;
 using UnityEngine;
 using UnityEngine.UI;
 
@@ -76,10 +75,14 @@ namespace Playground
 
         private string GetReceptionistHostFromArguments()
         {
-            var arguments = LaunchArguments.GetArguments();
+#if UNITY_ANDROID
+            var arguments = Improbable.Gdk.Mobile.Android.LaunchArguments.GetArguments();
             var hostIp =
                 CommandLineUtility.GetCommandLineValue(arguments, RuntimeConfigNames.ReceptionistHost, string.Empty);
             return hostIp;
+#else
+            return string.Empty;
+#endif
         }
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
@@ -5,6 +5,36 @@ namespace Improbable.Gdk.Core
 {
     public static class CommandLineUtility
     {
+        public static Dictionary<string, string> GetArguments()
+        {
+#if UNITY_ANDROID
+            try
+            {
+                var unityPlayer = new UnityEngine.AndroidJavaClass("com.unity3d.player.UnityPlayer");
+                var currentActivity = unityPlayer.GetStatic<UnityEngine.AndroidJavaObject>("currentActivity");
+
+                var intent = currentActivity.Call<UnityEngine.AndroidJavaObject>("getIntent");
+                var hasExtra = intent.Call<bool>("hasExtra", "arguments");
+
+                if (hasExtra)
+                {
+                    var extras = intent.Call<UnityEngine.AndroidJavaObject>("getExtras");
+                    var arguments = extras.Call<string>("getString", "arguments");
+                    UnityEngine.Debug.Log(arguments);
+                    return ParseCommandLineArgs(arguments.Split(' '));
+                }
+            }
+            catch (Exception e)
+            {
+                UnityEngine.Debug.LogException(e);
+            }
+
+            return new Dictionary<string, string>();
+#else
+            return ParseCommandLineArgs(Environment.GetCommandLineArgs());
+#endif
+        }
+
         public static T GetCommandLineValue<T>(Dictionary<string, string> commandLineDictionary, string configKey,
             T defaultValue)
         {
@@ -57,6 +87,11 @@ namespace Improbable.Gdk.Core
                 var flag = args[i];
                 if (flag.StartsWith("+"))
                 {
+                    if (i + 1 >= args.Count)
+                    {
+                        throw new ArgumentException($"Flag \"{flag}\" requires an argument");
+                    }
+
                     var flagArg = args[i + 1];
                     var strippedOfPlus = flag.Substring(1, flag.Length - 1);
                     config[strippedOfPlus] = flagArg;
--- a/workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
@@ -10,18 +10,19 @@ namespace Improbable.Gdk.Core
 #if UNITY_ANDROID
             try
             {
-                var unityPlayer = new UnityEngine.AndroidJavaClass("com.unity3d.player.UnityPlayer");
-                var currentActivity = unityPlayer.GetStatic<UnityEngine.AndroidJavaObject>("currentActivity");
-
-                var intent = currentActivity.Call<UnityEngine.AndroidJavaObject>("getIntent");
-                var hasExtra = intent.Call<bool>("hasExtra", "arguments");
-
-                if (hasExtra)
+                using (var unityPlayer = new UnityEngine.AndroidJavaClass("com.unity3d.player.UnityPlayer"))
+                using (var currentActivity = unityPlayer.GetStatic<UnityEngine.AndroidJavaObject>("currentActivity"))
+                using (var intent = currentActivity.Call<UnityEngine.AndroidJavaObject>("getIntent"))
                 {
-                    var extras = intent.Call<UnityEngine.AndroidJavaObject>("getExtras");
-                    var arguments = extras.Call<string>("getString", "arguments");
-                    UnityEngine.Debug.Log(arguments);
-                    return ParseCommandLineArgs(arguments.Split(' '));
+                    var hasExtra = intent.Call<bool>("hasExtra", "arguments");
+                    if (hasExtra)
+                    {
+                        using (var extras = intent.Call<UnityEngine.AndroidJavaObject>("getExtras"))
+                        {
+                            var arguments = extras.Call<string>("getString", "arguments").Split(' ');
+                            return ParseCommandLineArgs(arguments);
+                        }
+                    }
                 }
             }
             catch (Exception e)
@@ -89,7 +90,8 @@ namespace Improbable.Gdk.Core
                 {
                     if (i + 1 >= args.Count)
                     {
-                        throw new ArgumentException($"Flag \"{flag}\" requires an argument");
+                        throw new ArgumentException(
+                            $"Flag \"{flag}\" requires an argument\nArguments: \"{string.Join(", ", args)}\"");
                     }
 
                     var flagArg = args[i + 1];
--- a/workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
@@ -7,34 +7,9 @@ namespace Improbable.Gdk.Core
     {
         public static Dictionary<string, string> GetArguments()
         {
-#if UNITY_ANDROID
-            try
-            {
-                using (var unityPlayer = new UnityEngine.AndroidJavaClass("com.unity3d.player.UnityPlayer"))
-                using (var currentActivity = unityPlayer.GetStatic<UnityEngine.AndroidJavaObject>("currentActivity"))
-                using (var intent = currentActivity.Call<UnityEngine.AndroidJavaObject>("getIntent"))
-                {
-                    var hasExtra = intent.Call<bool>("hasExtra", "arguments");
-                    if (hasExtra)
-                    {
-                        using (var extras = intent.Call<UnityEngine.AndroidJavaObject>("getExtras"))
-                        {
-                            var arguments = extras.Call<string>("getString", "arguments").Split(' ');
-                            return ParseCommandLineArgs(arguments);
-                        }
-                    }
-                }
-            }
-            catch (Exception e)
-            {
-                UnityEngine.Debug.LogException(e);
-            }
-
-            return new Dictionary<string, string>();
-#else
             return ParseCommandLineArgs(Environment.GetCommandLineArgs());
-#endif
         }
+
 
         public static T GetCommandLineValue<T>(Dictionary<string, string> commandLineDictionary, string configKey,
             T defaultValue)
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/LaunchArguments.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/LaunchArguments.cs
@@ -1,0 +1,36 @@
+using System;
+using System.Collections.Generic;
+using Improbable.Gdk.Core;
+
+namespace Improbable.Gdk.Mobile.Android
+{
+    public static class LaunchArguments
+    {
+        public static Dictionary<string, string> GetArguments()
+        {
+            try
+            {
+                using (var unityPlayer = new UnityEngine.AndroidJavaClass("com.unity3d.player.UnityPlayer"))
+                using (var currentActivity = unityPlayer.GetStatic<UnityEngine.AndroidJavaObject>("currentActivity"))
+                using (var intent = currentActivity.Call<UnityEngine.AndroidJavaObject>("getIntent"))
+                {
+                    var hasExtra = intent.Call<bool>("hasExtra", "arguments");
+                    if (hasExtra)
+                    {
+                        using (var extras = intent.Call<UnityEngine.AndroidJavaObject>("getExtras"))
+                        {
+                            var arguments = extras.Call<string>("getString", "arguments").Split(' ');
+                            return CommandLineUtility.ParseCommandLineArgs(arguments);
+                        }
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                UnityEngine.Debug.LogException(e);
+            }
+
+            return new Dictionary<string, string>();
+        }
+    }
+}
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -1,0 +1,60 @@
+using System.Collections;
+using System.IO;
+using System.Text;
+using Improbable.Gdk.Tools;
+using UnityEditor;
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile
+{
+    public class LaunchMenu
+    {
+        [MenuItem("SpatialOS/Mobile/Launch on Device", false)]
+        private static void LaunchMobileClient()
+        {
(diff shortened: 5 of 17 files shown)


workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -64,5 +72,12 @@ private void TryConnect()
             workerConnector.ConnectionScreenController = this;
             workerConnector.TryConnect();
         }
+
+        private string GetIPFromExtras()
```
firtina-improbable: `FromExtras` is not particularly descriptive or helpful here, would `get receptionist host ip from command line` or something be better?

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -5,6 +5,36 @@ namespace Improbable.Gdk.Core
 {
     public static class CommandLineUtility
     {
+        public static Dictionary<string, string> GetArguments()
+        {
+#if UNITY_ANDROID
+            try
+            {
+                var unityPlayer = new UnityEngine.AndroidJavaClass("com.unity3d.player.UnityPlayer");
```
firtina-improbable: this needs to be within a `using`

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -5,6 +5,36 @@ namespace Improbable.Gdk.Core
...
+                var currentActivity = unityPlayer.GetStatic<UnityEngine.AndroidJavaObject>("currentActivity");
+
+                var intent = currentActivity.Call<UnityEngine.AndroidJavaObject>("getIntent");
+                var hasExtra = intent.Call<bool>("hasExtra", "arguments");
+
+                if (hasExtra)
+                {
+                    var extras = intent.Call<UnityEngine.AndroidJavaObject>("getExtras");
+                    var arguments = extras.Call<string>("getString", "arguments");
+                    UnityEngine.Debug.Log(arguments);
```
firtina-improbable: Debug log here can be removed

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -57,6 +87,11 @@ private static bool TryGetConfigValue<T>(Dictionary<string, string> dictionary,
                 var flag = args[i];
                 if (flag.StartsWith("+"))
                 {
+                    if (i + 1 >= args.Count)
+                    {
+                        throw new ArgumentException($"Flag \"{flag}\" requires an argument");
```
firtina-improbable: Can you print all arguments here within the exception? It will help the user debug it

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        [MenuItem("SpatialOS/Mobile/Launch on Device", false)]
+        private static void LaunchMobileClient()
+        {
+            EditorUtility.DisplayProgressBar("Launching Mobile Client", "Installing APK", 0.3f);
+
+            // Find apk to install
+            if (!TryGetApkPath("build", out var apkPath))
+            {
+                Debug.LogError("Could not find a build out Android binary to launch.");
```
firtina-improbable:
```suggestion
                Debug.LogError("Could not find a built out Android binary to launch.");
```

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -0,0 +1,60 @@
...
+                return;
```
firtina-improbable: Clear progress bar here (ideally within a finally clause of the whole method?)

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -206,6 +206,12 @@ public static void LaunchLocalDeployment()
             var command = Common.SpatialBinary;
             var commandArgs = "local launch";
 
+            var runtimeIp = GdkToolsConfiguration.GetOrCreateInstance().RuntimeIp;
+            if (!string.IsNullOrEmpty(runtimeIp))
+            {
+                commandArgs = $"{commandArgs} --runtime_ip={runtimeIp}";
```
firtina-improbable: Don't we also need to use the `mobile_launch.json` for this?

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -5,6 +5,36 @@ namespace Improbable.Gdk.Core
 {
     public static class CommandLineUtility
     {
+        public static Dictionary<string, string> GetArguments()
+        {
+#if UNITY_ANDROID
```
gkassabli: ifdefs like this in generic utility make me sad - in the long run we are going to have a lot of platform-dependent code where it shouldn't be.
The alternative is reworking our processing which would take some effort, so I'm fine with this for now, but we'll have to tackle that issue eventually.

workers/unity/Assets/Playground/Config/BuildConfiguration.asset
```diff
@@ -35,10 +35,10 @@ MonoBehaviour:
     - {fileID: 102900000, guid: 82ccd0fd96a2ba24f88d0c7d01592f2e, type: 3}
     LocalBuildConfig:
       BuildPlatforms: 32
-      BuildOptions: 0
+      BuildOptions: 1
```
jessicafalk: what are these options now?

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
jessicafalk: don't like to have that inside Core. we tried to move everything mobile related to its own package. Let's have a AdbPackage in the Android part of the mobile package

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Text;
+using Improbable.Gdk.Tools;
+using UnityEditor;
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile
+{
+    public class LaunchMenu
+    {
+        [MenuItem("SpatialOS/Mobile/Launch on Device", false)]
```
gkassabli: I would rephrase this to Launch on Android Device or similar - people might get confused and try to use this item to launch on iOS device

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -0,0 +1,60 @@
...
+                "-n", $"{bundleId}/com.unity3d.player.UnityPlayerActivity",
+                "-e", "\"arguments\"", $"\\\"{arguments.ToString()}\\\"");
+
+            EditorUtility.DisplayProgressBar("Launching Mobile Client", "Done", 1.0f);
+            EditorUtility.ClearProgressBar();
+        }
+
+        private static bool TryGetApkPath(string rootPath, out string apkPath)
+        {
+            foreach (var file in Directory.GetFiles(rootPath, "*.apk", SearchOption.AllDirectories))
```
jessicafalk: does a relative path also work correctly on mac?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            EditorUtility.DisplayProgressBar("Launching Mobile Client", "Launching Client", 0.9f);
+
+            // Optional arguments to be passed, same as standalone
+            // Use this to pass through the local ip to connect to
+            var runtimeIp = GdkToolsConfiguration.GetOrCreateInstance().RuntimeIp;
+            var arguments = new StringBuilder();
+            if (!string.IsNullOrEmpty(runtimeIp))
+            {
+                arguments.Append($"+receptionistHost {runtimeIp}");
```
jessicafalk: `+receptionistHost` -> use the default name we store in the RuntimeConfig (think that's what the class is called)

workers/unity/ProjectSettings/GraphicsSettings.asset
```diff
@@ -37,6 +37,7 @@ GraphicsSettings:
   - {fileID: 10770, guid: 0000000000000000f000000000000000, type: 0}
   - {fileID: 17000, guid: 0000000000000000f000000000000000, type: 0}
   - {fileID: 16000, guid: 0000000000000000f000000000000000, type: 0}
+  - {fileID: 16002, guid: 0000000000000000f000000000000000, type: 0}
```
jessicafalk: what is that?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: add priority as we did for all the other menu items

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
gkassabli: Correct me if I'm wrong, but mobile_launch.json is only for internal use - we expect external users to use their own config.
That said, allowing to pick launch configuration would be an awesome addition.

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: add where it is looking for the binary

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -0,0 +1,60 @@
...
+using UnityEngine;
+
+namespace Improbable.Gdk.Mobile
+{
+    public class LaunchMenu
+    {
+        [MenuItem("SpatialOS/Mobile/Launch on Device", false)]
+        private static void LaunchMobileClient()
+        {
+            EditorUtility.DisplayProgressBar("Launching Mobile Client", "Installing APK", 0.3f);
```
jessicafalk: probably also need a check whether the user actually has adb installed on their machine

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
zeroZshadow: `GetReceptionistHostFromArguments` ?

firtina-improbable: sgtm

workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
jessicafalk: Oh right, now I remember. This was also one of the downsides of using that: there is no way to define another launch config for mobile if we do that.

zeroZshadow: Lets leave that for later

workers/unity/Assets/Playground/Config/BuildConfiguration.asset
zeroZshadow: Debug mode for local, nothing for cloud

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
zeroZshadow: This is pretty standard to have? I really don't want to do even more packages just to avoid a standard C# flow

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
zeroZshadow: MSDN mentiones no differences. "Relative of absolute paths"

workers/unity/ProjectSettings/GraphicsSettings.asset
zeroZshadow: No clue, I assume default settings for android graphics mode we hadn't seen yet

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
gkassabli: Sorry, misunderstood the goal of the code on the first read.
I'd still vote for restructuring this code and having dedicated class that takes care of mobile/android argument parsing and have here only call that class under ifdef

workers/unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -157,7 +158,7 @@ PlayerSettings:
     tvOS: com.improbable.gdk
   buildNumber: {}
   AndroidBundleVersionCode: 1
-  AndroidMinSdkVersion: 16
+  AndroidMinSdkVersion: 21
```
gkassabli: Is this intentional?

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
jessicafalk: we already have a mobile package?

workers/unity/ProjectSettings/GraphicsSettings.asset
jessicafalk: revert it, if you don't know what it is?

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
jessicafalk: I'm still against this as we don't have any mobile code in core and don't want to introduce it here

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: we would still need to buttons as you need to know if you wanna start an ios or an android device

workers/unity/ProjectSettings/ProjectSettings.asset
zeroZshadow: I think I updated this to ensure Vulkan runs

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
zeroZshadow: Why not?

zeroZshadow: But we also already have this CommandLineUtility class for all come that works with the command line and parsing it. I don't want this to great a dependency on the mobile package.

firtina-improbable: I agree with @zeroZshadow here - this is a good place for "parse command line arguments for any platform", and it's pretty much the standard way of doing these kinds of things.

jessicafalk: As the core should only provide the bare minimum to ensure that you can communicate with SpatialOS. If you want to use mobile, you need the mobile feature module anyway, so let's try to keep all mobile related code in one place instead of spreading it all over the place.

(there is also a case for getting rid of this utility and inlining it in the defaultworkerconnector, but that's a different discussion)

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
firtina-improbable: But the question is what would the path be relative to, when you have the project open in Unity Editor. I think it's "likely the same - the project directory", but not 100% sure...

You can use https://docs.unity3d.com/ScriptReference/Application-dataPath.html and the build folder would be `Path.Combine(Application.dataPath, Path.Combine("..", "build"))` guaranteed for all platforms

workers/unity/ProjectSettings/ProjectSettings.asset
firtina-improbable: I guess it's not necessary any more since you were recommending not to use Vulkan as it's not great yet.

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
firtina-improbable: I mean for `private const string rootApkPath = "build";`

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
gkassabli: I'd argue for a middle ground here. On one hand, I don't think this belongs in mobile module, but rather in utility module - if we throw everything used in mobile in one module, we'll end up with a mobile module being a bag of logically unrelated things. On the other hand, this code looks a bit alien here - I would separate it in another method (like ExtractAndroidArguments) and call it here under ifdef. Potentially in the future we might want to create a separate class or dedicated mobile utility module

jessicafalk: while this sounds better, I don't think we should create a new module for one script. The mobile module is already very sparse and contains mainly utility functionality to make it easier, so I don't see a reason why we shouldn't have this method in there as well

jessicafalk: splitting functionality into different focused module is good, but not if the module is still really small and there is no added benefit in doing it so. We can think about refactoring things out when the module becomes too big

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
zeroZshadow: What menu priority should this have? The priorities list for the "other menu items" is current in the tools package, and it is internal.

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/LaunchArguments.cs
```diff
@@ -0,0 +1,36 @@
...
+                        {
+                            var arguments = extras.Call<string>("getString", "arguments").Split(' ');
+                            return CommandLineUtility.ParseCommandLineArgs(arguments);
+                        }
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                UnityEngine.Debug.LogException(e);
```
jessicafalk: when could an exception be thrown? do we just want to log it or actually throw it?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -0,0 +1,77 @@
...
+                // Find apk to install
+                var rootPath = Path.Combine(Application.dataPath, Path.Combine("..", rootApkPath));
+                if (!TryGetApkPath(rootPath, out var apkPath))
+                {
+                    Debug.LogError($"Could not find a built out Android binary in \"{rootPath}\" to launch.");
+                    return;
+                }
+
+                // Install apk on connected phone / emulator
+                RedirectedProcess.Run("adb", "install", "-r", apkPath);
```
jessicafalk: still no check if adb actually exists.

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                EditorUtility.DisplayProgressBar("Launching Mobile Client", "Done", 1.0f);
+            }
+            finally
+            {
+                EditorUtility.ClearProgressBar();
+            }
+        }
+
+        [MenuItem(MenuLaunchAndroid, true)]
```
jessicafalk: priority missing

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        [MenuItem(MenuLaunchAndroid, false, 73)]
+        private static void LaunchMobileClient()
+        {
+            try
+            {
+                EditorUtility.DisplayProgressBar("Launching Mobile Client", "Installing APK", 0.3f);
+
+                // Find apk to install
+                var rootPath = Path.Combine(Application.dataPath, Path.Combine("..", rootApkPath));
```
jessicafalk: that could be factored out into a static string, right? we need it for both menu items

workers/unity/ProjectSettings/GraphicsSettings.asset
jessicafalk: still didn't do it

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/LaunchArguments.cs
zeroZshadow: It can be thrown is this is run on an android version that has breaking changes on this api. Or a newer version of unity that breaks here.
Instead of breaking the entire launch button, it just breaks the auto connect.

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
zeroZshadow: It can't, `Application.dataPath` can only be set on runtime.
I could make it a static property however?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
zeroZshadow: Correct, rather out of scope of this initial implementation

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
zeroZshadow: The validation function needs no priority

workers/unity/ProjectSettings/GraphicsSettings.asset
zeroZshadow: Its the list of shaders that should always be included. Unity added this when I switched to android build. So I assume it wants it included?

jessicafalk: never seen that before, so maybe?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: yeah, some are in tools, some are in build system. Basically wherever you want it to show up

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: so what happens right now, if someone hasn't adb installed? cause this will troll ltierally everyone that sees that button and just wants to quickly check out android

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: we have priorities for all menu items

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: sure

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: plus I'm not sure if that will work on mac. I know that we had to use `osascript` for it to work

gkassabli: Yeah, this will most likely fail on Mac. Should we just show error that adb has to be installed on mac? Installing it here might be too convoluted. https://stackoverflow.com/questions/17901692/set-up-adb-on-mac-os-x describes common install ways - none of them look simple enough, though we might download and install it if we really want to

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -5,6 +5,36 @@ namespace Improbable.Gdk.Core
...
+                    return ParseCommandLineArgs(arguments.Split(' '));
+                }
+            }
+            catch (Exception e)
+            {
+                UnityEngine.Debug.LogException(e);
+            }
+
+            return new Dictionary<string, string>();
+#else
```
samiwh: Ideally pull this out into a method to have the `#ifdef` around the least code possible. Otherwise it becomes hard to see at a glance what this code can run on. As the other parse logic is already pulled out into a method it would also match that.

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -5,6 +5,36 @@ namespace Improbable.Gdk.Core
...
+                    UnityEngine.Debug.Log(arguments);
+                    return ParseCommandLineArgs(arguments.Split(' '));
+                }
+            }
+            catch (Exception e)
+            {
+                UnityEngine.Debug.LogException(e);
+            }
+
+            return new Dictionary<string, string>();
```
samiwh: Ideally pull this out into a method to have the #ifdef around the least code possible. Otherwise it becomes hard to see at a glance what this code can run on. As the other parse logic is already pulled out into a method it would also match that.

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
firtina-improbable: This is not a menu item though

jessicafalk: ok, nvm that. just looked up the documentation

zeroZshadow: Its not a menu item

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
zeroZshadow: This is what it was, but that was declined by @jessicafalk

zeroZshadow:
adb is no longer required to be on PATH. Now checks to make sure the Android SDK is present

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
zeroZshadow: Actually, what is this even about? This is an outdated code snipped that doesn't even exist anymore.

zeroZshadow:
It now throw an error if: no APK was build out / No AndroidSDK was installed (thus no adb)
It also uses the full path for adb, so it should no longer break if platform-tools isn't on path as it should be.

PR #646, opened 2018-12-06
Refactored connection configuration & added alpha locator flow

jessicafalk:
#### Description
Added the new alpha locator flow and refactored the Connection config as part of it.
Read in commit order:
1. Simplifying configurations necessary for connections. The validation we have been doing was very basic (does that string contain anything), so removed them
2. Updated Worker & WorkerConnectors and added new alpha locator flow
3. removing tests

When is which flow chosen for the default worker connector?
1. If we have a Login Token & a PIT -> Alpha Locator
2. If we have Steam Credentials or only a Login Token -> Locator
3. Else -> Receptionist

#### Tests
ran a cloud deployment, connected successfully

#### Documentation
none yet

#### Primary reviewers
@jamiebrynes7 @samiwh


workers/unity/Packages/com.improbable.gdk.core/Config/RuntimeConfig.cs
```diff
@@ -29,4 +31,40 @@ public static class RuntimeConfigNames
         public const string WorkerId = "workerId";
         public const string WorkerType = "workerType";
     }
+
+    /// <summary>
+    ///     Stores the configuration needed to connect via the Lcoator.
+    /// </summary>
```
jamiebrynes7: Given the docs focus.. could you add summary strings to the fields within these structs? 🙏

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
```diff
@@ -14,58 +16,151 @@ public abstract class DefaultWorkerConnector : WorkerConnector
         ///     Determines whether to connect via the locator.
         /// </summary>
         /// <returns>True, if should connect via the Locator, false otherwise.</returns>
-        protected override bool ShouldUseLocator()
+        protected override ConnectionService GetChosenService()
```
jamiebrynes7: Doc strings are incorrect here now.

jamiebrynes7: Also not a huge fan of `GetChosenService`. Maybe something like `GetConnectionService`?

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
```diff
@@ -14,58 +16,151 @@ public abstract class DefaultWorkerConnector : WorkerConnector
...
+                    : ConnectionService.Locator;
+            }
+
+            return ConnectionService.Receptionist;
+        }
+
+        protected override ConnectionParameters GetConnectionParameters(string workerType, ConnectionService service)
+        {
+            // UseExternalIp needs to be true when using the locator
+            var useExternalIp = service != ConnectionService.Receptionist || UseExternalIp;
```
jamiebrynes7: Hmm.. should we print a warning if we are overriding the user chosen value?

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
```diff
@@ -14,58 +16,151 @@ public abstract class DefaultWorkerConnector : WorkerConnector
...
+            var commandLineArguments = Environment.GetCommandLineArgs();
+            var commandLineArgs = CommandLineUtility.ParseCommandLineArgs(commandLineArguments);
+
+            var loginToken = CommandLineUtility.GetCommandLineValue(
+                commandLineArgs, RuntimeConfigNames.LoginToken, string.Empty);
+            var steamDeploymentTag = CommandLineUtility.GetCommandLineValue(
+                commandLineArgs, RuntimeConfigNames.SteamDeploymentTag, string.Empty);
+            var steamTicket = CommandLineUtility.GetCommandLineValue(
+                commandLineArgs, RuntimeConfigNames.SteamTicket, string.Empty);
+
```
jamiebrynes7: Given we can get to this point and have all the above values as `string.Empty`, is it worth doing a small validation step? I.e. - can we warn the user before it hits the actual connection flow and errors there?

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -76,91 +75,96 @@ private Worker(string workerType, Connection connection, ILogDispatcher logDispa
...
-        /// <param name="config">The Locator connection configuration.</param>
-        /// <param name="logger">The logger for this worker.</param>
-        /// <param name="origin">The origin of this worker in local Unity space.</param>
-        /// <returns>A task that returns a Worker when finished.</returns>
-        /// <exception cref="ConnectionFailedException">
-        ///     Thrown if the worker fails to connect.
-        /// </exception>
-        public static async Task<Worker> CreateWorkerAsync(LocatorConfig config,
-            Func<DeploymentList, string> deploymentListCallback, ILogDispatcher logger, Vector3 origin)
+        public static async Task<Worker> CreateWorkerAsync(
```
jamiebrynes7: Doc strings please

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -76,91 +75,96 @@ private Worker(string workerType, Connection connection, ILogDispatcher logDispa
...
+                Connection.ConnectAsync(config.ReceptionistHost, config.ReceptionistPort, config.WorkerId, connectionParameters))
             {
-                var deploymentList = await GetDeploymentList(locator);
+                return await TryToConnect(connectionFuture, connectionParameters.WorkerType, logger, origin);
+            }
+        }
 
-                var deploymentName = deploymentListCallback(deploymentList);
-                if (String.IsNullOrEmpty(deploymentName))
+        public static async Task<Worker> CreateWorkerAsync(
```
jamiebrynes7: Doc strings here as well

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -76,91 +75,96 @@ private Worker(string workerType, Connection connection, ILogDispatcher logDispa
...
                     }
 
-                    // A check is needed for the case that play mode is exited before the connection can complete.
-                    if (!Application.isPlaying)
+                    using (var connectionFuture = locator.ConnectAsync(deploymentName, connectionParameters, (_) => true))
                     {
-                        connection.Dispose();
-                        throw new ConnectionFailedException("Editor application stopped",
-                            ConnectionErrorReason.EditorApplicationStopped);
+                        var connection = await Task.Run(() => connectionFuture.Get());
```
jamiebrynes7: Can we not use the `TryToConnect` function here?

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -76,91 +75,96 @@ private Worker(string workerType, Connection connection, ILogDispatcher logDispa
...
                     }
+                }
+        }
+
 
-                    var worker = new Worker(config.WorkerType, connection, logger, origin);
-                    logger.HandleLog(LogType.Log, new LogEvent("Successfully created a worker")
-                        .WithField("WorkerId", worker.WorkerId));
-                    return worker;
+        public static async Task<Worker> CreateWorkerAsync(
```
jamiebrynes7: Can you guess what I'll write here?

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
```diff
@@ -7,26 +7,44 @@ public abstract class MobileWorkerConnector : WorkerConnector
...
-        protected override LocatorConfig GetLocatorConfig(string workerType)
+        protected override LocatorConfig GetLocatorConfig()
         {
             throw new System.NotImplementedException("The locator flow is currently not available for mobile workers.");
         }
 
-        protected override bool ShouldUseLocator()
+        protected override AlphaLocatorConfig GetAlphaLocatorConfig()
+        {
+            throw new System.NotImplementedException("The alpha locator flow is currently not available for mobile workers.");
```
jamiebrynes7: 😢

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
jessicafalk: could do, though that warning will always trigger when launching a client via locator

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
jessicafalk: soon (hopefully next week)

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
jessicafalk: yes we can

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
jamiebrynes7: I still think we should do _something_ here.

jessicafalk: missed that comment. Gonna add a simple check to ensure that either logintoken or steam tag & ticket are set

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
```diff
@@ -10,62 +12,151 @@ public abstract class DefaultWorkerConnector : WorkerConnector
         /// </summary>
         public bool UseExternalIp;
 
-        /// <summary>
-        ///     Determines whether to connect via the locator.
-        /// </summary>
-        /// <returns>True, if should connect via the Locator, false otherwise.</returns>
-        protected override bool ShouldUseLocator()
+        protected override ConnectionService GetConnectionService()
```
firtina-improbable: Since this is getting more complicated now, why not accept a command line argument for `connectionService`? Then we assert the parameters using that. The logic within here is too magical at this point.

(Can be done in another PR, shall I make a ticket for it?)

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
```diff
@@ -10,62 +12,151 @@ public abstract class DefaultWorkerConnector : WorkerConnector
...
+                    : ConnectionService.Locator;
+            }
+
+            return ConnectionService.Receptionist;
+        }
+
+        protected override ConnectionParameters GetConnectionParameters(string workerType, ConnectionService service)
+        {
+            // UseExternalIp needs to be true when using the locator
+            var useExternalIp = service != ConnectionService.Receptionist || UseExternalIp;
```
firtina-improbable: This should be `service == this || service == that` instead of `service != other /* this imples it's this or that */` (if we add a new method it'll be a bug)

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
```diff
@@ -10,62 +12,151 @@ public abstract class DefaultWorkerConnector : WorkerConnector
...
+                    LoginToken = new LoginTokenCredentials
+                    {
+                        Token = loginToken,
+                    },
+                    Steam = new SteamCredentials
+                    {
+                        DeploymentTag = steamDeploymentTag,
+                        Ticket = steamTicket,
+                    },
+                    ProjectName = CommandLineUtility.GetCommandLineValue(
```
firtina-improbable: What happens if project name is empty?

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -77,90 +76,108 @@ private Worker(string workerType, Connection connection, ILogDispatcher logDispa
...
-        /// <param name="origin">The origin of this worker in local Unity space.</param>
-        /// <returns>A task that returns a Worker when finished.</returns>
-        /// <exception cref="ConnectionFailedException">
-        ///     Thrown if the worker fails to connect.
-        /// </exception>
-        public static async Task<Worker> CreateWorkerAsync(ReceptionistConfig config, ILogDispatcher logger,
+        /// <param name="connectionFuture">The <see cref="Future{T}"/> of the <see cref="Connection"/> object that we use to connect to the SpatialOS Runtime.</param>
+        /// <param name="workerType">The type of the worker.</param>
+        /// <param name="logger">The logger used by the worker.</param>
+        /// <param name="origin">The position of the worker.</param>
```
firtina-improbable: Why was the text changed from "the origin of this worker in local Unity space" changed to "The position of the worker"? The latter is more ambiguous

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
jessicafalk: good point!

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
jessicafalk: The launcher does not pass that argument in... so we would need to check for both cases:
1. it is available
2. it is not available

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
jessicafalk: added a check

workers/unity/Packages/com.improbable.gdk.core/Worker/DefaultWorkerConnector.cs
firtina-improbable: Oh I see... will the launcher always be using a particular kind of service?

jessicafalk: it provides everything to use the alpha locator flow, but you can also use the normal locator flow

firtina-improbable: discussed offline, looks good

workers/unity/Packages/com.improbable.gdk.core/Config/RuntimeConfig.cs
```diff
@@ -29,4 +31,73 @@ public static class RuntimeConfigNames
...
+        /// A function that takes as input a list of available deployments and
+        /// returns the name of the deployment that we want to connect to via the Locator.
+        /// </summary>
+        public Func<DeploymentList, string> DeploymentListCallback;
+    }
+
+    /// <summary>
+    ///     Stores the configuration needed to connect via the Alpha Locator.
+    /// </summary>
+    public struct AlphaLocatorConfig
```
samiwh: Possibly better to swap these names around. If we know this one is definitely going to be the one we use going forward it might be better to call this the LocatorConfig and the other one the LegacyLocatorConfig.

samiwh: Or we change the names to say what kind of tokens they are using. We can change this later though so it's not that big a deal.

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
```diff
@@ -7,26 +7,44 @@ public abstract class MobileWorkerConnector : WorkerConnector
...
+            return new ConnectionParameters
+            {
+                WorkerType = workerType,
+                Network =
+                {
+                    ConnectionType = NetworkConnectionType.Tcp,
+                    UseExternalIp = true,
+                },
+                EnableProtocolLoggingAtStartup = false,
+                DefaultComponentVtable = new ComponentVtable(),
```
samiwh: This feels weird to be in the worker connector classes. I think it needs a larger refactor to get it out so there is no change to make for now, but it does feel weird.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -74,90 +76,108 @@ private Worker(string workerType, Connection connection, ILogDispatcher logDispa
...
-        /// <exception cref="ConnectionFailedException">
-        ///     Thrown if the worker fails to connect.
-        /// </exception>
-        public static async Task<Worker> CreateWorkerAsync(ReceptionistConfig config, ILogDispatcher logger,
+        /// <param name="connectionFuture">The <see cref="Future{T}"/> of the <see cref="Connection"/> object that we use to connect to the SpatialOS Runtime.</param>
+        /// <param name="workerType">The type of the worker.</param>
+        /// <param name="logger">The logger used by this worker.</param>
+        /// <param name="origin">The origin of this worker in the local Unity space.</param>
+        /// <returns>A <see cref="Task{TResult}"/> to run this method asyncally and retrieve the created <see cref="Worker"/> object upon connecting successfully.</returns>
+        private static async Task<Worker> TryToConnect(Future<Connection> connectionFuture,
```
samiwh: Nit - All async functions should have the world Async in them.

workers/unity/Packages/com.improbable.gdk.core/Config/RuntimeConfig.cs
jessicafalk: Main reason why I call them that way is because the Worker SDK is calling them like that. The current locator flow is still the standard locator, while the new locator is in the alpha namespace

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
jessicafalk: I don't find it too weird, as the worker connector takes care of specifying the exact connection logic, but I agree that we could probably separate it out a bit more. Gonna create a ticket to redesign this in the near future

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
jessicafalk: done

PR #648, opened 2018-12-06
Add more detailed profiling markers 

gkassabli:

-------

#### Description
This PR adds more detailed profiling markers that allow differentiating between component methods when profiling

#### Tests
Generate components with these changes -> Profile the project -> Ensure that different methods are displayed as separate entities

#### Documentation
Is this significant enough to go to change log?


paulbalaji:
probably need to change the branch name to run it through CI

samiwh:
FYI this will need to be redone in the near future as this is going away. However the performance characteristics will also be changing a lot so we may want to profile in different places anyway.

samiwh:
FYI this will need to be redone in the near future as this is going away. However the performance characteristics will also be changing a lot so we may want to profile in different places anyway.

gkassabli:
Closing this PR to create a new branch

PR #649, opened 2018-12-06
Add more detailed profiling markers

gkassabli:

-------

#### Description
This PR adds more detailed profiling markers that allow differentiating between component methods when profiling.
Older version of the PR - https://github.com/spatialos/gdk-for-unity/pull/648

#### Tests
Generate components with these changes -> Profile the project -> Ensure that different methods are displayed as separate entities

#### Documentation
Is this significant enough to go to change log?


samiwh:
What is it that is currently being profiled? As said before all these functions are going away, or changing dramatically, so profiling on this is of limited use mid to long term.

gkassabli:
I am using this to profile Mobile client in benchmark repository with 200 fake client scenarios.
I don't mind this living in my local feature branch, just thought that this could be a bit useful even in short term.

samiwh:
We can't really make any optimisations that won't just be immediately removed from it so it won't do much good merged in. Generally the largest bottlenecks are well known.

gkassabli:
Okay, let's close this for now and return back to it after the changes

PR #650, opened 2018-12-12
Subscriptions fixes

samiwh:
#### Description
Fixed a typo causing subscriptions to incorrectly register themselves as unavailable with subscription aggregates and not as available.

Fixed an issue with some subscriptions being set twice.

Fixed an issue with gameobjects being created later in the frame than the system enabling them.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Subscriptions/Systems/RequireLifecycleSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Subscriptions/Systems/RequireLifecycleSystem.cs
@@ -9,8 +9,7 @@ namespace Improbable.Gdk.Subscriptions
     // or the order of events should be decided by system order rather than in this system
     // would enable users changing readers more easily maybe
     [DisableAutoCreation]
-    [UpdateInGroup(typeof(SpatialOSReceiveGroup.InternalSpatialOSReceiveGroup))]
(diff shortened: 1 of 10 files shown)


workers/unity/Packages/com.improbable.gdk.core/Subscriptions/Systems/RequireLifecycleSystem.cs
```diff
@@ -9,8 +9,7 @@ namespace Improbable.Gdk.Subscriptions
     // or the order of events should be decided by system order rather than in this system
     // would enable users changing readers more easily maybe
     [DisableAutoCreation]
-    [UpdateInGroup(typeof(SpatialOSReceiveGroup.InternalSpatialOSReceiveGroup))]
-    [UpdateAfter(typeof(SpatialOSReceiveSystem))]
+    [UpdateInGroup(typeof(SpatialOSReceiveGroup.GameObjectInitializationGroup))]
```
andrew-improbable: Should this also have `[UpdateAfter(typeof(GameObjectInitializationSystem))]`?

samiwh: We can't do that at the moment this assembly doesn't know about that one.
discussed offline - as these are both supposed to be different assemblies from core it might be best to do this via groups rather than by systems directly

PR #654, opened 2018-12-17
QBI Helper v1

paulbalaji:

-------
Note: this is to aid the QBI feature only available with Runtime v2. 

#### Description
- `InterestTemplate` class to help users define `Interest` component in entity template
- `InterestQuery` class to help write interest queries 
- `Constraint` to more easily create constraints needed for interest queries

TODO:
- [x] remove examples from `PlayerTemplate`
- [x] remove schema added for examples
- [x] add tests
- [x] complete InterestBuilder tests

Example usage:
```csharp
var playerQuery = InterestQuery
    .Query(Constraint.All(
        Constraint.RelativeSphere(20),
        Constraint.Component<PlayerInfo.Component>()))
    .FilterResults(Position.ComponentId, PlayerInfo.ComponentId);

var miniMapQuery = InterestQuery
    .Query(Constraint.All(
        Constraint.RelativeBox(50, double.PositiveInfinity, 50),
        Constraint.Component<MinimapRepresentation.Component>()))
    .FilterResults(Position.ComponentId, MinimapRepresentation.ComponentId);

var interestTemplate = InterestHelper.Create()
    .AddQueries<PlayerControls.Component>(playerQuery, miniMapQuery);

var entityTemplate = new EntityTemplate();
...
entityTemplate.AddComponent(interestTemplate.ToSnapshot(), WorkerUtils.UnityGameLogic); 
```

#### Tests
yes

#### Documentation
- [x] changelog

feature module docs will be a separate PR

#### Primary reviewers


schema/playground/player_info.schema
```diff
@@ -0,0 +1,28 @@
+package playground;
+
+component PlayerInfo {
+   id = 2000;
+ 
+   int32 player_id = 1;
+}
+ 
+component PlayerControls {
```
samiwh: Is this not already a component?

schema/playground/player_info.schema
```diff
@@ -0,0 +1,28 @@
+package playground;
+
+component PlayerInfo {
```
samiwh: What other info might this have? What is is used for? Ideally, when you see a component on an entity you should know what functionality is being added.

schema/playground/player_info.schema
```diff
@@ -0,0 +1,28 @@
...
+   uint32 faction = 2;
+}
+
+component RedTeam {
+   id = 2004;
+}
+ 
+component BlueTeam {
+   id = 2005;
+}
```
samiwh: nit - new line

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Constraint.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    public class Constraint
+    {
+        public static ComponentInterest.QueryConstraint Sphere(double radius, Vector3 center)
+        {
+            return new ComponentInterest.QueryConstraint
+            {
+                AndConstraint = new List<ComponentInterest.QueryConstraint>(),
```
samiwh: Why do and and or need to tbe set here but not everything else? I thought you are only ever supposed to set one.

workers/unity/Packages/com.improbable.gdk.querybasedinterest/InterestComponent.cs
```diff
@@ -0,0 +1,54 @@
+using System.Collections.Generic;
+using Improbable.Gdk.Core;
+using Improbable.Worker.CInterop;
+using UnityEngine;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    public class InterestComponent
```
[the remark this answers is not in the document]
samiwh: Also there is `ComponentInterest` and `InterestComponent`.

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Interest.cs
```diff
@@ -0,0 +1,125 @@
...
+            return interest;
+        }
+        
+        public Interest Frequency(float frequency)
+        {
+            query.Frequency = frequency;
+            return this;
+        }
+        
+        public Interest Result<T>() where T : ISpatialComponentData
```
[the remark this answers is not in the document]
paulbalaji: Yeah I thought the same after writing it but left it in to hear other thoughts. Originally it was `params uint[] resultIds` without the generic - probably will revert to that.

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -32,5 +36,57 @@ public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improba
...
+                .AddQuery<PlayerControls.Component>(teamQuery);
+        }
+
+        private static void FrequencyExample()
+        {
+            var playerConstraint = Interest
+                .Query(Constraint.All(
+                    Constraint.RelativeSphere(20), 
+                    Constraint.Component<PlayerInfo.Component>()))
+                .Frequency(20)
```
samiwh: Is this in Hz?

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Constraint.cs
paulbalaji: Yup you're only meant to set one of the constraints. However, every other constraint is an Option<> (so they can be left out) but And/Or are List<>. It's something I want to ask runtime about.

schema/playground/player_info.schema
paulbalaji: Branched off master which does not have this as far as I can tell

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Constraint.cs
```diff
@@ -0,0 +1,139 @@
...
+                OrConstraint = new List<ComponentInterest.QueryConstraint>(),
+                ComponentConstraint = componentId
+            };
+        }
+        
+        public static ComponentInterest.QueryConstraint All(params ComponentInterest.QueryConstraint[] constraints)
```
samiwh: Any and all is nice. Keeps it consistent with the ecs queries. 👍

schema/playground/player_info.schema
paulbalaji: Yup - just added this/PlayerControls to recreate the examples given on the qbi docs page:
https://docs.improbable.io/reference/13.4/shared/reference/query-based-interest#examples

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
paulbalaji: yes it's Hz - could probably rename it something like `FrequencyHz` to make that clearer

also plan on adding docs/code comments to explain a lot of this further

workers/unity/Packages/com.improbable.gdk.querybasedinterest/InterestComponent.cs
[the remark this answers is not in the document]
paulbalaji: Agreed - left this for now but intention is to rename things a little bit to reduce the confusion

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
samiwh: Always put the units in the name if applicable. MaxFrequencyHz would be my preference (although Hz is the only bit I feel strongly about).

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Constraint.cs
samiwh: I think this is something we should deal with at serialisation. Otherwise we should have a private static `Empty` list somewhere that we set things too, if we know nothing will mutate it.

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Interest.cs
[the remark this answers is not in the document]
samiwh: The problem with params is that there is then an allocation in the conversion to List. However providing we are caching the results at startup I don't think it's that bad.

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -32,5 +36,57 @@ public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improba
...
+                .Query(Constraint.All(
+                    Constraint.RelativeBox(50, double.PositiveInfinity, 50),
+                    Constraint.Component<MinimapRepresentation.Component>()))
+                .Frequency(1)
+                .Result<Position.Component, MinimapRepresentation.Component>();
+
+            var interest = new InterestComponent()
+                .AddQuery<PlayerControls.Component>(playerConstraint)
+                .AddQuery<PlayerControls.Component>(miniMapConstraint);
+        }
```
firtina-improbable: Can you show how the interest would be added to a template?

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Constraint.cs
```diff
@@ -0,0 +1,139 @@
...
+using Improbable.Gdk.Core;
+using UnityEngine;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    public class Constraint
+    {
+        public static ComponentInterest.QueryConstraint Sphere(double radius, Vector3 center)
+        {
+            return new ComponentInterest.QueryConstraint
```
firtina-improbable: I'd have a method to return a pseudo "default(QueryConstraint)" (e.g. the empty lists set), use them here and then update the relevant fields e.g. 
```cs
var constraint = CreateDefault();
// constraint.SphereConstraint = ...;
```

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Interest.cs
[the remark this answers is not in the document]
firtina-improbable: How about a method like 
```
.WithResult<MyFirstComponent>()
.WithResult<MySecondComponent>()
.WithResult<MyThirdComponent>()
```
?

workers/unity/Packages/com.improbable.gdk.querybasedinterest/InterestComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+            return this;
+        }
+
+        public InterestComponent AddQuery<T>(ComponentInterest.Query interestQuery)
+            where T : ISpatialComponentData
+        {
+            var componentId = Dynamic.GetComponentId<T>();
+            if (!interest.TryGetValue(componentId, out var query))
+            {
+                return AddComponentInterest<T>(new ComponentInterest(new List<ComponentInterest.Query>
```
firtina-improbable: What happens in this branch? Does it add a component interest for the query? Is it a shortcut? Why?

workers/unity/Packages/com.improbable.gdk.querybasedinterest/InterestComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+            interest = new Dictionary<uint, ComponentInterest>();
+        }
+
+        public InterestComponent AddComponentInterest<T>(ComponentInterest componentInterest)
+            where T : ISpatialComponentData
+        {
+            var componentId = Dynamic.GetComponentId<T>();
+            if (interest.ContainsKey(componentId))
+            {
+                Debug.LogWarning($"Skipping component with id {componentId}.");
```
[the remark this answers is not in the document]
samiwh: It looks like it is skipping it, as all it can do it skip, merge, or replace. The other branch would replace and this one does nothing.

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Interest.cs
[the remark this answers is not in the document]
samiwh: Could do and it would be more general than what is there now. It's quite verbose though and it means you can't just save an array of components that different queries might need, if multiple queries need the same components. You could add some other type in to save a list created like that but that seems like adding epicycles.

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
paulbalaji: so once you have have your `var interest = new InterestComponent...`

just add that to your current template e.g.
```
var entityBuilder = ...
   ...
   ...
   .AddComponent(interest, workerAttribute);
```

there's an implicit conversion from `InterestComponent` to `ComponentData` that calls `CreateSchemaComponentData` under the hood

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Interest.cs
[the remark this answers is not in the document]
paulbalaji: the current implementation already lets you stack up the `.Result` bit as many times as you want so that's not an issue

workers/unity/Packages/com.improbable.gdk.querybasedinterest/InterestComponent.cs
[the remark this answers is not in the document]
paulbalaji: sweet that's a bug - should be returning straight away

...alternatively adding could just coalesce the two lists inside each `ComponentInterest`

yeah I think I'll do that instead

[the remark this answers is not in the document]
firtina-improbable: What's the difference between `AddComponentInterest` and `AddQuery`?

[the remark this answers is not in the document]
paulbalaji: `ComponentInterest` is a component with a list of queries. The `Interest` component is effectively a map from componentId --> list<queries>.  So `AddComponentInterest` effectively adds a list<query> to your interest.

`AddQuery` just adds a single query element, creating the underlying List<query> if one does not already exist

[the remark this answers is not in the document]
paulbalaji: Perhaps a more user-friendly way to do this is to replace `AddComponentInterest` with `AddQueries`

so we should only have `AddQuery` and `AddQueries`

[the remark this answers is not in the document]
firtina-improbable: It's still kind of hard to understand without an example, but it sounds like it should not be skipping, but could be an error or a merge instead.

Having parameter for the user will make it unreasonably complicated for a significant portion of users.

Don't give them choice unless it won't improve their lives greatly.

workers/unity/Packages/manifest.json
```diff
@@ -5,6 +5,7 @@
     "com.improbable.gdk.gameobjectcreation": "file:com.improbable.gdk.gameobjectcreation",
     "com.improbable.gdk.mobile": "file:com.improbable.gdk.mobile",
     "com.improbable.gdk.playerlifecycle": "file:com.improbable.gdk.playerlifecycle",
+    "com.improbable.gdk.querybasedinterest": "file:com.improbable.gdk.querybasedinterest",
```
jared-improbable: cc @ElleEss Is this definitely the name we want for this, forever?

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -32,5 +36,55 @@ public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improba
...
+                    Constraint.Component<PlayerInfo.Component>()))
+                .Filter(Position.ComponentId, PlayerInfo.ComponentId);
+
+            var miniMapQuery = Interest
+                .Query(Constraint.All(
+                    Constraint.RelativeBox(50, double.PositiveInfinity, 50),
+                    Constraint.Component<MinimapRepresentation.Component>()))
+                .Filter(Position.ComponentId, MinimapRepresentation.ComponentId);
+
+            var interest = new InterestBuilder()
```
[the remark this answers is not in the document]
paulbalaji: ah this is a remnant from earlier, will change 👍

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Constraint.cs
```diff
@@ -0,0 +1,124 @@
...
+            var constraint = Default();
+            constraint.SphereConstraint = new ComponentInterest.SphereConstraint
+            {
+                Center = new Coordinates(center.x, center.y, center.z),
+                Radius = radius
+            };
+            return constraint;
+        }
+
+        public static ComponentInterest.QueryConstraint Cylinder(double radius, Vector3 center)
```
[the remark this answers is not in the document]
paulbalaji: yup it's infinitely tall - probably should ask runtime why it's done like that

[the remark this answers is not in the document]
samiwh: I think it's because we tend to think of most load balancing in 2d terms. We tend to not have that might going in that axis and the query is cheaper this way.

workers/unity/Packages/manifest.json
jamiebrynes7: Bump on this ^

paulbalaji: @ElleEss another bump

test-project/Assets/EditmodeTests/Utility/InterestQueryTests.cs
```diff
@@ -0,0 +1,49 @@
+using Improbable.Gdk.QueryBasedInterest;
+using NUnit.Framework;
+
+namespace Improbable.Gdk.EditmodeTests.Utility
+{
+    [TestFixture]
+    public class InterestQueryTests
+    {
+        private static InterestQuery BasicQuery()
```
jamiebrynes7: Can be a property here as well imo

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Constraint.cs
```diff
@@ -0,0 +1,135 @@
...
+                    OrConstraint = EmptyList
+                };
+            }
+
+            return Default();
+        }
+
+        public static ComponentInterest.QueryConstraint Any(params ComponentInterest.QueryConstraint[] constraints)
+        {
+            if (constraints.Length > 0)
```
jamiebrynes7: You can just return the body of this if statement I think? 

If the length == 0 then the result ends up being the same as `Default()` 

Same for `All()` above.

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Improbable.Gdk.QueryBasedInterest.asmdef
```diff
@@ -0,0 +1,14 @@
...
+        "Improbable.Gdk.Generated",
+        "Unity.Entities",
+        "Unity.Entities.Hybrid",
+        "Unity.Mathematics"
+    ],
+    "optionalUnityReferences": [],
+    "includePlatforms": [],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false
+}
```
jamiebrynes7: Nit: new line

workers/unity/Packages/com.improbable.gdk.querybasedinterest/InterestQuery.cs
```diff
@@ -0,0 +1,47 @@
...
+                {
+                    Constraint = constraint,
+                    FullSnapshotResult = true,
+                    ResultComponentId = EmptyList
+                }
+            };
+            return interest;
+        }
+
+        public InterestQuery MaxFrequencyHz(float frequency)
```
jamiebrynes7: `SetMaxFrequencyHz` or `WithMaxFrequencyHz` ?

workers/unity/Packages/com.improbable.gdk.querybasedinterest/InterestQuery.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            query.Frequency = frequency;
+            return this;
+        }
+
+        public ComponentInterest.Query Filter(params uint[] resultComponentIds)
```
jamiebrynes7: `FilterResults` ?

jamiebrynes7: Something to suggest this selects the components that you want to receive as a result of the query

test-project/Assets/EditmodeTests/Utility/InterestConstraintTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            var constraint = Constraint.All(BasicConstraint());
+            Assert.True(constraint.AndConstraint.Count > 0);
+            Assert.True(constraint.OrConstraint.Count == 0);
+        }
+
+        [Test]
+        public void All_constraint_with_no_params_sets_AndOr_constraints_to_empty_list()
+        {
+            var constraint = Constraint.All();
```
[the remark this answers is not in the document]
paulbalaji: My thinking is that it is simpler to use `params cs[]` once and handle the empty array case, instead of doing `constraint c, params cs[]`

(similar with the queries case earlier)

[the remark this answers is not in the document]
jamiebrynes7: Its not clear what this as an API should do though

test-project/Assets/EditmodeTests/Utility/InterestBuilderTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            Assert.DoesNotThrow(() => BasicInterest()
+                .AddQueries<Position.Component>(BasicQuery(), BasicQuery())
+                .AddQuery<Position.Component>(BasicQuery()));
+        }
+
+        [Test]
+        public void AddQueries_should_not_change_dictionary_if_no_queries_given()
+        {
+            var interest = BasicInterest().AddQueries<Position.Component>().Build();
```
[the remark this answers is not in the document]
paulbalaji: not anymore

test-project/Assets/EditmodeTests/Utility/InterestConstraintTests.cs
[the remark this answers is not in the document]
paulbalaji: should be clearer now

also removed AddQuery<T> so all queries must be added via AddQueries<T>

test-project/Assets/EditmodeTests/Utility/InterestBuilderTests.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                initialQueryRadius = initialQuery.Constraint.RelativeSphereConstraint.Value.Radius;
+            });
+
+            var differentBasicQuery = DifferentBasicQuery;
+            double differentQueryRadius = 0;
+            Assert.DoesNotThrow(() =>
+            {
+                initialQueryRadius = DifferentBasicQuery.Constraint.RelativeSphereConstraint.Value.Radius;
+            });
```
jamiebrynes7: Why are we testing if accessing these properties throws?

test-project/Assets/EditmodeTests/Utility/InterestBuilderTests.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        [Test]
+        public void AddQueries_can_be_called_with_a_single_query()
+        {
+            Assert.DoesNotThrow(() => BasicInterest
+                .AddQueries<Position.Component>(BasicQuery));
+        }
+
+        [Test]
+        public void ReplaceQueries_clears_previous_query()
```
jamiebrynes7: This test is testing a bunch of different things - makes it hard to follow

test-project/Assets/EditmodeTests/Utility/InterestBuilderTests.cs
```diff
@@ -0,0 +1,154 @@
...
+                initialQueryRadius = replacedQuery.Queries[0].Constraint.RelativeSphereConstraint.Value.Radius;
+            });
+
+            Assert.AreEqual(1, replacedQuery.Queries.Count);
+            Assert.AreNotEqual(initialQueryRadius, replacedQueryRadius);
+            Assert.AreEqual(differentQueryRadius, replacedQueryRadius);
+        }
+
+        [Test]
+        public void ReplaceQueries_clears_previous_queries()
```
jamiebrynes7: Make the above test more like this one please

test-project/Assets/EditmodeTests/Utility/InterestBuilderTests.cs
```diff
@@ -0,0 +1,154 @@
...
+            => InterestQuery.Query(Constraint.RelativeSphere(10));
+
+        private static ComponentInterest.Query DifferentBasicQuery
+            => InterestQuery.Query(Constraint.RelativeSphere(20));
+
+        private static InterestBuilder BasicInterest
+            => InterestBuilder.Begin();
+
+        private static InterestBuilder ModifyInterest
+            => InterestBuilder.Modify(new Dictionary<uint, ComponentInterest>());
```
jamiebrynes7: At first glance, I've no idea what this `Modify` API does

test-project/Assets/EditmodeTests/Utility/InterestConstraintTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        [Test]
+        public void Component_constraint_sets_AndOr_constraints_to_empty_list()
+        {
+            var constraint = Constraint.Component(Position.ComponentId);
+            Assert.AreEqual(0, constraint.AndConstraint.Count);
+            Assert.AreEqual(0, constraint.OrConstraint.Count);
+        }
+
+        [Test]
+        public void Component_T__constraint_sets_AndOr_constraints_to_empty_list()
```
jamiebrynes7: Super nit: One too many `_` in `T__constraint`

test-project/Assets/EditmodeTests/Utility/InterestQueryTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            var query = BasicQuery.FilterResults();
+            Assert.AreEqual(0, query.ResultComponentId.Count);
+        }
+
+        [Test]
+        public void Filter_with_components_implies_FullSnapshotResult_is_null()
+        {
+            var query = BasicQuery.FilterResults(Position.ComponentId);
+            Assert.IsNull(query.FullSnapshotResult);
```
jamiebrynes7: Why is this `null` instead of false ?

test-project/Assets/EditmodeTests/Utility/InterestQueryTests.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        [Test]
+        public void Frequency_null_if_MaxFrequencyHz_never_called()
+        {
+            ComponentInterest.Query query = BasicQuery;
+            Assert.IsNull(query.Frequency);
+        }
+
+        [Test]
+        public void Filter_with_no_components_does_nothing()
```
jamiebrynes7: If so - why is this a possible API choice?

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,277 @@
...
+using UnityEngine.Assertions;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    /// <summary>
+    ///     Utility class to help construct QueryConstraint objects.
+    /// </summary>
+    public static class Constraint
+    {
+        private static readonly List<ComponentInterest.QueryConstraint> EmptyList
```
jamiebrynes7: The fact that this isn't a property scares me slightly. As its now setting a reference to this static list when its used.

If someone manually changes the list inside a `ComponentInterest.*Constraint` that was set to `EmptyList` it will mutate this list.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,277 @@
...
+        /// <param name="radius">
+        ///     Radius of the Sphere constraint.
+        /// </param>
+        /// <param name="center">
+        ///     Center of the Sphere constraint.
+        /// </param>
+        /// <returns>
+        ///     A QueryConstraint object.
+        /// </returns>
+        public static ComponentInterest.QueryConstraint Sphere(double radius, Vector3 center)
```
jamiebrynes7: Is it worth providing overrides with explicit `x`, `y`, `z` args?

jamiebrynes7: Also there's a loss of precision here where the center in schema is Coordinates (i.e. - doubles). This may be a problem

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestBuilder.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// <summary>
+        ///     Creates a new InterestBuilder object from the data of an existing Interest component.
+        /// </summary>
+        /// <param name="interest">
+        ///     The data of an existing Interest component.
+        /// </param>
+        /// <returns>
+        ///     A new InterestBuilder object.
+        /// </returns>
+        public static InterestBuilder Modify(Dictionary<uint, ComponentInterest> interest)
```
jamiebrynes7: This is kind of like a copy constructor. I'm not a hugeee fan of the name `Modify`.

Also note that the semantics here mean that you mutate the dictionary you pass into the constructor. That may not be what the API suggests (e.g. - a base set of interests which you append things to).

You'll need to do a deep copy of the dictionary to copy it

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestBuilder.cs
```diff
@@ -0,0 +1,177 @@
...
+            return interest;
+        }
+
+        /// <summary>
+        ///     Builds the Interest snapshot.
+        /// </summary>
+        /// <returns>
+        ///     A Interest.Snapshot object.
+        /// </returns>
+        public Interest.Snapshot Snapshot()
```
jamiebrynes7: `ToSnapshot` ?

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "com.improbable.gdk.querybasedinteresthelper",
+    "displayName": "SpatialOS GDK Query-based Interest Helper",
+    "version": "0.1.3",
```
jamiebrynes7: `0.1.4`

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/package.json
```diff
@@ -0,0 +1,11 @@
...
+    "unity": "2018.3",
+    "author": "Improbable Worlds Ltd",
+    "description": "SpatialOS GDK Query-based Interest Helper Module. Made by Improbable.",
+    "dependencies": {
+        "com.improbable.gdk.core":"0.1.3"
```
jamiebrynes7: `0.1.4`

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestBuilder.cs
paulbalaji: The intention here is to mutate the dictionary you pass in. A potential use case for that would be adding/removing a query depending on some external factor when the game is running. In which case you'd do something like:

```
InterestBuilder.Modify(InterestWriter.Data.ComponentInterest)
    .ReplaceQueries<Position.Component>(...)
    .GetInterest();
```

test-project/Assets/EditmodeTests/Utility/InterestQueryTests.cs
paulbalaji: it has since been removed

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestBuilder.cs
paulbalaji: thinking aloud - since it's modifying it directly the `GetInterest` method wouldn't be needed in the current implementation

however - is this the best way to give this sort of functionality to users?

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
paulbalaji: yes - that's a good idea

paulbalaji: hm you're right - probably best to pass in Coordinates and rely on the other extension methods if users want to convert from other types

test-project/Assets/EditmodeTests/Utility/InterestBuilderTests.cs
paulbalaji: my initial thought was that this property should exist in the Constraint - but really it's not the InterestBuilder tests' job to test this

test-project/Assets/EditmodeTests/Utility/InterestQueryTests.cs
paulbalaji: both null and false act as the 0 value - and since only one of the fields out of `result_component_ids` and `full_snapshot_result` needs to be set, I thought it would be better to explicitly set this to null

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
paulbalaji: hmm yeah that's not very nice, but the trade-off then is to create 2 new empty lists on each new Constraint made

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,388 @@
...
+using UnityEngine.Assertions;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    /// <summary>
+    ///     Utility class to help construct QueryConstraint objects.
+    /// </summary>
+    public static class Constraint
+    {
+        private static List<ComponentInterest.QueryConstraint> EmptyList
```
jared-improbable: If the intent is that you return an immutable list that nobody can ever mess up, why not return an `IReadOnlyList`?

paulbalaji: So if I return IReadOnlyList and call `.ToList()` when using them in the And/OrConstraint, under the hood that conversion would create a new empty list (using `_emptyArray`)

Would that be a good solution?

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestBuilder.cs
jamiebrynes7: Give the signature of the API I would expect this would seed my initial interest with the dictionary and then allow me to modify it, w/o modifying the original dictionary. 

Maybe an `InterestBuilder.From(GeneratedComponentName component)` if the intent is to mutate an existing component?

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Constraint.cs
```diff
@@ -0,0 +1,124 @@
+using System.Collections.Generic;
+using System.Linq;
+using Improbable.Gdk.Core;
+using UnityEngine;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    public static class Constraint
+    {
+        private static readonly List<ComponentInterest.QueryConstraint> EmptyList
```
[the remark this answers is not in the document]
paulbalaji: moving to:
https://github.com/spatialos/gdk-for-unity/pull/654#discussion_r252206624

workers/unity/Packages/com.improbable.gdk.querybasedinterest/Interest.cs
```diff
@@ -0,0 +1,48 @@
+using System.Collections.Generic;
+using Improbable.Gdk.Core;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    public class Interest
+    {
+        private static readonly List<uint> EmptyList = new List<uint>();
```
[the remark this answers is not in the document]
paulbalaji: moving to:
https://github.com/spatialos/gdk-for-unity/pull/654#discussion_r252206624

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestBuilder.cs
paulbalaji: so now there's 

`.Begin()` - fresh new interest builder

`.From(...)` - deep copy a given Interest component/dictionary and modify the copy

`.Mutate(...)` - modify the given Interest component/dictionary directly

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,388 @@
...
+using UnityEngine.Assertions;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    /// <summary>
+    ///     Utility class to help construct QueryConstraint objects.
+    /// </summary>
+    public static class Constraint
+    {
+        private static readonly IReadOnlyList<ComponentInterest.QueryConstraint> EmptyList
```
zeroZshadow: Please remove this and instantiate a new list everywhere instead of `ToList`

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,388 @@
...
+        /// <param name="centerY">
+        ///     Y coordinate of the center of the Sphere constraint.
+        /// </param>
+        /// <param name="centerZ">
+        ///     Z coordinate of the center of the Sphere constraint.
+        /// </param>
+        /// <returns>
+        ///     A QueryConstraint object.
+        /// </returns>
+        public static ComponentInterest.QueryConstraint Sphere(
```
zeroZshadow: Have this call into the Coordinates version instead to avoid code duplication

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,388 @@
...
+        /// <param name="centerY">
+        ///     Y coordinate of the center of the Cylinder constraint.
+        /// </param>
+        /// <param name="centerZ">
+        ///     Z coordinate of the center of the Cylinder constraint.
+        /// </param>
+        /// <returns>
+        ///     A QueryConstraint object.
+        /// </returns>
+        public static ComponentInterest.QueryConstraint Cylinder(
```
zeroZshadow: Ditto, code duplication

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,388 @@
...
+        /// <param name="centerY">
+        ///     Y coordinate of the center of the Box constraint.
+        /// </param>
+        /// <param name="centerZ">
+        ///     Z coordinate of the center of the Box constraint.
+        /// </param>
+        /// <returns>
+        ///     A QueryConstraint object.
+        /// </returns>
+        public static ComponentInterest.QueryConstraint Box(
```
zeroZshadow: duplication

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestHelper.cs
```diff
@@ -0,0 +1,220 @@
...
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    /// <summary>
+    ///     Utility class to help construct Interest component snapshots.
+    /// </summary>
+    public class InterestHelper
+    {
+        private readonly Dictionary<uint, ComponentInterest> interest;
+
+        private InterestHelper(Dictionary<uint, ComponentInterest> interest)
```
zeroZshadow: default value `null`, otherwise I wouldn't know that its a valid value to use. Saves passing it in in `Begin`

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestHelper.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// </returns>
+        public InterestHelper ReplaceQueries<T>(ComponentInterest.Query query,
+            params ComponentInterest.Query[] queries)
+            where T : ISpatialComponentData
+        {
+            var interestQueries = new List<ComponentInterest.Query>(queries.Length + 1) { query };
+            interestQueries.AddRange(queries);
+
+            var componentId = Dynamic.GetComponentId<T>();
+            if (!interest.ContainsKey(componentId))
```
zeroZshadow: Use a `TryGet` here to avoid the `interest[componentId]` lookup's below

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestHelper.cs
```diff
@@ -0,0 +1,220 @@
...
+            return new Interest.Snapshot { ComponentInterest = interest };
+        }
+
+        /// <summary>
+        ///     Returns the underlying data of an Interest component.
+        /// </summary>
+        /// <returns>
+        ///     A Dictionary<uint, ComponentInterest>.
+        /// </returns>
+        public Dictionary<uint, ComponentInterest> ToComponentInterest()
```
zeroZshadow: `ToComponentInterest` -> `AsComponentInterest` as this doesn't make a Copy

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Linq;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    /// <summary>
+    ///     Utility class to help construct ComponentInterest.Query objects.
+    /// </summary>
+    public class InterestQuery
+    {
+        private static readonly IReadOnlyList<uint> EmptyList = new List<uint>();
```
zeroZshadow: Remove empty list and instantiate lists on the spot instead.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestHelper.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// </summary>
+        /// <param name="interest">
+        ///     An existing Interest component.
+        /// </param>
+        /// <returns>
+        ///     An InterestHelper object.
+        /// </returns>
+        public static InterestHelper From(Interest.Component interest)
+        {
+            return new InterestHelper(interest.ComponentInterest
```
zeroZshadow: Dictionary has a copy constructor, use that instead

zeroZshadow: This does a ShallowCopy
Make sure it copies over the internal List of `Query` as well, and that the Query copy makes a deepcopy of its internals as well.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
samiwh: Agreed. Or the other way around. There is an interesting question around how this will mesh with coordinate remapping though. Doesn't need to be solved now but it is worth thinking about.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
paulbalaji: Initially this was done as a List to reduce allocation (because if we don't set an empty list, the field is left as null which then blows up later)

It became ReadOnly because Jamie made the point that it could be modified: https://github.com/spatialos/gdk-for-unity/pull/654#discussion_r251934226 

The tradeoff I see is:

allocate + ensure the lists can't be changed
or
minimise allocation, but run the risk that someone could change the list so it's not actually empty

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestHelper.cs
```diff
@@ -0,0 +1,220 @@
+using System.Collections.Generic;
+using System.Linq;
+using Improbable.Gdk.Core;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    /// <summary>
+    ///     Utility class to help construct Interest component snapshots.
+    /// </summary>
+    public class InterestHelper
```
samiwh: This name could be clearer. It seems like it is a way of creating a dictionary of specific queries. I generally wouldn't expect to instantiate something called InterestHelper and save them as archetypal queries. I would probably name that a noun.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestHelper.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// <summary>
+        ///     Mutates the given interest component.
+        /// </summary>
+        /// <param name="interest">
+        ///     An existing Interest component.
+        /// </param>
+        /// <returns>
+        ///     An InterestHelper object.
+        /// </returns>
+        public static InterestHelper Mutate(Interest.Component interest)
```
samiwh: What does mutate mean for these functions. I'm not sure, looking at the dictionary one especially, I would expect the given dictionary to be mutated or the one I'm calling this from.

paulbalaji: Begin/From/Mutate are just ways of initialising the helper 

begin -> new underlying dictionary 

from -> deepcopy and modify a copy of the dictionary (and use ToComponentInterest to get the new copy out at the end)

mutate -> use the reference and add/remove/clear queries from the given dictionary directly (so you don't have to return anything at the end)

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestHelper.cs
samiwh: This class generally looks very useful, although it might be worth comparing it to the planned extensions to the EntityTemplate to see if we can use the same pattern again. Broadly speaking that you have some base thing you create, then you create a new instance of it that makes a copy but with some modification. Rather than modifying the object directly. Don't know if that pattern is actually better but it might be worth a look.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,388 @@
...
+using Improbable.Gdk.Core;
+using UnityEngine;
+using UnityEngine.Assertions;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    /// <summary>
+    ///     Utility class to help construct QueryConstraint objects.
+    /// </summary>
+    public static class Constraint
```
samiwh: I worry this name can be confused with the single shot queries. Might be worth trying to come up with complimentary names for the two.

samiwh: I see we have a lot of methods that take a generated QueryConstraint as a parameter. Would it make more sense to have this as a non static class use that instead. Otherwise a user may try to use our helpers, not realise this is here and then end up constructing them from the base objects anyway.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
samiwh: You can't change the read only list but you also aren't using it so it's moot. You are just instantiating a new list each time through slightly obscured syntax right now. If we want to ensure a user can't change the list then we need to make sure the interface given to them is IReadOnlyList.

samiwh: Agreed that we don't want to give users a reference to a mutable internal object.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestTemplate.cs
```diff
@@ -0,0 +1,335 @@
...
+        /// <summary>
+        ///     Creates a new InterestTemplate object given an existing InterestTemplate.
+        /// </summary>
+        /// <param name="interest">
+        ///     An existing InterestTemplate.
+        /// </param>
+        /// <returns>
+        ///     An InterestTemplate object.
+        /// </returns>
+        public static InterestTemplate Create(InterestTemplate interestTemplate)
```
jamiebrynes7: As discussed we should rename these to `CopyFrom` or similar to make the intent explicit.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestTemplate.cs
```diff
@@ -0,0 +1,335 @@
...
+        /// <summary>
+        ///     Mutates the given interest component directly.
+        /// </summary>
+        /// <param name="interest">
+        ///     An existing Interest component.
+        /// </param>
+        /// <returns>
+        ///     An InterestTemplate object.
+        /// </returns>
+        public static InterestTemplate CreateReference(Interest.Component interest)
```
jamiebrynes7: `Borrow`? 

May be my Rust-lang sneaking in though

CHANGELOG.md
```diff
@@ -72,6 +72,10 @@
 - Added `RequestPlayerCreation` to manually request for player creation in `SendCreatePlayerRequestSystem`.
 - Added a menu item, navigate to **SpatialOS** > **Generate Dev Authentication Token**, to generate a TextAsset containing the [Development Authentication Token](https://docs.improbable.io/reference/latest/shared/auth/development-authentication).
 - Added the ability to mark a build target as `Required` which will cause builds to fail in the Editor if the prerequisite build support is not installed.
+- Added a new Query-based interest helper module, `com.improbable.gdk.querybasedinteresthelper`.
+    - `InterestTemplate` provides functionality to ergonomically add, replace and clear queries from an Interest component.
+    - `InterestQuery` enables simpler construction of interest queries.
```
jared-improbable: "simpler" than what?

test-project/Assets/EditmodeTests/Improbable.Gdk.EditModeTests.asmdef
```diff
@@ -1,6 +1,7 @@
 {
     "name": "Improbable.Gdk.EditModeTests",
     "references": [
+        "Improbable.Gdk.QueryBasedInterestHelper",
```
jared-improbable: At the risk of starting a bikeshed on naming, isn't the `Helper` suffix a bit redundant?

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,390 @@
...
+
+        private static List<ComponentInterest.QueryConstraint> ConstraintParamsToList(Constraint constraint,
+            params Constraint[] constraints)
+        {
+            var output = new List<ComponentInterest.QueryConstraint>(constraints.Length + 1)
+            {
+                constraint.constraint
+            };
+
+            foreach (var constraintParam in constraints)
```
jared-improbable: You should be able to use `AddRange` here

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
```diff
@@ -0,0 +1,390 @@
...
+        /// <param name="constraints">
+        ///     Further constraints for the list of conjunctions.
+        /// </param>
+        /// <remarks>
+        ///     At least one constraint must be provided to create an "All" QueryConstraint.
+        /// </remarks>
+        /// <returns>
+        ///     A Constraint object.
+        /// </returns>
+        public static Constraint All(Constraint constraint, params Constraint[] constraints)
```
jared-improbable: I'd suggest adding an overload for Any/All that takes an `IEnumerable<Constraint>` - I remember Klang giving feedback about this on the SDK version of this.
You lose the compile-time guarantee that you have to provide at least one constraint, but it does allow for more "native" use cases (say, constructing from a static List<> of constraints)

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <param name="resultComponentIds">
+        ///     Further IDs of components to return from the query results.
+        /// </param>
+        /// <remarks>
+        ///     At least one component ID must be provided.
+        /// </remarks>
+        /// <returns>
+        ///     An updated InterestQuery object.
+        /// </returns>
+        public InterestQuery FilterResults(uint resultComponentId, params uint[] resultComponentIds)
```
jared-improbable: As above, I'd suggest an `IEnumerable<uint>` overload

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestTemplate.cs
```diff
@@ -0,0 +1,335 @@
...
+        /// <typeparam name="T">
+        ///     Type of the authoritative component to add the interestQueries to.
+        /// </typeparam>
+        /// <remarks>
+        ///     At least one interestQuery must be provided to update the Interest component.
+        /// </remarks>
+        /// <returns>
+        ///     An InterestTemplate object.
+        /// </returns>
+        public InterestTemplate AddQueries<T>(InterestQuery query,
```
jared-improbable: Again, add an `IEnumerable<Query>` overload

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestTemplate.cs
```diff
@@ -0,0 +1,335 @@
...
+
+        private static List<ComponentInterest.Query> QueryParamsToList(InterestQuery interestQuery,
+            params InterestQuery[] interestQueries)
+        {
+            var output = new List<ComponentInterest.Query>(interestQueries.Length + 1)
+            {
+                interestQuery.AsComponentInterestQuery()
+            };
+
+            foreach (var constraintParam in interestQueries)
```
jared-improbable: Can `AddRange`

test-project/Assets/EditmodeTests/Improbable.Gdk.EditModeTests.asmdef
jamiebrynes7: Don't worry - everyone loves a good bikeshed 😉 

I think this was to differentiate from QBI itself.

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -0,0 +1,83 @@
+using System.Collections.Generic;
+
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    /// <summary>
+    ///     Utility class to help construct ComponentInterest.Query objects.
+    /// </summary>
+    public class InterestQuery
```
jamiebrynes7: You need to hide the default constructor on this as I can currently do `new InterestQuery()`

test-project/Assets/EditmodeTests/Improbable.Gdk.EditModeTests.asmdef
paulbalaji: ^yup what Jamie said

I discussed this with Laura when naming this, and this name gets the point across that you don't _need_ this module to use QBI

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Constraint.cs
paulbalaji: Don't think we can because we need to convert from the `Constraint` class to the `QueryConstraint` struct here

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestTemplate.cs
paulbalaji: same as here: https://github.com/spatialos/gdk-for-unity/pull/654#discussion_r271237927

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -70,14 +78,47 @@ public InterestQuery FilterResults(uint resultComponentId, params uint[] resultC
...
+        ///     if resultComponentIds is null or contains 0 elements.
+        /// </remarks>
+        /// <returns>
+        ///     An updated InterestQuery object.
+        /// </returns>
+        public InterestQuery FilterResults(IEnumerable<uint> resultComponentIds)
+        {
+            if (IsEnumerableNullOrEmpty(resultComponentIds))
+            {
+                Debug.LogWarning("At least one InterestQuery must be provided to filter a queries results.");
```
jamiebrynes7: Why do we throw exceptions in some places and only log warnings for others?

paulbalaji: We log warnings in `InterestTemplate` and `InterestQuery` because providing an IEnumerable with 0 elements can be safely continued with a no-op

However with `Constraint`, providing an IEnumerable with 0 elements into `Any` or `All` would mean that the constraint being set would be an empty list - which is equivalent to not having a constraint

The exception throwing is defensive in the case that an empty `QueryConstraint` is invalid to the runtime (chasing up with them to get clarity on this)

paulbalaji: @jamiebrynes7 got back from Runtime, if 0 Constraints are set then it is considered to be a user error

test-project/Assets/EditmodeTests/Utility/InterestTemplateTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        [Test]
+        public void AddQueries_can_be_called_multiple_times_on_same_component()
+        {
+            Assert.DoesNotThrow(() => EmptyInterest
+                .AddQueries<Position.Component>(BasicQuery, BasicQuery)
+                .AddQueries<Position.Component>(BasicQuery, BasicQuery));
+        }
+
+        [Test]
+        public void AddQueries_can_be_called_after_AddQueries()
```
jared-improbable: The name of this test may need a bit more clarity

test-project/Assets/EditmodeTests/Utility/InterestTemplateTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        [Test]
+        public void ClearQueries_clears_a_single_list_of_queries()
+        {
+            var interest = EmptyInterest
+                .AddQueries<Metadata.Component>(BasicQuery)
+                .AddQueries<Persistence.Component>(BasicQuery)
+                .AddQueries<Position.Component>(BasicQuery)
+                .ClearQueries<Metadata.Component>();
+
+            Assert.AreEqual(2, interest.AsComponentInterest().Keys.Count);
```
jared-improbable: Perhaps validate that the remaining queries are the expected ones?
Also, the test name seems to reference the internal implementation, it may need renaming.

test-project/Assets/EditmodeTests/Utility/InterestTemplateTests.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        [Test]
+        public void ClearQueries_can_be_used_with_empty_lists()
+        {
+            Assert.DoesNotThrow(() => EmptyInterest
+                .ClearQueries<Metadata.Component>());
+        }
+
+        [Test]
+        public void ClearAllQueries_clears_entire_dictionary()
```
jared-improbable: Test name references internal implementation (there are few more cases of this below, I won't repeat them all)

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -0,0 +1,117 @@
...
+namespace Improbable.Gdk.QueryBasedInterest
+{
+    /// <summary>
+    ///     Utility class to help construct ComponentInterest.Query objects.
+    /// </summary>
+    public class InterestQuery
+    {
+        private ComponentInterest.Query query;
+
+        // Hides the default constructor
```
jared-improbable: Redundant comment, I feel

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -0,0 +1,117 @@
...
+        private InterestQuery()
+        {
+        }
+
+        /// <summary>
+        ///     Creates an InterestQuery.
+        /// </summary>
+        /// <param name="constraint">
+        ///     A Constraint object defining the constraints of the query.
```
jared-improbable: You could make `Constraint` a `<cref>` to allow cross-linking to that type's reference docs (same for the return type)

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <summary>
+        ///     Sets the maximum frequency of the query.
+        /// </summary>
+        /// <param name="frequency">
+        ///     The maximum frequency to return query results.
+        /// </param>
+        /// <returns>
+        ///     An updated InterestQuery object.
+        /// </returns>
+        public InterestQuery WithMaxFrequencyHz(float frequency)
```
jared-improbable: Tag the Hz onto the field/parameters, too

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -0,0 +1,117 @@
...
+                    Constraint = constraint.AsQueryConstraint(),
+                    FullSnapshotResult = true,
+                    ResultComponentId = new List<uint>()
+                }
+            };
+            return interest;
+        }
+
+        /// <summary>
+        ///     Sets the maximum frequency of the query.
```
jared-improbable: Mention Hz here, too

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -0,0 +1,117 @@
...
+        ///     Defines what components to return in the query results.
+        /// </summary>
+        /// <param name="resultComponentId">
+        ///     First ID of a component to return from the query results.
+        /// </param>
+        /// <param name="resultComponentIds">
+        ///     Further IDs of components to return from the query results.
+        /// </param>
+        /// <remarks>
+        ///     At least one component ID must be provided.
```
jared-improbable: Isn't this redundant, given the signature of the function?

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
```diff
@@ -0,0 +1,117 @@
...
+        ///     if resultComponentIds is empty.
+        /// </remarks>
+        /// <returns>
+        ///     An updated InterestQuery object.
+        /// </returns>
+        public InterestQuery FilterResults(IEnumerable<uint> resultComponentIds)
+        {
+            if (!resultComponentIds.Any())
+            {
+                Debug.LogWarning("At least one InterestQuery must be provided to filter a query's results.");
```
jared-improbable: I can't find the conversation atm, what was the logic for deciding when we throw exceptions and when we log warnings?

paulbalaji: We log warnings in InterestTemplate and InterestQuery because providing an IEnumerable with 0 elements can be safely continued with a no-op

However with Constraint, providing an IEnumerable with 0 elements into Any or All would mean that the constraint being set would be an empty list - which is equivalent to not having a constraint set

The exception throwing is done because an empty QueryConstraint is considered an invalid by the runtime, so we want to stop users making that mistake as early as possible

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/InterestQuery.cs
paulbalaji: Perhaps yes, but I wanted to keep the remarks somewhat consistent between functions where this is implied by the signature and those that just take the `IEnumerable`

workers/unity/Packages/com.improbable.gdk.querybasedinteresthelper/Improbable.Gdk.QueryBasedInterestHelper.asmdef
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "Improbable.Gdk.QueryBasedInterestHelper",
+    "references": [
+        "Improbable.Gdk.Core",
+        "Improbable.Gdk.Generated",
+        "Unity.Entities",
```
zeroZshadow: What does it need all of these for?

PR #655, opened 2018-12-17
adding async method to send commands

jessicafalk:
#### Description
Added a potential solution on how to do task-based monobehaviour command senders. 
The generated code inside the `CommandSender` looks like this:
```
public async Task<Playground.CubeSpawner.SpawnCube.ReceivedResponse> SendSpawnCubeCommand(CubeSpawner.SpawnCube.Request request)
{
        var requestId = commandSender.SendCommand(request, entity);
        var taskSource = new TaskCompletionSource<Playground.CubeSpawner.SpawnCube.ReceivedResponse>();
        taskSystem.RegisterTask(requestId, taskSource);
        return await taskSource.Task;
}
```

#### Tests
tested locally and on cloud

#### Documentation
no documentation so far

#### Primary reviewers
@samiwh

### DIFF
--- a/workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
+++ b/workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
@@ -1,4 +1,3 @@
-using System.Threading.Tasks;
 using Improbable.Common;
 using Improbable.Gdk.Core;
 using Improbable.Worker.CInterop;
@@ -27,23 +26,7 @@ namespace Playground.MonoBehaviours
         private ILogDispatcher logDispatcher;
         private EntityId ownEntityId;
 
-        private void OnSpawnCubeResponse(CubeSpawner.SpawnCube.ReceivedResponse response)
-        {
-            if (response.StatusCode != StatusCode.Success)
-            {
-                //logDispatcher.HandleLog(LogType.Error, new LogEvent($"Spawn error: {response.Message}"));
-            }
-        }
-
-        private void OnDeleteSpawnedCubeResponse(CubeSpawner.DeleteSpawnedCube.ReceivedResponse response)
-        {
-            if (response.StatusCode != StatusCode.Success)
-            {
-                //logDispatcher.HandleLog(LogType.Error, new LogEvent($"Delete error: {response.Message}"));
-            }
-        }
-
-        private async void Update()
+        private void Update()
         {
             if (Input.GetKeyDown(KeyCode.Space))
             {
--- a/workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
+++ b/workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
@@ -1,3 +1,4 @@
+using System.Threading.Tasks;
 using Improbable.Common;
 using Improbable.Gdk.Core;
 using Improbable.Worker.CInterop;
@@ -42,26 +43,26 @@ namespace Playground.MonoBehaviours
             }
         }
 
-        private void Update()
+        private async void Update()
         {
             if (Input.GetKeyDown(KeyCode.Space))
             {
-                SendSpawnCubeCommand();
+                await SendSpawnCubeCommand();
             }
 
             if (Input.GetKeyDown(KeyCode.Backspace))
             {
-                SendDeleteCubeCommand();
+                await SendDeleteCubeCommand();
             }
         }
 
-        private void SendSpawnCubeCommand()
+        private async Task SendSpawnCubeCommand()
         {
             var request = new CubeSpawner.SpawnCube.Request(entityId, new Empty());
-            cubeSpawnerCommandSender.SendSpawnCubeCommand(request, OnSpawnCubeResponse);
+            await cubeSpawnerCommandSender.SendSpawnCubeCommand(request);
         }
 
-        private void SendDeleteCubeCommand()
+        private async Task SendDeleteCubeCommand()
         {
             var spawnedCubes = cubeSpawnerReader.Data.SpawnedCubes;
 
@@ -75,7 +76,7 @@ namespace Playground.MonoBehaviours
             {
                 CubeEntityId = spawnedCubes[0]
             });
-            cubeSpawnerCommandSender.SendDeleteSpawnedCubeCommand(request, OnDeleteSpawnedCubeResponse);
+            await cubeSpawnerCommandSender.SendDeleteSpawnedCubeCommand(request);
         }
     }
 }
--- a/workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
+++ b/workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
@@ -47,22 +47,23 @@ namespace Playground.MonoBehaviours
         {
             if (Input.GetKeyDown(KeyCode.Space))
             {
-                await SendSpawnCubeCommand();
+                SendSpawnCubeCommand();
             }
 
             if (Input.GetKeyDown(KeyCode.Backspace))
             {
-                await SendDeleteCubeCommand();
+                SendDeleteCubeCommand();
             }
         }
 
-        private async Task SendSpawnCubeCommand()
+        private async void SendSpawnCubeCommand()
         {
             var request = new CubeSpawner.SpawnCube.Request(entityId, new Empty());
-            await cubeSpawnerCommandSender.SendSpawnCubeCommand(request);
+            var response = await cubeSpawnerCommandSender.SendSpawnCubeCommand(request);
+            Debug.Log($"Message: {response.Message}, Request Id: {response.RequestId}");
         }
 
-        private async Task SendDeleteCubeCommand()
+        private async void SendDeleteCubeCommand()
         {
             var spawnedCubes = cubeSpawnerReader.Data.SpawnedCubes;
 
--- a/workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
+++ b/workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
@@ -1,6 +1,7 @@
+using System;
+using System.Threading.Tasks;
 using Improbable.Common;
 using Improbable.Gdk.Core;
-using Improbable.Worker.CInterop;
 using Improbable.Gdk.Subscriptions;
 using Unity.Entities;
 using UnityEngine;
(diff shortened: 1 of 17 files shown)


workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityCommandSenderReceiverGenerator.tt
```diff
@@ -286,6 +287,20 @@ namespace <#= qualifiedNamespace #>
...
         }
+
+        public async Task<<#= receivedCommandResponseType #>> Send<#= commandDetails.CommandName #>Command(<#= commandRequest #> request)
+        {
+            var requestId = commandSender.SendCommand(request, entity);
+            <#= receivedCommandResponseType #> response = default(<#= receivedCommandResponseType #>);
+            await Task.Run(() =>
+            {
+                while (!commandSender.TryGetReceivedResponseForRequestId(requestId, out response))
+                {
```
jessicafalk: we might want to add the option to add a `Task.Delay(..)` inbetween so that we are not polling all the time

jared-improbable: This busy waiting is definitely something to avoid.
Why does the `commandSender` not also have an `async` method to await the response (keep pushing async down as far as it makes sense.)
You could have an `AutoResetEvent` that is signaled when the command response op comes through and is stored in your table - http://dotnetpattern.com/threading-autoresetevent

jessicafalk: completely rewritten it after some more discussion. Now using `TaskCompletionSource` to get rid of `Task.Run` and the while loop

workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
```diff
@@ -42,26 +43,26 @@ private void OnDeleteSpawnedCubeResponse(CubeSpawner.DeleteSpawnedCube.ReceivedR
             }
         }
 
-        private void Update()
+        private async void Update()
         {
             if (Input.GetKeyDown(KeyCode.Space))
             {
-                SendSpawnCubeCommand();
+                await SendSpawnCubeCommand();
```
samiwh: I don't think you want to await this as it will stop this from reading all other input. You want the handling of the response to be after await.

workers/unity/Packages/com.improbable.gdk.core/Commands/WorldCommands/CreateEntity.cs
```diff
@@ -441,6 +444,17 @@ public List<ReceivedResponse> GetResponsesReceivedForEntity(Unity.Entities.Entit
                     return false;
                 }
 
+                public void RegisterResponseTask(long requestId, TaskCompletionSource<ReceivedResponse> task)
```
gkassabli: Not very familiar with C#/Unity threading model - are `RegisterResponseTask` and `AddResponse` always called from the same thread?
If not, is there a chance of race condition when we try to mutate `internalRequestIdToReceivedResponseTask` dictionary from different threads at the same time?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
```diff
@@ -42,26 +43,26 @@ private void OnDeleteSpawnedCubeResponse(CubeSpawner.DeleteSpawnedCube.ReceivedR
...
 
             if (Input.GetKeyDown(KeyCode.Backspace))
             {
-                SendDeleteCubeCommand();
+                await SendDeleteCubeCommand();
             }
         }
 
-        private void SendSpawnCubeCommand()
+        private async Task SendSpawnCubeCommand()
```
samiwh: Related to the above I think we can leave this is async void. But here we aren't actually doing anything with the response so there isn't a reason to await. Could just add a debug log if we want this is an example.

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityCommandSenderReceiverGenerator.tt
```diff
@@ -286,6 +289,14 @@ namespace <#= qualifiedNamespace #>
                 callbackSystem.RegisterCommandResponseCallback(requestId, wrappedCallback);
             }
         }
+
+        public async Task<<#= receivedCommandResponseType #>> Send<#= commandDetails.CommandName #>Command(<#= commandRequest #> request)
```
jared-improbable: Usually an API will provide a <Function>Async variant that returns the `Task<>` and allows users the ability to decide if they `await` or not, e.g. `SendCommandNameAsync`

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/Systems/CommandTaskSystem.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Improbable.Gdk.Subscriptions
+{
+    [DisableAutoCreation]
+    public class CommandTaskSystem : ComponentSystem
+    {
+        private readonly Dictionary<Type, ITaskManager> taskManagers =
+            new Dictionary<Type, ITaskManager>();
+
+        public void RegisterTask<T>(long requestId, TaskCompletionSource<T> task)
```
samiwh: It might be preferable to have this return a Task for a given ID. That way you can hide that it's actually a `TaskCompletionSource` that is being used. If at all possible, present a user with a basic `Task`.

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/TaskManagers/CommandTaskManager.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public void CompleteTasks()
+        {
+            var responses = commandSystem.GetResponses<T>();
+            foreach (var response in responses)
+            {
+                var requestId = response.GetRequestId();
+                if (registeredTasks.TryGetValue(requestId, out var task))
+                {
+                    task.SetResult(response);
```
samiwh: I think I'd expect this to set and exception in the case the command failed. Then we would probably have to have the reason and context as members of the exception.

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/TaskManagers/ITaskManager.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Improbable.Gdk.Subscriptions
+{
+    public interface ITaskManager
```
samiwh: nit that doesn't really matter for this - can likely make this internal

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityCommandSenderReceiverGenerator.tt
samiwh: I think this does this. If you call this without an await, you will just get a task object and continue synchronously (although it will deadlock if you call call `task.Result`. But that is a function of the task being run in the same thread rather than this function).

Agreed this should be called `SendCommandNameAsyc`.

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/Systems/CommandTaskSystem.cs
samiwh: Basically I think this might make more sense as an async command sender, as that can be used for more than just the subscriptions stuff. However for the purposes of this exercise it's not really important.

samiwh: The remaining issue with that is the lifecycle stuff. But that should be fixed by making it so that subscribed information is published at the right time. Rather than them registering with some central thing. (Not for this PR, it's a mistake in how subscriptions is currently setup).

samiwh: (Which is why the non-constraint callback managers are also setup like this)

workers/unity/Assets/Playground/Scripts/MonoBehaviours/CubeSpawnerInputBehaviour.cs
```diff
@@ -55,13 +39,14 @@ private void Update()
             }
         }
 
-        private void SendSpawnCubeCommand()
+        private async void SendSpawnCubeCommand()
         {
             var request = new CubeSpawner.SpawnCube.Request(entityId, new Empty());
-            cubeSpawnerCommandSender.SendSpawnCubeCommand(request, OnSpawnCubeResponse);
+            var response = await cubeSpawnerCommandSender.SendSpawnCubeCommand(request);
```
samiwh: If failures set an exception then this should have a try catch block.

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/TaskManagers/CommandTaskManager.cs
```diff
@@ -0,0 +1,46 @@
...
+        public CommandTaskManager(World world)
+        {
+            commandSystem = world.GetExistingManager<CommandSystem>();
+        }
+
+        public void RegisterTask(long requestId, TaskCompletionSource<T> task)
+        {
+            if (registeredTasks.ContainsKey(requestId))
+            {
+                task.SetException(new ArgumentException("Request Id is already registered."));
```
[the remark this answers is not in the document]
jessicafalk: good question... I guess you could register multiple tasks for a request id on this layer and then trigger all of them. 
though is this a behaviour that we would expect to happen as you usually only have the request id, if you have sent the command

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/Systems/CommandTaskSystem.cs
jessicafalk: oh yeah, I like the idea of hiding the `TaskCompletionSource` and agree that the lifecycle system, while convenient, makes it a bit more tricky to add custom functionality to it.

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/TaskManagers/CommandTaskManager.cs
jessicafalk: hm... I don't really agree with that. If we set this as an exception, it would mean that we except failed commands to throw instead of simply checking the status code. Seeing as even a bad network connection / timeout could result in a failed command, I don't think it's a good idea to throw every time. It is expected behaviour after all

jessicafalk:
Closing this for now as it was only made to validate the changes in the `experimental/core-refactor` and as a general proof of concept for async command senders.

PR #656, opened 2018-12-19
Add cmd option to choose il2cpp/mono in build tests

jessicafalk:

-------

#### Description
Added an optional parameter to enable switching the scripting backend to either il2cpp or mono. This is only exposed when building via cmd. When running it via the menu items in the editor the one that is active in the player settings will be chosen.

#### Tests
ran it locally. next step is to test it in ci. 

#### Documentation
no documentation needed

#### Primary reviewers
@firtina-improbable @paulbalaji

### DIFF
--- a/ci/bootstrap.sh
+++ b/ci/bootstrap.sh
@@ -13,6 +13,3 @@ if [ -d "${SHARED_CI_DIR}" ]; then
 fi
 
 git clone "${CLONE_URL}" "${SHARED_CI_DIR}"
-pushd "${SHARED_CI_DIR}"
-	git checkout feature/scripting-backend
-popd
--- a/ci/bootstrap.sh
+++ b/ci/bootstrap.sh
@@ -13,3 +13,6 @@ if [ -d "${SHARED_CI_DIR}" ]; then
 fi
 
 git clone "${CLONE_URL}" "${SHARED_CI_DIR}"
+pushd "${SHARED_CI_DIR}"
+	git checkout feature/scripting-backend
+popd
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -115,7 +129,24 @@ public static void BuildWorkerForEnvironment(string workerType, BuildEnvironment
...
+                    {
+                        Debug.Log($"Setting scripting backend to {scriptingBackend.Value}");
+                        PlayerSettings.SetScriptingBackend(buildTargetGroup, scriptingBackend.Value);
+                    }
+                    
+                    BuildWorkerForTarget(workerType, unityBuildTarget, buildOptions, targetEnvironment);
+                }
+                catch (Exception e)
+                {
+                    PlayerSettings.SetScriptingBackend(buildTargetGroup, activeScriptingBackend);
```
gkassabli: Should we move this to finally block? Seems a bit odd that we only restore scripting backend if we throw

ci/bootstrap.sh
```diff
@@ -13,3 +13,6 @@ if [ -d "${SHARED_CI_DIR}" ]; then
 fi
 
 git clone "${CLONE_URL}" "${SHARED_CI_DIR}"
+pushd "${SHARED_CI_DIR}"
```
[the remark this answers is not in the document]
paulbalaji: Plan is to keep it here for now to test with the shared-ci changes in TC. If that succeeds, then to merge shared-ci changes, remove this particular bootstrap.sh change, and re-run through TC before merging

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
jessicafalk: good spot, that was meant to be in the finally part

jessicafalk:
don't think this is mentioned anywhere in the docs right now, but will add the changelog to my other open pr :P

PR #657, opened 2018-12-19
Split the component send system from the reactive one

samiwh:
Split regular component updates from the reactive one. Which means we can now split out reactive components into their own feature module.


workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentSendSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            foreach (var replicator in componentReplicators)
+            {
+                Profiler.BeginSample("ExecuteReplication");
+                replicator.Handler.SendUpdates(replicator.Group, this, EntityManager, componentUpdateSystem);
+                Profiler.EndSample();
+            }
+        }
+
+        internal void AddComponentReplicator(IComponentReplicationHandler reactiveComponentReplicationHandler)
```
jessicafalk: should that variable be called`componentReplicationHandler`? assuming that this one is not handling reactive components

workers/unity/Packages/com.improbable.gdk.core/Systems/CustomSpatialOSSendSystem.cs
```diff
@@ -21,19 +21,19 @@ public abstract class CustomSpatialOSSendSystem<T> : ComponentSystem where T : I
 
         protected WorkerSystem WorkerSystem;
 
-        private SpatialOSSendSystem spatialOSSendSystem;
+        private ReactiveComponentSendSystem reactiveComponentSendSystem;
```
jessicafalk: do we need the reactive component send system for custom send systems?
Why not use the ComponentSendSystem?

workers/unity/Packages/com.improbable.gdk.core/Systems/ReactiveComponentSendSystem.cs
```diff
@@ -48,31 +49,34 @@ public bool TryRegisterCustomReplicationSystem(uint componentId)
 
         protected override void OnUpdate()
         {
+            var commandSystem = World.GetExistingManager<CommandSystem>();
```
jessicafalk: nit: move these variables after the " is connection null" check

workers/unity/Packages/com.improbable.gdk.core/Systems/ReactiveComponentSendSystem.cs
```diff
@@ -106,8 +109,8 @@ private void PopulateDefaultComponentReplicators()
         private struct ComponentReplicator
         {
             public uint ComponentId;
-            public ComponentReplicationHandler Handler;
-            public ComponentGroup UpdateGroup;
+            public IReactiveComponentReplicationHandler Handler;
+            public ComponentGroup EventGroup;
```
jessicafalk: Are events and commands still using reactive components?
if yes, are there plans to move those to non-reactive components as well?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentSenderGenerator.tt
```diff
@@ -0,0 +1,85 @@
...
+                foreach (var chunk in chunkArray)
+                {
+                    var entityIdArray = chunk.GetNativeArray(spatialOSEntityType);
+                    var componentArray = chunk.GetNativeArray(componentType);
+
+                    var authorityIndex = chunk.GetSharedComponentIndex(authorityType);
+
+                    if (!entityManager.GetSharedComponentData<ComponentAuthority>(authorityIndex).HasAuthority)
+                    {
+                        continue;
```
jessicafalk: should we print a warning here? or do we expect this to happen when e.g. authority just changed?

workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentSendSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+    [UpdateInGroup(typeof(SpatialOSSendGroup.InternalSpatialOSSendGroup))]
+    [UpdateBefore(typeof(ComponentUpdateSystem))]
+    public class ComponentSendSystem : ComponentSystem
+    {
+        // Can't access the generated component ID in Core code.
+        private const uint PositionComponentId = 54;
+
+        private Connection connection;
+
+        private readonly List<ComponentReplicator> componentReplicators =
```
jessicafalk: super nit: move the readonly variable before the connection variable

workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentSendSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            // The default replication system is removed, instead the custom one is responsible for replication.
+            return componentReplicators.Remove(componentReplicators.First(
+                componentReplicator => componentReplicator.ComponentId == componentId));
+        }
+
+        protected override void OnUpdate()
+        {
+            var componentUpdateSystem = World.GetExistingManager<ComponentUpdateSystem>();
```
jessicafalk: same as for the other system: should we move that to after the "is connection null" check

workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentSendSystem.cs
samiwh: Ah yep. This was copy pasted from the old ones. Will move it.

workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentSendSystem.cs
samiwh: Sure

workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentSendSystem.cs
samiwh: Copy paste error. Will change.

workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentSendSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+                {
+                    if (typeof(IComponentReplicationHandler).IsAssignableFrom(type) && !type.IsAbstract)
+                    {
+                        if (type.GetCustomAttribute(typeof(DisableAutoRegisterAttribute)) != null)
+                        {
+                            continue;
+                        }
+
+                        var handler =
+                            (IComponentReplicationHandler) Activator.CreateInstance(type);
```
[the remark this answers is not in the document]
samiwh: Yep. Will change

workers/unity/Packages/com.improbable.gdk.core/Systems/CustomSpatialOSSendSystem.cs
samiwh: Refactor error. Will fix.

workers/unity/Packages/com.improbable.gdk.core/Systems/ReactiveComponentSendSystem.cs
samiwh: The old reactive components are all still there if you enable them. What are non reactive components? Do you mean just have potentially empty static lists on things? If so that can be added as a feature module on top of the new stuff added in.

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentSenderGenerator.tt
samiwh: Thing the above should say - this check is applied per chunk, not per entity.

workers/unity/Packages/com.improbable.gdk.core/Systems/ReactiveComponentSendSystem.cs
jessicafalk: nvm that, I think I was slightly confused back then. If I understand it correctly, we now have two ways to process commands, right? One is via the command managers and the other one is via reactive components.

One question that I have regarding that: How would that work in an ECS system? I assume if we don't want to use reactive components, we would need to get a hold of the correct managers and query them all time to check whether we received a request / response

samiwh: Basically yes. It's not the ideal API but it's meant to be a low level API that we make other things from.

PR #660, opened 2018-12-20
anonymous auth for mobile

jessicafalk:
#### Description
- Added anonymous functionality to the workerconnector. 
- Adjusted scenes to allow to choose between connecting locally or via cloud. we have now 2 buttons to select whether you want to connect to a local or a cloud deployment
android

#### Tests
Tested on both iOS and Android. Works as expected

#### Documentation
docs will follow in a separate PR

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -87,6 +87,7 @@ public async Task Connect(string workerType, ILogDispatcher logger)
                 var origin = transform.position;
                 ConnectionDelegate connectionDelegate;
                 var chosenService = GetConnectionService();
+                Debug.Log(chosenService);
```
jared-improbable: Remove debug print

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
```diff
@@ -34,17 +39,72 @@ protected override ReceptionistConfig GetReceptionistConfig(string workerType)
...
+                {
+                    DevelopmentAuthenticationTokenId = DevelopmentAuthToken,
+                    PlayerId = $"Player-{Guid.NewGuid()}",
+                    DisplayName = "",
+                }
+            ).Get().Value;
+
+            if (!string.IsNullOrEmpty(pit.Error))
+            {
+                throw new ConnectionFailedException(pit.Error, ConnectionErrorReason.InvalidConfig);
```
jared-improbable: Is `InvalidConfig` actually a valid reason for failure in every case, here and below?
What if it's a different reason, for example if the service is down?

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
```diff
@@ -34,17 +39,72 @@ protected override ReceptionistConfig GetReceptionistConfig(string workerType)
...
+                }
+            ).Get();
+
+            if (!loginTokenRequestResult.HasValue)
+            {
+                throw new ConnectionFailedException("Did not receive any login tokens back.", 
+                    ConnectionErrorReason.InvalidConfig);
+            }
+
+            if (!loginTokenRequestResult.Value.Status.Equals(ConnectionStatusCode.Success))
```
jared-improbable: Why `! .Equals`, rather than `!=`?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -196,6 +207,67 @@ protected virtual string SelectDeploymentName(DeploymentList deployments)
...
+            }
+
+            return result.Value.PlayerIdentityToken;
+        }
+
+        protected virtual List<LoginTokenDetails> RetrieveLoginToken(string workerType, string playerIdentityToken)
+        {
+            var loginTokenRequestResult = DevelopmentAuthentication.CreateDevelopmentLoginTokensAsync(
+                RuntimeConfigDefaults.LocatorHost,
+                444,
```
gkassabli: DevelopmentPlayerIdentityTokenPort?

workers/unity/Assets/Playground/Scripts/Worker/AndroidClientWorkerConnector.cs
```diff
@@ -17,9 +19,11 @@ public class AndroidClientWorkerConnector : MobileWorkerConnector, IMobileConnec
         [SerializeField] private GameObject level;
 
         private GameObject levelInstance;
+        private ConnectionService connectionService;
 
-        public async void TryConnect()
+        public async void TryConnect(ConnectionService connectionService)
```
jamiebrynes7: I think @samiwh noted on this in another PR - but should async methods have async in the name. I.e. - `TryConnectAsync`

workers/unity/Assets/Playground/Scripts/Worker/iOSClientWorkerConnector.cs
```diff
@@ -55,6 +59,26 @@ protected override string GetHostIp()
 #endif
         }
 
+        protected override string GetPlayerId()
```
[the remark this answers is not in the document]
jamiebrynes7: There's definitely some code duplication here (and in the Android one) that could be factored down to the parent class

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -196,6 +207,67 @@ protected virtual string SelectDeploymentName(DeploymentList deployments)
             return null;
         }
 
+        protected virtual string SelectLoginToken(List<LoginTokenDetails> loginTokens)
+        {
+            return null;
+        }
```
jamiebrynes7: Need docstrings on the new `protected virtual` methods in the class

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -186,6 +187,16 @@ await Worker.CreateWorkerAsync(GetAlphaLocatorConfig(), connectionParameters, lo
...
 
+        protected virtual string GetPlayerId()
+        {
+            return null;
+        }
+
+        protected virtual string GetDisplayName()
+        {
+            return null;
+        }
```
jamiebrynes7: I know this similar to other functions defined here - but I feel we should either provide a reasonable default implementation or make these fully `abstract`.

Returning null as a default implementation seems likely to break things later on.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -196,6 +207,67 @@ protected virtual string SelectDeploymentName(DeploymentList deployments)
...
+                {
+                    DevelopmentAuthenticationTokenId = authToken,
+                    PlayerId = playerId,
+                    DisplayName = displayName,
+                }
+            ).Get();
+
+            if (!result.HasValue)
+            {
+                throw new ConnectionFailedException("Did not receive a player identity token.",
```
jamiebrynes7: This seems like the wrong exception to throw here and elsewhere in this function and `RetrieveLoginToken`.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -196,6 +207,67 @@ protected virtual string SelectDeploymentName(DeploymentList deployments)
...
+
+            if (!loginTokenRequestResult.HasValue)
+            {
+                throw new ConnectionFailedException("Did not receive any login tokens back.",
+                    ConnectionErrorReason.InvalidConfig);
+            }
+
+            if (loginTokenRequestResult.Value.Status != ConnectionStatusCode.Success)
+            {
+                throw new ConnectionFailedException($"Failed to retrieve login token, " +
```
jamiebrynes7: I'd rather have a slightly longer line length than the string concatenation here.

Mild preference though

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/LaunchArguments.cs
```diff
@@ -27,7 +27,7 @@ public static class LaunchArguments
             }
             catch (Exception e)
             {
-                UnityEngine.Debug.LogException(e);
+                UnityEngine.Debug.LogWarning($"Failed to retrieve launch arguments: {e}");
```
jamiebrynes7: Exception -> Warning? What's the context here?

jessicafalk: this will always fail, if you run this in the editor after switching to Android as current platform

therefore, it is expected behaviour to fail depending on what exactly you do and it shouldn't throw an exception imo

jamiebrynes7: Should this function even be able to be called in the editor then if it will always fail?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
jessicafalk: would you prefer to have a new type of exception that is only related to the anonymous auth flow?

jamiebrynes7: Hmmm - either that or a new element in the `ConnectionErrorReason` like `AuthenticationFailed` or similar.

jamiebrynes7:
Also changelog 😛

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/LaunchArguments.cs
jessicafalk: fair point. we could also change the part where it's called to only get called if android and not editor. or wrap this part here around an `#if UNITY_EDITOR`, which I would slightly prefer as it makes it safe to call for users that are not familiar with the code.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
jessicafalk: talked offline, already had added that element in the enum, but we agree that the anon auth flow is big enough to create a new exception for this part of the connection flow. gonna change that

spatialos.json
```diff
@@ -1,8 +1,8 @@
 {
     "name": "unity_gdk",
     "project_version": "0.0.1",
-    "sdk_version": "13.5.0-b7795-fdd6a-WORKER-SNAPSHOT",
+    "sdk_version": "13.5.0",
```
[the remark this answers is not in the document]
zeroZshadow: Does this not break our tracking?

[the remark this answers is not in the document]
jessicafalk: as discussed internally: as long as we are not doing a new release it should be ok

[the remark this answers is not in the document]
jamiebrynes7: Yup - ticket to track: https://improbableio.atlassian.net/browse/UTY-1580

workers/unity/Packages/com.improbable.gdk.core/Exceptions/AuthenticationFailedException.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    ///     Represents an error that occurs when the local state of a SpatialOS entity is not valid.
+    /// </summary>
+    internal class AuthenticationFailedException : Exception
```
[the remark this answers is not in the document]
jessicafalk: cause we also had internal for other exceptions, just copy-pasted it. but can change it to public

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -27,17 +28,20 @@ public void Awake()
             if (!string.IsNullOrEmpty(hostIp))
             {
                 ipAddressInput.text = hostIp;
-                TryConnect();
```
gkassabli: Is this intentional change? If I understand correctly this was here to automatically connect to host locally when we launch with command line arguments

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/LaunchArguments.cs
```diff
@@ -8,6 +8,7 @@ public static class LaunchArguments
     {
         public static Dictionary<string, string> GetArguments()
         {
+#if !UNITY_EDITOR
```
gkassabli: Was this malfunctioning in Unity Editor?

jessicafalk: Yes, as jamie and I were discussing in one of the previous threads in this PR: you can set your current platform to Android in the Editor, which will cause this code to be run. However using these java objects will only succeed in an actual android device/emulator. You end up with a nullpointer exception. 

So either
- we don‘t throw an exception
- we don‘t run this code in the editor

We decided that not running it seems like the better option, because this code shouldn‘t fail on an actual android device

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
jessicafalk: This is intentional, should have probably commented that earlier. Now that we can either connect to a local or a cloud deployment, we don‘t want to try and immediately connect to a local deployment all the time. Sure, we could remove the ip in the config, but I think this is a nicer solution as the other one would mean that you would have to look up for local ip again

gkassabli: Fine with me if this is explicit change.
I'd point out that this was only triggered from launch menu, where I'd expect people mostly launch for local, but I understand the reasoning.
I think this is another pointer that we need to rethink how we implement client launch configuration so that it is more transparent and processing logic is easier to separate out.

workers/unity/Packages/com.improbable.gdk.core/Exceptions/AuthenticationFailedException.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Improbable.Gdk.Core
+{
+    /// <summary>
+    ///     Represents an error that occurs when the local state of a SpatialOS entity is not valid.
```
jared-improbable: This doc doesn't seem to reflect the exception name

workers/unity/Packages/com.improbable.gdk.mobile/Android/Utility/LaunchArguments.cs
jared-improbable: Would it be better to wrap this in `Application.isEditor` rather than compiling it out?

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
jessicafalk: I'm not too sure about this. As it is not required to upload the android clients to the cloud deployment, I could imagine that people want to quickly iterate and test changes and try them out in a running cloud deployment.  I simply don't think we should try to immediately connect, if there are now two ways to connect, but this is all code in the playground so developers will likely use a different interface to decide how to connect.

Not sure I understand what you mean by the second point. Do you mean how we decide whether to connect to a local or a cloud deployment?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -196,6 +215,89 @@ protected virtual string SelectDeploymentName(DeploymentList deployments)
...
+                    DisplayName = displayName,
+                }
+            ).Get();
+
+            if (!result.HasValue)
+            {
+                throw new AuthenticationFailedException("Did not receive a player identity token.");
+            }
+
+            if (!string.IsNullOrEmpty(result.Value.Error))
```
jamiebrynes7: I think you should only check for `null` as the underlying C API returns a null pointer if there is no error. Not sure how this gets transformed into the C# struct. cc @davedissian

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -196,6 +215,89 @@ protected virtual string SelectDeploymentName(DeploymentList deployments)
...
+            {
+                throw new AuthenticationFailedException(result.Value.Error);
+            }
+
+            return result.Value.PlayerIdentityToken;
+        }
+
+        /// <summary>
+        ///     Retrieves the login tokens of all active deployments that the player
```
jamiebrynes7: Nit: `Retrieves the login tokens for all active deployments...`

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
gkassabli: Agree, this approach is more flexible.
For my second point I have a general feeling that we can improve the launch configuration and processing both in tooling and in Playground as a sample, but I don't have concrete suggestions as of the moment. Will try to think more about this and maybe come up with something more concrete.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
jessicafalk: Checked with David, we can actually just use the status code

PR #662, opened 2018-12-21
Prerequisites for Deployment Manager

davedissian:

-------

#### Description
Essentially, modified `RedirectedProcess` to support running a process asynchronously

#### Tests
Testing manually.

#### Documentation
n/a

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -1,13 +1,22 @@
...
 using Improbable.Gdk.Tools.MiniJSON;
 using UnityEngine;
 using Debug = UnityEngine.Debug;
 
 namespace Improbable.Gdk.Tools
 {
+    public class RedirectedProcessResult
+    {
+        public int ExitCode;
+        public List<string> Stdout;
```
zeroZshadow: nit: Why a list over a Stack?

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -95,6 +104,89 @@ void OnReceived(object sender, DataReceivedEventArgs args)
...
+                {
+                    if (process == null)
+                    {
+                        throw new Exception(
+                            $"Failed to run {info.FileName} {info.Arguments}\nIs the .NET Core SDK installed?");
+                    }
+
+                    process.EnableRaisingEvents = true;
+
+                    var processStandardOutput = new List<string>();
```
jared-improbable: A `StringBuilder` will be much more efficient.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
davedissian: No reason. I'll change it.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
davedissian: Yeah agreed (that's what used to be there). However, the design decision I made here is that I wanted to make sure that separate lines of output were in a list rather than CRLF characters in the string itself. As it's not exactly expensive to convert a collection of strings to a single string with `string.Join` (or whatever it is) for general output sizes of 10's of lines of output, I figured it was worth it, to avoid having to split a single string into multiple strings later.

jared-improbable: Why do you need individual lines of text?

davedissian: In the deployment manager thing I wrote in a different PR, I need to invoke a binary using the Platform SDK to obtain a list of deployments. This binary (an invocation of `dotnet run`) outputs a list of deployment_name deployment_id pairs, and it's far easier to process this without having to split across newlines.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
davedissian: Wait, this probably shouldn't be a stack. FILO semantics don't really make sense for log messages.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -95,6 +104,89 @@ void OnReceived(object sender, DataReceivedEventArgs args)
...
+                    var processStandardError = new List<string>();
+
+                    void OnStandardOutput(object sender, DataReceivedEventArgs args)
```
samiwh: Trying to get every C# feature into this function :P.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -95,6 +104,89 @@ void OnReceived(object sender, DataReceivedEventArgs args)
...
+
+                    void OnStandardError(object sender, DataReceivedEventArgs args)
+                    {
+                        if (!string.IsNullOrEmpty(args.Data))
+                        {
+                            if (redirectStderr)
+                            {
+                                Debug.LogError(args.Data);
+                            }
+                            lock (processStandardOutput)
```
samiwh: Can also set the process sync object to do this. I think I mildly prefer the lock though.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -1,13 +1,22 @@
...
 using System.IO;
 using System.Text;
+using System.Threading.Tasks;
 using Improbable.Gdk.Tools.MiniJSON;
 using UnityEngine;
 using Debug = UnityEngine.Debug;
 
 namespace Improbable.Gdk.Tools
 {
+    public class RedirectedProcessResult
```
samiwh: Nit - I think this might make more sense as a struct, although it makes almost no difference.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -95,6 +104,89 @@ void OnReceived(object sender, DataReceivedEventArgs args)
...
+                        }
+                    }
+
+                    process.OutputDataReceived += OnStandardOutput;
+                    process.ErrorDataReceived += OnStandardError;
+
+                    process.BeginOutputReadLine();
+                    process.BeginErrorReadLine();
+
+                    process.WaitForExit();
```
gkassabli: I am not 100% familiar with async handling in c#, but I am curious if WaitForExit would actually make this task async or just block the thread.
Should we use something like described in https://stackoverflow.com/questions/470256/process-waitforexit-asynchronously instead?

davedissian: This is already in an async context (see Task.Run above), so it wouldn't block the main thread.

PR #668, opened 2019-01-04
[UTY-1541]Adding functionality to start iOS device client from Unity menu item

gkassabli:

-------

#### Description
This PR adds functionality to run ios builds on device with pre-filled ip.
Implemented functionality assumes having pre-built .ipa iOS archive. Running is only supported on device: running on iOS simulator will be added in a separate diff

#### Tests
Ensure Spatial Unity menu has "Launch mobile client/iOS device" item => ensure that launches the client => ensure client is connection to predefined ip address

#### Documentation
https://github.com/spatialos/gdk-for-unity/pull/678

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Added
+- Added support for launching an iOS client from an iOS device from the Unity Editor. Navigate to **SpatialOS** > **Launch mobile client** > **iOS device** to try it out
+
 ### Changed
 
 - Changed `RedirectedProcess` to have Builder-like API.
--- a/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
+++ b/workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
@@ -1,4 +1,5 @@
 using Improbable.Gdk.Core;
+using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.UI;
 
@@ -98,14 +99,20 @@ namespace Playground
 
         private string GetReceptionistHostFromArguments()
         {
+            Dictionary<string, string> arguments = null;
 #if UNITY_ANDROID
-            var arguments = Improbable.Gdk.Mobile.Android.LaunchArguments.GetArguments();
+            arguments = Improbable.Gdk.Mobile.LaunchArguments.GetAndroidArguments();
+#elif UNITY_IOS
+            arguments = Improbable.Gdk.Mobile.LaunchArguments.GetiOSArguments();
+#endif
+            if (arguments == null)
+            {
+                return string.Empty;
+            }
+
             var hostIp =
                 CommandLineUtility.GetCommandLineValue(arguments, RuntimeConfigNames.ReceptionistHost, string.Empty);
             return hostIp;
-#else
-            return string.Empty;
-#endif
         }
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -115,10 +115,11 @@ namespace Improbable.Gdk.Mobile
                     .RedirectOutputOptions(OutputRedirectBehaviour.None).Run();
                 var existsOnDevice = appList.Contains(bundleId);
                 var existsLocally = !string.IsNullOrEmpty(ipaPath);
-                
-                // If .ipa exists - we install/upgrade device app and launch it
-                // If .ipa doesn't exist, but the bunlde is installed on device - we launch existing app with parameters
-                // If .ipa doesn't exist and bundle isn't found on device - we fail with an error.
+
+                // If .ipa exists and the bundle is installed - we upgrade device app and launch it
+                // If .ipa exists, but isn't installed - we install device app and launch it
+                // If .ipa doesn't exist, but the bundle is installed - we launch existing app with parameters
+                // If .ipa doesn't exist and bundle isn't found - we fail with an error.
                 if (!existsLocally && !existsOnDevice)
                 {
                     Debug.LogError($"Could not find an app on device or built out iOS .ipa archive in \"{AbsoluteAppBuildPath}\" to launch.");
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using System.Linq;
 using System.Text;
@@ -117,6 +118,7 @@ namespace Improbable.Gdk.Mobile
                 if (!existsLocally && !existsOnDevice)
                 {
                     Debug.LogError($"Could not find an app on device or built out iOS .ipa archive in \"{AbsoluteAppBuildPath}\" to launch.");
+                    return;
                 }
 
                 if (existsLocally)
@@ -131,9 +133,17 @@ namespace Improbable.Gdk.Mobile
                     }
                 }
 
-                // Wait until the app is installed
+                // Wait until the app has finished installing
+                DateTime timeout = DateTime.Now.AddSeconds(5);
                 while (RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-l").Run() != 0)
                 {
+                    if (DateTime.Now > timeout)
+                    {
+                        Debug.LogError($"Device communication error. Please ensure it is connected");
+                        return;
+                    }
+
+                    System.Threading.Thread.Sleep(500);
                 }
 
                 EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Launching Client", 0.9f);
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -1,4 +1,5 @@
 using System.IO;
+using System.Linq;
 using System.Text;
 using Improbable.Gdk.Core;
 using Improbable.Gdk.Tools;
@@ -9,13 +10,15 @@ namespace Improbable.Gdk.Mobile
 {
     public static class LaunchMenu
     {
-        private const string rootApkPath = "build";
-        private static string AbsoluteApkPath => Path.GetFullPath(Path.Combine(Application.dataPath, Path.Combine("..", rootApkPath)));
+        private static string AbsoluteAppBuildPath => Path.GetFullPath(Path.Combine(Application.dataPath, "..", "build"));
+        private static string LibIDeviceInstallerBinary => Common.DiscoverLocation("ideviceinstaller");
+        private static string LibIDeviceDebugBinary => Common.DiscoverLocation("idevicedebug");
 
         private const string MenuLaunchAndroid = "SpatialOS/Launch mobile client/Android Device";
+        private const string MenuLaunchiOSDevice = "SpatialOS/Launch mobile client/iOS Device";
 
         [MenuItem(MenuLaunchAndroid, false, 73)]
-        private static void LaunchMobileClient()
+        private static void LaunchAndroidClient()
         {
             try
             {
@@ -29,12 +32,13 @@ namespace Improbable.Gdk.Mobile
 
                 var adbPath = Path.Combine(sdkRootPath, "platform-tools", "adb");
 
-                EditorUtility.DisplayProgressBar("Launching Mobile Client", "Installing APK", 0.3f);
+                EditorUtility.DisplayProgressBar("Launching Android Client", "Installing APK", 0.3f);
 
                 // Find apk to install
-                if (!TryGetApkPath(AbsoluteApkPath, out var apkPath))
+                var apkPath = Directory.GetFiles(AbsoluteAppBuildPath, "*.apk", SearchOption.AllDirectories).FirstOrDefault();
+                if (apkPath == string.Empty)
                 {
-                    Debug.LogError($"Could not find a built out Android binary in \"{AbsoluteApkPath}\" to launch.");
+                    Debug.LogError($"Could not find a built out Android binary in \"{AbsoluteAppBuildPath}\" to launch.");
                     return;
                 }
 
@@ -53,7 +57,7 @@ namespace Improbable.Gdk.Mobile
                     return;
                 }
 
-                EditorUtility.DisplayProgressBar("Launching Mobile Client", "Launching Client", 0.9f);
+                EditorUtility.DisplayProgressBar("Launching Android Client", "Launching Client", 0.9f);
 
                 // Optional arguments to be passed, same as standalone
                 // Use this to pass through the local ip to connect to
@@ -70,8 +74,6 @@ namespace Improbable.Gdk.Mobile
                     .WithArgs("shell", "am", "start", "-S", "-n", $"{bundleId}/com.unity3d.player.UnityPlayerActivity",
                         "-e", "\"arguments\"", $"\\\"{arguments.ToString()}\\\"")
                     .Run();
-
-                EditorUtility.DisplayProgressBar("Launching Mobile Client", "Done", 1.0f);
             }
             finally
             {
@@ -79,16 +81,65 @@ namespace Improbable.Gdk.Mobile
             }
         }
 
-        private static bool TryGetApkPath(string rootPath, out string apkPath)
+        [MenuItem(MenuLaunchiOSDevice, false, 74)]
+        private static void LaunchiOSDeviceClient()
         {
-            foreach (var file in Directory.GetFiles(rootPath, "*.apk", SearchOption.AllDirectories))
+            try
             {
-                apkPath = file;
-                return true;
-            }
+                // Ensure needed tools are installed
+                if (string.IsNullOrEmpty(LibIDeviceInstallerBinary))
+                {
+                    Debug.LogError("Could not find ideviceinstaller tool. Please ensure it is installed. " +
+                        "See https://github.com/libimobiledevice/ideviceinstaller fore more details.");
+                    return;
+                }
 
-            apkPath = string.Empty;
-            return false;
+                if (string.IsNullOrEmpty(LibIDeviceDebugBinary))
+                {
+                    Debug.LogError("Could not find idevicedebug tool. Please ensure libimobiledevice is installed. " +
+                        "See https://helpmanual.io/help/idevicedebug/ for more details.");
+                    return;
+                }
+
+                EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Installing archive", 0.3f);
+
+                // Find archive to install
+                var ipaPath = Directory.GetFiles(AbsoluteAppBuildPath, "*.ipa", SearchOption.AllDirectories).FirstOrDefault();
+                if (string.IsNullOrEmpty(ipaPath))
+                {
+                    Debug.LogError($"Could not find a built out iOS .ipa archive in \"{AbsoluteAppBuildPath}\" to launch.");
+                    return;
+                }
+
+                if (RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-i", ipaPath).Run() != 0)
+                {
+                    Debug.LogError("Error while installing .ipa archive to the device. Please check the log for details about the error.");
+                    return;
+                }
+
+                EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Launching Client", 0.9f);
+
+                // Get chosen ios package id and launch
+                var bundleId = PlayerSettings.GetApplicationIdentifier(BuildTargetGroup.iOS);
+
+                // Optional arguments to be passed, same as standalone
+                // Use this to pass through the local ip to connect to
+                var runtimeIp = GdkToolsConfiguration.GetOrCreateInstance().RuntimeIp;
+                var arguments = new StringBuilder();
+                if (!string.IsNullOrEmpty(runtimeIp))
+                {
+                    arguments.Append($"-e SPATIALOS_ARGUMENTS=\"+{RuntimeConfigNames.ReceptionistHost} {runtimeIp}\"");
+                }
+
+                RedirectedProcess.Command(LibIDeviceDebugBinary).WithArgs(arguments.ToString(), "run", bundleId)
+                    .RedirectOutputOptions(OutputRedirectBehaviour.RedirectStdOut | OutputRedirectBehaviour.RedirectStdErr)
+                    .ReturnImmediately()
+                    .Run();
+            }
+            finally
+            {
+                EditorUtility.ClearProgressBar();
+            }
         }
     }
 }
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -103,25 +103,40 @@ namespace Improbable.Gdk.Mobile
 
                 EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Installing archive", 0.3f);
 
+                // Get chosen ios package id
+                var bundleId = PlayerSettings.GetApplicationIdentifier(BuildTargetGroup.iOS);
+                string appList = null;
+
                 // Find archive to install
                 var ipaPath = Directory.GetFiles(AbsoluteAppBuildPath, "*.ipa", SearchOption.AllDirectories).FirstOrDefault();
-                if (string.IsNullOrEmpty(ipaPath))
+                RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-l")
+                    .AddAccumulatedOutputProcessing(output => { appList = output; })
+                    .RedirectOutputOptions(OutputRedirectBehaviour.None).Run();
+                var existsOnDevice = appList.Contains(bundleId);
+                var existsLocally = !string.IsNullOrEmpty(ipaPath);
+                if (!existsLocally && !existsOnDevice)
                 {
-                    Debug.LogError($"Could not find a built out iOS .ipa archive in \"{AbsoluteAppBuildPath}\" to launch.");
-                    return;
+                    Debug.LogError($"Could not find an app on device or built out iOS .ipa archive in \"{AbsoluteAppBuildPath}\" to launch.");
                 }
 
-                if (RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-i", ipaPath).Run() != 0)
+                if (existsLocally)
                 {
-                    Debug.LogError("Error while installing .ipa archive to the device. Please check the log for details about the error.");
-                    return;
+                    if (RedirectedProcess.Command(LibIDeviceInstallerBinary)
+                            .WithArgs((existsOnDevice ? "-g" : "-i"), $"\"{ipaPath}\"").Run() !=
+                        0)
+                    {
+                        Debug.LogError(
+                            "Error while installing .ipa archive to the device. Please check the log for details about the error.");
+                        return;
+                    }
+                }
+
+                // Wait until the app is installed
+                while (RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-l").Run() != 0)
+                {
                 }
 
                 EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Launching Client", 0.9f);
-
-                // Get chosen ios package id and launch
-                var bundleId = PlayerSettings.GetApplicationIdentifier(BuildTargetGroup.iOS);
-
                 // Optional arguments to be passed, same as standalone
                 // Use this to pass through the local ip to connect to
                 var runtimeIp = GdkToolsConfiguration.GetOrCreateInstance().RuntimeIp;
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -115,6 +115,10 @@ namespace Improbable.Gdk.Mobile
                     .RedirectOutputOptions(OutputRedirectBehaviour.None).Run();
                 var existsOnDevice = appList.Contains(bundleId);
                 var existsLocally = !string.IsNullOrEmpty(ipaPath);
+                
+                // If .ipa exists - we install/upgrade device app and launch it
+                // If .ipa doesn't exist, but the bunlde is installed on device - we launch existing app with parameters
+                // If .ipa doesn't exist and bundle isn't found on device - we fail with an error.
                 if (!existsLocally && !existsOnDevice)
                 {
                     Debug.LogError($"Could not find an app on device or built out iOS .ipa archive in \"{AbsoluteAppBuildPath}\" to launch.");
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -134,16 +134,10 @@ namespace Improbable.Gdk.Mobile
                 }
 
                 // Wait until the app has finished installing
-                DateTime timeout = DateTime.Now.AddSeconds(5);
-                while (RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-l").Run() != 0)
+                if (RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-l").RedirectOutputOptions(OutputRedirectBehaviour.None).Run() != 0)
                 {
-                    if (DateTime.Now > timeout)
-                    {
-                        Debug.LogError($"Device communication error. Please ensure it is connected");
-                        return;
-                    }
-
-                    System.Threading.Thread.Sleep(500);
+                    Debug.LogError($"Device communication error. Please ensure it is connected");
+                    return;
                 }
 
                 EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Launching Client", 0.9f);
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Utility/LaunchArguments.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Utility/LaunchArguments.cs
@@ -3,11 +3,13 @@ using System.Collections.Generic;
 using Improbable.Gdk.Core;
 using UnityEngine;
 
-namespace Improbable.Gdk.Mobile.Android
+namespace Improbable.Gdk.Mobile
 {
     public static class LaunchArguments
     {
-        public static Dictionary<string, string> GetArguments()
+        private const string LaunchArgumentsEnvKey = "SPATIALOS_ARGUMENTS";
+
+        public static Dictionary<string, string> GetAndroidArguments()
         {
             if (Application.isEditor)
             {
@@ -36,7 +38,30 @@ namespace Improbable.Gdk.Mobile.Android
                 Debug.LogException(e);
             }
 
-            return new Dictionary<string, string>();
+            return null;
+        }
+
+        public static Dictionary<string, string> GetiOSArguments()
+        {
+            if (Application.isEditor)
+            {
+                return new Dictionary<string, string>();
+            }
+
+            try
+            {
+                var argumentsEnvVar = System.Environment.GetEnvironmentVariable(LaunchArgumentsEnvKey);
+                if (argumentsEnvVar != null)
+                {
+                    return CommandLineUtility.ParseCommandLineArgs(argumentsEnvVar.Split(' '));
+                }
+            }
+            catch (Exception e)
+            {
+                UnityEngine.Debug.LogException(e);
+            }
+
+            return null;
         }
     }
 }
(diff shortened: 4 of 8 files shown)


workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -65,24 +71,126 @@ private static void LaunchMobileClient()
...
+                var runtimeIp = GdkToolsConfiguration.GetOrCreateInstance().RuntimeIp;
+                var arguments = new StringBuilder();
+                if (!string.IsNullOrEmpty(runtimeIp))
+                {
+                    arguments.Append($"-e SPATIALOS_ARGUMENTS=\"+{RuntimeConfigNames.ReceptionistHost} {runtimeIp}\"");
+                }
+
+                StartiOSDeviceProcess(arguments.ToString(), bundleId);
+
+                EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Done", 1.0f);
```
jared-improbable: Redundant since it'll be cleared immediately anyway

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -65,24 +71,126 @@ private static void LaunchMobileClient()
...
             }
+            finally
+            {
+                EditorUtility.ClearProgressBar();
+            }
+        }
+
+        private static string TryGetArchivePath(string rootPath, string archivePattern)
+        {
+            foreach (var file in Directory.GetFiles(rootPath, archivePattern, SearchOption.AllDirectories))
```
jared-improbable: This could just be replaced with `return Directory.GetFiles(rootPath, archivePattern, SearchOption.AllDirectories).FirstOrDefault();`
What happens if there are multiple archives in the `rootPath`?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -65,24 +71,126 @@ private static void LaunchMobileClient()
...
+            process.ErrorDataReceived += OnReceived;
+
+            process.BeginOutputReadLine();
+            process.BeginErrorReadLine();
+
+            process.EnableRaisingEvents = true;
+            process.Exited += (sender, args) =>
+            {
+                process.Dispose();
+                process = null;
```
jared-improbable: `process = null` is redundant

workers/unity/Packages/com.improbable.gdk.mobile/Utility/LaunchArguments.cs
```diff
@@ -30,7 +32,25 @@ public static class LaunchArguments
...
+        }
+
+        public static Dictionary<string, string> GetiOSArguments()
+        {
+            try
+            {
+                var argumentsEnvVar = System.Environment.GetEnvironmentVariable(LaunchArgumentsEnvKey);
+                if (argumentsEnvVar != null)
+                {
+                    return CommandLineUtility.ParseCommandLineArgs(argumentsEnvVar.Split(' '));
```
jared-improbable: What are the odds of having a quoted string with spaces as part of the arguments? Like `--path "Applications\Probably a bad idea"`?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -1,21 +1,26 @@
...
+using Debug = UnityEngine.Debug;
 
 namespace Improbable.Gdk.Mobile
 {
     public static class LaunchMenu
     {
         private const string rootApkPath = "build";
-        private static string AbsoluteApkPath => Path.GetFullPath(Path.Combine(Application.dataPath, Path.Combine("..", rootApkPath)));
+        private static string AbsoluteAppBuildPath => Path.GetFullPath(Path.Combine(Application.dataPath, Path.Combine("..", rootApkPath)));
+        public static string IDeviceInstallerBinary => Common.DiscoverLocation("ideviceinstaller");
```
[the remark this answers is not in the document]
gkassabli: Is LibIDeviceInstallerBinary fine?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -65,24 +71,126 @@ private static void LaunchMobileClient()
                     "-n", $"{bundleId}/com.unity3d.player.UnityPlayerActivity",
                     "-e", "\"arguments\"", $"\\\"{arguments.ToString()}\\\"");
 
-                EditorUtility.DisplayProgressBar("Launching Mobile Client", "Done", 1.0f);
+                EditorUtility.DisplayProgressBar("Launching Android Client", "Done", 1.0f);
```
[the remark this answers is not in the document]
gkassabli: Yeah, I was wondering if it was needed, will remove this

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -65,24 +71,126 @@ private static void LaunchMobileClient()
...
-        private static bool TryGetApkPath(string rootPath, out string apkPath)
+        [MenuItem(MenuLaunchiOSDevice, false, 74)]
+        private static void LaunchiOSDeviceClient()
         {
-            foreach (var file in Directory.GetFiles(rootPath, "*.apk", SearchOption.AllDirectories))
+            try
             {
-                apkPath = file;
-                return true;
+                if (IDeviceInstallerBinary == string.Empty)
```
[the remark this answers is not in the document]
gkassabli: 👍

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -65,24 +71,126 @@ private static void LaunchMobileClient()
...
+                {
+                    Debug.LogError("Could not find ideviceinstaller tool. Please ensure it is installed.");
```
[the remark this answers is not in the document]
gkassabli: Sure, we can point to https://helpmanual.io/help/idevicedebug/

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -65,24 +71,126 @@ private static void LaunchMobileClient()
...
+
+                if (RedirectedProcess.Run(IDeviceInstallerBinary, "-i", ipaPath) != 0)
+                {
+                    Debug.LogError("Error while installing archive to the device. Please check the log for details about the error.");
+                    return;
+                }
+
+                EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Launching Client", 0.9f);
+
+                // Get chosen android package id and launch
```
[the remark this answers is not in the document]
gkassabli: Yup, leftover from the function above

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
gkassabli: Will just pick first of them, as we assume unity project only contains one Android/iOS archive. Ideally we would want to allow people to select archive to launch, but I think this functionality should suffice for now

workers/unity/Packages/com.improbable.gdk.mobile/Utility/LaunchArguments.cs
gkassabli: Right now it's a non case - we control what is passed here as environment variable and we only pass ip address and the key for the argument. We'll have to improve this in the future once there is a chance we pass strings with spaces

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -64,25 +71,115 @@ private static void LaunchMobileClient()
...
+                {
+                    Debug.LogError("Error while installing archive to the device. Please check the log for details about the error.");
+                    return;
+                }
 
-                EditorUtility.DisplayProgressBar("Launching Mobile Client", "Done", 1.0f);
+                EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Launching Client", 0.9f);
+
+                // Get chosen ios package id and launch
+                var bundleId = PlayerSettings.GetApplicationIdentifier(BuildTargetGroup.Android);
```
jessicafalk: should that be android?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -64,25 +71,115 @@ private static void LaunchMobileClient()
...
+                // Optional arguments to be passed, same as standalone
+                // Use this to pass through the local ip to connect to
+                var runtimeIp = GdkToolsConfiguration.GetOrCreateInstance().RuntimeIp;
+                var arguments = new StringBuilder();
+                if (!string.IsNullOrEmpty(runtimeIp))
+                {
+                    arguments.Append($"-e SPATIALOS_ARGUMENTS=\"+{RuntimeConfigNames.ReceptionistHost} {runtimeIp}\"");
+                }
+
+                StartiOSDeviceProcess(arguments.ToString(), bundleId);
```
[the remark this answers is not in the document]
gkassabli: RedirectProcess waits for the end of process and processes output only after that. With idevicedebug we want to process output and show it to Unity as soon as we get it

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
gkassabli: Nope :) Definitely iOS, nice catch!

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -64,25 +71,110 @@ private static void LaunchMobileClient()
...
+
+        [MenuItem(MenuLaunchiOSDevice, false, 74)]
+        private static void LaunchiOSDeviceClient()
+        {
+            try
+            {
+                // Ensure needed tools are installed
+                if (string.IsNullOrEmpty(LibIDeviceInstallerBinary))
+                {
+                    Debug.LogError("Could not find ideviceinstaller tool. Please ensure it is installed." +
```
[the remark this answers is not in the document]
gkassabli: 👍

workers/unity/Assets/Playground/Scripts/UI/ConnectionScreenController.cs
```diff
@@ -75,14 +76,20 @@ private void TryConnect()
...
+            Dictionary<string, string> arguments = null;
 #if UNITY_ANDROID
-            var arguments = Improbable.Gdk.Mobile.Android.LaunchArguments.GetArguments();
+            arguments = Improbable.Gdk.Mobile.LaunchArguments.GetAndroidArguments();
+#elif UNITY_IOS
+            arguments = Improbable.Gdk.Mobile.LaunchArguments.GetiOSArguments();
+#endif
+            if (arguments == null)
+            {
+                return string.Empty;
```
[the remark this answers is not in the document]
gkassabli: Sorry, not sure I understand what is default value in this context.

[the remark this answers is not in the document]
jamiebrynes7: `RuntimeConfigDefaults.ReceptionistHost`. Although at a second look I'm not sure if that entirely makes sense in context here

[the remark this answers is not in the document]
jessicafalk: no, it doesn't. in this case return an empty string would be the correct thing to do so that users enter the ip adress themselves

workers/unity/Packages/com.improbable.gdk.mobile/Utility/LaunchArguments.cs
```diff
@@ -30,7 +32,25 @@ public static class LaunchArguments
...
 
-            return new Dictionary<string, string>();
+            return null;
+        }
+
+        public static Dictionary<string, string> GetiOSArguments()
+        {
+            try
+            {
+                var argumentsEnvVar = System.Environment.GetEnvironmentVariable(LaunchArgumentsEnvKey);
```
jessicafalk: do the same change that we did for Android: checking whether we are in the editor as that means we can just return immediately. In this case, it should be fine as we are not calling any java objects, but just to be consistent

gkassabli: 👍

jessicafalk:
please repoint to `develop` :)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Added option to build workers out via IL2CPP in the cmd.
 - Added an example of handling disconnect for mobile workers.
 - Added support for launching an Android client from the Editor over ADB.
+- Added `Launch Mobile client` menu to `SpatialOS` menu in Unity Editor. `Android device` and `iOS device` menu items allow launching corresponding mobile clients to connect to local deployment.
```
jessicafalk: multiple things:
- this needs to be under the `Unreleased` header
- rephrase that entry to
```suggestion
- Added support for launching an iOS client from an iOS device from the Unity Editor. Navigate to **SpatialOS** > **Launch mobile client** > **iOS device** to try it out.
```

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -1,21 +1,27 @@
+using System.Diagnostics;
```
jessicafalk: is this import used?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -1,21 +1,27 @@
...
 using System.IO;
+using System.Linq;
 using System.Text;
 using Improbable.Gdk.Core;
 using Improbable.Gdk.Tools;
 using UnityEditor;
 using UnityEngine;
+using Debug = UnityEngine.Debug;
```
jessicafalk: I assume that happened because of the System.Diagnostics import, shouldn't be necessary to define Debug

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -1,21 +1,27 @@
...
 
 namespace Improbable.Gdk.Mobile
 {
     public static class LaunchMenu
     {
-        private const string rootApkPath = "build";
-        private static string AbsoluteApkPath => Path.GetFullPath(Path.Combine(Application.dataPath, Path.Combine("..", rootApkPath)));
+        private const string rootBuildPath = "build";
```
jessicafalk: it's only used in one variable, please inline :)

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -1,21 +1,27 @@
...
+        private static string AbsoluteAppBuildPath => Path.GetFullPath(Path.Combine(Application.dataPath, Path.Combine("..", rootBuildPath)));
```
jessicafalk: you can use one Path.Combine instead of multiple. it takes an arbitrary number of string arguments

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -259,10 +259,11 @@ private static void Zip(string zipAbsolutePath, string basePath, bool useCompres
         {
             using (new ShowProgressBarScope($"Package {basePath}"))
             {
-                RedirectedProcess.Run(Common.SpatialBinary, "file", "zip",
-                    $"--output=\"{Path.GetFullPath(zipAbsolutePath)}\"",
-                    $"--basePath=\"{Path.GetFullPath(basePath)}\"", "\"**\"",
-                    $"--compression={useCompression}");
+                RedirectedProcess.Command(Common.SpatialBinary)
```
jessicafalk: just realized that that used to be part of another PR and was merged in here. Could we please not do this until absolutely necessary? I know you are used to stacks diffs (they are nice with the right infrastructure :) ), but here it just gets more difficult to review a PR when the scope of it suddenly changes

in this case the refactor of the redirectedprocess part could have been merged into the master /develop branch just fine without this part :)

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -66,28 +73,76 @@ private static void LaunchMobileClient()
...
+                    return;
+                }
+
+                EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Installing archive", 0.3f);
+
+                // Find archive to install
+                var ipaPath = Directory.GetFiles(AbsoluteAppBuildPath, "*.ipa", SearchOption.AllDirectories).FirstOrDefault();
+                if (string.IsNullOrEmpty(ipaPath))
+                {
+                    Debug.LogError($"Could not find a built out iOS archive in \"{AbsoluteAppBuildPath}\" to launch.");
```
jessicafalk: should we specifiy here that we are explicitly looking for an `ipa`?

workers/unity/Packages/com.improbable.gdk.mobile/Utility/LaunchArguments.cs
```diff
@@ -36,7 +38,30 @@ public static class LaunchArguments
...
+        public static Dictionary<string, string> GetiOSArguments()
+        {
+            if (Application.isEditor)
+            {
+                return new Dictionary<string, string>();
+            }
+
+            try
+            {
+                var argumentsEnvVar = System.Environment.GetEnvironmentVariable(LaunchArgumentsEnvKey);
```
jessicafalk: just checking (after I had all these weird issues with environment variables): does that work properly?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
gkassabli: Probably not. Let me remove it

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
gkassabli: Good point!

workers/unity/Packages/com.improbable.gdk.mobile/Utility/LaunchArguments.cs
gkassabli: yeah, this works as expected apparently as we pass it to the process directly

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +74,87 @@ private static void LaunchMobileClient()
...
                 EditorUtility.ClearProgressBar();
             }
         }
 
-        private static bool TryGetApkPath(string rootPath, out string apkPath)
+        [MenuItem(MenuLaunchiOSDevice, false, 74)]
+        private static void LaunchiOSDeviceClient()
         {
-            foreach (var file in Directory.GetFiles(rootPath, "*.apk", SearchOption.AllDirectories))
+            try
```
jessicafalk: this try block doesn't have a catch block afterwards. Do we even need to wrap it in a try block?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +74,87 @@ private static void LaunchMobileClient()
...
+                        Debug.LogError(
+                            "Error while installing .ipa archive to the device. Please check the log for details about the error.");
+                        return;
+                    }
+                }
 
-            apkPath = string.Empty;
-            return false;
+                // Wait until the app is installed
+                while (RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-l").Run() != 0)
```
jessicafalk: this looks sketchy. why do we need to wrap it in a while loop?
is it possible for the user to do something funky, e.g. disconnect the device, close the simulator so that Unity becomes completely unresponsive until you force quit it?

gkassabli: Fair point. Let me add a timeout here so that we don't hold off indefinitely

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
gkassabli: It contains `finally` block to remove progress bar. Otherwise we might end up with non-dismissable bar when we have an exception

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
gkassabli: Okay, so after some digging in the code I found out that `ideviceinstaller -l`(list) command has waiting for device built in, so we don't have to mess with timeout. Instead we can just call list command and when it returns we're ready to run.
I tested this extensively on slow device with FPS starter project to ensure this works as expected and didn't see any issues.

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +75,90 @@ private static void LaunchMobileClient()
...
+                string appList = null;
+
+                // Find archive to install
+                var ipaPath = Directory.GetFiles(AbsoluteAppBuildPath, "*.ipa", SearchOption.AllDirectories).FirstOrDefault();
+                RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-l")
+                    .AddAccumulatedOutputProcessing(output => { appList = output; })
+                    .RedirectOutputOptions(OutputRedirectBehaviour.None).Run();
+                var existsOnDevice = appList.Contains(bundleId);
+                var existsLocally = !string.IsNullOrEmpty(ipaPath);
+                if (!existsLocally && !existsOnDevice)
```
jessicafalk: would separate that out into two check:
first: check if ipaPath exists: if not, print an error directly related to that and mention that you first need to build it
second: check if the app exists on the device: if not, print an errror mentioning the installation has failed. 

Is it possible to retrieve the reason of why it could have failed?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +75,90 @@ private static void LaunchMobileClient()
...
+                {
+                    Debug.LogError($"Could not find an app on device or built out iOS .ipa archive in \"{AbsoluteAppBuildPath}\" to launch.");
+                    return;
+                }
+
+                if (existsLocally)
```
jessicafalk: at this point we shouldn't need to check for it. It should be clear that the installation has succeeded and we can run the app

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -111,6 +115,16 @@ public RedirectedProcess AddErrorProcessing(Action<string> errorProcessor)
             return this;
         }
 
+        /// <summary>
+        ///     Adds accumulated output processing
+        /// </summary>
+        /// <param name="errorProcessor">Processing action for error output.</param>
+        public RedirectedProcess AddAccumulatedOutputProcessing(Action<string> errorProcessor)
```
jessicafalk: `errorProcessor` seems like the wrong name :)

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -111,6 +115,16 @@ public RedirectedProcess AddErrorProcessing(Action<string> errorProcessor)
             return this;
         }
 
+        /// <summary>
+        ///     Adds accumulated output processing
+        /// </summary>
+        /// <param name="errorProcessor">Processing action for error output.</param>
```
jessicafalk: this definition seems wrong

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -111,6 +115,16 @@ public RedirectedProcess AddErrorProcessing(Action<string> errorProcessor)
             return this;
         }
 
+        /// <summary>
+        ///     Adds accumulated output processing
```
jessicafalk:
```suggestion
        ///     Adds the specified action to a list of processors. These actions are called at the end of running the specified process and passes in as an argument the complete log of that process.
```

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -135,113 +158,134 @@ public int Run()
...
             };
 
-            using (var process = Process.Start(info))
+            var process = Process.Start(info);
+
+            if (process == null)
             {
-                if (process == null)
+                throw new Exception(
+                    $"Failed to run {info.FileName} {info.Arguments}\nIs the .NET Core SDK installed?");
```
jessicafalk: do we always need the .Net Core SDK when running a process?

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -135,113 +158,134 @@ public int Run()
...
-                    {
-                        outputString = ProcessSpatialOutput(outputString);
-                    }
+                    Debug.Log(outputString);
+                }
 
-                    if (string.IsNullOrEmpty(outputString))
+                if (outputLog != null)
+                {
+                    lock (outputLog)
```
jessicafalk: why the lock?

jessicafalk: ah, because it's used both in `OutputDataReceived` and `OutputDataReceived`?

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
gkassabli: Yup, copypasta. Thanks for the catch, I'll update this

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
gkassabli: hmmm, I recall making those changes. Probably lost them during some nasty rebase :(

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
gkassabli: This was here a while ago, so I don't have a clue :)
My guess is that we use this in a few places to run .NET related processes.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
gkassabli: Let me update other processors as well along to add more clarity

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
gkassabli: The logic here is following:
If .ipa exists - we install/upgrade device app and launch it
If .ipa doesn't exist, but the bunlde is installed on device - we launch existing app with parameters
If .ipa doesn't exist and bundle isn't found on device - we fail with an error.

Let me add a comment in code explaining it.

All important output is redirected to the console, so that it is visible to developer.

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
gkassabli: See the comment about for the logic intended here.
The idea is that developer doesn't have to build .ipa every time they want to start the app with new config.

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +75,94 @@ private static void LaunchMobileClient()
...
+                // Find archive to install
+                var ipaPath = Directory.GetFiles(AbsoluteAppBuildPath, "*.ipa", SearchOption.AllDirectories).FirstOrDefault();
+                RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-l")
+                    .AddAccumulatedOutputProcessing(output => { appList = output; })
+                    .RedirectOutputOptions(OutputRedirectBehaviour.None).Run();
+                var existsOnDevice = appList.Contains(bundleId);
+                var existsLocally = !string.IsNullOrEmpty(ipaPath);
+                
+                // If .ipa exists - we install/upgrade device app and launch it
+                // If .ipa doesn't exist, but the bunlde is installed on device - we launch existing app with parameters
```
jamiebrynes7: Super nit: `bunlde` -> `bundle`

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +75,94 @@ private static void LaunchMobileClient()
...
+                // If .ipa doesn't exist and bundle isn't found on device - we fail with an error.
+                if (!existsLocally && !existsOnDevice)
+                {
+                    Debug.LogError($"Could not find an app on device or built out iOS .ipa archive in \"{AbsoluteAppBuildPath}\" to launch.");
+                    return;
+                }
+
+                if (existsLocally)
```
jamiebrynes7: Doesn't this imply that we reinstall the app every single time, even if the operation is not required? 

Not sure how long this normally takes, but if it takes a while - it may be worth looking to check if its a new IPA or the same one that's installed.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -56,11 +57,14 @@ public class RedirectedProcess
         private string workingDirectory = Path.GetFullPath(Path.Combine(Application.dataPath, ".."));
         private readonly List<Action<string>> outputProcessors = new List<Action<string>>();
         private readonly List<Action<string>> errorProcessors = new List<Action<string>>();
+        private readonly List<Action<string>> accumulatedOutputProcessors = new List<Action<string>>();
 
         private OutputRedirectBehaviour outputRedirectBehaviour =
             OutputRedirectBehaviour.ProcessSpatialOutput |
             OutputRedirectBehaviour.RedirectAccumulatedOutput;
 
+        private bool returnImmediately = false;
```
jamiebrynes7: Redundant initialization

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
jamiebrynes7: Yeah this isn't a valid error message given this method

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
gkassabli: Not exactly, in a line below, where we actually do install, we use `-g` option if the archive both exists on device and locally, which triggers app "upgrade" instead of direct reinstall.
Let me add this to the comment.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
gkassabli: Fair point, this function has changed significantly since this message was added, I'll generalise the message

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: does this app upgrade always work? I know that for Android, if an app has been built from different machines / different signatures, the upgrade will fail

gkassabli: Dug into the code a bit and it looks like `update` and `install` commands are being delivered to Apple system services, where it is hard to understand what is the logic applied.
More general, .ipa installations happen from time to time on iOS, when the app with the same bundle id has different entitlements (access to system services) or in some scenarios with codesigning differences. Though in my experience, those are more likely to happen when we install directly, not try to upgrade the app and the common workaround is to just remove existing app.
Considering that, I would imagine using `upgrade` if the bundle is already present won't make our workflows more fragile, but will save a bit of time every now and then

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +75,95 @@ private static void LaunchMobileClient()
...
+            try
             {
-                apkPath = file;
-                return true;
-            }
+                // Ensure needed tools are installed
+                if (string.IsNullOrEmpty(LibIDeviceInstallerBinary))
+                {
+                    Debug.LogError("Could not find ideviceinstaller tool. Please ensure it is installed. " +
+                        "See https://github.com/libimobiledevice/ideviceinstaller fore more details.");
```
jessicafalk: typo:
```suggestion
                        "See https://github.com/libimobiledevice/ideviceinstaller for more details.");
```

After testing this pr and going to this page, I'm actually more confused. Do I actually need to clone their repo and build it?
I simply used `brew` to install it, which seems to work. but we also don't specify which version of it we actually need / have tested against

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +75,95 @@ private static void LaunchMobileClient()
...
+                    return;
+                }
+
+                if (string.IsNullOrEmpty(LibIDeviceDebugBinary))
+                {
+                    Debug.LogError("Could not find idevicedebug tool. Please ensure libimobiledevice is installed. " +
+                        "See https://helpmanual.io/help/idevicedebug/ for more details.");
```
jessicafalk: same here: this is the help manual telling you how idevicedebug works in the command line. It doesn't tell you how to install it

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +75,95 @@ private static void LaunchMobileClient()
...
+                    return;
+                }
+
+                EditorUtility.DisplayProgressBar("Launching iOS Device Client", "Installing archive", 0.3f);
+
+                // Get chosen ios package id
+                var bundleId = PlayerSettings.GetApplicationIdentifier(BuildTargetGroup.iOS);
+                string appList = null;
```
jessicafalk: no need to assign it to null

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +75,95 @@ private static void LaunchMobileClient()
...
+
+                // Find archive to install
+                var ipaPath = Directory.GetFiles(AbsoluteAppBuildPath, "*.ipa", SearchOption.AllDirectories).FirstOrDefault();
+                RedirectedProcess.Command(LibIDeviceInstallerBinary).WithArgs("-l")
+                    .AddAccumulatedOutputProcessing(output => { appList = output; })
+                    .RedirectOutputOptions(OutputRedirectBehaviour.None).Run();
+                var existsOnDevice = appList.Contains(bundleId);
```
jessicafalk: could this throw a null pointer exception, if there is no output at all -> applist stays null? better to assign it to the empty string in this case

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -70,25 +75,95 @@ private static void LaunchMobileClient()
...
+                if (!existsLocally && !existsOnDevice)
+                {
+                    Debug.LogError($"Could not find an app on device or built out iOS .ipa archive in \"{AbsoluteAppBuildPath}\" to launch.");
+                    return;
+                }
+
+                if (existsLocally)
+                {
+                    if (RedirectedProcess.Command(LibIDeviceInstallerBinary)
+                            .WithArgs((existsOnDevice ? "-g" : "-i"), $"\"{ipaPath}\"").Run() !=
```
jessicafalk: could the upgrade fail, if it was previously built on a different machine, with a different provisioning profile? I know that we had that issue with Android

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Added
+- Added support for launching an iOS client from an iOS device from the Unity Editor. Navigate to **SpatialOS** > **Launch mobile client** > **iOS device** to try it out
```
jessicafalk:
```suggestion
- Added support for launching an iOS client from an iOS device using the Unity Editor. Navigate to **SpatialOS** > **Launch mobile client** > **iOS device** to try it out.
```

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
jessicafalk: tested it: yes, it can fail. we need to have an additional check for that

PR #675, opened 2019-01-09
Add OWNERS files for automatic labeling

jamiebrynes7:
#### Description
Add OWNERS files to relevant parts of the repo to allow PROW to automatically label PRs that touch those directories. We can optionally add "approvers" which roughly correlate to technical owners if we wanted.


ci/OWNERS
```diff
@@ -0,0 +1,2 @@
+labels: 
+  - "A: maintenance"
```
petemounce: I recommend using `A/thing` - more concise, leaves room for hierarchy if you decide later you want that, and is parseable as if it were a file-path via various stdlib support.

workers/unity/Packages/com.improbable.gdk.mobile/OWNERS.meta
```diff
@@ -0,0 +1,7 @@
+fileFormatVersion: 2
```
petemounce: What is this?

jamiebrynes7: A Unity metadata file

ci/OWNERS
jamiebrynes7: Hmmm hierarchy is nice - but I find the `/` notation harder to read at a glance 😕 

What would be advantage of the latter?

PR #681, opened 2019-01-11
Update GameObjectCreatorFromMetadata.cs

filod:
it's nice to have a chance to modify prefab before Instantiate.

Since you are not accept public PR, you can take this as a feedback and close it.


jessicafalk:
Thank you for your contribution!

We like the idea and have discussed this internally. Instead of passing in an extra parameter, we are considering making the methods in the `GameObjectCreatorFromMetadata` virtual. This would allow you to inherit from the `GameObjectCreatorFromMetadata` and modify it based on whatever changes you would like to make.

What do you think about this?

filod:
That will do too.

jessicafalk:
I created https://github.com/spatialos/gdk-for-unity/issues/687 to track progress of this feature.
Closing this PR for now.

PR #683, opened 2019-01-16
Fix underflow memory allocation crash in transform sync.

zeroZshadow:
#### Description
Fix a crash bug in the Transform sync feature module.
Includes a slight optimization to avoid dynamic buffer resizes.

#### Tests
Ran the playground project locally and went through the same steps which previously caused the crash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,11 @@
 - `Improbable.Gdk.EntityTemplate` is now mutable and exposes a set of APIs to add, remove, and replace component snapshots
     - This replaces the `Improbable.Gdk.Core.EntityBuilder` class.
     - These changes also allow you to reuse an `EntityTemplate` more than once.
+
 - Upgraded the project to be compatible with `2018.3.2f1`.
 - Upgraded the entities package to `0.0.12-preview.21`
 - Disabled protocol logging on Linux workers to prevent crashes. This will be reverted once the underlying issue is fixed.
-- Fixed Memory crash in TransformSynchronization module.
-
+- Fixed an issue in the TransformSynchronization module where an integer underflow would cause a memory crash.
 ### Fixed
 
 - `Clean all workers` now cleans worker configs in addition to built-out workers.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,11 @@
 - `Improbable.Gdk.EntityTemplate` is now mutable and exposes a set of APIs to add, remove, and replace component snapshots
     - This replaces the `Improbable.Gdk.Core.EntityBuilder` class.
     - These changes also allow you to reuse an `EntityTemplate` more than once.
-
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - `Improbable.Gdk.EntityTemplate` is now mutable and exposes a set of APIs to add, remove, and replace component snapshots
     - This replaces the `Improbable.Gdk.Core.EntityBuilder` class.
     - These changes also allow you to reuse an `EntityTemplate` more than once.
+- Fixed Memory crash in TransformSynchronization module.
```
jamiebrynes7: Nit: capital M in memory.

Also maybe a better message is:

`Fixed a crash in the TransformSynchronization module where an integer overflow would cause a memory crash`

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/InterpolateTransformSystem.cs
```diff
@@ -100,23 +100,25 @@ protected override void OnUpdate()
...
-                    if (remoteTickDifference == 0)
+                    if (transformToInterpolateTo.PhysicsTick <= lastTickId)
                     {
                         continue;
                     }
 
-                    uint ticksToFill =
-                        math.max((uint) (transformToInterpolateTo.PhysicsTick - lastTickId), 1);
-                    for (uint j = 0; j < ticksToFill - 1; ++j)
+                    var remoteTickDifference = (int) (transformToInterpolateTo.PhysicsTick - lastTickId);
```
jamiebrynes7: What are the types of `PhysicsTick` and `lastTickId`? 

If they are both `uint`s can this still cause a problem if `PhysicsTick` > `Int.MaxPositive` ?

CHANGELOG.md
paulbalaji: `Fixed an issue in the TransformSynchronization module where an integer overflow would cause a memory crash`

jamiebrynes7: Wot he said ^^^

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/InterpolateTransformSystem.cs
zeroZshadow: Verified that this will be fine

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Upgraded the project to be compatible with `2018.3.2f1`.
 - Upgraded the entities package to `0.0.12-preview.21`
 - Disabled protocol logging on Linux workers to prevent crashes. This will be reverted once the underlying issue is fixed.
+- Fixed an issue in the TransformSynchronization module where an integer underflow would cause a memory crash.
```
jamiebrynes7: This should go in the fixed section

PR #685, opened 2019-01-16
Fix mobile client read access to PlayerTemplate

gkassabli:
#### Description
Previous changes have broken the mobile clients as they didn't have access to player

#### Tests
Run mobile client and ensure it is connecting correctly

#### Primary reviewers


jamiebrynes7:
Make sure that the cube and spinner ones are good as well 👍

PR #686, opened 2019-01-16
[UTY-1613] Gracefully handle disconnect in Playground

gkassabli:
#### Description
We didn't have any specific handling for disconnect on mobile. Whenever the server was killed, mobile client worker is going to get to weird stale state where it doesn't technically hang, but nothing was happening and the client wasn't interactive.
This PR adds the code to restore Playground to "Ready to connect" mode

#### Tests
Run Playground locally and ensure that when server is disconnected we are ready to connect again ->
After the server is back, we should be able to reconnect

### DIFF
--- a/workers/unity/Assets/Playground/Scripts/Camera/InitCameraSystem.cs
+++ b/workers/unity/Assets/Playground/Scripts/Camera/InitCameraSystem.cs
@@ -37,6 +37,14 @@ namespace Playground
 
                 Cursor.lockState = CursorLockMode.Locked;
             }
+
+            // Disable system after first run.
+            Enabled = false;
+        }
+
+        protected override void OnDestroyManager()
+        {
+            Cursor.lockState = CursorLockMode.None;
         }
     }
 }
--- a/workers/unity/Assets/Playground/Scripts/Camera/InitCameraSystem.cs
+++ b/workers/unity/Assets/Playground/Scripts/Camera/InitCameraSystem.cs
@@ -36,10 +36,10 @@ namespace Playground
                 PostUpdateCommands.AddComponent(entity, CameraComponentDefaults.Transform);
 
                 Cursor.lockState = CursorLockMode.Locked;
-            }
 
-            // Disable system after first run.
-            Enabled = false;
+                // Disable system after first run.
+                Enabled = false;
+            }
         }
 
         protected override void OnDestroyManager()
--- a/workers/unity/Assets/Playground/Scripts/UI/InitUISystem.cs
+++ b/workers/unity/Assets/Playground/Scripts/UI/InitUISystem.cs
@@ -45,5 +45,10 @@ namespace Playground
             // Disable system after first run.
             Enabled = false;
         }
+
+        protected override void OnDestroyManager()
+        {
+            UnityEngine.Object.Destroy(UIComponent.Main.gameObject);
+        }
     }
 }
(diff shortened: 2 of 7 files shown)


jamiebrynes7:
What's up with the branch target?

workers/unity/Assets/Playground/Scripts/Camera/InitCameraSystem.cs
```diff
@@ -37,6 +37,14 @@ protected override void OnUpdate()
 
                 Cursor.lockState = CursorLockMode.Locked;
             }
+
+            // Disable system after first run.
+            Enabled = false;
```
gkassabli: This isn't related to PR, but I think this was missing here. The code in the system isn't really intended to be ran multiple times

jessicafalk: this might disable the system before it is actually run depending on when the components are added to the entities

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -339,7 +339,7 @@ protected static string CreateNewWorkerId(string workerType)
         }
 
 
-        private void OnDisconnected(string reason)
+        protected virtual void OnDisconnected(string reason)
```
jessicafalk: instead of making this protected, just register to the event directly on the worker: 

e.g. `Worker.OnDisconnect +=ConnectionScreenController.OnDisconnected `

workers/unity/Assets/Playground/Scripts/Camera/InitCameraSystem.cs
gkassabli: Good point. This should be fixed if moved to the cycle above, right?

gkassabli:
@jamiebrynes7 this is temporary to exclude unrelated code in mobile-clients-fix branch until it lands. Will be switched to master then

workers/unity/Assets/Playground/Scripts/UI/InitUISystem.cs
```diff
@@ -45,5 +45,10 @@ protected override void OnUpdate()
             // Disable system after first run.
             Enabled = false;
         }
+
+        protected override void OnDestroyManager()
+        {
+            UnityEngine.Object.Destroy(UIComponent.Main.gameObject);
```
jessicafalk: use the `UnityObjectDestroyer`

gkassabli: Interesting, didn't know about that one. Not related, but I am curious whey do we use DestoyImmediate in tests

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
gkassabli: Makes sense

workers/unity/Assets/Playground/Scripts/Camera/InitCameraSystem.cs
jamiebrynes7: Yeah if its in the `for` loop it should be good 👍

gkassabli:
Verified this works as expected with KCP on master.

gkassabli:
> LGTM. To check - if you try to reconnect, does everything work as normal?

Yup, after enabling connection back reconnect works as expected

PR #693, opened 2019-01-17
Switching to KCP on mobile worker + use small connection timeout

gkassabli:
#### Description
This PR switches mobile module to use KCP instead of TCP + adds meaningful connection timeout. Timeout seem to not have effect at the moment

#### Tests
Run mobile local deployment -> Run mobile client -> Ensure client is connected correctly to the deployment

#### Documentation
Do we need to change any docs for this? I wasn't able to find any relevant reference.

### DIFF
--- a/mobile_launch.json
+++ b/mobile_launch.json
@@ -1,5 +1,5 @@
 {
-  "template": "small",
+  "template": "small_entity_db_v2",
   "world": {
     "chunkEdgeLengthMeters": 50,
     "snapshots": {
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
@@ -21,7 +21,20 @@ namespace Improbable.Gdk.Mobile
                 {
                     ConnectionType = NetworkConnectionType.Kcp,
                     UseExternalIp = true,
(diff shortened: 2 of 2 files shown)


mobile_launch.json
```diff
@@ -1,5 +1,5 @@
 {
-  "template": "small",
+  "template": "small_entity_db_v2",
```
jessicafalk: was that intentional?

workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
```diff
@@ -19,8 +19,9 @@ protected override ConnectionParameters GetConnectionParameters(string workerTyp
                 WorkerType = workerType,
                 Network =
                 {
-                    ConnectionType = NetworkConnectionType.Tcp,
+                    ConnectionType = NetworkConnectionType.Kcp,
                     UseExternalIp = true,
+                    ConnectionTimeoutMillis = 1000,
```
jessicafalk: we will need to do some tests on what would be a good timeout, but 1s might be too short. let's set it to 10s for now

mobile_launch.json
gkassabli: Yes, this is a V2 template that supports KCP. Old one didn't allow KCP connection

gkassabli:
:) was actually wandering if we need it

PR #670, opened 2019-01-07
Adding an option to launch ios simulator client

gkassabli:
**Contributions**: We are not currently taking public contributions - see our [contributions](https://docs.improbable.io/unity/alpha/contributing) policy. However, we are accepting issues and we do want your [feedback](../README.md#give-us-feedback).

-------

#### Description
This PR adds a menu item to launch ios simulator client with host IP to connect to and the scripts to execute that.

#### Tests
Run Unity -> ensure an item to launch iOS simulator is available -> ensure it triggers simulator with configured IP address and tries connection to the host

#### Documentation
Will be added to changelog and documentation

PR #701, opened 2019-01-22
Rework RedirectProccess to have Builder API

gkassabli:
#### Description
Following the https://github.com/spatialos/gdk-for-unity/pull/670/files I've tried the idea to rework RedirectedProcess to have Builder API

#### Tests
Ensure the code using RedirectedProcess runs the same

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
+++ b/workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
@@ -259,10 +259,10 @@ namespace Improbable.Gdk.BuildSystem
         {
             using (new ShowProgressBarScope($"Package {basePath}"))
             {
-                RedirectedProcess.CommandWithArgs(Common.SpatialBinary, "file", "zip",
-                    $"--output=\"{Path.GetFullPath(zipAbsolutePath)}\"",
-                    $"--basePath=\"{Path.GetFullPath(basePath)}\"", "\"**\"",
-                    $"--compression={useCompression}")
+                RedirectedProcess.Command(Common.SpatialBinary)
+                    .WithArgs("file", "zip", $"--output=\"{Path.GetFullPath(zipAbsolutePath)}\"",
+                        $"--basePath=\"{Path.GetFullPath(basePath)}\"", "\"**\"",
+                        $"--compression={useCompression}")
                     .Run();
             }
         }
--- a/workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
+++ b/workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
@@ -259,10 +259,11 @@ namespace Improbable.Gdk.BuildSystem
         {
             using (new ShowProgressBarScope($"Package {basePath}"))
             {
-                RedirectedProcess.Run(Common.SpatialBinary, "file", "zip",
+                RedirectedProcess.CommandWithArgs(Common.SpatialBinary, "file", "zip",
                     $"--output=\"{Path.GetFullPath(zipAbsolutePath)}\"",
                     $"--basePath=\"{Path.GetFullPath(basePath)}\"", "\"**\"",
-                    $"--compression={useCompression}");
+                    $"--compression={useCompression}")
+                    .Run();
             }
         }
     }
--- a/workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
@@ -18,7 +18,7 @@ namespace Improbable.Gdk.Tools
     }
 
     [Flags]
-    public enum ProcessOutputOptions
+    public enum OutputRedirectBehaviour
     {
         /// <summary>
         ///   <para>No redirected output, only custom outputProcessors are used</para>
@@ -57,9 +57,9 @@ namespace Improbable.Gdk.Tools
         private readonly List<Action<string>> outputProcessors = new List<Action<string>>();
         private readonly List<Action<string>> errorProcessors = new List<Action<string>>();
 
-        private ProcessOutputOptions outputOptions =
-            ProcessOutputOptions.ProcessSpatialOutput |
-            ProcessOutputOptions.RedirectAccumulatedOutput;
+        private OutputRedirectBehaviour outputRedirectBehaviour =
+            OutputRedirectBehaviour.ProcessSpatialOutput |
+            OutputRedirectBehaviour.RedirectAccumulatedOutput;
 
         /// <summary>
         ///     Creates the redirected process for the command.
@@ -114,10 +114,10 @@ namespace Improbable.Gdk.Tools
         /// <summary>
         ///     Adds custom processing for error output of process.
         /// </summary>
-        /// <param name="options">Options for redirecting process output to Debug.Log().</param>
-        public RedirectedProcess RedirectOutputOptions(ProcessOutputOptions options)
+        /// <param name="redirectBehaviour">Options for redirecting process output to Debug.Log().</param>
+        public RedirectedProcess RedirectOutputOptions(OutputRedirectBehaviour redirectBehaviour)
         {
-            outputOptions = options;
+            outputRedirectBehaviour = redirectBehaviour;
             return this;
         }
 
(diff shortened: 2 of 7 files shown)


workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -257,10 +257,11 @@ private static void Zip(string zipAbsolutePath, string basePath, bool useCompres
         {
             using (new ShowProgressBarScope($"Package {basePath}"))
             {
-                RedirectedProcess.Run(Common.SpatialBinary, "file", "zip",
+                RedirectedProcess.CommandWithArgs(Common.SpatialBinary, "file", "zip",
```
jamiebrynes7: I'd prefer an explicit:

```
RedirectedProgress.Command(Common.SpatialBinary)
	.WithArgs(...)
```

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -23,15 +23,46 @@ public class RedirectedProcessResult
     /// </summary>
     public class RedirectedProcess
     {
+        private string RunningCommand = string.Empty;
+        private string[] Arguments;
+        private string WorkingDirectory = Path.GetFullPath(Path.Combine(Application.dataPath, ".."));
+        private Action<string> OutputProcessor;
+        private Action<string> ErrorProcessor;
```
jamiebrynes7: `private` variables should be camel case.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -23,15 +23,46 @@ public class RedirectedProcessResult
...
         ///     Runs the redirected process and waits for it to return.
         /// </summary>
         /// <param name="command">The filename to run.</param>
         /// <param name="arguments">Parameters that will be passed to the command.</param>
         /// <returns>The exit code.</returns>
-        public static int Run(string command, params string[] arguments)
+        // public static int Run(string command, params string[] arguments)
+        // {
+        //     return RunIn(Path.GetFullPath(Path.Combine(Application.dataPath, "..")), command, arguments);
+        // }
```
jamiebrynes7: Commented code.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -64,21 +96,35 @@ public static int RunIn(string workingDirectory, string command, params string[]
 
                 var processOutput = new StringBuilder();
 
-                void OnReceived(object sender, DataReceivedEventArgs args)
+                void OnReceived(string output)
```
jamiebrynes7: These probably shouldn't be local methods here. As we allocate the string builder regardless of whether we end up using this

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -107,83 +168,49 @@ void OnReceived(object sender, DataReceivedEventArgs args)
...
         ///     Runs the redirected process and returns a task which can be waited on.
         /// </summary>
-        /// <param name="workingDirectory">The directory to run the command from.</param>
-        /// <param name="command">The filename to run.</param>
-        /// <param name="arguments">Parameters that will be passed to the command.</param>
         /// <param name="redirectStdout">Redirect standard output to Debug.Log</param>
         /// <param name="redirectStderr">Redirect standard error to Debug.LogError</param>
         /// <returns>A task which would return the exit code and output.</returns>
-        public static async Task<RedirectedProcessResult> RunInAsync(string workingDirectory, string command, string[] arguments, bool redirectStdout = false, bool redirectStderr = false)
+        public async Task<RedirectedProcessResult> RunInAsync(bool redirectStdout = false, bool redirectStderr = false)
```
samiwh: Shouldn't this be RunAsync and the In part of it is folded back into whether or not that option has been set?

gkassabli: Kept the name as is - I'll rename it

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -107,83 +168,49 @@ void OnReceived(object sender, DataReceivedEventArgs args)
...
         ///     Runs the redirected process and returns a task which can be waited on.
         /// </summary>
-        /// <param name="workingDirectory">The directory to run the command from.</param>
-        /// <param name="command">The filename to run.</param>
-        /// <param name="arguments">Parameters that will be passed to the command.</param>
         /// <param name="redirectStdout">Redirect standard output to Debug.Log</param>
         /// <param name="redirectStderr">Redirect standard error to Debug.LogError</param>
         /// <returns>A task which would return the exit code and output.</returns>
-        public static async Task<RedirectedProcessResult> RunInAsync(string workingDirectory, string command, string[] arguments, bool redirectStdout = false, bool redirectStderr = false)
+        public async Task<RedirectedProcessResult> RunAsync(bool redirectStdout = false, bool redirectStderr = false)
```
jared-improbable: `bool`-based APIs generally smell a bit. Prefer an `enum` as flags here instead. 
For example, `[Flags] enum { RedirectStdErr, RedirectStdOut, RedirectAll }`
This avoid call sites that look like `RunAsync(true, true) <-- huh?` or the verbose `RunAsync(redirectStdout: true, redirectStderr: true)`.
Also, as a general point, `bools` tend to multiply and encourage "customizing" functions that should really be split out into their own separate area of concern. I've seen monstrosities like `thing.DoUpdate(false, 7, true, true)` more times than I can count.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -107,83 +168,49 @@ void OnReceived(object sender, DataReceivedEventArgs args)
...
-            var info = new ProcessStartInfo(command, string.Join(" ", arguments))
-            {
-                CreateNoWindow = true,
-                RedirectStandardError = true,
-                RedirectStandardOutput = true,
-                UseShellExecute = false,
-                WorkingDirectory = workingDirectory
-            };
-
             return await Task.Run(() =>
```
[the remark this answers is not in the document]
gkassabli: Left this part as is. cc @davedissian

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
gkassabli: Good point, let me rework this API

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -17,31 +17,114 @@ public class RedirectedProcessResult
         public List<string> Stderr;
     }
 
+    [Flags]
+    public enum ProcessOutputOptions
+    {
+        /// <summary>
+        ///   <para>No redirected output, only custom outputProcessors are used</para>
+        /// </summary>
+        None = 0,
```
jamiebrynes7: `NoRedirect` ?

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -60,44 +143,101 @@ public static int RunIn(string workingDirectory, string command, params string[]
...
+                {
+                    return process.ExitCode;
+                }
+
                 // Ensure that the first line of the log is something useful in the Unity editor console.
-                var trimmedOutput = processOutput.ToString().Trim();
+                var trimmedOutput = outputLog.ToString().TrimStart();
 
-                if (!string.IsNullOrEmpty(trimmedOutput))
+                if (string.IsNullOrEmpty(trimmedOutput))
```
jamiebrynes7: Nit - Only need to check for `empty` here

jamiebrynes7:
Oh also! Changelog 😉 

Its technically a public API that someone may be using.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
gkassabli: This is set of flags, not direct enum, so `.NoRedirect | .RedirectStdOut` is a valid assignment, though doesn't make a lot of sense semantically. None also doesn't look great in this case, but I am not sure what would be a good naming

jamiebrynes7: Alternatively you could rename the enum to `RedirectBehaviour` in which case `None` makes more sense 💯

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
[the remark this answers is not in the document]
jared-improbable: Then you definitely want to remove the `await` here

PR #703, opened 2019-01-23
Use the new Runtime by default

paulbalaji:
#### Description
Update launch configurations to use the new Runtime.

#### Tests
ran locally + in cloud

#### Documentation
n/a

no I didn't - thank you kind PR template

#### Primary reviewers


default_launch.json
```diff
@@ -16,26 +16,35 @@
       }
     ]
   },
+  "load_balancing": {
+    "layer_configurations": [
+      {
+        "layer": "UnityGameLogic",
+        "hex_grid": {
+          "num_workers": 2
```
jamiebrynes7: This is a different configuration to the previous JSON

paulbalaji: there's no direct equivalent for the `singleton` strategy used before

I could switch this to POI with one point at the origin?

PR #704, opened 2019-01-23
Add CI Launch Script

jamiebrynes7:
#### Description
Adds a CI script to build, upload, and launch a deployment for QA testing during the release process.

#### Tests
Ran the new pipeline through this branch - it worked! 

#### Documentation
Will add this to the release process.

N/A

#### Primary reviewers


ci/upload-launch.sh
```diff
@@ -0,0 +1,22 @@
...
+source ".shared-ci/scripts/profiling.sh"
+
+.shared-ci/scripts/upload-assemblies.sh ${PREFIX}
+
+markStartOfBlock "Launching deployment"
+
+# Get first 8 characters of current git hash.
+GIT_HASH="$(git rev-parse HEAD | cut -c1-8)"
+
+spatial cloud launch "${PREFIX}_${GIT_HASH}" cloud_launch.json "${PREFIX}_${GIT_HASH}" --snapshot=snapshots/default.snapshot
```
jamiebrynes7: In retrospect - pulling the prefix + git hash out to a shared function may be preferable. Given that its used here & upload-assemblies.

gkassabli: Yup, having it used in both scripts without clear dependency looks quite fragile. Can we instead just pass the full name to upload-assemblies?

PR #705, opened 2019-01-24
Release 0.1.4 Prep

jamiebrynes7:
#### Description
- Changelog updated.
- Package versions updated.

Note removed the steam auth flow from the changelog as the WIP PR for that was dropped.

#### Tests
Release QA in progress.

#### Documentation
Ye

**Did you remember a changelog entry?**
Of course.

#### Primary reviewers
If your change will take a long time to review, you can name at most two primary reviewers who are ultimately responsible for reviewing this request. @ mention them.

PR #706, opened 2019-01-24
Build System Tweaks

jamiebrynes7:
#### Description
A last minute addition to the `0.1.4` release ~_grumble grumble_~ - adjusting the build system to skip builds that you don't have build support for (without cancelling the other builds). 

For example - if you build all workers without mobile support installed you would get the following messages in the console.

errors
all workers build

Also the static `Build` method still throws if build support is missing (such that the process _actually_ returns an error exit code) - so that behavior is unchanged.

#### Tests
Tested a build locally and via command line.

#### Documentation

Yes

#### Primary reviewers
@jessicafalk @jared-improbable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Updated the `MobileWorkerConnector` to use the KCP network protocol by default.
 - Changed the `mobile_launch.json` config to use the new Runtime.
 - Updated all the launch configs to use the new Runtime.
+- Changed the way builds are processed in the Unity Editor. If you don't have build support for a particular target, that specific build target will be skipped rather than the whole build process cancelled.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - Updated the `MobileWorkerConnector` to use the KCP network protocol by default.
 - Changed the `mobile_launch.json` config to use the new Runtime.
 - Updated all the launch configs to use the new Runtime.
+- Changed the way builds are processed in the Unity Editor. If you don't have build support for a particular target, that specific build target will be skipped rather than the whole build process cancelled.
```
gkassabli: Can we drop the first sentence and just say something like "Changed the Unity Editor building process to skip builds with missing targets"?

jamiebrynes7: I might be cheeky and change it in #705 so I don't have to rerun CI here 😛

jamiebrynes7: Unless people have code concerns

PR #707, opened 2019-01-25
Removed the concept of a temporarily removed entity

samiwh:
No longer a concept of temporarily removed entities. Should fix the initial add - remove - add problem.


samiwh:
The alternative fix is to make the reactive component systems also check for temporarily removed entities.

PR #716, opened 2019-01-29
Fix Package Traversal Exception

jamiebrynes7:
#### Description
In our `package.json` parsing code, if a package did not have a `dependencies` field, it would throw exceptions despite this being a valid package declaration.

This could cause code generation to fail as @polartron experienced. 

#### Tests
Added a mock package with no `dependencies` field. Ran code generation, it was all good.

#### Documentation

N/A

Yes

#### Primary reviewers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/Common.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/Common.cs
@@ -95,7 +95,12 @@ namespace Improbable.Gdk.Tools
             try
             {
                 var package = Json.Deserialize(File.ReadAllText(filePath, Encoding.UTF8));
-                return ((Dictionary<string, object>) package["dependencies"]).ToDictionary(kv => kv.Key,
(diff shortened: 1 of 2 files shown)


workers/unity/Packages/com.improbable.gdk.tools/Common.cs
```diff
@@ -95,7 +95,12 @@ public static string GetPackagePath(string packageName)
             try
             {
                 var package = Json.Deserialize(File.ReadAllText(filePath, Encoding.UTF8));
-                return ((Dictionary<string, object>) package["dependencies"]).ToDictionary(kv => kv.Key,
+                if (!package.TryGetValue("dependencies", out var dependenciesJson))
```
zeroZshadow: If you replace `var` with `Dictionary<string, object>`, we might be able to do without the cast below?

jamiebrynes7: You can't do the cast in the out parameter declaration sadly

PR #726, opened 2019-01-31
Setup Docs Bypass for Buildkite

jamiebrynes7:
#### Description
This PR introduces the same functionality for bypassing building code for `docs/*` branches. The strategy is slightly different for Buildkite than TeamCity. 

In TeamCity we had a separate pipeline that ran separate steps that reported the exact same status check - this allowed us to mark one status check as required which could be fulfilled by either pipeline.

In Buildkite however, separate pipelines will report different status checks automatically (based on pipeline name, slug, etc.). So instead, we enable all steps on the `premerge` pipeline and just early exit depending on whether we are a docs branch. 

If you have a `docs/*` branch: 

- "build" step will early exit
- "verify docs only changes" step will run
- "lint docs" will run

If you have a non `docs/*` branch:

- "build" step will early exit
- "verify docs only changes" step will run
- "lint docs" step will run

Note the change in `docs-branch-verification.sh` is required as buildkite checks out a detached commit instead of a branch. Also fake configuration to make Git happy when performing the merge.

#### Tests
Ran Buildkite with this branch and a `docs/test` branch branched off of this. Behaviour was as expected.

Need to test that TeamCity has unchanged behaviour.

#### Documentation
N/A - Some guide to CI which has yet to be written

N/A

#### Primary reviewers


jamiebrynes7:
Tested on TeamCity - existing functionality is preserved!

PR #729, opened 2019-02-01
Run Build & Test in Parallel on Buildkite

jamiebrynes7:
#### Description
Run all build and test steps in parallel. In the best case this reduces the CI time from ~15 minutes to 5 minutes. This does require _more_ agents available, the case where everything is sequential (on one agent), this method will be slower. However, we expect to have more than 1 agent available for a build. 

- Added `build.sh` which acts as a passthrough to the shared-ci version which correctly sets up the environment and sets the worker type, etc.
- Made `test.sh` able to be run from standalone on the CI agent.
- Removed `build-test.sh` from Buildkite altogether!
- Early out for building iOS on Windows.

#### Tests
Ran on Buildkite

#### Documentation
N/A

#### Primary reviewers

### DIFF
--- a/.buildkite/premerge.steps.yaml
+++ b/.buildkite/premerge.steps.yaml
@@ -40,36 +40,31 @@ steps:
       - logs/**/*
 
   - label: "build :android:"
-    command: >
-      bash -c 'ci/build.sh "$0" "$1" "$2"' "AndroidClient" "local" "mono"
+    command: "bash -c 'ci/build.sh \"$0\" \"$1\" \"$2\"' \"AndroidClient\" \"local\" \"mono\""
     <<: *common # This folds the YAML named anchor into this step. Overrides, if any, should follow, not precede.
     artifact_paths:
       - logs/**/*
 
   - label: "build :iphone:"
-    command: > 
-      bash -c 'ci/build.sh "$0" "$1" "$2"' "iOSClient" "local" "mono"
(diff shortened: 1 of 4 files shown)


jamiebrynes7:
Nice bonus of this is that you can see which build failed by looking at the GitHub checks UI 😄

.buildkite/premerge.steps.yaml
```diff
@@ -25,14 +25,66 @@ common: &common
...
+    artifact_paths:
+      - logs/**/*
+
+  - label: "build :android:"
+    command: bash -c ci/build.sh
+    <<: *common
+    artifact_paths:
+      - logs/**/*
+    env: 
+      WORKER_TYPE: "AndroidClient"
```
gkassabli: Curious why do we pass this through env vars not command arguments?

jamiebrynes7: It gets very ugly due to `bash -c`. See https://github.com/spatialos/gdk-for-unity/blob/617b98dbcff3618b32c8326ba6d8abde7ba9df14/.buildkite/premerge.steps.yaml for an example of what it would look like.

TL;DR - Parameters passed in the `command` field are taken as parameters to the `bash` command, not the invocation of the script. So you must interpolate the arguments from bash into the invokation of the script. This also means you have to do some nasty YAML escaping to get everything running nicely & edit`trigger.sh` to turn off Buildkite trying to interpolate the variables (which is does when you call `buildkite-agent upload pipeline`).

Actually that wasn't very TL;DR 😬

PR #732, opened 2019-02-04
Set systems to always run

samiwh:
Some systems didn't have the tag to always run


zeroZshadow:
NIT: Attributes are in a different order in each file

PR #733, opened 2019-02-04
Add release QA pipeline for buildkite

jamiebrynes7:
#### Description

- Adding a Buildkite pipeline to replicate the TeamCity one.
- Uses the artifacts functionality to distribute the build load.
- Added `spatial prepare-to-run` at the end of `build.sh` to ensure the schema.descriptor is built. Launching fails without this. Longer term once we move to FPL we will have better control over this.
- Added a build annotation that surfaces the deployment URL. Uses regex magic to extract it from the spatial log. If anyone has a better suggestion please let me know.

#### Tests
Ran the pipeline! See --> https://buildkite.com/improbable/gdk-for-unity-release-qa/builds/15

#### Documentation
N/A - Existing docs will be updated when we migrate to Buildkite fully.

N/A

#### Primary reviewers
@jared-improbable @jessicafalk


ci/upload-launch.sh
```diff
@@ -6,17 +6,34 @@ cd "$(dirname "$0")/../"
 
 PREFIX="playground"
 
-ci/build-test.sh
-
 source ".shared-ci/scripts/profiling.sh"
 source ".shared-ci/scripts/pinned-tools.sh"
 
+if [[ -n "${BUILDKITE}" ]]; then
```
gkassabli: We might want to add `isBuildKite` check if we intend to use this extensively

jamiebrynes7: Probably not worth it - all these checks can be removed once we are off TeamCity. Especially as it doesn't reduce the verbosity much at all

PR #737, opened 2019-02-05
Schema Bundle Migration

jamiebrynes7:
#### Description
This PR contains the work required to migrate the code generator from the current AST to schema bundles.

This involves wiring up the `UnityxxxDetails` classes to a new data format. In the process, I'm cutting out the `UnityxxxDefinition` middle man as its no longer required.

This _will_ be a pretty big PR as its tricky to do this incrementally. 

PR is kinda commit ordered, at least it should help break up the review.  

#### Tests
Will test against exhaustive schema.

#### Documentation
N/A - Purely internal, allows for things like schema annotations and paves the way for modular codegen.

Will do so in the final PR.

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Model/SchemaBundleV1/Common.cs
```diff
@@ -1,13 +1,73 @@
...
+        {
+            return new Identifier
+            {
+                QualifiedName = qualifiedName,
+                Name = qualifiedName.Split(".").Last(),
+                Path = qualifiedName.Split(".").ToList()
+            };
+        }
+
+        public bool IsChild(Identifier identifier)
```
jared-improbable: It feels weird to have `IsChild` and `FromQualifiedName` in something that (IMO) should be a very pure data container. Can you refactor so that hierarchy is decided from outside of this class?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Utils/Formatting.cs
```diff
@@ -0,0 +1,59 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Improbable.Gdk.CodeGeneration.Utils
+{
+    public static class Formatting
```
jared-improbable: Can you refactor these to take `Identifiers`? (I mean to do this in the dotnetcore version)

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
```diff
@@ -0,0 +1,144 @@
...
+{
+    public class DetailsStore
+    {
+        public IReadOnlyDictionary<Identifier, UnityTypeDetails> Types { get; }
+        public IReadOnlyDictionary<Identifier, UnityEnumDetails> Enums { get; }
+        public IReadOnlyDictionary<Identifier, UnityComponentDetails> Components { get; }
+        public readonly ReadOnlyDictionary<Identifier, bool> BlittableMap;
+        public IReadOnlyList<string> SchemaFiles { get; }
+
+        public static readonly HashSet<string> NonBlittableSchemaTypes = new HashSet<string> { "String", "Bytes" };
```
jared-improbable: Use the constants for these types?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
```diff
@@ -0,0 +1,144 @@
...
+using Improbable.Gdk.CodeGeneration.Model.SchemaBundleV1;
+
+namespace Improbable.Gdk.CodeGenerator
+{
+    public class DetailsStore
+    {
+        public IReadOnlyDictionary<Identifier, UnityTypeDetails> Types { get; }
+        public IReadOnlyDictionary<Identifier, UnityEnumDetails> Enums { get; }
+        public IReadOnlyDictionary<Identifier, UnityComponentDetails> Components { get; }
+        public readonly ReadOnlyDictionary<Identifier, bool> BlittableMap;
```
jared-improbable: You could also use a `HashSet` and `Contains`

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
```diff
@@ -0,0 +1,144 @@
...
+                if (field.Map != null || field.List != null || field.Option != null)
+                {
+                    return false;
+                }
+
+                var innerSingularType = field.Singular.Type;
+
+                if (innerSingularType.UserType != null)
+                {
+                    if (!blittableMap.TryGetValue(Identifier.FromQualifiedName(innerSingularType.UserType.QualifiedName), out var isFieldBlittable))
```
jared-improbable: Why can't you use the `Identifier` directly?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityTypeDetails.cs
```diff
@@ -1,26 +1,67 @@
...
-            CapitalisedName = Formatting.SnakeCaseToCapitalisedCamelCase(typeDefinition.Name);
-            CamelCaseName = Formatting.SnakeCaseToCamelCase(typeDefinition.Name);
-            FullyQualifiedSpatialTypeName =
-                "global::" + Formatting.CapitaliseQualifiedNameParts(typeDefinition.QualifiedName);
-            FullyQualifiedTypeName = "global::" + Formatting.CapitaliseQualifiedNameParts(typeDefinition.QualifiedName);
+            CapitalisedName = typeDefinitionRaw.Identifier.Name;
+            CamelCaseName = Formatting.PascalCaseToCamelCase(CapitalisedName);
+            FullyQualifiedTypeName =
+                $"global::{Formatting.CapitaliseQualifiedNameParts(typeDefinitionRaw.Identifier.QualifiedName)}";
+            FullyQualifiedSpatialTypeName = "Not sure if needed";
```
jared-improbable: 😮

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Model/SchemaBundleV1/Common.cs
jamiebrynes7: The `IsChild` method can be moved somewhere else easily.

The `FromQualifiedName` is required as type references in commands and events are only given a FQN, not a fully identifier struct. So when generating event/command code, I need to access the actual type --> create an identifier.

I can move this out of the class definition into something else but will end up with a `IdentifierFactory.FromQualifiedName(string fqn)` or similar.

---

The alternative is to key all the types on the qualified name instead, but sort of defeats the purpose of an Identifier. On reflection - this should probably be an FR.

For reference the JSON of the event definition: 

```
{
      "identifier": {
       "qualifiedName": "playground.CubeColor.change_color",
       "name": "change_color",
       "path": [
        "playground",
        "CubeColor",
        "change_color"
       ]
      },
      "eventIndex": 1,
      "type": {
       "type": {
        "qualifiedName": "playground.ColorData"
       }
      },
```

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Utils/Formatting.cs
jamiebrynes7: I think so - will need to check we don't call them anywhere without an `Identifier`

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
jamiebrynes7: Yes you are correct.

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
jamiebrynes7: There isn't an identifier inside the type definitions:

```
     {
      "identifier": {
       "qualifiedName": "playground.SpinnerColor.color",
       "name": "color",
       "path": [
        "playground",
        "SpinnerColor",
        "color"
       ]
      },
      "fieldId": 1,
      "transient": false,
      "singularType": {
       "type": {
        "enum": {
         "qualifiedName": "playground.Color"
        }
       }
      },
      "annotations": []
     }
```

jamiebrynes7: Same issue as discussed above - will raise an FR

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityTypeDetails.cs
jamiebrynes7: 😬

Forgot to annotate with a TODO!

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Utils/Formatting.cs
```diff
@@ -0,0 +1,59 @@
...
+                .ToArray();
+            return string.Join(".", parts);
+        }
+
+        public static string SnakeCaseToPascalCase(string text)
+        {
+            return ToPascalCase(text.Split(new[] { "_" }, StringSplitOptions.RemoveEmptyEntries));
+        }
+
+        private static string ToPascalCase(IEnumerable<string> parts)
```
gkassabli: Does this need to be separate? It is only used in SnakeCaseToPascalCase which is quite simple on itself

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Model/SchemaBundleV1/Common.cs
gkassabli: Does `a.isChild(b)` means that `a` is child of `b` of that `b` is child of `a`?
For me it reads like in this function it is latter, though semantically I would expect this to be the former, or am I reading this wrong?

jamiebrynes7: Moved `IsChild` to the calling site and `FromQualifiedName` to `CommonDetailsUtils.CreateIdentifier`

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Utils/Formatting.cs
jamiebrynes7: Discussed offline - this will be dependent on the fate of the `Identifier`. Will leave for now.

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityEnumDetails.cs
```diff
@@ -1,18 +1,25 @@
...
-        
-        public UnityEnumDetails(EnumDefinitionRaw enumDefinition)
+        public List<(uint, string)> Values;
+
+        public Identifier Identifier;
+
+        public UnityEnumDetails(EnumDefinitionRaw enumDefinitionRaw)
         {
-            this.enumDefinition = enumDefinition;
+            TypeName = enumDefinitionRaw.EnumIdentifier.Name;
```
gkassabli: Do we need to convert case here?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityFieldDetails.cs
```diff
@@ -1,44 +1,68 @@
...
-            IsBlittable = isBlittable;
+            PascalCaseName = Formatting.SnakeCaseToPascalCase(field.Identifier.Name);
+            CamelCaseName = Formatting.PascalCaseToCamelCase(PascalCaseName);
+            FieldNumber = field.FieldId;
 
-            if (rawFieldDefinition.IsOption())
+            if (field.Option != null)
             {
-                fieldType = new OptionFieldType(rawFieldDefinition.optionType, enumSet);
+                IsBlittable = false;
```
gkassabli: It feels like we're duplicating the code for blittable check here. Any way we can merge them into a single function?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Model/SchemaBundleV1/Bundle.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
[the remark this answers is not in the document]
jamiebrynes7: Nope!

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Utils/Formatting.cs
jamiebrynes7: Will merge them 👍

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityEnumDetails.cs
jamiebrynes7: Nope - the names of types are already PascalCase in the `bundle.json`. Fields and event/command names are snake_case

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityFieldDetails.cs
jamiebrynes7: Hmm maybe.. the problem I'm having is that the methods in `DetailsStore` are used in the context of - "we are building a blittable map for all identifiers" vs "we are doing a lookup in that completed map". 

So for things like option, map, list, that's fine - but when you go to check the `UserType` of a singular type, you end up accessing a different data type (a mutable dictionary in the `DetailsStore` internals, and an `ImmutableHashSet` in the `FieldDetails`).

There is likely a way to factor this out nicely - probably by using the `Identifier` of the `Field` to make entries into the `ImmutableHashSet` so its always just a lookup in the `UnityFieldDetails`

gkassabli: Can we have one function that takes a Map/HashSet of already identified Types and just pass incomplete or complete one depending on the context where it is called?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            // Populate all enums.
+            foreach (var enumm in bundle.BundleContents.EnumDefinitions)
+            {
+                blittableMap.Add(enumm.EnumIdentifier, true);
+            }
+
+            var typesToTraverse = new Queue<TypeDefinitionRaw>(bundle.BundleContents.TypeDefinitions);
+
+            while (typesToTraverse.Count > 0)
```
gkassabli: Is there a chance we would go into infinite loop here if two types are referencing each other?

jamiebrynes7: No as the only way that can happen is through a reference type which marks it as non-blittable immediately. (I believe)

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityFieldDetails.cs
jamiebrynes7: Refactored this in https://github.com/spatialos/gdk-for-unity/pull/737/commits/89f563292bca13d6b32a787649e8d765e3e7699d PTAL 😄

gkassabli: Love it!

jamiebrynes7:
Also - no diff on the test generated code! Which is good! 🎉

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
```diff
@@ -117,35 +118,45 @@ private void PopulateBlittableMaps()
 
         private bool? CheckBlittable(IEnumerable<Field> fields)
```
jared-improbable: A field is either blittable, or not - it feels over-complicated to have 3 potential states for a result.
I want to reiterate that storing a mapping of `name->bool` also feels over-complicated when presence or absence in a `HashSet` represents the same information more succinctly.

jamiebrynes7: Will change this to a DFS instead of BFS

jamiebrynes7: After going down the rabbit hole a bit, and discussing this offline - a solution that involves dropping the tri-state logic would require building a DAG representation of the types, where nodes are types, and edges represent a field that is a user type (dependency between types). You then check nodes where (all dependencies already checked = true) and insert into a hashset if its blittable.

At the end of the day, this solution is probably more complex and is easier to get wrong (imo), than the naive solution currently implemented.

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/GdkCodeGenerator.csproj
```diff
@@ -19,18 +19,14 @@
     <EmbeddedResource Include="Partials\Improbable.Coordinates" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.2" />
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
-    <PackageReference Include="NUnit" Version="3.10.0" />
-    <PackageReference Include="NUnit3TestAdapter" Version="3.10.0" />
-    <PackageReference Include="NunitXml.TestLogger" Version="1.1.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
+    <PackageReference Include="Mono.Options" Version="5.3.0.1" />
```
jessicafalk: do we need this new dependency?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/UnityTypeMappings.cs
```diff
@@ -1,45 +1,29 @@
...
+                { BuiltInSchemaTypes.BuiltInSint32, "int" },
+                { BuiltInSchemaTypes.BuiltInSint64, "long" },
+                { BuiltInSchemaTypes.BuiltInFixed32, "uint" },
+                { BuiltInSchemaTypes.BuiltInFixed64, "ulong" },
+                { BuiltInSchemaTypes.BuiltInSfixed32, "int" },
+                { BuiltInSchemaTypes.BuiltInSfixed64, "long" },
+                { BuiltInSchemaTypes.BuiltInBool, "BlittableBool" },
+                { BuiltInSchemaTypes.BuiltInString, "string" },
+                { BuiltInSchemaTypes.BuiltInBytes, "byte[]" },
+                { BuiltInSchemaTypes.BuiltInEntityId, "global::Improbable.Gdk.Core.EntityId" }
```
jessicafalk: what happened to the vector and coordinate types?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityEventDetails.cs
```diff
@@ -1,19 +1,21 @@
...
 
 namespace Improbable.Gdk.CodeGenerator
 {
     public class UnityEventDetails
     {
-        public string EventName => Formatting.SnakeCaseToCapitalisedCamelCase(eventDefinition.Name);
-        public string CamelCaseEventName => Formatting.SnakeCaseToCamelCase(eventDefinition.Name);
-        public string FqnPayloadType => CommonDetailsUtils.GetCapitalisedFqnTypename(eventDefinition.RawType.TypeName);
-        public uint EventIndex => eventDefinition.EventIndex;
+        public string EventName { get; }
```
jessicafalk: do we need them to be properties? they could also just be fields with a readonly attribute

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityFieldDetails.cs
```diff
@@ -1,44 +1,68 @@
-using System.Collections.Generic;
-using Improbable.CodeGeneration.Model;
-using Improbable.CodeGeneration.Utils;
+using System;
+using System.Collections.ObjectModel;
+using Improbable.Gdk.CodeGeneration.Model.SchemaBundleV1;
```
jessicafalk: `SchemaBundleV1` - V1? is this generated?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/GdkCodeGenerator.csproj
jamiebrynes7: I'm baffled because this was definitely used before - it may have been bundled in the old `CodeGenerator` dependency 🤔

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/UnityTypeMappings.cs
jamiebrynes7: They are handled by the normal generation process (i.e. - they have an `Identifier` and are in the lookup table)

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityEventDetails.cs
jamiebrynes7: They could be - the properties give a bit more flexibility imo

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/UnityFieldDetails.cs
jamiebrynes7: This matches the bundle specification - I guess the implication is that at some point we have have `SchemaBundleV2` and they could exist side by side

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/CodeGeneration/Model/SchemaBundleV1/Common.cs
```diff
@@ -1,13 +1,57 @@
+using System;
 using System.Collections.Generic;
+using System.Linq;
```
jessicafalk: do we need that?

jamiebrynes7: Its used in:

```
public string PackagePath => string.Join(".", Path.Take(Path.Count - 1));
```

PR #738, opened 2019-02-07
Subscription component check

samiwh:
Fixed an issue where we didn't check the component existed before checking it's authority state.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentUpdateManagerGenerator.tt
+++ b/workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentUpdateManagerGenerator.tt
@@ -350,7 +350,7 @@ namespace <#= qualifiedNamespace #>
             {
                 if (!entityIdToAuthority.TryGetValue(entityId, out var authority))
                 {
-                    throw new ArgumentException("Entity not in view");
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentUpdateManagerGenerator.tt
```diff
@@ -350,7 +350,7 @@ namespace <#= qualifiedNamespace #>
             {
                 if (!entityIdToAuthority.TryGetValue(entityId, out var authority))
                 {
-                    throw new ArgumentException("Entity not in view");
+                    throw new ArgumentException("Can not check authority; entity-component not in view.");
```
jessicafalk: can we add which entity id and component combination is failing?

jamiebrynes7: Yeah that would be useful. This could be spun out into a new exception with ids as fields so you don't have to string interpolate here

samiwh: Making a ticket for this. There are quite a few exceptions that need to be changed and I suspect that whatever type we make for this will need to change subtly in the next couple of weeks.

PR #739, opened 2019-02-07
Lifecycle API validation

zeroZshadow:
#### Description
Re-implemented the Lifecycle systems using the new CommandSystem

#### Tests
Tested locally, runningly kicking off a cloud build to double check.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
@@ -12,23 +12,13 @@ namespace Improbable.Gdk.PlayerLifecycle
     [UpdateInGroup(typeof(SpatialOSUpdateGroup))]
     public class HandleCreatePlayerRequestSystem : ComponentSystem
     {
-        private struct CreatePlayerData
+        private CommandSystem commandSystem;
+
+        protected override void OnCreateManager()
         {
-            public readonly int Length;
-            [ReadOnly] public ComponentDataArray<PlayerCreator.CommandRequests.CreatePlayer> CreatePlayerRequests;
-            [ReadOnly] public ComponentDataArray<WorldCommands.CreateEntity.CommandSender> CreateEntitySender;
+            base.OnCreateManager();
+            commandSystem = World.GetExistingManager<CommandSystem>();
         }
-
-        [Inject] private CreatePlayerData createPlayerData;
-
-        private struct EntityCreationResponseData
-        {
-            public readonly int Length;
-            [ReadOnly] public ComponentDataArray<WorldCommands.CreateEntity.CommandResponses> CreateEntityResponses;
-            [ReadOnly] public ComponentDataArray<PlayerCreator.CommandResponders.CreatePlayer> CreatePlayerResponders;
-        }
-
-        [Inject] private EntityCreationResponseData entityCreationResponseData;
 
         private class PlayerCreationRequestContext
         {
@@ -37,58 +27,70 @@ namespace Improbable.Gdk.PlayerLifecycle
 
         protected override void OnUpdate()
         {
-            for (var i = 0; i < createPlayerData.Length; i++)
+            HandlePlayerCreateRequests();
+            HandlePlayerCreateEntityResponses();
+        }
+
+        private void HandlePlayerCreateRequests()
+        {
+            if (PlayerLifecycleConfig.CreatePlayerEntityTemplate == null)
             {
-                var requests = createPlayerData.CreatePlayerRequests[i].Requests;
-                var createEntitySender = createPlayerData.CreateEntitySender[i];
-
-                foreach (var request in requests)
-                {
-                    if (PlayerLifecycleConfig.CreatePlayerEntityTemplate == null)
-                    {
-                        throw new InvalidOperationException(Errors.PlayerEntityTemplateNotFound);
-                    }
-
-                    var playerEntity = PlayerLifecycleConfig.CreatePlayerEntityTemplate(request.CallerWorkerId,
-                        request.Payload.Position);
-                    createEntitySender.RequestsToSend.Add(new WorldCommands.CreateEntity.Request
-                    (
-                        playerEntity,
-                        context: new PlayerCreationRequestContext
-                        {
-                            createPlayerRequest = request
-                        }
-                    ));
-                }
+                throw new InvalidOperationException(Errors.PlayerEntityTemplateNotFound);
             }
 
-            for (var i = 0; i < entityCreationResponseData.Length; ++i)
+            var requests = commandSystem.GetRequests<PlayerCreator.CreatePlayer.ReceivedRequest>();
+            foreach (var request in requests)
             {
-                var entityCreationResponses = entityCreationResponseData.CreateEntityResponses[i];
-                var responder = entityCreationResponseData.CreatePlayerResponders[i];
+                var playerEntityTemplate = PlayerLifecycleConfig.CreatePlayerEntityTemplate(
+                    request.CallerWorkerId,
+                    request.Payload.Position
+                );
 
-                foreach (var receivedResponse in entityCreationResponses.Responses)
+                var entityRequest = new WorldCommands.CreateEntity.Request
+                (
+                    playerEntityTemplate,
+                    context: new PlayerCreationRequestContext
+                    {
+                        createPlayerRequest = request
+                    }
+                );
+
+                commandSystem.SendCommand(entityRequest);
+            }
+        }
+
+        private void HandlePlayerCreateEntityResponses()
+        {
+            var responses = commandSystem.GetResponses<WorldCommands.CreateEntity.ReceivedResponse>();
+            foreach (var response in responses)
+            {
+                if (!(response.Context is PlayerCreationRequestContext requestContext))
                 {
-                    if (!(receivedResponse.Context is PlayerCreationRequestContext requestContext))
-                    {
-                        // Ignore non-player entity creation requests
-                        continue;
-                    }
+                    // Ignore non-player entity creation requests
+                    continue;
+                }
 
-                    if (receivedResponse.StatusCode != StatusCode.Success || !receivedResponse.EntityId.HasValue)
-                    {
-                        responder.ResponsesToSend.Add(new PlayerCreator.CreatePlayer
-                            .Response(requestContext.createPlayerRequest.RequestId,
-                                $"Failed to create player: \"{receivedResponse.Message}\""));
 
-                        continue;
-                    }
+                if (response.StatusCode != StatusCode.Success || !response.EntityId.HasValue)
+                {
+                    var responseFailed = new PlayerCreator.CreatePlayer.Response(
+                        requestContext.createPlayerRequest.RequestId,
+                        $"Failed to create player: \"{response.Message}\""
+                    );
 
-                    responder.ResponsesToSend.Add(new PlayerCreator.CreatePlayer
-                        .Response(requestContext.createPlayerRequest.RequestId, new CreatePlayerResponseType
+                    commandSystem.SendResponse(responseFailed);
+                }
+                else
+                {
+                    var responseSuccess = new PlayerCreator.CreatePlayer.Response(
+                        requestContext.createPlayerRequest.RequestId,
+                        new CreatePlayerResponseType
                         {
-                            CreatedEntityId = receivedResponse.EntityId.Value
-                        }));
+                            CreatedEntityId = response.EntityId.Value
+                        }
+                    );
+
+                    commandSystem.SendResponse(responseSuccess);
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
```diff
@@ -12,83 +12,85 @@ namespace Improbable.Gdk.PlayerLifecycle
...
 
-                    if (receivedResponse.StatusCode != StatusCode.Success || !receivedResponse.EntityId.HasValue)
-                    {
-                        responder.ResponsesToSend.Add(new PlayerCreator.CreatePlayer
-                            .Response(requestContext.createPlayerRequest.RequestId,
-                                $"Failed to create player: \"{receivedResponse.Message}\""));
 
-                        continue;
-                    }
+                if (response.StatusCode != StatusCode.Success || !response.EntityId.HasValue)
```
jamiebrynes7: When would the status code be success but no entity id is returned?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -14,76 +12,77 @@ public class SendCreatePlayerRequestSystem : ComponentSystem
...
-                    if (receivedResponse.StatusCode == StatusCode.AuthorityLost)
-                    {
-                        PostUpdateCommands.AddComponent(responseData.Entities[i], new ShouldRequestPlayerTag());
-                    }
-                    else if (receivedResponse.StatusCode != StatusCode.Success)
+                    if (workerSystem.TryGetEntity(response.EntityId, out var entity))
                     {
-                        logDispatcher.HandleLog(LogType.Error, new LogEvent(
-                            $"Create player request failed: {receivedResponse.Message}"));
+                        PostUpdateCommands.AddComponent(entity, new ShouldRequestPlayerTag());
```
jamiebrynes7: This is not the same.

Previously, it used the ECS entity that sent the command (the worker entity), now it uses the Entity Id of the entity which we _sent_ the command to.

Given the state of the world at startup - you will probably never have this entity checked out. 

This should instead add the tag to the worker entity again.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
```diff
@@ -1,101 +1,87 @@
...
-                    {
-                        if (response.StatusCode == StatusCode.Success)
-                        {
-                            responded = true;
-                            break;
-                        }
-                    }
+                var key = (spatialId, entity);
+                respondedCache.TryGetValue(key, out var responded);
+                respondedCache[key] = responded | (response.StatusCode == StatusCode.Success);
```
jamiebrynes7: Don't you lose information if you receive two heartbeat responses in the same frame of: success, failed? 

Not that this is very likely and not very problematic?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
```diff
@@ -23,25 +24,20 @@ public class SendPlayerHeartbeatRequestSystem : ComponentSystem
...
-                },
-                Any = Array.Empty<ComponentType>(),
-                None = Array.Empty<ComponentType>()
-            };
+            group = GetComponentGroup(
+                ComponentType.ReadOnly<PlayerHeartbeatServer.ComponentAuthority>(),
+                ComponentType.Create<HeartbeatData>(),
+                ComponentType.ReadOnly<SpatialEntityId>()
+            );
+            group.SetFilter(new PlayerHeartbeatServer.ComponentAuthority(true));
```
jamiebrynes7: I didn't know you could filter in `OnCreateManager`

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/HandlePlayerHeartbeatResponseSystem.cs
zeroZshadow: It does, but since the goal is not to actually count heartbeats, but just ensure we keep the entity around while there is a connection, this was a simple way to ensure there are no false positives

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
samiwh: Neither did I. Have we verified that this works?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
```diff
@@ -12,83 +12,86 @@ namespace Improbable.Gdk.PlayerLifecycle
...
-                var requests = createPlayerData.CreatePlayerRequests[i].Requests;
-                var createEntitySender = createPlayerData.CreateEntitySender[i];
+                throw new InvalidOperationException(Errors.PlayerEntityTemplateNotFound);
+            }
 
-                foreach (var request in requests)
-                {
-                    if (PlayerLifecycleConfig.CreatePlayerEntityTemplate == null)
+            var requests = commandSystem.GetRequests<PlayerCreator.CreatePlayer.ReceivedRequest>();
+            foreach (var request in requests)
```
samiwh: This is about to be broken by moving to spans. Would be better to make this a regular for like, as for component updates.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
```diff
@@ -12,83 +12,86 @@ namespace Improbable.Gdk.PlayerLifecycle
...
+        private void HandlePlayerCreateEntityResponses()
+        {
+            var responses = commandSystem.GetResponses<WorldCommands.CreateEntity.ReceivedResponse>();
+            foreach (var response in responses)
             {
-                var entityCreationResponses = entityCreationResponseData.CreateEntityResponses[i];
-                var responder = entityCreationResponseData.CreatePlayerResponders[i];
-
-                foreach (var receivedResponse in entityCreationResponses.Responses)
+                if (!(response.Context is PlayerCreationRequestContext requestContext))
```
samiwh: This raises an interesting idea on whether we should somehow "type" world commands.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
samiwh: This is guaranteed not to happen. So to answer the todo: we can just do the status check.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerHeartbeat/SendPlayerHeartbeatRequestSystem.cs
zeroZshadow: Fixed the filtering in the last commit. It does work, just not with the ArchetypeChunk API.

samiwh: Filtering doesn't work with the chunk API anywhere though. The question is if it works on OnCreateManager. Looking at the ChangeFilter stuff it looks like that has to be done each time you get query the group.

zeroZshadow: The code was changed to use the proper API, and confirmed to work with the filter set in OnCreateManager.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
jamiebrynes7: @samiwh Should this be done before merging? Or fixed when the change happens?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
```diff
@@ -12,83 +11,85 @@ namespace Improbable.Gdk.PlayerLifecycle
...
+                        requestContext.createPlayerRequest.RequestId,
+                        new CreatePlayerResponseType
                         {
-                            CreatedEntityId = receivedResponse.EntityId.Value
-                        }));
+                            CreatedEntityId = response.EntityId.Value
+                        }
+                    );
+
+                    commandSystem.SendResponse(responseSuccess);
```
jamiebrynes7: Nit: This can be refactored to:

```
var requestId = requestContext.createPlayerRequest.RequestId;

var response = response.StatusCode == StatusCode.Success 
	? new PlayerCreator.CreatePlayer(requestId, new CreatePlayerResponseType { CreatedEntityId = response.EntityId.Value })
	: new PlayerCreator.CreatePlayer(requestId, $"Failed to create player: \"{response.Message}\"");

commandSystem.SendResponse(response);
```

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -14,75 +12,74 @@ public class SendCreatePlayerRequestSystem : ComponentSystem
...
+
+            playerSpawnGroup = GetComponentGroup(
+                ComponentType.ReadOnly<ShouldRequestPlayerTag>()
+            );
         }
 
         protected override void OnUpdate()
         {
-            for (int i = 0; i < newEntityData.Length; ++i)
+            if (!initializationGroup.IsEmptyIgnoreFilter)
```
jamiebrynes7: Didn't we find out this was somewhat expensive? I can't exactly recall

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
zeroZshadow: Thats way too big for a conditional operator

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
zeroZshadow: I'll double check, but I remember `Length` doing this under the hood anyways.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
jamiebrynes7: I disagree, but its fine either way

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
zeroZshadow: Did a check. `IsEmptyIgnoreFilter` will go through all matchingt archetypes until it hits an archetype that is NOT empty. This means that most of the times it will go through all archetypes.
The `GetEntityArray` function that we try to avoid by doing this check, also goes through all matching archetypes to count the amount of entities it matches. This makes the early out rather useless since we only grab 1 Array. Removing the check.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/HandleCreatePlayerRequestSystem.cs
zeroZshadow: Rebased on spans API

zeroZshadow:
@samiwh is the `ref readonly var X = ref A[i]` needed? What is the difference with just `var X = A[i]`?

PR #742, opened 2019-02-08
Improve the local launch error message as per UTY-1275

oblm:
#### Description
A more helpful local launch error message.
#### Tests
None
#### Documentation
This is (arguably) documentation.
#### Primary reviewers
@paulbalaji


workers/unity/Packages/com.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -259,7 +259,7 @@ public static void LaunchLocalDeployment()
                     return;
                 }
 
-                var message = $"Spatial local launch logfile: {latestLogFile.FullName}";
+                var message = $"For more information, check the Spatial local launch logfile: {latestLogFile.FullName}";
```
paulbalaji: nit: `Spatial` --> `spatial`

PR #744, opened 2019-02-11
Add pre exit hook for Buildkite

jamiebrynes7:
#### Description
Pre exit hook to kill all dangling processes in Buildkite.

Written in Powershell and uses `handle64` to find processes. I went with a blacklist approach as blindly killing processes touching the directory didn't end too well 😬 

Currently kills all `dotnet`, `Unity`, and `UnityPackageManager` processes

### DIFF
--- a/.buildkite/hooks/pre-exit
+++ b/.buildkite/hooks/pre-exit
@@ -4,5 +4,5 @@ set -euo pipefail
 
 # The premerge step runs on linux agents!
 if [[ ${BUILDKITE_AGENT_META_DATA_OS} == "windows" ]]; then
-    powershell ci/purge.ps1
+    powershell -NoProfile -NonInteractive ci/purge.ps1
 fi
--- a/.buildkite/hooks/pre-exit
+++ b/.buildkite/hooks/pre-exit
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+[[ -n "${DEBUG:-}" ]] && set -x
+
+# The premerge step runs on linux agents!
+if [[ ${BUILDKITE_AGENT_META_DATA_OS} == "windows" ]]; then
+    powershell ci/purge.ps1
+fi
--- a/ci/purge.ps1
+++ b/ci/purge.ps1
@@ -4,30 +4,30 @@
 # to find all processes of a particular name which have file locks on our build (project) directory.
 # We can then attempt to kill each of these processes if any exist.
 
-$SCRIPT_LOCATION = $MyInvocation.MyCommand.Path
-$SCRIPT_DIR = Split-Path $SCRIPT_LOCATION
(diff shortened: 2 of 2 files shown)


ci/purge.ps1
```diff
@@ -0,0 +1,29 @@
+# This script is used to kill all dangling procceses in the pre-exit hook.
+
+$SCRIPT_LOCATION = $MyInvocation.MyCommand.Path
+$SCRIPT_DIR = Split-Path $SCRIPT_LOCATION
+cd "${SCRIPT_DIR}/.."
+
+function Kill-Dangling-Processes {
+	param( [string]$ProcessName )
+
+	echo "Looking for ${ProcessName} processes to kill..."
```
jared-improbable: Why not use the `Get-Process` cmdlet? https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-process?view=powershell-6
That gives you back objects, and means you don't have to do regex and string splitting.

jamiebrynes7: The idea is to get processes handles that are locking files in the build directory - from a cursory look I don't think `Get-Process` supports that behaviour.

jared-improbable: Aha, that intent isn't clear from comments or obvious command flags

jamiebrynes7: Will add some comments about what this script does 👍

jamiebrynes7: PTAL 😄

ci/purge.ps1
```diff
@@ -0,0 +1,33 @@
...
+	${DIR}=$(pwd).Path
+	${OUT}=$(handle64 -accepteula -p "${ProcessName}.exe" ${DIR})
```
petemounce: I recommend avoiding using process-names as the only filter, since when Windows nodes have more than one buildkite-agent process running builds, this will interfere with other running builds.

(It's only an accident of fate that has limited BK to one process for Windows nodes - we haven't built the multi-agent per node support yet on Windows, but will, though you'll still have the option of how many to run).

jamiebrynes7: Discussed offline - this does currently filter by build directory and shouldn't wreak havoc on a multi-agent per node setup.

.buildkite/hooks/pre-exit
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+[[ -n "${DEBUG:-}" ]] && set -x
+
+# The premerge step runs on linux agents!
+if [[ ${BUILDKITE_AGENT_META_DATA_OS} == "windows" ]]; then
+    powershell ci/purge.ps1
```
petemounce: If you add
* `-NoProfile` the ps will execute slightly faster
* `-NonInteractive` there will be no possibility of interactive prompts blocking

Both are good practise.

PR #745, opened 2019-02-11
Add Schema Annotations Support

jamiebrynes7:
#### Description
Add annotations support for schema. 

- Regenerated example JSON
- Added `AnnotationRaw` class
- Renamed `UserType` -> `FullyQualifiedReference` as it was a more appropriate name.
- Applied a bugfix where blittable checks would be incorrect for components defined with the `data` schema syntax

PR is in commit order.

Will update exhaustive test schema in another PR to not pollute this one. See #746 

#### Tests
Ran the one and only unit test 💯 

#### Documentation
N/A

#### Primary reviewers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
@@ -105,9 +105,24 @@ namespace Improbable.Gdk.CodeGenerator
 
             foreach (var component in bundle.BundleContents.ComponentDefinitions)
             {
-                var fields = component.Data != null
-                    ? bundle.BundleContents.TypeDefinitions.First(type => type.Identifier.QualifiedName == component.Data.QualifiedName).Fields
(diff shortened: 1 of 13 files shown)


workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/src/Generation/Model/Details/DetailsStore.cs
```diff
@@ -105,7 +105,11 @@ private void PopulateBlittableMaps()
 
             foreach (var component in bundle.BundleContents.ComponentDefinitions)
             {
-                var blittableCheckResult = CheckBlittable(component.Fields);
+                var fields = component.Data != null
+                    ? bundle.BundleContents.TypeDefinitions.First(type => type.Identifier.QualifiedName == component.Data.QualifiedName).Fields
```
zeroZshadow: Can `First` have no result? If so, will this throw?

jamiebrynes7: If it has no result the bundle is invalid. 

We could probably handle this more gracefully however.

PR #749, opened 2019-02-12
Feature/command list consistency

samiwh:
Make lists use spans to be consistent with updates

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Collections/ReceivedMessageListExtensions.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Collections/ReceivedMessageListExtensions.cs
@@ -7,7 +7,7 @@ namespace Improbable.Gdk.Core
     {
         // binary search for the command for given request ID
         // invariant: lower <= target <= upper
-        public static int GetResponse<T>(this ReceivedMessageList<T> list, long requestId)
+        public static int GetResponseIndex<T>(this ReceivedMessageList<T> list, long requestId)
             where T : struct, IReceivedCommandResponse
         {
             long targetId = requestId;
@@ -21,17 +21,17 @@ namespace Improbable.Gdk.Core
 
                 long id = list[current].GetRequestId();
 
-                if (id == targetId)
-                {
-                    return current;
-                }
-                else if (id > targetId)
+                if (id > targetId)
                 {
                     upper = current - 1;
                 }
                 else if (id < targetId)
                 {
                     lower = current + 1;
+                }
+                else
+                {
+                    return current;
                 }
             }
 
--- a/workers/unity/Packages/com.improbable.gdk.core/Collections/ReceivedMessageListExtensions.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Collections/ReceivedMessageListExtensions.cs
@@ -1,13 +1,48 @@
+using Improbable.Gdk.Core.Commands;
+
 namespace Improbable.Gdk.Core
 {
     // todo consider exposing the private methods - the upper limit one can be used to find the last event/auth change
     internal static class ReceivedMessageListExtensions
     {
+        // binary search for the command for given request ID
+        // invariant: lower <= target <= upper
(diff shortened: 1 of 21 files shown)


workers/unity/Packages/com.improbable.gdk.core/Collections/ReceivedMessageListExtensions.cs
```diff
@@ -1,13 +1,48 @@
...
+            int lower = 0;
+            int upper = list.Count - 1;
+
+            while (upper >= lower)
+            {
+                var current = (lower + upper) / 2;
+
+                long id = list[current].GetRequestId();
+
+                if (id == targetId)
```
zeroZshadow: Instead of checking for equality, simply return as the last `else` statement. Saves a comparison.

```
if (id > targetId)
  upper = current - 1;
else if (id < targetId)
  lower = current + 1;
else
  return current;
```

workers/unity/Packages/com.improbable.gdk.core/Collections/ReceivedMessageListExtensions.cs
```diff
@@ -1,13 +1,48 @@
+using Improbable.Gdk.Core.Commands;
+
 namespace Improbable.Gdk.Core
 {
     // todo consider exposing the private methods - the upper limit one can be used to find the last event/auth change
     internal static class ReceivedMessageListExtensions
     {
+        // binary search for the command for given request ID
+        // invariant: lower <= target <= upper
+        public static int GetResponse<T>(this ReceivedMessageList<T> list, long requestId)
```
zeroZshadow: `GetResponseIndex` ?

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/CalllbackManagers/ResponseCallbackManager.cs
```diff
@@ -20,10 +20,10 @@ public ResponseCallbackManager(World world)
         public void InvokeCallbacks()
         {
             var responses = commandSystem.GetResponses<T>();
-            foreach (var response in responses)
+            for (int i = 0; i < responses.Count; ++i)
             {
-                callbacks.InvokeAll(response.GetRequestId(), response);
-                callbacks.RemoveAllCallbacksForIndex(response.GetRequestId());
+                callbacks.InvokeAll(responses[i].GetRequestId(), responses[i]);
```
zeroZshadow: Put `responses[i]` into a variable before using to avoid indexing multiple times.

workers/unity/Packages/com.improbable.gdk.core/Collections/ReceivedMessageListExtensions.cs
zeroZshadow: Would this be better as an `internal` function?

samiwh: I believe the class scope makes this internal. But it's also public function on an internal collection.

workers/unity/Packages/com.improbable.gdk.core/Collections/ReceivedMessageListExtensions.cs
samiwh: Yes. Copy past laziness.

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/CalllbackManagers/ResponseCallbackManager.cs
samiwh: Sure. Although I'd be very surprised if this isn't optimised out.

samiwh: The bigger performance problem here is the callbacks doing things by value probably. But that change can be made later.

PR #751, opened 2019-02-13
[WIP] Upgrade to Unity 2018.3.5f1

zeroZshadow:
#### Description
Upgraded gdk project to 2018.3.5f1

#### Tests
Ran locally and in the cloud

#### Documentation
Not yet, testing buildkite


jamiebrynes7:
Can you remove the code in https://github.com/spatialos/gdk-for-unity/blob/master/workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs where we `#if UNITY_STANDALONE_LINUX` and test turning protocol logging on in Linux? 

This was one of the workarounds for the libcpp linking issues and would be nice if we can remove this

zeroZshadow:
Will do the protocol logging fix in a separate PR

PR #752, opened 2019-02-13
Enable worker entity linking

samiwh:
#### Description
Re enabled the old worker gameobject flow

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
@@ -48,8 +48,7 @@ namespace Improbable.Gdk.GameObjectCreation
 
         private readonly GameObject workerGameObject;
 
-        public GameObjectInitializationSystem(IEntityGameObjectCreator gameObjectCreator,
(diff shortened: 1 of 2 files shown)


samiwh:
I don't really like this way of doing things, but it means it's non breaking for the moment.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -47,15 +46,24 @@ private struct RemovedEntitiesData
         private readonly IEntityGameObjectCreator gameObjectCreator;
         private EntityGameObjectLinker linker;
 
-        public GameObjectInitializationSystem(IEntityGameObjectCreator gameObjectCreator)
+        private readonly GameObject workerGameObject;
+
+        public GameObjectInitializationSystem(IEntityGameObjectCreator gameObjectCreator,
+            GameObject workerGameObject = null)
```
jamiebrynes7: Any reason this has a default value of `null` ?

samiwh: Not really while the helper exists. Can remove.

PR #755, opened 2019-02-14
Feature/clean up reflection warnings

samiwh:
Removed injection code where it wasn't needed and suppressed warnings were it was.


workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -70,37 +46,25 @@ protected override void OnDestroyManager()
...
 
-        // todo replace this with something that polls for entities added and removed
         protected override unsafe void OnUpdate()
         {
-            for (var i = 0; i < addedEntitiesData.Length; i++)
+            foreach (var entityId in entitySystem.GetEntitiesAdded())
             {
-                var entity = addedEntitiesData.Entities[i];
-                var spatialEntityId = addedEntitiesData.SpatialEntityIds[i].EntityId;
+                workerSystem.TryGetEntity(entityId, out var entity);
```
jamiebrynes7: Is this intended to discard the result of the `TryGet`? 

If so is it worth exposing a `GetEntity` method which doesn't have the out param and will actually throw if the entity doesn't exist?

jamiebrynes7: Such that the use cases of the methods break down into: 

- `GetEntity`: Entity will be there and its an error if its not.
- `TryGetEntity`: Entity may be there and I can handle the case where its not.

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -70,37 +46,25 @@ protected override void OnDestroyManager()
...
-                PostUpdateCommands.AddComponent(entity, new InitializedEntitySystemState
-                {
-                    EntityId = spatialEntityId
-                });
             }
 
-            EntityId* entitiesToRemove = stackalloc EntityId[removedEntitiesData.Length];
-            for (var i = 0; i < removedEntitiesData.Length; i++)
+            var removedEntities = entitySystem.GetEntitiesRemoved();
+            foreach (var entityId in removedEntities)
```
jamiebrynes7: I enjoy how much more concise this is now 💯

workers/unity/Packages/com.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
samiwh: Yes and maybe. The entity definitely exists at this point, however mid term I'd rather look at removing the dependency on ecs entities. GetEntity should be a method though.

samiwh: Basically I didn't touch that stuff too much as I suspect we may want to change it a lot.

PR #763, opened 2019-02-18
Make Types, Enums, and Snapshots Serializable.

### LINKED ISSUE #677: Feature Request: Serializable structs from generated code
#### Description

I want to be able to define an entity template in a scriptableobject, making it easy for myself or designers to tweak values of components and creating new entity templates without touching any code.

## Expected behaviour

Adding the following code to a script or a scriptable object should make it editable in the inspector. The same should be possible for any component you generate.

`public Position.Snapshot positionComponent;`

image

## Current behaviour
`public Position.Snapshot positionComponent;`

Does not display anything and can't be serialized.

## Possible solution
Adding [Serializable] to the component snapshot struct and all types that can be used and is serializable in a component.

image
image

zeroZshadow:
#### Description
Added the `[Serializable]` tag to all Types, Enums, and Snapshots.
Fixes #677 

#### Tests
Created a temporary `ScriptableObject` with public fields of the now serializable types, and confirmed that they indeed shop up in the editor.

Small caveat, the Nullable/Option types do NOT show up, as Unity does not render them by default.
This could possible be worked around using generated inspectors, but will not be trivial.

#### Documentation
NOT YET, post release to avoid conflics.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentDataGenerator.tt
+++ b/workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentDataGenerator.tt
@@ -194,6 +194,7 @@ var fieldDetails = fieldDetailsList[i]; #>
             }
         }
 
+        [System.Serializable]
         public struct Snapshot : ISpatialComponentSnapshot
         {
             public uint ComponentId => <#= componentDetails.ComponentId #>;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Added
+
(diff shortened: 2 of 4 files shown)


paulbalaji:
please link the UTY ticket for generating the nullable/option type inspectors

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityComponentDataGenerator.tt
```diff
@@ -154,6 +154,7 @@ var fieldDetails = fieldDetailsList[i]; #>
 <# } #>
         }
 
+        [Serializable]
```
jamiebrynes7: My consistency 😢

zeroZshadow:
Ignore the git madness, I based of the wrong branch

jamiebrynes7:
Please add changelog or else it will be forgotten.

PR #768, opened 2019-02-20
WIP Ignore DotSettings files

samiwh:
Ignore dotsettings files for rider 2018.3


.gitignore
```diff
@@ -25,6 +25,7 @@ workers/unity/*.sln
 *.svd
 *.DotSettings.user
 _ReSharper.Caches
+*.DotSettings
```
jared-improbable: We use these to keep code styles in sync

jamiebrynes7: Yeah this isn't an acceptable solution imo

PR #770, opened 2019-02-20
Separate View

jamiebrynes7:
#### Description
Implement the basic implementation of a separate view without performance optimizations. Verified in a debugger that the data and authority looks sensible.

#### Tests
Manual verification.

#### Documentation
N/A - Will need to be fleshed out and likely be hidden behind a `Worker` 

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.core/View/View.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        public bool HasEntity(EntityId entityId)
+        {
+            return entities.Contains(entityId);
+        }
+
+        // TODO: Make ref readonly when we have a dictionary type with ref indexing semantics.
+        public T GetComponent<T>(EntityId entityId) where T : struct, ISpatialComponentSnapshot
+        {
+            if (!HasEntity(entityId))
```
samiwh: Should make this has component

workers/unity/Packages/com.improbable.gdk.core/View/View.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                throw new ArgumentException($"The view does not have entity with Entity ID: {entityId.Id}");
+            }
+
+            return ((IViewComponentStorage<T>) typeToViewStorage[typeof(T)]).GetAuthority(entityId.Id);
+        }
+
+        public bool IsAuthoritative<T>(EntityId entityId) where T : struct, ISpatialComponentSnapshot
+        {
+            if (!HasComponent<T>(entityId))
```
samiwh: Could just make this GetAuthority<> == Authoritative.
Which also raises if we need this. It doesn't seem bad to have it in now but we might want to make it a hashset of things that have authority or something that can be returned.

workers/unity/Packages/com.improbable.gdk.core/View/View.cs
jamiebrynes7: Done

workers/unity/Packages/com.improbable.gdk.core/View/View.cs
jamiebrynes7: Hmmm those semantics are slightly different.

Currently:

- `GetAuthority<T>` throws if you don't have the entity/component in your view.
- `IsAuthoritative<T>` just returns false if you don't have the entity/component in your view.

We can unify these by allowing `GetAuthority` to return `NotAuthoritative` if the entity/component isn't in your view and doing as you suggest.

Thoughts?

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/ViewStorageGenerator.tt
```diff
@@ -0,0 +1,102 @@
...
+
+                var authorityChanges = ((IDiffAuthorityStorage) storage).GetAuthorityChanges();
+                for (var i = 0; i < authorityChanges.Count; i++)
+                {
+                    var authorityChange = authorityChanges[i];
+                    authorityStates[authorityChange.EntityId.Id] = authorityChange.Authority;
+                }
+            }
+
+            private void ApplyUpdate(long entityId, Update update)
```
samiwh: The Update should probably be passed by `ref`. As it's not readonly it shouldn't use `in`.

jamiebrynes7: Done

PR #771, opened 2019-02-20
Improved error visibility when code generation breaks.

zeroZshadow:
#### Description
Improved error visibility when code generation breaks.
Fixed possible deadlock in GenerateCode
Drive through clean-up of TransformAllTask and RedirectedProcess.

#### Tests
Broke a TT file and tested how the Editor reacts when generating code.
Tested how the alterations work when loading a clean project for the first time.


jamiebrynes7:
Why are you targetting `experimental/core-refactor` ?

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -100,11 +105,22 @@ private static void Generate()
...
                         .WithArgs(ConstructArgs(projectPath, schemaCompilerPath, workerJsonPath))
+                        .RedirectOutputOptions(OutputRedirectBehaviour.None)
+                        .AddErrorProcessing(Debug.LogError)
+                        .AddOutputProcessing(ProcessStdOut)
                         .Run();
 
                     if (exitCode != 0)
                     {
-                        Debug.LogError("Failed to generate code.");
+                        if (!Application.isBatchMode)
```
jared-improbable: Would it still be useful to output the above line to the log as well as the extra output from below, just to sum up?

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -27,10 +27,15 @@ internal static class GenerateCode
         private static readonly string StartupCodegenMarkerFile =
             Path.GetFullPath(Path.Combine("Temp", "ImprobableCodegen.marker"));
 
+        private static readonly Regex dotnetRegex = new Regex(
+            @"(?<file>[\w\\\.]+)\((?<line>\d+),(?<col>\d+)\): (?<type>\w+) (?<code>\w+): (?<message>[\s\S]+)",
```
jared-improbable: May be useful to write out a string literal example of the message format in a comment above

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/Improbable.TextTemplating/TransformAllTask.cs
```diff
@@ -28,42 +28,49 @@ public class TransformAllTask : Task
...
-                TemplateGenerator templateGenerator = new TemplateGenerator();
-                Encoding utF8 = Encoding.UTF8;
-                templateGenerator.Imports.AddRange((IEnumerable<string>) this.Imports);
-                string language;
-                string[] references;
-                templateGenerator.PreprocessTemplate(itemSpec, withoutExtension, this.ClassNameSpace, outputFile, utF8,
-                    out language, out references);
+                var itemSpec = inputFile.ItemSpec;
+                var withoutExtension = Path.GetFileNameWithoutExtension(itemSpec);
+                var path2 = $"{withoutExtension}.cs";
```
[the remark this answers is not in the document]
jamiebrynes7: Seems reasonable - we explicitly ignore the generated files in the `csproj` though so the benefit of formatting, etc. isn't as large.

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
jamiebrynes7: Yes please - anywhere there's a non-trivial Regex it needs really good documentation.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -177,13 +177,22 @@ public int Run()
 
                     foreach (var outputProcessor in outputProcessors)
                     {
-                        outputProcessor(outputString);
+                        try
+                        {
+                            outputProcessor?.Invoke(outputString);
```
jamiebrynes7: If the `Action` is null - we should catch that when you add the processor rather than doing `action?.Invoke()`

jamiebrynes7: Which will silently pass over

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -208,11 +218,25 @@ public int Run()
...
+                        }
                     }
                 };
 
                 process.EnableRaisingEvents = true;
+
+                if (!process.Start())
+                {
+                    throw new Exception(
+                        $"Failed to run {info.FileName} {info.Arguments}\nIs the .NET Core SDK installed?");
```
jamiebrynes7: Error message doesn't make sense. It may not be a `dotnet` process

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
zeroZshadow: The application itself already prints that it failed, the double log was rather weird

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
zeroZshadow: I can add one, but its rather long and will break our formatting rules, is that ok?

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
zeroZshadow: This was already there

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
zeroZshadow: I'll remove the null check and just let it fail, the exception will explain nicely.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
jamiebrynes7: Please fix it then 😉

jamiebrynes7:
Will rebase and merge this into `develop` once #788 lands

PR #774, opened 2019-02-21
Quick fix of a wrong query type

zeroZshadow:
Unblocks upgrade path to entities.24

#### Description
Quick fix of a wrong ComponentType used in a query. Doesn't change anything functionally, but stops Entities .24 packages from throwing as this has an added check.


jamiebrynes7:
Do we have a ticket for the entities upgrade? 

Also: 

> experimental/core-refactor 

😡

PR #777, opened 2019-02-21
Feature/modular sending

samiwh:
#### Description
Added a connection handler. Components and component commands now send and receive from this.

PR #778, opened 2019-02-22
Diff Builder and Reader Writer Tests

jamiebrynes7:
#### Description
This PR introduces the start of a mocking/testing framework (pending other changes) along with an example of how to use it with some basic reader/writer injection tests.

- DiffBuilder for creating diffs in a nice way (this likely will be wrapped into a `MockConnectionHandler : IConnectionHandler` when that lands @samiwh) 
  - Note that this is in the test-project as there _seems_ to be a Unity bug re asmdefs and test assemblies. When the above is done the Mock could live in the Core? 
- `DynamicConvert` class for generically converting snapshots -> updates. This can be expanded in the future to a full conversion suite.
- Reader Writer tests for injection using the above utilities.

Will rebase this PR once #777 is merged

#### Tests
Ran them tests.

#### Documentation
Some of this will want to be documented (particular for users testing scenarios). Not yet though.

#### Primary reviewers
@samiwh


jamiebrynes7:
Superseded by #789

PR #782, opened 2019-02-25
Feature/expose metrics logging

samiwh:
Metrics and logs and now be sent and received.


workers/unity/Packages/com.improbable.gdk.core/Worker/ViewDiff.cs
```diff
@@ -238,12 +243,39 @@ public void AddEntityQueryResponse(WorldCommands.EntityQuery.ReceivedResponse re
             worldCommandsReceivedStorage.AddResponse(response);
         }
 
+        public void AddLogMessage(string message, LogLevel level)
+        {
+            logsReceived.Add(new LogMessageReceived(message, level));
```
jamiebrynes7: This does mean we can't use the same diff object for sending & receiving - although that's probably okay

samiwh: We already can't because of commands for now :(.

PR #786, opened 2019-02-25
Use new shared-ci script to build workers

paulbalaji:
#### Description
- replaced `build.sh` with `.shared-ci/scripts/build.sh`
- using `uploadAssemblies()` function in `.shared-ci/scripts/pinned-tools.sh` instead of manually setting `$ASSEMBLY_NAME` and calling toolbelt
- `pre-exit` --> `post-command` hook

#### Tests
ran through buildkite against `feature/unify-build-scripts` shared-ci branch

#### Documentation
n/a

#### Primary reviewers


ci/bootstrap.sh
```diff
@@ -12,4 +12,4 @@ if [ -d "${SHARED_CI_DIR}" ]; then
     rm -rf "${SHARED_CI_DIR}"
 fi
 
-git clone "${CLONE_URL}" "${SHARED_CI_DIR}"
+git clone "${CLONE_URL}" "${SHARED_CI_DIR}" --branch feature/unify-build-scripts
```
paulbalaji: this change will be removed before the final merge

ci/launch.sh
```diff
@@ -17,9 +15,7 @@ else
     ci/build-test.sh
 fi
 
-setAssemblyName "${PREFIX}"
-
-spatial cloud upload "${ASSEMBLY_NAME}" --log_level=debug --force --enable_pre_upload_check=false
+uploadAssembly "${ASSEMBLY_PREFIX}" "${PROJECT_NAME}"
```
jessicafalk: will that work? I don't see you specifying either of the two variables anywhere

paulbalaji: these are both defined in the buildkite step that calls this script

.buildkite/hooks/post-command
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
samiwh: usual style is to separate these out.

samiwh: Do we not want -x

samiwh: ah just seen below. Do the two paths both get run?

paulbalaji: I kept the style from the pre-exit hook that we had before

PR #787, opened 2019-02-25
Separate the ecs view from other systems

samiwh:
#### Description
The ecs view is now updated from its own system.
The component update system now uses the view rather than the update mangers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentUpdateSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentUpdateSystem.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using Improbable.Worker.CInterop;
 using Unity.Entities;
@@ -7,16 +6,9 @@ namespace Improbable.Gdk.Core
 {
     [DisableAutoCreation]
     [AlwaysUpdateSystem]
-    [UpdateInGroup(typeof(SpatialOSSendGroup.InternalSpatialOSSendGroup))]
     public class ComponentUpdateSystem : ComponentSystem
     {
-        private readonly SerializedMessagesToSend serializedMessagesToSend = new SerializedMessagesToSend();
         private WorkerSystem worker;
-
-        private readonly List<IComponentManager> managers = new List<IComponentManager>();
-
-        private readonly Dictionary<uint, IComponentManager> componentIdToManager =
-            new Dictionary<uint, IComponentManager>();
 
         public void SendUpdate<T>(T update, EntityId entityId) where T : ISpatialComponentUpdate
         {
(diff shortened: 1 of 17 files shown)


workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentUpdateSystem.cs
```diff
@@ -81,45 +73,17 @@ public List<EntityId> GetComponentsRemoved(uint componentId)
...
-        {
-            if (!componentIdToManager.TryGetValue(componentId, out var manager))
-            {
-                throw new ArgumentException("Component ID not recognized");
-            }
-
-            return manager.GetInitialComponents();
-        }
-
         public bool HasComponent(uint componentId, EntityId entityId)
```
jamiebrynes7: Do these APIs make sense here anymore?

jamiebrynes7: This one and `GetAuthority`

workers/unity/Packages/com.improbable.gdk.core/Systems/EcsViewSystem.cs
```diff
@@ -0,0 +1,135 @@
+using System;
+using System.Collections.Generic;
+using Unity.Entities;
+using UnityEngine.Profiling;
+
+namespace Improbable.Gdk.Core
+{
+    [DisableAutoCreation]
+    [AlwaysUpdateSystem]
+    public class EcsViewSystem : ComponentSystem
```
jamiebrynes7: Entity Component System View System

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
```diff
@@ -69,65 +58,11 @@ protected override void OnUpdate()
...
-
-            Profiler.BeginSample("OnRemoveEntity");
-            EntityManager.DestroyEntity(worker.EntityIdToEntity[entityId]);
-            worker.EntityIdToEntity.Remove(entityId);
-            Profiler.EndSample();
-        }
 
         private void OnDisconnect(string reason)
         {
             EntityManager.AddSharedComponentData(worker.WorkerEntity,
```
jamiebrynes7: It feels like this belongs in the `EcsViewSystem`

samiwh: Oops missed that. It just seemed reasonable to me last night the the worker would do something on disconnect. We should also make the disconnect event work outside the ecs.

workers/unity/Packages/com.improbable.gdk.core/Systems/ComponentUpdateSystem.cs
samiwh: Discussed offline. Yes for the moment but it's an ongoing task to decide if this will remain the case.

PR #791, opened 2019-02-26
Fixed an error when linking a gameobject to an invalid entity

samiwh:
Fixed being able to link against a worker entity

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Improbable.Gdk.Core;
 using Unity.Entities;
 using UnityEngine;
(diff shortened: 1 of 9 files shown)


workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
```diff
@@ -51,19 +52,17 @@ public EntityGameObjectLinker(World world)
...
-            if (!workerSystem.TryGetEntity(entityId, out var entity))
+            var linkedComponent = gameObject.GetComponent<LinkedEntityComponent>();
+            if (linkedComponent == null)
             {
-                throw new ArgumentException("Entity not in view");
+                linkedComponent = gameObject.AddComponent<LinkedEntityComponent>();
             }
-
-            if (gameObjectToComponentsAdded.TryGetValue(gameObject, out var added))
+            else if (linkedComponent.IsValid)
```
jamiebrynes7: What does it mean if the `linkedComponent` is not valid?

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
```diff
@@ -127,6 +99,11 @@ public EntityGameObjectLinker(World world)
             }
 
             gameObjectToInjectors.Add(gameObject, injectors);
+
+            if (entityId.IsValid())
```
jamiebrynes7: Why is this conditional on the EntityId being valid? 

Is there a reason to prevent GameObject components being pushed onto the worker entity?

jamiebrynes7: Ah I see - its to stop the exception from being thrown.

jamiebrynes7: Is this a regression on the core v1 implementation?

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
samiwh: Means it isn't linked to a gameobject.

samiwh: It could do with a better name.

PR #798, opened 2019-02-28
fps upgrade fixes

samiwh:
Changes found while upgrading the fps starter project

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
@@ -72,7 +72,6 @@ namespace Improbable.Gdk.Subscriptions
             linkedGameObjects.Add(gameObject);
 
             linkedComponent.IsValid = true;
-            linkedComponent.Linker = this;
             linkedComponent.EntityId = entityId;
             linkedComponent.World = world;
             linkedComponent.Worker = workerSystem;
@@ -106,6 +105,7 @@ namespace Improbable.Gdk.Subscriptions
             }
         }
 
+        // todo remove components from entity
         public void UnlinkGameObjectFromEntity(EntityId entityId, GameObject gameObject)
         {
             if (!entityIdToGameObjects.TryGetValue(entityId, out var gameObjectSet))
@@ -129,10 +129,9 @@ namespace Improbable.Gdk.Subscriptions
             }
 
             var linkComponent = gameObject.GetComponent<LinkedEntityComponent>();
-            // todo check if this can happen - right now I think it can on application quit
             if (linkComponent != null)
             {
-                linkComponent.Invalidate();
+                linkComponent.IsValid = false;
             }
 
             foreach (var injector in injectors)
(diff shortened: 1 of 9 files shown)


workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
```diff
@@ -72,10 +72,10 @@ public EntityGameObjectLinker(World world)
             linkedGameObjects.Add(gameObject);
 
             linkedComponent.IsValid = true;
-            linkedComponent.Linker = this;
             linkedComponent.EntityId = entityId;
             linkedComponent.World = world;
             linkedComponent.Worker = workerSystem;
+            linkedComponent.Linker = this;
```
jamiebrynes7: Why? Does the ordering here actually matter?

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
```diff
@@ -129,10 +134,13 @@ public void UnlinkGameObjectFromEntity(EntityId entityId, GameObject gameObject)
...
             var linkComponent = gameObject.GetComponent<LinkedEntityComponent>();
-            // todo check if this can happen - right now I think it can on application quit
             if (linkComponent != null)
             {
-                linkComponent.Invalidate();
+                linkComponent.IsValid = false;
+                linkComponent.EntityId = new EntityId(0);
+                linkComponent.World = null;
+                linkComponent.Worker = null;
+                linkComponent.Linker = null;
```
jamiebrynes7: This and the linking action feels like an:

`Init()` and `Deinit()` pair of methods on the linkComponent.

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/RequiredSubscriptionsInjector.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using Improbable.Gdk.Core;
+using UnityEngine;
```
jamiebrynes7: Is this actually needed?

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
```diff
@@ -30,11 +30,6 @@ protected override void OnCreateManager()
 
         protected override void OnUpdate()
         {
-            if (!worker.ConnectionHandler.IsConnected())
-            {
-                return;
-            }
```
jamiebrynes7: Why is this moved into the implementation of the `IConnectionHandler`?

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
samiwh: Side effect

samiwh: The ordering does not matter

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
samiwh: To prevent the case where you read the same diff again and again. When it says to disconnect it could also clear everything else. However you could have an implementation that clears stuff up in later frames.

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/EntityGameObjectLinker.cs
samiwh: Discussed offline. Agreed but will merge this in for now.

PR #799, opened 2019-02-28
Add Required Build Flag

jamiebrynes7:
#### Description
Small addition to the build system which allows a user to mark a build target as "Required". The context for this change is that in #706 we allowed builds in the Editor to be skipped if the build support was not installed. The rationale for this change was that users should not have to install Android or iOS support if they do not want to and previously clicking `Build > All Workers` in the menu would fail if they were not installed.

This raised a different problem where if a user did not install Linux Build Support they wouldn't be able to deploy as the Editor build would allow the build to pass. This PR rectifies this by allowing a user to mark a build as `Required` such that if the build support _is not_ installed, it will fail instead of skipping.

This allows us the flexibility to say that you can skip iOS and Android targets if they aren't installed, but you **need** certain build targets to succeed (Linux + UnityGameLogic) for example. 

#### Tests
Tried checking & unchecking the tick box and building via the menu. Behaviour was as expected.

#### Documentation
Probably should be - the build system needs good docs all over.

Yes

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -183,16 +183,28 @@ public static void Clean()
...
+                Debug.LogWarning($"Skipping {buildTarget} because build support is not installed in the Unity Editor and the build target is not marked as 'Required'.");
+
                 return;
             }
 
             Debug.Log(
                 $"Building \"{buildTarget}\" for worker platform: \"{workerType}\", environment: \"{targetEnvironment}\"");
 
-            var spatialOSBuildConfiguration = BuildConfig.GetInstance();
+
```
jessicafalk: nit: new line

workers/unity/Packages/com.improbable.gdk.buildsystem/Configuration/BuildConfig.cs
```diff
@@ -54,8 +54,8 @@ private void ResetToDefault()
                             WorkerBuildData.GetCurrentBuildTargetConfig()),
                     CloudBuildConfig = new BuildEnvironmentConfig(
                         WorkerBuildData.AllBuildTargets,
-                        new BuildTargetConfig(BuildTarget.StandaloneOSX, BuildOptions.Development, true),
-                        new BuildTargetConfig(BuildTarget.StandaloneWindows64, BuildOptions.Development, true))
+                        new BuildTargetConfig(BuildTarget.StandaloneOSX, BuildOptions.Development, true, false),
```
jared-improbable: We're entering the dreaded "bool bool bool" API territory here. At the very least let's use named parameters.

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -183,16 +183,28 @@ public static void Clean()
...
             if (!WorkerBuildData.BuildTargetsThatCanBeBuilt[buildTarget])
             {
-                Debug.LogWarning($"Skipping {buildTarget} because support is not installed in the Unity Editor.");
+                var config = spatialOSBuildConfiguration.GetEnvironmentConfigForWorker(workerType, targetEnvironment);
+                var target = config.BuildTargets.First(targetConfig => targetConfig.Target == buildTarget);
+
+                if (target.Required)
+                {
+                    throw new BuildFailedException(
+                        $"Build failed for {workerType}. Cannot build for {buildTarget} because build support is not installed in the Unity Editor.");
```
jared-improbable: "Cannot build for *required* {buildTarget}" - gives a hint about the criteria.

workers/unity/Packages/com.improbable.gdk.buildsystem/Configuration/BuildConfig.cs
jamiebrynes7: This could easily be a tri-state enum: 

```
Disabled
Enabled
Required
```

jamiebrynes7: Discussed offline with multiple people - will leave as bools, but will add named parameters

PR #806, opened 2019-03-03
Add Application.isPlaying check to WorkerConnector

jessicafalk:
#### Description
Adding an `Application.isPlaying`check before calling `HandleWorkerConnectionEstablished` as this can result in spawning countless error messages, if the Editor play mode is exited at the wrong time. 
this occured in the `SimualtedPlayerWorkerConnector` which tried to access its transform component after already being destroyed. 
A second PR in the FPS starter project will be opened to add additional checks as that behaviour occured in multiple places. (though all related to the `HandleWorkerConnectionEstablished`)

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
@@ -117,10 +117,12 @@ namespace Improbable.Gdk.Core
 
                 Worker.OnDisconnect += OnDisconnected;
 
-                HandleWorkerConnectionEstablished();
-
-                World.Active = World.Active ?? Worker.World;
-                ScriptBehaviourUpdateOrder.UpdatePlayerLoop(World.AllWorlds.ToArray());
+                if (Application.isPlaying)
+                {
+                    HandleWorkerConnectionEstablished();
+                    World.Active = World.Active ?? Worker.World;
+                    ScriptBehaviourUpdateOrder.UpdatePlayerLoop(World.AllWorlds.ToArray());
+                }
             }
             catch (Exception e)
             {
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -117,10 +117,12 @@ await Worker.CreateWorkerAsync(GetAlphaLocatorConfig(workerType), connectionPara
 
                 Worker.OnDisconnect += OnDisconnected;
 
-                HandleWorkerConnectionEstablished();
-
-                World.Active = World.Active ?? Worker.World;
-                ScriptBehaviourUpdateOrder.UpdatePlayerLoop(World.AllWorlds.ToArray());
+                if (Application.isPlaying)
```
jamiebrynes7: If it isn't playing.. is there anything we should clean up?

samiwh: This could also be a check for not playing that returns, which is what happens elsewhere in the same file.

samiwh: However as the connection has been made by this point we will need to dispose of that to free native memory.

jessicafalk: will add that. 

also going to make `OnDestroy` protected as I just realized that we are not disposing the sim player coordinator correctly anymore in the other PR I made

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -54,12 +54,12 @@ public abstract class WorkerConnector : MonoBehaviour, IDisposable
...
 
         // Important run in this step as otherwise it can interfere with the the domain unloading logic.
-        private void OnApplicationQuit()
+        protected void OnApplicationQuit()
         {
             Dispose();
         }
 
-        private void OnDestroy()
+        protected void OnDestroy()
```
jessicafalk: made those two methods protected as you might want to override them in user code

PR #807, opened 2019-03-04
Get commands received by index

samiwh:
#### Description
Command requests can now be queried by entity ID and responses by request ID. This is to match the component update API that allows searching updates by entity ID.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
@@ -96,8 +96,7 @@ namespace Improbable.Gdk.Core
             var responseIndex = createEntityResponses.GetResponseIndex(requestId);
             if (responseIndex < 0)
             {
-                return new ReceivedMessagesSpan<WorldCommands.CreateEntity.ReceivedResponse>(createEntityResponses, 0,
-                    0);
+                return ReceivedMessagesSpan<WorldCommands.CreateEntity.ReceivedResponse>.Empty();
             }
 
             return new ReceivedMessagesSpan<WorldCommands.CreateEntity.ReceivedResponse>(createEntityResponses,
@@ -116,8 +115,7 @@ namespace Improbable.Gdk.Core
             var responseIndex = deleteEntityResponses.GetResponseIndex(requestId);
             if (responseIndex < 0)
             {
-                return new ReceivedMessagesSpan<WorldCommands.DeleteEntity.ReceivedResponse>(deleteEntityResponses, 0,
-                    0);
+                return ReceivedMessagesSpan<WorldCommands.DeleteEntity.ReceivedResponse>.Empty();
             }
 
             return new ReceivedMessagesSpan<WorldCommands.DeleteEntity.ReceivedResponse>(deleteEntityResponses,
@@ -136,8 +134,7 @@ namespace Improbable.Gdk.Core
             var responseIndex = reserveEntityIdsResponses.GetResponseIndex(requestId);
             if (responseIndex < 0)
             {
-                return new ReceivedMessagesSpan<WorldCommands.ReserveEntityIds.ReceivedResponse>(
-                    reserveEntityIdsResponses, 0, 0);
+                return ReceivedMessagesSpan<WorldCommands.ReserveEntityIds.ReceivedResponse>.Empty();
             }
 
             return new ReceivedMessagesSpan<WorldCommands.ReserveEntityIds.ReceivedResponse>(reserveEntityIdsResponses,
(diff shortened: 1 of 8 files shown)


workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
```diff
@@ -71,5 +83,114 @@ public void AddResponse(WorldCommands.EntityQuery.ReceivedResponse response)
...
+            if (!createEntitySorted)
+            {
+                createEntityResponses.Sort(comparer);
+                createEntitySorted = true;
+            }
+
+            var responseIndex = createEntityResponses.GetResponseIndex(requestId);
+            if (responseIndex < 0)
+            {
+                return new ReceivedMessagesSpan<WorldCommands.CreateEntity.ReceivedResponse>(createEntityResponses, 0,
```
zeroZshadow: Could we have an `ReceivedMessagesSpan<T>.Empty();`? Similar to the `Array<T>.Empty();`

workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
```diff
@@ -71,5 +83,114 @@ public void AddResponse(WorldCommands.EntityQuery.ReceivedResponse response)
...
+            IDiffCommandResponseStorage<WorldCommands.CreateEntity.ReceivedResponse>.GetResponse(long requestId)
+        {
+            if (!createEntitySorted)
+            {
+                createEntityResponses.Sort(comparer);
+                createEntitySorted = true;
+            }
+
+            var responseIndex = createEntityResponses.GetResponseIndex(requestId);
+            if (responseIndex < 0)
```
zeroZshadow: Is this valid behaviour to run into? Should it display any log?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
```diff
@@ -20,12 +21,23 @@ public class WorldCommandsReceivedStorage : ICommandDiffStorage
...
+        private bool reserveEntityIdsSorted;
+        private bool entityQueriesSorted;
+
         public void Clear()
         {
             createEntityResponses.Clear();
             deleteEntityResponses.Clear();
             reserveEntityIdsResponses.Clear();
             entityQueryResponses.Clear();
+            createEntitySorted = false;
```
zeroZshadow: Can the `MessageList` not keep track of its sorted state?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
```diff
@@ -71,5 +83,114 @@ public void AddResponse(WorldCommands.EntityQuery.ReceivedResponse response)
         {
             return new ReceivedMessagesSpan<WorldCommands.EntityQuery.ReceivedResponse>(entityQueryResponses);
         }
+
+        ReceivedMessagesSpan<WorldCommands.CreateEntity.ReceivedResponse>
+            IDiffCommandResponseStorage<WorldCommands.CreateEntity.ReceivedResponse>.GetResponse(long requestId)
+        {
+            if (!createEntitySorted)
+            {
+                createEntityResponses.Sort(comparer);
```
zeroZshadow: Is Sorting required before a `GetResponseIndex`? If so, shouldn't `GetResponseIndex` take care of that?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
samiwh: Done

workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
samiwh: Yes this just means you are polling for a response that hasn't arrived yet. There is a question as to whether or not we want this API or a different one.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
samiwh: Yes it can. That is something that I'd like to add in. This was the minimal change to keep it consistent. What might happen is that we make these apply things in sorted order and that can turned off if you don't need it.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorldCommandsReceivedStorage.cs
samiwh: Yes and it can but it checking if something is sorted is an O(n) operation so I'd rather not do it each time.

samiwh: This basically just does a binary search.

PR #808, opened 2019-03-04
Update the view when sending an update. Disable loopback

samiwh:
#### Description
Sending an update will update the view. This does not make the more intrusive changes, like separating the writer send api from the ecs one. Also does not deal with the potential `in` copy issue.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/View/View.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/View/View.cs
@@ -24,6 +24,7 @@ namespace Improbable.Gdk.Core
                         var instance = (IViewStorage) Activator.CreateInstance(type);
 
                         typeToViewStorage.Add(instance.GetSnapshotType(), instance);
+                        typeToViewStorage.Add(instance.GetUpdateType(), instance);
                         componentIdToViewStorage.Add(instance.GetComponentId(), instance);
 
                         viewStorages.Add(instance);
@@ -32,7 +33,7 @@ namespace Improbable.Gdk.Core
             }
         }
 
-        public void ApplyDiff(ViewDiff diff)
+        internal void ApplyDiff(ViewDiff diff)
         {
             var entitiesAdded = diff.GetEntitiesAdded();
             foreach (var entity in entitiesAdded)
@@ -51,6 +52,12 @@ namespace Improbable.Gdk.Core
                 // Resolve this with an actual diff!
                 storage.ApplyDiff(diff);
             }
+        }
+
+        internal void UpdateComponent<T>(EntityId entityId, in T update) where T : struct, ISpatialComponentUpdate
+        {
+            var storage = (IViewComponentUpdater<T>) typeToViewStorage[typeof(T)];
+            storage.ApplyUpdate(entityId.Id, in update);
         }
 
         public HashSet<EntityId> GetEntityIds()
--- a/workers/unity/Packages/com.improbable.gdk.core/View/View.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/View/View.cs
@@ -139,7 +139,6 @@ namespace Improbable.Gdk.Core
 
             foreach (var storage in viewStorages)
             {
-                // Resolve this with an actual diff!
                 storage.ApplyDiff(diff);
             }
         }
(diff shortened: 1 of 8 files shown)


workers/unity/Packages/com.improbable.gdk.core/Worker/SerializedMessagesToSend.cs
```diff
@@ -112,7 +115,7 @@ public void SendAll(Connection connection, CommandMetaData commandMetaData)
             for (int i = 0; i < updates.Count; ++i)
             {
                 ref readonly var update = ref updates[i];
-                connection.SendComponentUpdate(update.EntityId, update.Update);
+                connection.SendComponentUpdate(update.EntityId, update.Update, updateParams);
```
jamiebrynes7: Isn't this a semantic breaking change?

workers/unity/Packages/com.improbable.gdk.core/View/View.cs
```diff
@@ -136,5 +122,26 @@ public bool IsAuthoritative(EntityId entityId, uint componentId)
...
+
+            var entitiesRemoved = diff.GetEntitiesRemoved();
+            foreach (var entity in entitiesRemoved)
+            {
+                entities.Remove(entity);
+            }
+
+            foreach (var storage in viewStorages)
+            {
+                // Resolve this with an actual diff!
```
zeroZshadow: Nit: No need for the !

workers/unity/Packages/com.improbable.gdk.core/Worker/SerializedMessagesToSend.cs
```diff
@@ -37,8 +37,11 @@ public class SerializedMessagesToSend
         private readonly List<IComponentSerializer> componentSerializers = new List<IComponentSerializer>();
         private readonly List<ICommandSerializer> commandSerializers = new List<ICommandSerializer>();
 
+        private readonly UpdateParameters updateParams = new UpdateParameters();
+
         public SerializedMessagesToSend()
         {
+            updateParams.Loopback = ComponentUpdateLoopback.None;
```
zeroZshadow: Can this not be set in the object's constructor?

workers/unity/Packages/com.improbable.gdk.core/View/View.cs
samiwh: I just copied this over.

workers/unity/Packages/com.improbable.gdk.core/Worker/SerializedMessagesToSend.cs
samiwh: It doesn't have a non default constructor.

workers/unity/Packages/com.improbable.gdk.core/View/View.cs
samiwh: @jamiebrynes7 what is this comment for

workers/unity/Packages/com.improbable.gdk.core/Worker/SerializedMessagesToSend.cs
samiwh: Actually that is only half true. We don't reapply the update to the component, but we do add it to the updates received list. So these might need to stay :(. We currently have nothing internal depending on this behaviour but a user might. Basically monobehaviours should have an immediate update or we should keep a list of updates sent as well as received.

workers/unity/Packages/com.improbable.gdk.core/View/View.cs
jamiebrynes7: Leftover comment that I forgot to remove. Can remove

workers/unity/Packages/com.improbable.gdk.core/Worker/SerializedMessagesToSend.cs
samiwh: Hmm will keep the loopback for the moment although we really should get rid of it.

workers/unity/Packages/com.improbable.gdk.core/Worker/SerializedMessagesToSend.cs
zeroZshadow: If its a struct you can just use named initialisation.

samiwh: It's a class. It has to be set after it's definition.

PR #809, opened 2019-03-04
Pass arbitrary data to CreatePlayerRequestType

paulbalaji:
TL;DR
- Adds ability for arbitrary data to be passed into `CreatePlayerRequestType`
- Adds ability to disable automatic player creation as soon as worker connects

#### Description
- remove `ShouldRequestPlayerTag`
- `SendCreatePlayerRequestSystem`
	- add `RequestPlayerCreation` function
- `PlayerLifecycleConfig`
	- added `AutoRequestPlayerCreation` boolean to determine whether to automatically add send player creation request as soon as worker is connected
	- `GetPlayerEntityTemplateDelegate` has an extra field of type `Option<byte[]>` for the serialized arguments sent in the player creation module
- `PlayerLifecycleHelper`
	- added `SerializeArguments` and `DeserializeArguments<T>` methods to let users serialize/deserialize arbitrary params using `BinaryFormatter` (sorry Jamie)

#### Tests
tested in editor

#### Documentation
n/a

yes

#### Primary reviewers


samiwh:
This was a suggested change but I don't think it was agreed upon as better than just adding a string.

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -3,13 +3,21 @@
...
     public static class PlayerTemplate
     {
-        public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improbable.Vector3f position)
+        public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improbable.Vector3f position,
+            Option<byte[]> playerCreationArguments)
         {
+            if (playerCreationArguments.HasValue)
+            {
+                var test = PlayerLifecycleHelper.DeserializeParams<PlayerCreationParams>(playerCreationArguments.Value);
+                UnityEngine.Debug.Log(test.PlayerName);
```
zeroZshadow: No debug logs

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -3,13 +3,21 @@
...
 {
     public static class PlayerTemplate
     {
-        public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improbable.Vector3f position)
+        public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improbable.Vector3f position,
+            Option<byte[]> playerCreationArguments)
         {
+            if (playerCreationArguments.HasValue)
+            {
+                var test = PlayerLifecycleHelper.DeserializeParams<PlayerCreationParams>(playerCreationArguments.Value);
```
zeroZshadow: Can this fail?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Components/ShouldRequestPlayerTag.cs
```diff
@@ -2,7 +2,9 @@
 
 namespace Improbable.Gdk.PlayerLifecycle
 {
-    public struct ShouldRequestPlayerTag : IComponentData
+    public struct ShouldRequestPlayerTag : ISharedComponentData
```
zeroZshadow: Now no longer a tag, as it has data

zeroZshadow: Why is this a shared component now?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -21,6 +23,43 @@ public static class PlayerLifecycleHelper
...
+                using (var memoryStream = new MemoryStream())
+                {
+                    var binaryFormatter = new BinaryFormatter();
+                    binaryFormatter.Serialize(memoryStream, playerCreationArguments);
+                    return memoryStream.ToArray();
+                }
+            }
+            catch
+            {
+                UnityEngine.Debug.LogWarning("Unable to serialize player creation arguments.");
```
zeroZshadow: This hides the exception

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -21,6 +23,43 @@ public static class PlayerLifecycleHelper
             template.AddComponent(owningComponent, serverAccess);
         }
 
+        public static byte[] SerializeParams<T>(T playerCreationArguments)
```
zeroZshadow: Where is `T` used?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -1,11 +1,13 @@
...
+using System.IO;
+using System.Runtime.Serialization.Formatters.Binary;
 using Improbable.Gdk.Core;
 using Improbable.PlayerLifecycle;
 using Unity.Entities;
 
 namespace Improbable.Gdk.PlayerLifecycle
 {
-    public static class PlayerLifecycleHelper
+    public class PlayerLifecycleHelper
```
zeroZshadow: Why is this no longer static?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Components/ShouldRequestPlayerTag.cs
paulbalaji: IComponentData must be blittable, so I can't have a byte[] field (or a string for that matter)

ran into the same issue as defined here:
https://forum.unity.com/threads/trying-to-store-a-string-in-a-blittable-format.544283/

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
paulbalaji: yeah I fixed this locally but waiting to push until I incorporate PR feedback

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
paulbalaji: ah was trying to debug something, forgot to change it back

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
paulbalaji: hmm it will silently fail but return the default `PlayerCreationParams` as if nothing bad happened

I think it's worth changing serialize/deserialize to return a bool to indicate success, passing the outback back with an `out` param

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -44,26 +51,17 @@ protected override void OnUpdate()
             {
                 if (PlayerLifecycleConfig.AutoRequestPlayerCreation)
```
zeroZshadow: Make this if the outer if

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -31,10 +35,13 @@ protected override void OnCreateManager()
                 ComponentType.ReadOnly<WorkerEntityTag>(),
                 ComponentType.ReadOnly<OnConnected>()
             );
+        }
 
-            playerSpawnGroup = GetComponentGroup(
-                ComponentType.ReadOnly<CreatePlayerRequestTrigger>()
-            );
+        public void TriggerPlayerCreation(Vector3 spawnPosition = default, byte[] serializedArguments = null)
```
zeroZshadow: `Trigger` -> `Queue` ?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -21,6 +23,47 @@ public static class PlayerLifecycleHelper
...
+                    var binaryFormatter = new BinaryFormatter();
+                    memoryStream.Write(serializedArguments, 0, serializedArguments.Length);
+                    memoryStream.Seek(0, SeekOrigin.Begin);
+                    deserializedArguments = (T) binaryFormatter.Deserialize(memoryStream);
+                    return true;
+                }
+            }
+            catch (Exception e)
+            {
+                UnityEngine.Debug.LogError($"Unable to deserialize player creation arguments. {e.Message}");
```
jessicafalk: this should probably throw an exception as there is no way to recover from this

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -3,13 +3,21 @@
...
 using Improbable.Gdk.TransformSynchronization;
+using Improbable.PlayerLifecycle;
 
 namespace Playground
 {
     public static class PlayerTemplate
     {
-        public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improbable.Vector3f position)
+        public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improbable.Vector3f position,
+            Option<byte[]> playerCreationArguments)
```
samiwh: Why option? `byte[]` can be null

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -25,7 +33,8 @@ public static EntityTemplate CreatePlayerEntityTemplate(string workerId, Improba
             PlayerLifecycleHelper.AddPlayerLifecycleComponents(template, workerId, clientAttribute,
                 WorkerUtils.UnityGameLogic);
 
-            template.SetReadAccess(WorkerUtils.UnityClient, WorkerUtils.UnityGameLogic, WorkerUtils.AndroidClient, WorkerUtils.iOSClient);
+            template.SetReadAccess(WorkerUtils.UnityClient, WorkerUtils.UnityGameLogic, WorkerUtils.AndroidClient,
```
samiwh: I didn't know we had this line :S. This means this only works if you have the same setup as playground.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -21,6 +23,47 @@ public static class PlayerLifecycleHelper
             template.AddComponent(owningComponent, serverAccess);
         }
 
+        public static bool SerializeArguments(Object playerCreationArguments, out byte[] serializedArguments)
```
samiwh: We are we providing a helper for this? What if the user wants to control their own serialisation? Also note that our types are not marked as serializable as we expect that to happen via schema objects.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Schema/player_creator.schema
```diff
@@ -1,9 +1,15 @@
...
 import "improbable/vector3.schema";
 
+type PlayerCreationParams
+{
+    string player_name = 1;
+}
+
 type CreatePlayerRequestType
 {
     improbable.Vector3f position = 1;
```
samiwh: This shouldn't be a 3f

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -31,43 +35,54 @@ protected override void OnCreateManager()
                 ComponentType.ReadOnly<WorkerEntityTag>(),
                 ComponentType.ReadOnly<OnConnected>()
             );
+        }
 
-            playerSpawnGroup = GetComponentGroup(
-                ComponentType.ReadOnly<ShouldRequestPlayerTag>()
-            );
+        public void QueuePlayerCreationRequest(Vector3 spawnPosition = default, byte[] serializedArguments = null)
```
samiwh: This doesn't seem to queue anything

samiwh: Presumably we only allow a client to request a single player?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
samiwh: I think it should be up to the caller to do this. We should just have byte[] as an argument and leave it at that. If we want to do this I would much rather make it structured data though and provide a way of making this a schema type.

workers/unity/Assets/Playground/Scripts/OneTimeInitialisation.cs
```diff
@@ -23,6 +23,7 @@ private static void Init()
 
             // Setup template to use for player on connecting client
             PlayerLifecycleConfig.CreatePlayerEntityTemplate = PlayerTemplate.CreatePlayerEntityTemplate;
+            PlayerLifecycleConfig.AutoRequestPlayerCreation = true;
```
samiwh: Why here rather than on the addition of the systems to the world?

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
paulbalaji: Given that arguments are optional, it might be nicer for users to check `HasValue` rather than `== null`

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
zeroZshadow: Our types are marked as serializable since last week.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -31,43 +35,54 @@ protected override void OnCreateManager()
...
             // The fix for this is more sophisticated server side handling of requests
             var responses = commandSystem.GetResponses<PlayerCreator.CreatePlayer.ReceivedResponse>();
+
             for (var i = 0; i < responses.Count; i++)
             {
                 ref readonly var response = ref responses[i];
                 if (response.StatusCode == StatusCode.AuthorityLost)
                 {
-                    PostUpdateCommands.AddComponent(response.SendingEntity, new ShouldRequestPlayerTag());
+                    QueuePlayerCreationRequest();
```
jamiebrynes7: This doesn't persist the arguments from the failed request

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
paulbalaji: Yes, clients can only request a single player

The `Queue` name was initially `Trigger`, but it doesn't exactly trigger a player creation request

perhaps it could renamed to `RequestPlayerCreation`?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
paulbalaji: Martijn merged in a change to make types, enums and snapshots serializable

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -31,43 +35,54 @@ protected override void OnCreateManager()
...
+            playerCreationRequested = true;
+            this.spawnPosition = Vector3f.FromUnityVector(spawnPosition);
+            this.serializedArguments = serializedArguments;
         }
 
         protected override void OnUpdate()
         {
-            var initEntities = initializationGroup.GetEntityArray();
-            for (var i = 0; i < initEntities.Length; ++i)
+            if (PlayerLifecycleConfig.AutoRequestPlayerCreation)
```
jamiebrynes7: If I want to auto create a player at startup - but with arguments, is this useful for me? And if not, is this misleading as an option?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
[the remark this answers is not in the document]
paulbalaji: C# object

Nothing in the method body relies on knowing the type of `playerCreationArguments`, so why would it need to be generic?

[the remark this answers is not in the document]
jamiebrynes7: This boxes structs currently. Use lowercase `object` so it doesn't clash with `UnityEngine.Object`

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
jamiebrynes7: Agree with @samiwh 100%

schema/playground/player_creation_arguments.schema
```diff
@@ -0,0 +1,6 @@
+package playground;
+
+type PlayerCreationParams
```
jamiebrynes7: Biggest nit of all time - file name = `player_creation_arguments`, type name is `PlayerCreationParams` 😛

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
paulbalaji: you can now define what default position/serialized arguments to use with auto-player creation

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
```diff
@@ -14,8 +14,16 @@ public static class PlayerTemplate
         {
             if (playerCreationArguments.HasValue)
             {
-                var deserializationSuccess = PlayerLifecycleHelper.DeserializeArguments<PlayerCreationParams>(
-                    playerCreationArguments.Value, out var deserializedArguments);
+                try
+                {
+                    var deserializedArguments = PlayerLifecycleHelper.DeserializeArguments<PlayerCreationParams>(
```
jamiebrynes7: Again - if we aren't using this, it doesn't belong in Playground. Rather unit or integration tests

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -43,24 +45,14 @@ public static bool SerializeArguments(Object playerCreationArguments, out byte[]
...
-                    return true;
-                }
-            }
-            catch (Exception e)
+            using (var memoryStream = new MemoryStream())
             {
-                UnityEngine.Debug.LogError($"Unable to deserialize player creation arguments. {e.Message}");
-                deserializedArguments = default;
-                return false;
+                var binaryFormatter = new BinaryFormatter();
```
jamiebrynes7: There's now asymmetry between this and the Serialize. One throws, one catches the exception and logs

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -90,9 +82,15 @@ public static bool IsOwningWorker(SpatialEntityId entityId, World workerWorld)
             return worker.Connection.GetWorkerId() == ownerId;
         }
 
-        public static void AddClientSystems(World world)
+        public static void AddClientSystems(World world, bool autoRequestPlayerCreation = true,
+            Vector3 spawnPosition = default,
+            byte[] serializedArguments = null)
```
jamiebrynes7: Can we assume that we will know the arguments at this time? (Immediately after the worker connects)

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -37,29 +36,39 @@ protected override void OnCreateManager()
...
+
         protected override void OnUpdate()
         {
             if (PlayerLifecycleConfig.AutoRequestPlayerCreation)
             {
                 var initEntities = initializationGroup.GetEntityArray();
                 for (var i = 0; i < initEntities.Length; ++i)
                 {
-                    QueuePlayerCreationRequest();
+                    RequestPlayerCreation();
```
jamiebrynes7: This can just be:

```
if (!initializationGroup.IsEmptyFilter)
{
	RequestPlayerCreation();
}
```

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -37,29 +36,39 @@ protected override void OnCreateManager()
...
                 }
             }
 
             if (playerCreationRequested)
             {
                 var request = new CreatePlayerRequestType
                 {
-                    Position = spawnPosition
+                    Position = Vector3f.FromUnityVector(spawnPosition)
```
jamiebrynes7: I still think we shouldn't mandate you add a position - the client shouldn't be telling the server where it should spawn anyway

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
paulbalaji: ah neat! didn't know that was a thing

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
paulbalaji: ok, so now both will throw and it's up to the user to handle these exceptions (probably for the better as caller has context)

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
paulbalaji: There could be some default arguments that a user may set once and never use again that are known at this point

jamiebrynes7: Hmmm - I'm not sure what the best API here is, especially since this splits the API across different places. 

Use cases:
- Automatic player spawning with preset arguments
- Automatic player spawning with dynamic (Is this a valid use case? This would require the user to ensure that they set the arguments _before_ the player lifecycle systems ran. In which case why do not a manual player spawn after you set the arguments)
- Manual player spawning with preset arguments
- Manual player spawning with dynamic arguments

paulbalaji: discussed offline:
* auto-spawning will not send any serialized arguments
* `RequestPlayerCreation` will take the serialized arguments directly (instead of having a different place to set it)
* keep the private `byte[]` field to cache the requested arguments for retries

workers/unity/Assets/Playground/Config/PlayerTemplate.cs
samiwh: They feel like they are the same thing. The difference is that option is a non standard collection.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
samiwh: Request would be more in keeping with what it does. Except I would expect that to do something immediately and that multiple calls to Request... would not cancel each other out.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Schema/player_creator.schema
```diff
@@ -3,12 +3,12 @@ import "improbable/vector3.schema";
 
 type CreatePlayerRequestType
 {
-    improbable.Vector3f position = 1;
+    bytes serialized_arguments = 1;
```
paulbalaji: :trollface:

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -37,15 +37,8 @@ protected override void OnCreateManager()
...
         {
             serializedArgumentsCache = serializedArguments;
-
-            var request = new CreatePlayerRequestType();
-            if (serializedArguments != null)
-            {
-                request.SerializedArguments = serializedArguments;
-            }
-
+            var request = new CreatePlayerRequestType(serializedArguments);
```
jamiebrynes7: I don't think we handle null assignments in the code generator. As in we don't translate null to `new byte[] { }`

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Schema/player_creator.schema
```diff
@@ -3,12 +3,12 @@ import "improbable/vector3.schema";
...
 }
 
 type CreatePlayerResponseType
 {
-    EntityId created_entity_id = 3;
+    EntityId created_entity_id = 1;
```
jamiebrynes7: 👏

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
paulbalaji: serialization of `CreatePlayerRequestType` calls `AddBytes` which has a special case if the value is null :)

(double checked it works when testing locally)

PR #813, opened 2019-03-05
Fix prefab processor

zeroZshadow:
#### Description
The prefab processor was not yet using the new Prefab editing API's, causing issues when upgrading from the SDK to the GDK.

This small PR fixes this issue, and adds some additional logging to identify problematic prefabs

#### Tests
Tested with playground

### DIFF
--- a/workers/unity/Assets/Playground/Resources/Prefabs/Common/Cube.prefab
+++ b/workers/unity/Assets/Playground/Resources/Prefabs/Common/Cube.prefab
@@ -1,21 +1,11 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
---- !u!1001 &100100000
-Prefab:
-  m_ObjectHideFlags: 1
-  serializedVersion: 2
-  m_Modification:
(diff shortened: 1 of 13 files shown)


workers/unity/Assets/Playground/Resources/Prefabs/Common/Cube.prefab
```diff
@@ -34,9 +24,10 @@ GameObject:
   m_IsActive: 1
 --- !u!4 &4711983361604210
 Transform:
-  m_ObjectHideFlags: 1
+  m_ObjectHideFlags: 0
   m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInternal: {fileID: 100100000}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
```
jessicafalk: these changes: do we need to make them for the other projects as well?

zeroZshadow: This happens automatically the first time you run play. So we can do that once things merge in

PR #822, opened 2019-03-11
Multithreaded connection handler

samiwh:
#### Description
A multi threaded connection handler.
Offloads serialisation and deserialisation work to other threads.
Worker is still set to use the single threaded one.

Good entry point for reading is the `MultiThreadedSpatialOSConnectionHandler`.

Currently the deserialsation takes the simplest form it can, locking on an oplist and releasing the lock when completely done. There are various ways this could potentially increase latency, however these are expected to be rare. If they do cause a problem there are two other implementations made but not in this PR. One of which has a more complicated non-standard lock, the other of which is lock free. Until the performance tests are merged and we have uses cases that cause latency issue we should not make it more complicated than it needs to be though,


workers/unity/Packages/com.improbable.gdk.core/Worker/CommandMetaData.cs
```diff
@@ -4,31 +4,42 @@
...
 
         private List<CommandIds> requestsToRemove = new List<CommandIds>();
 
         public CommandMetaData()
         {
-            var types = ReflectionUtility.GetNonAbstractTypes(typeof(ICommandMetaDataStorage));
-            foreach (var type in types)
+            if (storageTypes == null)
+            {
+                storageTypes = ReflectionUtility.GetNonAbstractTypes(typeof(ICommandMetaDataStorage));
```
jamiebrynes7: Could use the `??` operator to assign, but probably not worth

workers/unity/Packages/com.improbable.gdk.core/Worker/ComponentOpDeserializer.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+
+            deserializer.AddComponentToDiff(op, viewDiff);
+        }
+
+        public static void DeserializeAndApplyComponentUpdate(ComponentUpdateOp op, ViewDiff viewDiff, uint updateId)
+        {
+            if (!ComponentIdToComponentDeserializer.TryGetValue(op.Update.ComponentId, out var deserializer))
+            {
+                throw new ArgumentException($"Can not deserialize component with ID {op.Update.ComponentId}");
```
jamiebrynes7: Nit: Component update rather than component

workers/unity/Packages/com.improbable.gdk.core/Worker/ComponentOpDeserializer.cs
```diff
@@ -0,0 +1,128 @@
...
+            ++updateId;
+        }
+
+        public static void DeserializeAndApplyCommandRequestReceived(CommandRequestOp op, ViewDiff viewDiff)
+        {
+            if (!CommandIdsToCommandDeserializer.TryGetValue(
+                (op.Request.ComponentId, op.Request.SchemaData.Value.GetCommandIndex()),
+                out var deserializer))
+            {
+                throw new ArgumentException($"Can not deserialize component with ID {op.Request.ComponentId}");
```
jamiebrynes7: Probably want to print which command as well. 

Same for the response received below.

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/MockConnectionHandler.cs
```diff
@@ -1,4 +1,6 @@
+using System;
 using Improbable.Worker.CInterop;
+using JetBrains.Annotations;
```
jamiebrynes7: Please remove this.

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/MultithreadedSpatialOSConnectionHandler/ThreadedSerializationHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+            serializedMessagesToSendPool.Return(serializedMessagesToSend);
+        }
+
+        private void SerializeMessages()
+        {
+            // Block until a new message is available to serialize.
+            while (messagesToSerialize.TryTake(out var messagesToSend, -1))
+            {
+                // If adding is complete then there is no reason to deserialize the message.
+                if (messagesToSerialize.IsAddingCompleted)
```
jamiebrynes7: The name of this is slightly odd - afaict its just a signal to the thread that its time to shut down.

workers/unity/Packages/com.improbable.gdk.core/Worker/CommandMetaData.cs
samiwh: I mildly prefer the explicit check as we normally don't want to perform an assignment, but I'm not too precious about this.

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/MultithreadedSpatialOSConnectionHandler/ThreadedSerializationHandler.cs
samiwh: The `IsAddingComplete`? That's part of `BlockingCollection<>` rather than something I made. We can add another explicit signal instead but then we have to lock around it.

samiwh: It stops you being able to add more messages to the queue and cause another flag to be set when it's then emptied.

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/MockConnectionHandler.cs
samiwh: Will do.

workers/unity/Packages/com.improbable.gdk.core/Worker/ComponentOpDeserializer.cs
samiwh: Sure

workers/unity/Packages/com.improbable.gdk.core/Worker/ComponentOpDeserializer.cs
samiwh: Sure

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/MultithreadedSpatialOSConnectionHandler/ThreadedSerializationHandler.cs
jamiebrynes7: In which case - forgot I said that 😬

PR #823, opened 2019-03-11
Harvest android binaries to artifacts

petemounce:
... for subsequent steps to run device-farm tests against.

#### Description
Need the APK to be inside artifacts, so that we can use it within subsequent testing steps.

#### Tests
Nothing - the BK UI will show whether the artifact is uploaded, or not.

#### Documentation
Does not apply.

#### Primary reviewers
@jessicafalk @jamiebrynes7


petemounce:
Success: 

```
2019-03-13 22:53:00 INFO   Uploading artifact b584d47c-4d58-4745-9daf-0e83f19ca35e workers\unity\build\worker\AndroidClient@Android\AndroidClient@Android.apk (52496987 bytes)
```

PR #824, opened 2019-03-11
Sort while deserialising

samiwh:
#### Description
Change lists are now created sorted rather than doing this lazily. Configuration should be added to specify what to do on a component basis.

This is not optimised an any way but it can be offloaded onto a deserialisation thread as is.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Collections/MessageList.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Collections/MessageList.cs
@@ -45,6 +45,40 @@ namespace Improbable.Gdk.Core
             ++Count;
         }
 
+        public void Insert(int index, in T item)
+        {
+            if (items.Length <= Count)
+            {
+                int targetLength = items.Length == 0 ? 4 : items.Length * 2;
+                var temp = new T[targetLength];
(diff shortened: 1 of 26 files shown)


workers/unity/Packages/com.improbable.gdk.core/Collections/MessageList.cs
```diff
@@ -45,6 +45,39 @@ public void Add(in T item)
...
+                Array.Copy(items, index, items, index + 1, Count - index);
+                items[index] = item;
+            }
+
+            ++Count;
+        }
+
+        public void InsertSorted(in T item, IComparer<T> comparer)
+        {
+            var index = Array.BinarySearch(items, 0, Count, item, comparer);
```
jamiebrynes7: The way this function works is excellent

jamiebrynes7: > The index of the specified value in the specified array, if value is found; otherwise, a negative number. If value is not found and value is less than one or more elements in array, the negative number returned is the bitwise complement of the index of the first element that is larger than value. If value is not found and value is greater than all elements in array, the negative number returned is the bitwise complement of (the index of the last element plus 1).

For those interested

PR #828, opened 2019-03-12
Fix a race on component update loopbacks

jamiebrynes7:
#### Description
Temporary fix until we handle loopbacks properly. cc @samiwh 

Will make a ticket for this.

#### Tests
Ran it

#### Documentation
N/A

#### Primary reviewers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/ViewStorageGenerator.tt
+++ b/workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/ViewStorageGenerator.tt
@@ -72,12 +72,6 @@ namespace <#= qualifiedNamespace #>
                     componentData[entity.Id] = new Snapshot();
                 }
 
-                foreach (var entity in storage.GetComponentsRemoved())
-                {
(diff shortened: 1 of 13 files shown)


workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/ViewStorageGenerator.tt
```diff
@@ -91,6 +85,13 @@ namespace <#= qualifiedNamespace #>
                     var authorityChange = authorityChanges[i];
                     authorityStates[authorityChange.EntityId.Id] = authorityChange.Authority;
                 }
+
+                // TODO: When loopbacks are properly handled, this should be done before updates are applied.
```
samiwh: There's a race that this doesn't solve.
If using the multithreaded connection handler, an oplist could be placed on the deserialisation queue that contains a remove authority + remove entity for some entity E.
That tick, while this is still being deserialised, the worker thinks it's still authoritative and tries to send an update. Then the oplist containing the update loopback gets added back to deserialisation queue.

So on tick 0 nothing happens,
tick 1 the worker is told the entity is removed
tick 2 the worker receives it's own component updates for that entity via a loopback

PR #831, opened 2019-03-12
Made the option type readonly + minor changes

samiwh:
#### Description
Speculative change to make option readonly.
Removed the requirement to for references to not be null

Some slight changes to the tests

WIP - write changes in release notes


zeroZshadow:
Re-implemented using #1110

PR #835, opened 2019-03-13
Reactive Systems now use concurrent jobs for chunk gathering

zeroZshadow:
#### Description
Reactive Systems now use concurrent jobs for chunk gathering
ReactiveComponentSendSystem order of sending has been changed to better cache usage
Local UnityGameLogic build is now a Development build

#### Tests
Ran playground project using editor and build out workers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/ReactiveComponents/Systems/ReactiveComponentSendSystem.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/ReactiveComponents/Systems/ReactiveComponentSendSystem.cs
@@ -2,7 +2,9 @@ using System;
 using System.Collections.Generic;
 using System.Reflection;
 using Improbable.Gdk.Core;
+using Unity.Collections;
 using Unity.Entities;
+using Unity.Jobs;
 using UnityEngine.Profiling;
 
 namespace Improbable.Gdk.ReactiveComponents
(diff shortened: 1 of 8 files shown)


workers/unity/Packages/com.improbable.gdk.core/ReactiveComponents/Systems/ReactiveComponentSendSystem.cs
```diff
@@ -36,17 +40,58 @@ protected override void OnUpdate()
                 return;
             }
 
-            var commandSystem = World.GetExistingManager<CommandSystem>();
+            ReplicateEvents();
+            ReplicateCommands();
```
jamiebrynes7: Is it worth grouping the gathering jobs from these together and waiting once for them to complete? 

I.e. - 

`var chunks = GatherChunks();`
`ReplicateEvents(ref chunks);`
`ReplicateCommands(ref chunks);`

Or is the extra bookkeeping not worth it?

samiwh: Given the change that we now check each chunk for the relevant components we could also try things like getting all chunks with a spatailos entity ID in them. I don't know if that would actually be good for everything, but for command senders that might actually be better.

PR #843, opened 2019-03-15
WIP (should be a draft but github is a troll) Put together a slightly different worker model

samiwh:
#### Description
Mocked up what a csharp gdk worker might look like


samiwh:
Closing to see if I can open it as a draft.

PR #844, opened 2019-03-15
Put together a slightly different worker model

samiwh:
#### Description
Mocked up what a csharp GDK worker might look like.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandlerBuilder.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandlerBuilder.cs
@@ -1,0 +1,101 @@
+using System;
+using System.Threading.Tasks;
+using Improbable.Worker.CInterop;
+
+namespace Improbable.Gdk.Core
+{
+    public class SpatialOSConnectionHandlerBuilder : IConnectionHandlerBuilder
+    {
+        public enum ThreadingMode
+        {
+            SingleThreaded,
+            MultiThreaded
+        }
+
+        private ReceptionistConfig? receptionistConfig;
+        private LocatorConfig? locatorConfig;
+        private AlphaLocatorConfig? alphaLocatorConfig;
+
+        private ConnectionParameters connectionParameters;
+
+        private ThreadingMode threadingMode = ThreadingMode.SingleThreaded;
+
+        public SpatialOSConnectionHandlerBuilder SetConfig(ReceptionistConfig config)
+        {
+            CheckExistingConfigAndThrow();
+            receptionistConfig = config;
+            return this;
+        }
+
+        public SpatialOSConnectionHandlerBuilder SetConfig(LocatorConfig config)
+        {
+            CheckExistingConfigAndThrow();
+            locatorConfig = config;
+            return this;
+        }
+
+        public SpatialOSConnectionHandlerBuilder SetConfig(AlphaLocatorConfig config)
+        {
+            CheckExistingConfigAndThrow();
+            alphaLocatorConfig = config;
+            return this;
+        }
+
+        public SpatialOSConnectionHandlerBuilder SetThreadingMode(ThreadingMode mode)
+        {
+            threadingMode = mode;
+            return this;
+        }
+
+        public SpatialOSConnectionHandlerBuilder SetConnectionParameters(ConnectionParameters parameters)
+        {
+            connectionParameters = parameters;
+            return this;
+        }
+
+        public async Task<IConnectionHandler> CreateAsync()
+        {
+            if (!(receptionistConfig.HasValue || locatorConfig.HasValue || alphaLocatorConfig.HasValue))
+            {
+                throw new InvalidOperationException("Must set exactly one of ReceptionistConfig, LocatorConfig, or AlphaLocatorConfig");
+            }
+
+            if (connectionParameters == null)
+            {
+                throw new InvalidOperationException("ConnectionParameters must be provided");
+            }
+
+            Connection connection;
+            if (receptionistConfig.HasValue)
+            {
+                connection = await ConnectionUtils.CreateConnectionAsync(receptionistConfig.Value, connectionParameters);
+            }
+            else if (locatorConfig.HasValue)
+            {
+                connection = await ConnectionUtils.CreateConnectionAsync(locatorConfig.Value, connectionParameters);
+            }
+            else
+            {
+                connection = await ConnectionUtils.CreateConnectionAsync(alphaLocatorConfig.Value, connectionParameters);
+            }
+
+            switch (threadingMode)
+            {
+                case ThreadingMode.SingleThreaded:
+                    return new SpatialOSConnectionHandler(connection);
+                case ThreadingMode.MultiThreaded:
+                    return new MultiThreadedSpatialOSConnectionHandler(connection);
+                default:
+                    throw new ArgumentOutOfRangeException();
+            }
+        }
+
+        private void CheckExistingConfigAndThrow()
+        {
+            if (receptionistConfig.HasValue || locatorConfig.HasValue || alphaLocatorConfig.HasValue)
+            {
+                throw new InvalidOperationException("Must set exactly one of ReceptionistConfig, LocatorConfig, or AlphaLocatorConfig");
+            }
+        }
+    }
+}
--- a/workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionUtils.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionUtils.cs
@@ -1,0 +1,135 @@
+using System.Threading.Tasks;
+using Improbable.Worker.CInterop;
+using AlphaLocator = Improbable.Worker.CInterop.Alpha.Locator;
+
+namespace Improbable.Gdk.Core
+{
+    public static class ConnectionUtils
+    {
+        /// <summary>
+        ///     Connects to the SpatialOS Runtime via the Receptionist service and creates a <see cref="Connection" /> object
+        ///     asynchronously.
+        /// </summary>
+        /// <param name="config">
+        ///     The <see cref="ReceptionistConfig" /> object stores the configuration needed to connect via the
+        ///     Receptionist Service.
+        /// </param>
+        /// <param name="connectionParameters">The <see cref="ConnectionParameters" /> storing </param>
+        /// <returns>
+        ///     A <see cref="Task{TResult}" /> to run this method asynchronously and retrieve the created
+        ///     <see cref="Connection" /> object upon connecting successfully.
+        /// </returns>
+        public static async Task<Connection> CreateConnectionAsync(
+            ReceptionistConfig config,
+            ConnectionParameters connectionParameters)
+        {
+            using (var connectionFuture =
+                Connection.ConnectAsync(config.ReceptionistHost, config.ReceptionistPort, config.WorkerId,
+                    connectionParameters))
+            {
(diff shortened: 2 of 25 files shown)


workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandlerBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        public enum ThreadingMode
+        {
+            SingleThreaded,
+            MultiThreaded
+        }
+
+        private ReceptionistConfig? receptionistConfig;
+        private LocatorConfig? locatorConfig;
+        private AlphaLocatorConfig? alphaLocatorConfig;
```
jamiebrynes7: Wish we have C# unions :confused: 

Alternatively - perfect use case for https://github.com/mcintyre321/OneOf but sadly external dependencies :sob:

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionUtils.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// </returns>
+        public static async Task<Connection> CreateConnectionAsync(
+            LocatorConfig parameters,
+            ConnectionParameters connectionParameters)
+        {
+            using (var locator = new Locator(parameters.LocatorHost, parameters.LocatorParameters))
+            {
+                var deploymentList = await GetDeploymentList(locator);
+
+                var deploymentName = parameters.DeploymentListCallback(deploymentList);
```
jamiebrynes7: Hmmm - we should allow for the possibility that this callback can be asynchronous. 

I.e - user displays to the client and they select a deployment

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -33,244 +24,67 @@ public class Worker : IDisposable
...
-        /// <param name="origin">The origin of this worker in the local Unity space.</param>
-        /// <returns>
-        ///     A <see cref="Task{TResult}" /> to run this method asyncally and retrieve the created <see cref="Worker" />
-        ///     object upon connecting successfully.
-        /// </returns>
-        private static async Task<Worker> TryToConnectAsync(Future<Connection> connectionFuture,
-            string workerType,
-            ILogDispatcher logger,
-            Vector3 origin)
+        public static async Task<Worker> CreateWorkerAsync(IConnectionHandlerBuilder connectionHandlerBuilder, string workerType,
```
jamiebrynes7: It feels like the connection builder should be extended into a `WorkerBuilder`. I.e. - 

```
var worker = await new WorkerBuilder(workerType)
	.WithConnectionConfig(connectionConfig)
	.WithLogger(logDispatcher)
	.AsMultithreaded()
	.ConnectAsync();
```

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -23,51 +23,31 @@ public static class PlayerLifecycleHelper
...
-            if (!entityManager.HasComponent<OwningWorker.Component>(entity))
+            if (!updateSystem.HasComponent(OwningWorker.ComponentId, entityId))
             {
                 return false;
             }
 
-            var ownerId = entityManager.GetComponentData<OwningWorker.Component>(entity).WorkerId;
 
-            return worker.Connection.GetWorkerId() == ownerId;
+            var ownerId = updateSystem.GetComponent<OwningWorker.Snapshot>(entityId).WorkerId;
```
jamiebrynes7: Why would you grab this off the updateSystem rather than the view?

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionUtils.cs
jessicafalk: we probably don't want that. the login token that is provided by the launcher will most likely time out by the time the user selects a deployment. 
If users want to do that, they need to add quite a bit of additional logic and will most likely need to use the alpha locator together with dev auth flow or their own auth server.

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionUtils.cs
```diff
@@ -0,0 +1,135 @@
...
+                        ConnectionErrorReason.DeploymentNotFound);
+                }
+
+                return deploymentList.Value;
+            }
+        }
+
+        private static string GetConnectionFailureReason(Connection connection)
+        {
+            return connection.GetConnectionStatusCodeDetailString() ?? "No reason found for connection failure";
```
jessicafalk: can this actually be null? 
what about the status code?

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -88,32 +86,27 @@ public async Task Connect(string workerType, ILogDispatcher logger)
                 }
 
                 var origin = transform.position;
-                ConnectionDelegate connectionDelegate;
                 var chosenService = GetConnectionService();
                 var connectionParameters = GetConnectionParameters(workerType, chosenService);
+                SpatialOSConnectionHandlerBuilder connectionHandlerBuilder = new SpatialOSConnectionHandlerBuilder();
```
jessicafalk: now that we have the `ConnectionHandlerBuilder`, it would be really nice to get rid of the `WorkerConnector`. If we can make it a `WorkerBuilder` as Jamie suggested, I think we can get rid of this and simply provide sample implementations of how to use it for PC vs Mobile. 
This would allow us to finally get rid of the many levels of inheritance we have right now

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerInWorld.cs
```diff
@@ -0,0 +1,85 @@
...
+
+namespace Improbable.Gdk.Core
+{
+    public class WorkerInWorld : Worker
+    {
+        /// <summary>
+        ///     The origin of the worker in global Unity space.
+        /// </summary>
+        public readonly Vector3 Origin;
+
```
jessicafalk: nit: empty line

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandlerBuilder.cs
samiwh: We can make this a union if we really want to. I wasn't changing any of this code though, just rearranging it.

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionUtils.cs
samiwh: This was just rearranging what was there. If we can't select a deployment without a token and we can't keep a token after selecting a deployment that is a problem though.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
samiwh: The view isn't currently public. The update system is how you access it.

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionUtils.cs
samiwh: It shouldn't be. I didn't write this though, just moved it I think.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
samiwh: I wanted to keep Unity and non Unity separate. The worker connector is glue between the worker and Unity, so I suspect that some form of it will still be useful.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
samiwh: I didn't do that because you might want different types of connection handler. The IConnectionHandlerBuilder interface is basically a type safe way of transferring ownership.

The current connection handler builder is to build the standard one you might one when running spatial. If you want a test one or a remoting one you want different options. We could write one massive builder that can encompass every useful permutation but that seems like a bad idea even if possible.

samiwh: We could instead set it up so that there is a single connection handler with a pipeline of behaviours and we feed it in different way (ie what it considers the simulation can be rewritten). Then we can potentially just have a single builder and merge it with the worker creation process.

I have looked into doing this and have some rough design notes if you want to go over them. (It's also probably what I'd do if starting now)

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandlerBuilder.cs
samiwh: Also I think we can get rid of alpha locator though.

jessicafalk: nope, not of the alpha locator :P we could get rid of the locator, if  you wanna reduce complexity.

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionUtils.cs
jessicafalk: I agree, I had many discussions about this...

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
jamiebrynes7: That's bizarre

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandlerBuilder.cs
jamiebrynes7: Discussed offline - can't get rid of the locator.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
jamiebrynes7: Pls send design notes

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
samiwh: It's basically a stepping stone. Update system was written for something else. Made sense as an interface for a bit, and now doesn't. It's mostly why the class wasn't mentioned in the release notes. It would be best to remove it and make a new thing because it's just a badly named interface at the moment.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
samiwh: Currently writing a doc for a slightly different version. Can go through that and work out how it might work in C# too.

jamiebrynes7:
Rebased the branch - logging appears to be broken and some functionality that was added since this PR was opened needs to be added back in

jamiebrynes7:
Superseded by #981

PR #859, opened 2019-03-25
Player lifecycle improvements

paulbalaji:
#### Description
- remove hardcoded player creator entity id (UTY-1687)
- if no player creator entity ids recorded, query for them
- when sending player creation request, send to a random player creation entity
- cleaned up schema (UTY-1626)
- improved `AddPlayerLifecycleComponents` signature, constructing client access attribute in the helper (UTY-1431)
- tidied up playground PlayerTemplate
- updated default snapshot to have more than one player creator

#### Tests
- [x] local playground
- [x] cloud playground

#### Documentation
- [x] changelog
- [x] https://github.com/spatialos/gdk-for-unity/pull/869

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,18 +35,29 @@ protected override void OnCreateManager()
...
+            var playerCreatorCount = playerCreatorEntityIds.Count;
+            if (playerCreatorCount > 0)
+            {
+                commandSystem.SendCommand(new PlayerCreator.CreatePlayer.Request(
+                    playerCreatorEntityIds[Random.Range(0, playerCreatorCount)],
+                    new CreatePlayerRequestType(serializedArgumentsCache)));
+                return true;
+            }
+
+            Debug.LogWarning("Unable to send player creation request: no player creator entities found yet.");
```
jamiebrynes7: I don't think this warning is particularly useful here - as a user I'd rather you queued my request and send it when it was found.

This warning should be thrown (as an error) if the query fails to find any matching entities (more than once).

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,18 +35,29 @@ protected override void OnCreateManager()
             commandSystem = World.GetExistingManager<CommandSystem>();
             logDispatcher = workerSystem.LogDispatcher;
 
+            playerCreatorEntityIds = new List<EntityId>();
```
jamiebrynes7: This can just be done in the declaration

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -51,8 +68,48 @@ private void RetryCreatePlayerRequest()
         protected override void OnUpdate()
         {
             if (PlayerLifecycleConfig.AutoRequestPlayerCreation && !initializationGroup.IsEmptyIgnoreFilter)
+            {
+                sendAutoPlayerCreationRequest = true;
```
jamiebrynes7: Why is this conditional on `!initializationGroup.IsEmptyIgnoreFilter` ?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -51,8 +68,48 @@ private void RetryCreatePlayerRequest()
...
+            }
+
+            if (playerCreatorEntityIds.Count == 0)
```
jamiebrynes7: Invert the if to reduce nesting. If this is `0` I don't think we should even attempt to send the request?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -51,8 +68,48 @@ private void RetryCreatePlayerRequest()
...
+            {
+                if (!playerCreatorQueryId.HasValue)
+                {
+                    playerCreatorQueryId = commandSystem.SendCommand(new WorldCommands.EntityQuery.Request
+                    {
+                        EntityQuery = new EntityQuery
```
jamiebrynes7: Can cache this query as it doesn't change

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
paulbalaji: to prevent against the case where `AutoRequestPlayerCreation` is true but the worker hasn't connected yet

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
paulbalaji: atm we don't as there's a return at the end of this `if`

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
jamiebrynes7: That should never be the case - these systems are always created after the worker is created. Ideally they should be internal and only accessible through the helpers

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
jamiebrynes7: Ah right - still would be cleaner if inverted

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -11,16 +14,26 @@ namespace Improbable.Gdk.PlayerLifecycle
...
         private ILogDispatcher logDispatcher;
 
-        private ComponentGroup initializationGroup;
-
         private byte[] serializedArgumentsCache;
 
+        private bool playerRequestQueued;
+        private int playerCreationAttempts = 0;
+
+        private int playerCreatorQueryAttempts = 0;
```
jamiebrynes7: Unnecessary initialization

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -11,16 +14,26 @@ namespace Improbable.Gdk.PlayerLifecycle
...
+        private long? playerCreatorQueryId;
+
+        private readonly List<EntityId> playerCreatorEntityIds = new List<EntityId>();
+
+        private readonly EntityQuery playerCreatorQuery = new EntityQuery
+        {
+            Constraint = new ComponentConstraint(PlayerCreator.ComponentId),
+            ResultType = new SnapshotResultType()
```
jamiebrynes7: Isn't the other result type more appropriate? We don't care about the data within the entity - just the ID

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -11,16 +14,26 @@ namespace Improbable.Gdk.PlayerLifecycle
...
 
-        private ComponentGroup initializationGroup;
-
         private byte[] serializedArgumentsCache;
 
+        private bool playerRequestQueued;
+        private int playerCreationAttempts = 0;
+
+        private int playerCreatorQueryAttempts = 0;
+        private long? playerCreatorQueryId;
```
jamiebrynes7: Nit - just for clarity lets call this `playerCreatorEntityQueryId`

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,48 +42,116 @@ protected override void OnCreateManager()
...
 
-            for (var i = 0; i < responses.Count; i++)
-            {
-                ref readonly var response = ref responses[i];
-                if (response.StatusCode == StatusCode.AuthorityLost)
+                for (var i = 0; i < responses.Count; i++)
                 {
-                    RetryCreatePlayerRequest();
+                    ref readonly var response = ref responses[i];
+                    if (response.StatusCode == StatusCode.AuthorityLost)
```
jamiebrynes7: Can you ever get `AuthorityLost` back from a world command..? That doesn't make a ton of sense

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,48 +42,116 @@ protected override void OnCreateManager()
...
+                    {
+                        RetryCreatePlayerRequest();
+                    }
+                    else if (response.StatusCode != StatusCode.Success)
```
jamiebrynes7: Surely a timeout status code warrants a retry? I'd recommend expanding this into a switch-case and grouping retry-able failures into a block that fall through into the Retry invocation.

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,48 +42,116 @@ protected override void OnCreateManager()
...
+                        logDispatcher.HandleLog(LogType.Error, new LogEvent(
+                            $"Create player request failed: {response.Message}"
+                        ));
+                    }
                 }
-                else if (response.StatusCode != StatusCode.Success)
+            }
+            else
+            {
+                var entityQueryResponses = commandSystem.GetResponses<WorldCommands.EntityQuery.ReceivedResponse>();
```
jamiebrynes7: Small nit we can refactor these so Update is just: 

```csharp
if (playerCreatorEntityIDs.Count >0) 
{
	HandleCreateEntity();
}
else
{
	HandleEntityQueryResponses();
}
```

or similar for legibilities sake

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,48 +42,116 @@ protected override void OnCreateManager()
...
+                for (var i = 0; i < entityQueryResponses.Count; i++)
                 {
-                    logDispatcher.HandleLog(LogType.Error, new LogEvent(
-                        $"Create player request failed: {response.Message}"
-                    ));
+                    ref readonly var response = ref entityQueryResponses[i];
+                    if (response.RequestId == playerCreatorQueryId)
```
jamiebrynes7: Can invert the if to reduce nesting

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,48 +42,116 @@ protected override void OnCreateManager()
...
+                    {
+                        if (response.StatusCode == StatusCode.Success && response.Result != null)
```
jamiebrynes7: I think one of the contracts from the Worker SDK is that if the response is a success, the Result cannot be null

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,48 +42,116 @@ protected override void OnCreateManager()
...
+        private void QueryForPlayerCreators()
+        {
+            playerCreatorQueryId = commandSystem.SendCommand(new WorldCommands.EntityQuery.Request
+            {
+                EntityQuery = playerCreatorQuery
+            });
+            ++playerCreatorQueryAttempts;
         }
 
         public void RequestPlayerCreation(byte[] serializedArguments = null)
```
jamiebrynes7: If a user calls this inbetween a retry attempt - we end up overriding the parameters? 

user says: send request with params A
request A1 sent
request A1 fails 
request A2 sent
user says: send request with params B
request A2 fails
request B1 sent
--- request A lost to the void ---

Is that acceptable? 

Should we allow users to call this multiple times? Should we retry request A N times until final failure and then do request B or do they run concurrently? 

Not sure on the use cases here

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
paulbalaji: Using the `CountResultType` only returns the number of entities that match the query

to get the entity IDs we have to get the full dictionary of results and ignore the values

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
paulbalaji: this isn't a world command - it's our own `CreatePlayer` command that'll be handled by whatever worker is authoritative over the player creator entity

the `CreateEntity` world command is different (and is handled by a server worker)

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
paulbalaji: I've updated it now so that once a request with args A has been requested, any further requests (ie B in your example) will not be considered until A succeeds, fails concretely or finishes retrying.

I reckon doing this + giving the option to manually enable/disable retries + perhaps giving a callback option for success/failure is probably more flexible and better for the user

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
jamiebrynes7: That's annoying 😕

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -10,17 +10,14 @@ namespace Improbable.Gdk.PlayerLifecycle
     public static class PlayerLifecycleHelper
     {
         public static void AddPlayerLifecycleComponents(EntityTemplate template,
-            string workerId,
-            string clientAccess,
+            string clientWorkerId,
```
jamiebrynes7: While we are here - shall we add XML docs to this?

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,49 +45,157 @@ protected override void OnCreateManager()
...
+                playerCreatorEntityQueryId = null;
+
+                if (response.StatusCode == StatusCode.Success)
+                {
+                    playerCreatorEntityIds.AddRange(response.Result.Keys);
+                }
+                else if (playerCreatorQueryRetries < PlayerLifecycleConfig.MaxPlayerCreatorQueryRetries)
+                {
+                    logDispatcher.HandleLog(LogType.Warning, new LogEvent(
+                        $"Retrying player creator query, attempt {playerCreatorQueryRetries}."
```
jamiebrynes7: This will log `attempt 0` the first time it is ran 😉

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Changed
```
jamiebrynes7: Add an entry under `Added`, the `EntityTemplate` static helper

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
...
+
+- Tidied up the player lifecycle module:
+  - It now dynamically queries for player creator entities, and sends requests to a random one each time. This removes the reliance on a hardcoded player creator Entity ID.
+  - Removed the `Type` suffix from player lifecycle schema types.
+  - Removed `clientAccess` from `AddPlayerLifecycleComponents` signature, now constructing the client access attribute within the helper.
```
jamiebrynes7: Add this to a `Breaking Changes` section

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Changed
+
+- Tidied up the player lifecycle module:
```
jamiebrynes7: Can probably drop this heading

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,49 +45,171 @@ protected override void OnCreateManager()
...
-            var request = new CreatePlayerRequestType(serializedArguments);
-            var createPlayerRequest = new PlayerCreator.CreatePlayer.Request(playerCreatorEntityId, request);
-            commandSystem.SendCommand(createPlayerRequest);
+            playerCreationRequestQueued = true;
+        }
+
+        private void SendCreatePlayerRequest()
+        {
+            playerCreationRequestId = commandSystem.SendCommand(new PlayerCreator.CreatePlayer.Request(
+                playerCreatorEntityIds[Random.Range(0, playerCreatorEntityIds.Count)],
```
zeroZshadow: This might need a comment

workers/unity/Packages/com.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -29,49 +45,171 @@ protected override void OnCreateManager()
...
                 ref readonly var response = ref responses[i];
-                if (response.StatusCode == StatusCode.AuthorityLost)
+                if (response.RequestId != playerCreationRequestId)
                 {
-                    RetryCreatePlayerRequest();
+                    continue;
                 }
-                else if (response.StatusCode != StatusCode.Success)
+
+                playerCreationRequestId = null;
```
zeroZshadow: This might need a comment as to what this does (clear the nullable id, and why)

PR #862, opened 2019-03-25
Tooling PackageManager refactor

zeroZshadow:
#### Description
Refactored the Tooling package to no longer have a hardcoded dependency on `file:` paths in the manifest.
This allows support for published unity packages in the future.

#### Tests
New approach works with playground, test project, and fps

#### Documentation
No - Internal tooling change

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/Common.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/Common.cs
@@ -67,20 +67,21 @@ namespace Improbable.Gdk.Tools
         public static string GetPackagePath(string packageName)
         {
             // Get package info
-            var request = Client.Search(packageName);
+            var request = Client.List(true);
             while (!request.IsCompleted)
             {
                 // Wait for the request to complete
             }
 
             var result = request.Result;
+            var package = result.FirstOrDefault(info => info.name.Equals(packageName));
 
-            if (request.Status != StatusCode.Success || result.Length <= 0)
+            if (package == null)
             {
-                throw new Exception($"Count not find '{packageName}', is it in your projects' manifest?");
+                throw new Exception($"Count not find '{packageName}', is it in your projects' manifest?\n{request.Error.message}");
             }
 
-            return result[0].resolvedPath;
+            return package.resolvedPath;
         }
 
         /// <summary>
--- a/workers/unity/Packages/com.improbable.gdk.tools/Common.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/Common.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Text;
 using Improbable.Gdk.Tools.MiniJSON;
 using UnityEditor;
+using UnityEditor.PackageManager;
 using UnityEngine;
 
 namespace Improbable.Gdk.Tools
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/com.improbable.gdk.tools/Common.cs
```diff
@@ -66,65 +66,32 @@ internal static string GetThisPackagePath()
         /// </summary>
         public static string GetPackagePath(string packageName)
         {
-            var manifest = ParseDependencies(ManifestPath);
-
-            if (!manifest.TryGetValue(packageName, out var path))
+            // Get package info
+            var request = Client.List(true);
+            while (!request.IsCompleted)
```
jared-improbable: Yikes, what a terrible API. Busy-waiting the main thread...

Also, use a named parameter so people can tell what `true` means without reading the docs.

workers/unity/Packages/com.improbable.gdk.tools/Common.cs
```diff
@@ -66,65 +66,32 @@ internal static string GetThisPackagePath()
...
             {
-                throw new Exception($"The project manifest must reference '{packageName}'.");
+                // Wait for the request to complete
             }
 
-            if (!path.StartsWith("file:"))
-            {
-                throw new Exception($"The '{packageName}' package must exist on disk.");
-            }
+            var result = request.Result;
```
jamiebrynes7: Nit - can just be wrapped into the line below - there aren't any other usages of `result`

workers/unity/Packages/com.improbable.gdk.tools/Common.cs
```diff
@@ -66,65 +66,32 @@ internal static string GetThisPackagePath()
...
+            var package = result.FirstOrDefault(info => info.name.Equals(packageName));
 
-            path = path.Replace("file:", string.Empty);
-
-            if (Path.IsPathRooted(path))
+            if (package == null)
             {
-                // A "rooted path" is an absolute path, therefore it will point directly at the package.
-                return path;
+                throw new Exception($"Count not find '{packageName}', is it in your project's manifest?\n{request.Error.message}");
```
jamiebrynes7: Count --> Could

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -227,31 +228,21 @@ private static void CopySchema()
...
-                    }
+                    // Wait for the request to complete
                 }
 
-                var schemaSources = packages.Where(SchemaPathExists)
-                    .ToDictionary(kv => kv.Key, GetSchemaPath);
+                var packages = request.Result;
+                var schemaSources = packages.ToDictionary(package => package.name,
+                        package => Path.Combine(package.resolvedPath, "Schema"))
+                    .Where(kv => Directory.Exists(kv.Value));
```
jamiebrynes7: Can we do the filtering before creating the dictionary?

workers/unity/Packages/com.improbable.gdk.tools/Common.cs
zeroZshadow: It needs to be blocking, and sadly does not allow `await`, its pretty much instantly complete in offline mode anyways.

Will do the named variables on all calls there

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
zeroZshadow: That becomes tricky because I need the Schema directory, so I'd have to do that path combination twice. I think this will work fine.

PR #869, opened 2019-03-28
Improve Player Lifecycle Docs

paulbalaji:
#### Description
- expand and update player lifecycle docs
- updated player lifecycle code samples to make use of the new `EntityTemplate.GetWorkerAccessAttribute(workerId)` method
- drive-by update to use the snapshot constructors in the player entity template examples

TODO:
- [x] feature module overview
- [x] how to install package
- [x] how to setup basic spawning
- [x] explain heartbeating
	- [x] overview
	- [x] initialisation
	- [x] server send heartbeat request
	- [x] client send heartbeat response
	- [x] server handle response OR kill client
- [x] how to customise things
	- [x] manually call RequestPlayerCreation
	- [x] pass in arbitrary arguments
	- [x] pass in a callback
	- [x] what the different `PlayerLifecycleConfig` variables represent/do: https://github.com/spatialos/gdk-for-unity/pull/893

#### Tests
- [x] improbadoc serve

#### Documentation
this is the docs

#### Primary reviewers
If your change will take a long time to review, you can name at most two primary reviewers who are ultimately responsible for reviewing this request. @ mention them.

PR #873, opened 2019-03-28
Add callback to player creation

paulbalaji:
#### Description
This feature is necessary to fully use the player lifecycle module in the FPS starter project.

Added the option to set a callback when requesting player creation. This is then invoked when a create player request receives a response back, passing the response in as an argument.

#### Tests
with callback:
- [x] works with fps locally
- [x] works with fps in the cloud

without callback:
- [x] works with playground locally
- [x] works with playground in the cloud

#### Documentation
- [x] xml docs
- [x] changelog
- [x] https://github.com/spatialos/gdk-for-unity/pull/869

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.playerlifecycle/Config/PlayerLifecycleConfig.cs
```diff
@@ -11,22 +11,19 @@ public static class PlayerLifecycleConfig
         /// <summary>
         ///     The time in seconds between player heartbeat requests.
         /// </summary>
-        public const float PlayerHeartbeatIntervalSeconds = 5f;
+        public static float PlayerHeartbeatIntervalSeconds = 5f;
```
jamiebrynes7: Why `const` -> `static` ?

paulbalaji: This is to allow users to change these variables in-code, outside of this class

PR #888, opened 2019-04-01
Redirected process changes required for deployment launcher fixes

jamiebrynes7:
#### Description
* Refactor `RedirectedProcess` to support running async with a cancellation token.
* Also fixed a bug where if you invoked `RedirectedProcess.Command()` on a non-main thread Unity would throw an exception.

#### Tests
Ran things. It seemed to work just fine.

#### Documentation
Needs release note.

#### Primary reviewers

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
@@ -105,7 +105,10 @@ namespace Improbable.Gdk.Tools
                     case RuntimePlatform.LinuxEditor:
                     case RuntimePlatform.OSXEditor:
                         // Ensure the schema compiler is executable.
-                        var _ = RedirectedProcess.Command("chmod").WithArgs("+x", $"\"{schemaCompilerPath}\"").Run();
+                        var _ = RedirectedProcess.Command("chmod")
+                            .WithArgs("+x", $"\"{schemaCompilerPath}\"")
+                            .InDirectory(Path.GetFullPath(Path.Combine(Application.dataPath, "..")))
+                            .Run();
                         break;
                     default:
                         throw new PlatformNotSupportedException(
--- a/workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
@@ -104,8 +104,7 @@ namespace Improbable.Gdk.Tools
                         break;
                     case RuntimePlatform.LinuxEditor:
                     case RuntimePlatform.OSXEditor:
-                        // Ensure the schema compiler is executable.
-                        var _ = RedirectedProcess.Command("chmod")
+                        RedirectedProcess.Command("chmod")
                             .WithArgs("+x", $"\"{schemaCompilerPath}\"")
                             .InDirectory(Path.GetFullPath(Path.Combine(Application.dataPath, "..")))
                             .Run();
--- a/workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
+++ b/workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
@@ -93,6 +93,10 @@ namespace Improbable.Gdk.Tools
         /// <summary>
         ///     Adds custom processing for regular output of process.
         /// </summary>
+        /// <remarks>
+        ///     The <see cref="outputProcessor"/> callback will be ran on a different thread to the one which
+        ///     registered it.
+        /// </remarks>
         /// <param name="outputProcessor">Processing action for regular output.</param>
         public RedirectedProcess AddOutputProcessing(Action<string> outputProcessor)
         {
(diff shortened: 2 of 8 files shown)


workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -105,7 +105,10 @@ private static void Generate()
                     case RuntimePlatform.LinuxEditor:
                     case RuntimePlatform.OSXEditor:
                         // Ensure the schema compiler is executable.
-                        var _ = RedirectedProcess.Command("chmod").WithArgs("+x", $"\"{schemaCompilerPath}\"").Run();
+                        var _ = RedirectedProcess.Command("chmod")
```
jared-improbable: Is the `var _` really needed?

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -126,117 +127,108 @@ public RedirectedProcess RedirectOutputOptions(OutputRedirectBehaviour redirectB
...
 
-                if (!process.Start())
+            using (process)
+            {
+                await Task.Run(() =>
                 {
-                    throw new Exception(
-                        $"Failed to run {info.FileName} {info.Arguments}");
-                }
+                    while (!process.HasExited)
```
jared-improbable: This polling approach is an anti-pattern.
Use the `Process.Exited` event and a `TaskCompletionSource` to signal completion, and the `CancellationToken`'s `Register` method to register a callback that kills the process on cancellation.

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -126,117 +127,108 @@ public RedirectedProcess RedirectOutputOptions(OutputRedirectBehaviour redirectB
...
+                }
 
-                    if (outputLog != null)
-                    {
-                        lock (outputLog)
-                        {
-                            outputLog.AppendLine(ProcessSpatialOutput(errorString));
-                        }
-                    }
+                var task = RunWithCancel(token.Value);
```
jared-improbable: Perhaps make the delegate `async` and use `await` here?

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -266,41 +258,121 @@ public int Run()
...
-                AddErrorProcessing(error =>
+                }
+
+                foreach (var outputProcessor in outputProcessors)
                 {
-                    lock (processStandardOutput)
+                    try
                     {
-                        processStandardError.Add(error);
+                        outputProcessor.Invoke(outputString);
```
jared-improbable: It should be noted in the doc-strings for registering processors that they'll be invoked on a different thread than the registering/main thread.

workers/unity/Packages/com.improbable.gdk.tools/GenerateCode.cs
jamiebrynes7: Probably not - will driveby it

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
jamiebrynes7: Yeah that's a good shout

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
jamiebrynes7: PTAL and double check my work 😄

workers/unity/Packages/com.improbable.gdk.tools/RedirectedProcess.cs
```diff
@@ -126,131 +133,96 @@ public RedirectedProcess RedirectOutputOptions(OutputRedirectBehaviour redirectB
...
+            var taskCompletionSource = new TaskCompletionSource<int>();
 
-                if (outputLog == null)
-                {
-                    return process.ExitCode;
-                }
+            // Register a handler to process the output when it exits.
+            process.Exited += (sender, args) =>
+            {
+                taskCompletionSource.SetResult(process.ExitCode);
```
jared-improbable: What happens in here when the token is cancelled and the process exits due to `Kill()`?

jamiebrynes7: The Exited event should fire and I would _hope_ you get a non-zero exit code. I can test this however.

I also realized this code may be race-y: 

- You can try to kill the process after it ends which will throw.
- The process could end before the `Exited` handler is set which may result in it never being thrown.

The former is easier to fix that the latter which will require some structural changes.

jared-improbable: The latter I'm not sure is an issue - may need a bit of research

jamiebrynes7: `process.Start()` was called in `SetupProcess()` which means it could immediately exit before we register the handler? Either way re-ordering these fixes the issue

jamiebrynes7:
We can extend it in the future with some sorts of functionality goodies.

PR #892, opened 2019-04-02
Expand documentation for the transform synchronization feature module

jamiebrynes7:
#### Description

Removed the previous documentation, we now have: 
- Installation instructions
- How it works and why we want it
- Setting up transform synchronization
- Documentation for each strategy

#### Tests
`improbadoc serve` 

#### Documentation
Yes

#### Primary reviewers


docs/modules/core-and-feature-module-overview.md
```diff
@@ -54,24 +54,9 @@ Call this during [entity template creation]({{urlRoot}}/reference/concepts/entit
...
-Call this to implement the transform synchronization module, adding all the necessary systems to all workers.
-
-* `AddTransformSynchronizationComponents(entityTemplate, writeAccess, location = default(Vector3), velocity = default(Vector3))` - in the repository [here](https://github.com/spatialos/gdk-for-unity/tree/master/workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformSynchronizationHelper.cs).<br/>
-Call this to add the SpatialOS components used by the transform synchronization module to your SpatialOS entity template. <br/>
-You can optionally pass in a `rotation`, `location` or `velocity`.
-
-Call this during [entity templates creation]({{urlRoot}}/reference/concepts/entity-templates).
-
-Find out more in the [Transform synchronization feature module]({{urlRoot}}/modules/transform-feature-module) documentation.
+This feature module contains functionality that will automatically synchronize your entities' transform. See our [Transform Synchronization documentation]({{urlRoot}}/modules/transform-sync/overview) for more details on installation and usage.
```
jessicafalk:
```suggestion
This feature module contains functionality that automatically synchronizes your entities' transform component. See our [Transform Synchronization documentation]({{urlRoot}}/modules/transform-sync/overview) for more details on installation and usage.
```

docs/modules/transform-sync/how-it-works.md
```diff
@@ -0,0 +1,37 @@
+# How the Transform Synchronization Feature Module works
+
+## What is Transform
+
+There are two concepts of the location of a SpatialOS entity:
```
jessicafalk: probably needs a note that this is not the same as the Transform component offered by Unity, though it does do some conversion between unity and spatialos

docs/modules/transform-sync/how-it-works.md
```diff
@@ -0,0 +1,37 @@
...
+
+* The _transform_. This contains a location, rotation, and physics tick and is used by worker-instances to run physics simulations or render entities for a client.
+
+* The _position_. This only contains a location. This is defined by the standard library component `Improbable.Position` and is used by the SpatialOS load balancer to determine which worker should be delegated authority over a given entity-component.
+
+<%(#Expandable title="Why do we need two locations?")%>
```
jessicafalk: this is confusing: we use location for both
1. the position
2. as the two components which both contain a location as well

we definitely should not name them the same

docs/modules/transform-sync/how-it-works.md
```diff
@@ -0,0 +1,37 @@
...
+3. **Atomicity.** The `Transform` component contains more than just the location. It also contains the rotation and physics tick. If these were on a separate component, you lose the guarantee that all three (location, rotation, physics tick) are update atomically.<br/><br/>Two updates sent in the same frame from one worker are _not_ guaranteed to be received on the same frame in another.
+<%(/Expandable)%>
+
+## How do my entities' transform get synchronized
+
+The behaviour of the Transform Synchronization Feature Modules differs depending on whether your worker-instance is authoritative over the `Transform` _and_ `Position` components. This behaviour can be adjusted with [transform synchronization strategies]({{urlRoot}}/modules/transform-sync/strategies), but broadly goes as follows:
```
jessicafalk: we don't seem to explain anything here?

jamiebrynes7: The subsections explain it. The expandable should probably be moved below.

docs/modules/transform-sync/how-it-works.md
```diff
@@ -0,0 +1,63 @@
+# How the Transform Synchronization Feature Module works
+
+## What is transform
+
+The transform of an entity describes its location and rotation. Worker-instances generally use it to run physics simulations or render entities for a client. Note that the SpatialOS representation of transform is _not_ the same as Unity's representation of transform.
+
+In the Transform Synchronization Feature Module, we represent the Transform as:
+
+```
+package improbable.transform;
```
samiwh: These are implementation details that might change at any time and users aren't supposed to know.

docs/modules/transform-sync/how-it-works.md
```diff
@@ -0,0 +1,63 @@
+# How the Transform Synchronization Feature Module works
+
+## What is transform
+
+The transform of an entity describes its location and rotation. Worker-instances generally use it to run physics simulations or render entities for a client. Note that the SpatialOS representation of transform is _not_ the same as Unity's representation of transform.
```
samiwh: Orientation is more accurate.
Also possibly velocity.

docs/modules/transform-sync/how-it-works.md
samiwh: In fact this is only what it was in the public branch at release. My internal branch had a different representation that was intended to be released later.

jamiebrynes7: Its not like its actually hidden in any way

docs/modules/transform-sync/how-it-works.md
jamiebrynes7: The first sentence refers to transform as a concept, I guess the idea was to present it in a way that's familiar to Unity users - i.e. - the Unity transform component has position and rotation immediately available - then move into how we treat transform 

How would you structure this section?

docs/modules/transform-sync/how-it-works.md
```diff
@@ -0,0 +1,63 @@
...
+
+<%(#Expandable title="Why not reuse the <code>Improbable.Position</code> SpatialOS component?")%>
+There are a few reasons why you might want different `Transform` and `Position` components.
+
+1. **Separation of responsibilies.** The `Improbable.Position` component is the load balancer's representation of location and the `TransformInternal` component is the workers' representation of location. This allows you to abstract the concept of location from the load balancer. A gameplay programmer would only need to be aware of the `TransformInternal` component without having to know about the load balancer at all.
+2. **Save bandwidth.** The `Position` component represents location with 3 doubles. The `TransformInternal` component represents location as 3 floats, which is the Unity native representation. This means that, ignoring rotation, a position update is twice as large as a transform update.<br/><br/>If the `TransformInternal` component has a high frequency update rate and the `Position` component has a low frequency update rate then you have a net bandwidth saving over updating just the `Position` component at a high frequency!
```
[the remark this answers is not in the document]
jamiebrynes7: Hmmm this was done intentionally to split it up otherwise its a big blob of text. It does look a little odd though 😕

docs/modules/transform-sync/setup-transform.md
```diff
@@ -0,0 +1,74 @@
...
+When authority is re-gained the Rigidbody returns to the state it was in before.
+<%(/Expandable)%>
+
+## Add transform strategies to the `TransformSynchronization` MonoBehaviour
+
+The behaviour of the Transform Synchronization Feature Module is dictated by which strategies are attached to the `TransformSynchronization` MonoBehaviour. Each strategy can be classified as either a "receive strategy" or a "send strategy". See our complete documentation on the [transform strategies]({{urlRoot}}/modules/transform-sync/strategies) to learn what each of them do.
+
+You may need to create instances of these strategies as they are scriptable objects. Select **Assets** > **Create** > **SpatialOS** > **Transforms** to see the available strategies.
+
+### Recommended configuration
```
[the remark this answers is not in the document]
jamiebrynes7: I'm slightly reluctant given that you have to create the scriptable objects yourself. 

My worry is if we do that users will immediately try to copy the configuration to find that they don't have the scriptable objects created yet.

docs/modules/transform-sync/how-it-works.md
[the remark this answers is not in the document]
paulbalaji: Surely the big blob would still be roughly the same size as the first point?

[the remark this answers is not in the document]
jamiebrynes7: That may be the case now - will change

docs/modules/transform-sync/how-it-works.md
samiwh: I doesn't matter that much. If we think that is easier to understand for unity users then it's better.

docs/modules/transform-sync/how-it-works.md
samiwh: It does mean that this will likely go out of sync. Also eventually it will be incomprehensible to someone who doesn't know how it works already.

jamiebrynes7: It does increase the likelihood of that (until we can include or similar)

PR #897, opened 2019-04-05
Refresh code generator section

jamiebrynes7:
#### Description
Removed code samples that: weren't particularly useful, and likely to go stale. Removed a lot of wordiness.

At this point, I kind of doubt the usefulness of this page outside of the type mappings.

#### Tests
Local improbadoc serve.

### DIFF
--- a/docs/reference/concepts/code-generation.md
+++ b/docs/reference/concepts/code-generation.md
@@ -3,7 +3,7 @@
 
 The code generator uses `.schema` files to generate C# code that you can use to interact with SpatialOS in Unity. See the [schemalang reference](https://docs.improbable.io/reference/latest/shared/schema/introduction#schema-introduction) for details on how to create schema components.
 
-The code generator runs when you:
+The code generator runs when you do one of the following:
 
 * Open your Unity Editor
 * Select **SpatialOS** > **Generate Code** from the Unity Editor menu
--- a/docs/reference/concepts/code-generation.md
+++ b/docs/reference/concepts/code-generation.md
@@ -35,7 +35,7 @@ The code generator also maps [schema collection types](https://docs.improbable.i
 
 ## User defined types
 
-The code generator generates a C# struct for each [user defined type in schema](https://docs.improbable.io/reference/latest/shared/schema/reference#user-defined-types). The generated struct is annotated with the [`System.Serializable` attribute](https://docs.unity3d.com/ScriptReference/Serializable.html). The struct has a constructor with a parameter per schema field.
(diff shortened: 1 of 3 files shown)


docs/reference/concepts/code-generation.md
```diff
@@ -1,69 +1,19 @@
...
-# Code generation
- _This document relates to both the [MonoBehaviour and ECS workflows]({{urlRoot}}/reference/workflows/which-workflow)._
+<%(TOC max="2")%>
+# Code generator
 
-> Note this document refers to ECS components but it is also relevant if you are using the MonoBehaviour workflow. You don't need to know about ECS to use component generation.
+The code generator uses `.schema` files to generate C# code that you can use to interact with SpatialOS in Unity. See the [schemalang reference](https://docs.improbable.io/reference/latest/shared/schema/introduction#schema-introduction) for details on how to create schema components.
 
-The code generator uses `.schema` files to generate components that the Unity ECS can understand. See the [schemalang guide](https://docs.improbable.io/reference/latest/shared/schema/introduction#schema-introduction) for details on how to create schema components.
+The code generator runs when you:
```
paulbalaji: might just be me, but this reads as though it runs only when you do all of the following

jamiebrynes7: Can make it more explicit: `... runs when you do one of the following:`

docs/reference/concepts/code-generation.md
```diff
@@ -75,66 +25,38 @@ Each primitive type in schemalang corresponds to a type in the SpatialOS GDK for
...
 | --------------------- | :-----------------------------------------------: |
 | `map<K, V>`           | `System.Collections.Generic.Dictionary<K, V>`     |
 | `list<T>`             | `System.Collections.Generic.List<T>`              |
-| `option<T>`           | `Improbable.Gdk.Core.Option<T>`                              |
+| `option<T>`           | `Improbable.Gdk.Core.Option<T>`                   |
 
+## User defined types
 
-### Custom types
+The code generator generates a C# struct for each [user defined type in schema](https://docs.improbable.io/reference/latest/shared/schema/reference#user-defined-types). The generated struct is annotated with the [`System.Serializable` attribute](https://docs.unity3d.com/ScriptReference/Serializable.html). The struct has a constructor with a parameter per schema field.
```
jessicafalk:
```suggestion
The code generator generates a C# struct for each [user defined type in schema](https://docs.improbable.io/reference/latest/shared/schema/reference#user-defined-types). The generated struct is annotated with the [`System.Serializable` attribute](https://docs.unity3d.com/ScriptReference/Serializable.html) and has a constructor with a parameter per schema field.
```

just don't like to start two sentences with "the struct ..."

PR #899, opened 2019-04-05
Upgrade Playground API

zeroZshadow:
#### Description
Added the ability to NOT compile reactive components, using the `DISABLE_REACTIVE_COMPONENTS` scriptable define symbol.
The playground no longer uses reactive components.

#### Tests
Playground remains equal in functionality with our without the define.
FPS Breaks when using the symbol, as it still uses reactive components.

#### Documentation
Added a changelog


workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerSystem.cs
```diff
@@ -94,7 +94,6 @@ protected override void OnCreateManager()
             base.OnCreateManager();
             var entityManager = World.GetOrCreateManager<EntityManager>();
             WorkerEntity = entityManager.CreateEntity(typeof(OnConnected), typeof(WorkerEntityTag));
-            Debug.Log("WORKER ENTITY CREATED");
```
jamiebrynes7: 💯

PR #903, opened 2019-04-10
Fixing name clashes between packages and component / type names

jessicafalk:
#### Description
e.g. 
```
package name;

component Name
{
    id = 200;
    float name = 1;
}
```

would lead to the code generator generating invalid code. Added `global::` to all the namespaces to make sure it knows which namespace to use.

This fixes most of the cases, however if we have a component and an event / command with the same name, this is not fixed. This will require fixes in a different place and will be done at a later point.

#### Tests
added test schema to the test project and tested it locally.

#### Documentation
no need

#### Primary reviewers
@jamiebrynes7

### DIFF
--- a/test-project/Assets/.Schema/improbable/gdk/tests/same_names.schema
+++ b/test-project/Assets/.Schema/improbable/gdk/tests/same_names.schema
@@ -1,0 +1,7 @@
+package name;
+
+component Name
+{
+    id = 200;
+    float name = 1;
+}
--- a/test-project/Assets/.Schema/improbable/gdk/tests/same_names.schema
+++ b/test-project/Assets/.Schema/improbable/gdk/tests/same_names.schema
@@ -1,6 +1,6 @@
 package name;
 
-type Name {
+type Test {
     float color = 1;
 }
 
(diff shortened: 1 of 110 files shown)


test-project/Assets/.Schema/improbable/gdk/tests/same_names.schema
```diff
@@ -0,0 +1,7 @@
+package name;
+
+component Name
+{
+    id = 200;
+    float name = 1;
+}
```
jamiebrynes7: Nit: new line

workers/unity/Packages/com.improbable.gdk.tools/.CodeGenerator/GdkCodeGenerator/Templates/UnityReferenceTypeProviderGenerator.tt
```diff
@@ -7,7 +7,7 @@
     var commandDetailsList = GetCommandDetailsList();
     var generatedHeader = CommonGeneratorUtils.GetGeneratedHeader();
     var generator = new UnityReferenceTypeProviderContent();
-    var componentNamespace = qualifiedNamespace + "." + componentDetails.ComponentName;
+    var componentNamespace = "global::" + qualifiedNamespace + "." + componentDetails.ComponentName;
```
zeroZshadow: Could this use string interpolation?

Issue #780, opened 2019-02-22
Add back support for win x86


judah4:
The SDK supported windows x86 builds but the current GDK does not without some modifications. To fix it the package com.improbable.gdk.tools needs some modifications.

In `.DownloadCoreSdk`, the program.cs imports needs this line added to pull the right sdk version for x86 machines.
```CSharp
new Package(tempPath, "worker_sdk", "core-dynamic-x86-win32", $"{nativeDependenciesPath}/Windows/x86", new List<string> {"include", "CoreSdkDll.lib"}),
```

Alongside that, `Plugins/PluginPostprocessor.cs` needs to be updated to not use the x86 dll in the editor and instead keep using the x86_64 one.
```CSharp
new PluginCompatibilitySetting(
                PluginType.Core,
                BuildTarget.StandaloneWindows,
                CpuType.X86,
                compatibleWithEditor: false
                ),
```

With this change the x86 windows client will not crash and work as intended! I'd do a pull request myself if allowed.

jamiebrynes7:
The PR for this has been merged and will be released in `0.2.2`! 🎉

PR #909, opened 2019-04-12
Add support for win x86 builds

paulbalaji:
#### Description
Adding support for win x86 builds, as requested by `judah4` https://github.com/spatialos/gdk-for-unity/issues/780.

also drive-by `pragma` additions to reduce warnings in editor

#### Tests
- [x] editor unaffected
- [x] local x86 client works
- [x] cloud x86 client works

#### Documentation
added changelog entry

#### Primary reviewers


workers/unity/Assets/Playground/Config/BuildConfiguration.asset
```diff
@@ -30,7 +30,7 @@ MonoBehaviour:
     CloudBuildConfig:
       BuildTargets:
       - Options: 0
-        Enabled: 0
+        Enabled: 1
```
jamiebrynes7: QQ when you build a windows x86 client is it still zipped into `WorkerType@Windows.zip` ? Or will be it something like `WorkerType@Windows86.zip` ?

paulbalaji: x86 and x64 clients share the same zip name `WorkerType@Windows.zip`

need to think about how to best ensure we test the x64 clients in release-qa

paulbalaji: it now will be `WorkerType@Windows86.zip`

PR #922, opened 2019-04-30
Load dev token asset by default

jessicafalk:
#### Description
Instead of having to enter the dev auth token into an inspector field, we should load an asset when needed.
Need to update the other projects as well.

#### Tests
tested locally and in cloud.

#### Documentation
will be updated soonish

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
     - Pulled out the `UiStateManager` from the `BuildConfigEditor` into this assembly and made it public.
 - Exceptions thrown in user-code callbacks no longer cause other callbacks scheduled for that frame to not fire. Instead, the exceptions are caught and logged with Debug.LogException.
 - Upgraded the Worker SDK version to `13.7.1`.
-- The Development Authentication Token will by default be loaded searching for a `DevAuthToken.txt` asset in any of your Resources folder.
+- Updated the default method of loading a Development Authentication Token to search for a `DevAuthToken.txt` asset at the root of any `Resources` folder.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,12 @@
     - Pulled out the `UiStateManager` from the `BuildConfigEditor` into this assembly and made it public.
 - Exceptions thrown in user-code callbacks no longer cause other callbacks scheduled for that frame to not fire. Instead, the exceptions are caught and logged with Debug.LogException.
 - Upgraded the Worker SDK version to `13.7.1`.
+- The Development Authentication Token will by default be loaded searching for a `DevAuthToken.txt` asset in any of your Resources folder.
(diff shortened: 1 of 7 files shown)


workers/unity/Packages/com.improbable.gdk.mobile/Worker/MobileWorkerConnector.cs
```diff
@@ -50,6 +50,7 @@ protected override LocatorConfig GetLocatorConfig()
 
         protected override AlphaLocatorConfig GetAlphaLocatorConfig(string workerType)
         {
+            LoadDevAuthToken();
```
jamiebrynes7: It feels strange to _need_ to call this but it doesn't return anything. I'd rather it does something like: 

```
var token = LoadDevAuthToken();
var pit = GetDevelopmentPlayerIdentitityToken(token, ...);
...
```

Also we should actually throw an error/exception if the asset _doesn't_ exist which is an invariant of the method.

---

If a user wants to load the dev auth token another way, they can simply write their own method and call that instead.

jamiebrynes7: Especially as we have the pattern of 

```
var thing = GetThing();
```

for the rest of the flow, these could equally be class variables, but (rightly) aren't

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -222,6 +220,22 @@ protected virtual string SelectDeploymentName(DeploymentList deployments)
...
+        protected virtual string GetDevAuthToken()
+        {
+            var textAsset = Resources.Load<TextAsset>("DevAuthToken");
+            if (textAsset != null)
+            {
+                return textAsset.text.Trim();
+            }
+            else
+            {
+                throw new MissingReferenceException("Unable to find DevAuthToken.txt in the Resources folder.");
```
jamiebrynes7: Maybe tell them how to fix it?

CHANGELOG.md
```diff
@@ -14,10 +14,12 @@
     - Pulled out the `UiStateManager` from the `BuildConfigEditor` into this assembly and made it public.
 - Exceptions thrown in user-code callbacks no longer cause other callbacks scheduled for that frame to not fire. Instead, the exceptions are caught and logged with Debug.LogException.
 - Upgraded the Worker SDK version to `13.7.1`.
+- The Development Authentication Token will by default be loaded searching for a `DevAuthToken.txt` asset in any of your Resources folder.
```
paulbalaji: this could probably be clearer to be something along the lines of: 
```suggestion
- Updated the default method of loading a Development Authentication Token to search for a `DevAuthToken.txt` asset at the root of any `Resources` folder.
```

PR #952, opened 2019-05-20
Fixed world being disposed multiple times

jessicafalk:
#### Description
fix for [UTY-1870](https://improbableio.atlassian.net/browse/UTY-1870)

Unity seems to dispose the World as well whenever you stop your editor application, so we need to check first before we dispose of the World to ensure we are not trying to dispose it multiple times.

Also did a quick rename of a parameter to ensure that code and XML documentation are consistent.

#### Tests
ran it in the editor, now it only throws the `Editor application stopped` error instead of a `World is already disposed' and a connection error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
-- Fixed a bug where the `World` would get disposed multiple times, if you stopped the application inside the Editor while the worker still gets created.
+- Fixed a bug where a worker's `World` could get disposed multiple times if you stopped the application inside the Editor while the worker is being created.
 
 ## `0.2.2` - 2019-05-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Upgraded the project to be compatible with `2019.1.3f1`.
 
+### Fixed
+
+- Fixed a bug where the `World` would get disposed multiple times, if you stopped the application inside the Editor while the worker still gets created.
+
 ## `0.2.2` - 2019-05-15
 
 ### Breaking Changes
--- a/workers/unity/Packages/com.improbable.gdk.core/Utility/TaskUtility.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Utility/TaskUtility.cs
@@ -1,0 +1,23 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Improbable.Gdk.Core
+{
+    public static class TaskUtility
+    {
+        public static async Task<T> WithCancellation<T>(this Task<T> task, CancellationToken cancellationToken)
+        {
+            var tcs = new TaskCompletionSource<bool>();
+            using (cancellationToken.Register(s => ((TaskCompletionSource<bool>) s).TrySetResult(true), tcs))
+            {
+                if (task != await Task.WhenAny(task, tcs.Task))
+                {
+                    throw new OperationCanceledException(cancellationToken);
+                }
+            }
+
+            return task.Result;
+        }
+    }
+}
--- a/workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
+++ b/workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
@@ -121,7 +121,7 @@ namespace Improbable.Gdk.Core
         ///     Connects to the SpatialOS Runtime via the Receptionist service and creates a <see cref="Worker" /> object
         ///     asynchronously.
         /// </summary>
-        /// <param name="config">
+        /// <param name="parameters">
         ///     The <see cref="ReceptionistConfig" /> object stores the configuration needed to connect via the
         ///     Receptionist Service.
         /// </param>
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- Fixed a bug where the `World` would get disposed multiple times, if you stopped the application inside the Editor while the worker still gets created.
```
```suggestion
- Fixed a bug where a worker's `World` could get disposed multiple times if you stopped the application inside the Editor while the worker is being created.
```

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -268,7 +268,11 @@ private void AddCoreSystems()
 
         public void Dispose()
         {
-            World?.Dispose();
+            if (World != null && World.IsCreated)
+            {
+                World?.Dispose();
```
jamiebrynes7: No need for the null conditional operator since we explicitly do the null check 😄

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -118,6 +118,8 @@ await Worker.CreateWorkerAsync(GetAlphaLocatorConfig(workerType), connectionPara
                 if (!Application.isPlaying)
                 {
                     Dispose();
+                    throw new ConnectionFailedException("Editor application stopped",
```
jamiebrynes7: What happens to the task that's being await'ed after you stop in the Editor? Is it cancelled or does it persist?

jessicafalk: as soon as you throw an exception, it goes into the catch block and prints an error. So it's not cancelled, but also only logs the error and doesn't do anything else anymore.

only if the application is still playing will it call `HandleWorkerConnectionFailure`,

jamiebrynes7: The task has finished by the time we get here as this is part of the continuation of 

```
 Worker = await ConnectWithRetries(connectionDelegate, MaxConnectionAttempts, logger, workerType);
```

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -33,6 +34,8 @@ public abstract class WorkerConnector : MonoBehaviour, IDisposable
 
         private List<Action<Worker>> workerConnectedCallbacks = new List<Action<Worker>>();
 
+        private CancellationTokenSource tokenSource = new CancellationTokenSource();
```
jamiebrynes7: Make sure to dispose this! 😄

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -54,6 +57,8 @@ public abstract class WorkerConnector : MonoBehaviour, IDisposable
         // Important run in this step as otherwise it can interfere with the the domain unloading logic.
         protected void OnApplicationQuit()
         {
+            Debug.Log("cancelling");
```
jamiebrynes7: Debug log

workers/unity/Packages/com.improbable.gdk.core/Utility/TaskUtility.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static class TaskUtility
+    {
+        public static async Task<T> WithCancellation<T>(this Task<T> task, CancellationToken cancellationToken)
+        {
+            var tcs = new TaskCompletionSource<bool>();
+            using (cancellationToken.Register(s => ((TaskCompletionSource<bool>) s).TrySetResult(true), tcs))
+            {
+                if (task != await Task.WhenAny(task, tcs.Task))
+                {
+                    throw new OperationCanceledException(cancellationToken);
```
jamiebrynes7: We don't seem to catch this anywhere?

jessicafalk: it's getting caught in the `WorkerConnector`. Not specifically this exception, but it will just be propagated up to fail the whole worker creation

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -163,7 +168,9 @@ private Worker(string workerType, Connection connection, ILogDispatcher logDispa
         public static async Task<Worker> CreateWorkerAsync(
             LocatorConfig parameters,
             ConnectionParameters connectionParameters,
-            ILogDispatcher logger, Vector3 origin)
+            ILogDispatcher logger,
+            Vector3 origin,
+            CancellationToken token)
```
jamiebrynes7: Technical this is a breaking change as this is part of the public API.. add a small note to the Changelog about this 😄

PR #961, opened 2019-05-29
Move runtime IP to EditorPref / DevAuthToken to PlayerPref

paulbalaji:
#### Description

- move runtime ip to an editor pref
- move DAT to player pref
	- [this ties a DAT to a company/project_name](https://docs.unity3d.com/ScriptReference/PlayerPrefs.html)
	- only create DAT.txt if enabled in GdkToolsConfiguration
	- pass in DAT as argument when starting android client for cloud
	- if DAT argument passed in, player pref is set
	- if no DAT argument exists / previously set as player pref - read from DAT.txt and save to player prefs

#### Tests

- local launch works fine
- android cloud launch via editor works fine
- no DAT.txt generated unless specifically requested

#### Documentation

- [x] changelog
- [ ] update screenshots/docs where relevant

#### Primary reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@
 ### Changed
 
 - Upgraded the project to be compatible with `2019.1.3f1`.
+- Moved Runtime IP from the `GdkToolsConfiguration.json` to an Editor Preference.
+- Moved Dev Auth Token to a PlayerPref.
+    - Added a setting in `GdkToolsConfiguration` to let users configure whether a DevAuthToken.txt should be generated or not.
+    - When launching Android cloud clients from the Editor, the DevAuthToken is now passed in as a command line argument.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,9 @@
 ### Changed
 
 - Upgraded the project to be compatible with `2019.1.3f1`.
-- Moved Runtime IP from the `GdkToolsConfiguration.json` to an Editor Preference.
-- Moved Dev Auth Token to a PlayerPref.
-    - Added a setting in `GdkToolsConfiguration` to let users configure whether a DevAuthToken.txt should be generated or not.
+- Moved Runtime IP from the `GdkToolsConfiguration.json` to the Editor Preferences.
+- Moved Dev Auth Token to the Player Preferences.
+    - Added a setting in `GdkToolsConfiguration` to let users configure whether a `DevAuthToken.txt` should be generated or not.
     - When launching Android cloud clients from the Editor, the DevAuthToken is now passed in as a command line argument.
 
 ### Fixed
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -16,7 +16,6 @@ namespace Improbable.Gdk.Mobile
         private const string MenuLaunchAndroidLocal = "/Android for local";
         private const string MenuLaunchAndroidCloud = "/Android for cloud";
 
-
         [MenuItem(MenuLaunchMobile + MenuLaunchAndroidLocal, false, 73)]
         private static void LaunchAndroidLocal()
         {
@@ -94,14 +93,14 @@ namespace Improbable.Gdk.Mobile
                     var gdkToolsConfig = GdkToolsConfiguration.GetOrCreateInstance();
 
                     // Return error if no DevAuthToken is set AND fails to generate new DevAuthToken.
-                    if (!EditorPrefs.HasKey(RuntimeConfigNames.DevAuthTokenKey) && !DevAuthTokenUtils.Generate())
+                    if (!PlayerPrefs.HasKey(RuntimeConfigNames.DevAuthTokenKey) && !DevAuthTokenUtils.Generate())
                     {
                         Debug.LogError("Failed to generate a Dev Auth Token to launch mobile client.");
                         return;
                     }
 
-                    // Returns error if DevAuthToken.txt does not exists AND fails to save new DevAuthToken.txt.
-                    // Ignores if GdkToolsConfiguration does not require DevAuthToken to be saved as text file
+                    // Prints a warning if DevAuthToken.txt does not exist, if GdkToolsConfiguration is configured
+                    // to save a DAT to file.
                     if (gdkToolsConfig.SaveDevAuthTokenToFile && !File.Exists(gdkToolsConfig.DevAuthTokenFilepath))
                     {
                         Debug.LogWarning("Launching Android client without a DevAuthToken.txt asset.");
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -99,13 +99,6 @@ namespace Improbable.Gdk.Mobile
                         return;
                     }
 
-                    // Prints a warning if DevAuthToken.txt does not exist, if GdkToolsConfiguration is configured
-                    // to save a DAT to file.
-                    if (gdkToolsConfig.SaveDevAuthTokenToFile && !File.Exists(gdkToolsConfig.DevAuthTokenFilepath))
-                    {
-                        Debug.LogWarning("Launching Android client without a DevAuthToken.txt asset.");
-                    }
-
                     arguments.Append($"+{RuntimeConfigNames.DevAuthTokenKey} {DevAuthTokenUtils.DevAuthToken} ");
                 }
 
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -90,6 +90,26 @@ namespace Improbable.Gdk.Mobile
                 else
                 {
                     arguments.Append($"+{RuntimeConfigNames.Environment} {RuntimeConfigDefaults.CloudEnvironment} ");
+
+                    var gdkToolsConfig = GdkToolsConfiguration.GetOrCreateInstance();
+
+                    // Return error if no DevAuthToken is set AND fails to generate new DevAuthToken.
+                    if (!EditorPrefs.HasKey(RuntimeConfigNames.DevAuthTokenKey) && !DevAuthTokenUtils.Generate())
+                    {
+                        Debug.LogError("Failed to generate a Dev Auth Token to launch mobile client.");
+                        return;
+                    }
+
+                    // Returns error if DevAuthToken.txt does not exists AND fails to save new DevAuthToken.txt.
+                    // Ignores if GdkToolsConfiguration does not require DevAuthToken to be saved as text file
+                    if (gdkToolsConfig.SaveDevAuthTokenToFile &&
+                        (!File.Exists(gdkToolsConfig.DevAuthTokenFilepath) || !DevAuthTokenUtils.SaveTokenToFile()))
+                    {
+                        Debug.LogError("Failed to save DevAuthToken.txt asset.");
+                        return;
+                    }
+
+                    arguments.Append($"+{RuntimeConfigNames.DevAuthTokenKey} {DevAuthTokenUtils.DevAuthToken}");
                 }
 
                 // Get chosen android package id and launch
--- a/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
+++ b/workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
@@ -102,11 +102,9 @@ namespace Improbable.Gdk.Mobile
 
                     // Returns error if DevAuthToken.txt does not exists AND fails to save new DevAuthToken.txt.
                     // Ignores if GdkToolsConfiguration does not require DevAuthToken to be saved as text file
-                    if (gdkToolsConfig.SaveDevAuthTokenToFile &&
-                        (!File.Exists(gdkToolsConfig.DevAuthTokenFilepath) || !DevAuthTokenUtils.SaveTokenToFile()))
+                    if (gdkToolsConfig.SaveDevAuthTokenToFile && !File.Exists(gdkToolsConfig.DevAuthTokenFilepath))
                     {
-                        Debug.LogError("Failed to save DevAuthToken.txt asset.");
-                        return;
+                        Debug.LogWarning("Launching Android client without a DevAuthToken.txt asset.");
                     }
 
                     arguments.Append($"+{RuntimeConfigNames.DevAuthTokenKey} {DevAuthTokenUtils.DevAuthToken}");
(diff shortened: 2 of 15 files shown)


workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -90,6 +90,26 @@ private static void LaunchAndroid(bool shouldConnectLocally)
...
+                    if (!EditorPrefs.HasKey(RuntimeConfigNames.DevAuthTokenKey) && !DevAuthTokenUtils.Generate())
+                    {
+                        Debug.LogError("Failed to generate a Dev Auth Token to launch mobile client.");
+                        return;
+                    }
+
+                    // Returns error if DevAuthToken.txt does not exists AND fails to save new DevAuthToken.txt.
+                    // Ignores if GdkToolsConfiguration does not require DevAuthToken to be saved as text file
+                    if (gdkToolsConfig.SaveDevAuthTokenToFile &&
+                        (!File.Exists(gdkToolsConfig.DevAuthTokenFilepath) || !DevAuthTokenUtils.SaveTokenToFile()))
```
paulbalaji: by this point it's too late to save the file

remove the .txt creation but include the file check + print a warning

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -33,18 +43,48 @@ private static void Generate()
...
             }
             catch (Exception e)
             {
                 Debug.LogError($"Unable to generate Dev Auth Token. {e.Message}");
-                return;
+                return false;
             }
 
+            Debug.Log($"Saving token {devAuthToken} to Editor Preferences.");
+            EditorPrefs.SetString(DevAuthTokenKey, devAuthToken);
```
paulbalaji: this is across all projects - so need to think about how to identify the right one for the project

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -90,6 +89,24 @@ private static void LaunchAndroid(bool shouldConnectLocally)
...
+                    }
+
+                    // Prints a warning if DevAuthToken.txt does not exist, if GdkToolsConfiguration is configured
+                    // to save a DAT to file.
+                    if (gdkToolsConfig.SaveDevAuthTokenToFile && !File.Exists(gdkToolsConfig.DevAuthTokenFilepath))
+                    {
+                        Debug.LogWarning("Launching Android client without a DevAuthToken.txt asset.");
+                    }
+
+                    arguments.Append($"+{RuntimeConfigNames.DevAuthTokenKey} {DevAuthTokenUtils.DevAuthToken}");
```
jessicafalk: that needs an empty space at the end in case we want to append more arguments later on and ensure that everything still works

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -9,16 +9,26 @@ namespace Improbable.Gdk.Tools
 {
     public static class DevAuthTokenUtils
     {
+        public static string DevAuthToken => PlayerPrefs.GetString(DevAuthTokenKey);
```
jessicafalk: do we want to store them in the PlayerPrefs or the EditorPrefs (I genuinely don't know whether that actually amkes a difference)

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -9,16 +9,26 @@ namespace Improbable.Gdk.Tools
...
+
+        private static string DevAuthTokenAssetPath =>
+            Path.Combine("Assets", GdkToolsConfiguration.GetOrCreateInstance().DevAuthTokenDir, "DevAuthToken.txt");
```
jessicafalk: can also use `Application.dataPath` instead of hardcoding it to "Assets"

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -9,16 +9,26 @@ namespace Improbable.Gdk.Tools
...
+
         private static readonly string JsonDataKey = "json_data";
+        private static readonly string ErrorKey = "error";
         private static readonly string TokenSecretKey = "token_secret";
+        private static readonly string DevAuthTokenKey = "devAuthTokenSecret";
```
jessicafalk: maybe prepend the key names to indicate whether they are used to get a value from json or put it into player preferences

jessicafalk: why not use the key that you already defined in the `RuntimeConfig` class

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -9,16 +9,26 @@ namespace Improbable.Gdk.Tools
...
 
-        [MenuItem("SpatialOS/Generate Dev Authentication Token", false, MenuPriorities.GenerateDevAuthToken)]
-        private static void Generate()
+        private const string DevAuthMenuPrefix = "SpatialOS/Dev Authentication Token";
+        private const string DevAuthMenuGenerateToken = "/Generate Token";
+        private const string DevAuthMenuClearToken = "/Clear Token";
+
+        [MenuItem(DevAuthMenuPrefix + DevAuthMenuGenerateToken, false, MenuPriorities.GenerateDevAuthToken)]
+        public static bool Generate()
```
jessicafalk: rename to `TryGenerate` to indicate that we are returning a bool?

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -33,18 +43,48 @@ private static void Generate()
...
+                }
             }
             catch (Exception e)
             {
                 Debug.LogError($"Unable to generate Dev Auth Token. {e.Message}");
-                return;
+                return false;
             }
 
+            Debug.Log($"Saving token {devAuthToken} to Editor Preferences.");
```
jessicafalk: on second thought: let's not print out the token secret in case people share editor logs with each other. Just say "Saving token to ..."
also, you are saving it to the Player Preferences, not the Editor preferences

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -33,18 +43,48 @@ private static void Generate()
...
+            PlayerPrefs.SetString(DevAuthTokenKey, devAuthToken);
+
+            return !gdkToolsConfiguration.SaveDevAuthTokenToFile || SaveTokenToFile();
```
jessicafalk: nit: would personally prefer to have that written out as an if-statement for readability.

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -57,13 +97,21 @@ private static void Generate()
...
                 Debug.LogError($"Unable to save Dev Auth Token asset. {e.Message}");
-                return;
+                return false;
             }
 
             Debug.Log($"Saving token {devAuthToken} to {devAuthTokenFilePath}.");
-            AssetDatabase.ImportAsset(
-                Path.Combine("Assets", gdkToolsConfiguration.DevAuthTokenDir, "DevAuthToken.txt"),
-                ImportAssetOptions.ForceUpdate);
+            AssetDatabase.ImportAsset(DevAuthTokenAssetPath, ImportAssetOptions.ForceUpdate);
```
jessicafalk: same comment here about the devauthtoken that should not be printed :)

workers/unity/Packages/com.improbable.gdk.tools/GdkToolsConfiguration.cs
```diff
@@ -13,9 +14,12 @@ public class GdkToolsConfiguration
         public List<string> SchemaSourceDirs = new List<string>();
         public string CodegenOutputDir;
         public string DescriptorOutputDir;
-        public string RuntimeIp;
         public string DevAuthTokenDir;
         public int DevAuthTokenLifetimeDays;
+        public bool SaveDevAuthTokenToFile;
+
+        internal string RuntimeIpEditorPrefKey = "RuntimeIp";
+        public string RuntimeIp => EditorPrefs.GetString(RuntimeIpEditorPrefKey);
```
jessicafalk: here we are using EditorPrefs instead of PlayerPrefs?

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
paulbalaji: good catch

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
paulbalaji: `AssetDatabase` requires we use relative paths - so I had to hardcode the "Assets"

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
paulbalaji: the tools package doesn't have a dependency on core so I had to define them separately

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
paulbalaji: EditorPrefs are for all projects with a specific version of the Unity editor - so if a user generated a DAT in project X on 2019.1, the DAT would exist for them in project Z on 2019.1

PlayerPrefs are tied to the company/product name of the project - so different projects would have different DATs (assuming the product names in the project are different)

workers/unity/Packages/com.improbable.gdk.tools/GdkToolsConfiguration.cs
paulbalaji: yes because we'd expect the RuntimeIp to be system wide and not per project

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
 ### Changed
 
 - Upgraded the project to be compatible with `2019.1.3f1`.
+- Moved Runtime IP from the `GdkToolsConfiguration.json` to an Editor Preference.
```
jessicafalk:
```suggestion
- Moved Runtime IP from the `GdkToolsConfiguration.json` to the Editor Preferences.
```

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
...
+- Moved Dev Auth Token to a PlayerPref.
```
jessicafalk:
```suggestion
- Moved Dev Auth Token to the Player Preferences.
```

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
...
+    - Added a setting in `GdkToolsConfiguration` to let users configure whether a DevAuthToken.txt should be generated or not.
```
jessicafalk:
```suggestion
    - Added a setting in `GdkToolsConfiguration` to let users configure whether a `DevAuthToken.txt` should be generated or not.
```

workers/unity/Packages/com.improbable.gdk.mobile/Editor/LaunchMenu.cs
```diff
@@ -90,6 +89,24 @@ private static void LaunchAndroid(bool shouldConnectLocally)
...
+                    {
+                        Debug.LogError("Failed to generate a Dev Auth Token to launch mobile client.");
+                        return;
+                    }
+
+                    // Prints a warning if DevAuthToken.txt does not exist, if GdkToolsConfiguration is configured
+                    // to save a DAT to file.
+                    if (gdkToolsConfig.SaveDevAuthTokenToFile && !File.Exists(gdkToolsConfig.DevAuthTokenFilepath))
+                    {
+                        Debug.LogWarning("Launching Android client without a DevAuthToken.txt asset.");
```
jessicafalk: would remove this check (or rephrase it quite a bit). cause the text file might have existed when the worker was built and for some reason deleted later. Additionally, we throw an error when starting the app, if we can't find the dev auth token. that should be enough

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -33,18 +43,53 @@ private static void Generate()
...
+        {
+            var gdkToolsConfiguration = GdkToolsConfiguration.GetOrCreateInstance();
+            var devAuthTokenFullDir = gdkToolsConfiguration.DevAuthTokenFullDir;
+            var devAuthTokenFilePath = gdkToolsConfiguration.DevAuthTokenFilepath;
+
+            if (!PlayerPrefs.HasKey(PlayerPrefDevAuthTokenKey))
+            {
+                // Given we call SaveTokenToFile after successfully generating a Dev Auth Token,
+                // we should never see the following warning.
+                Debug.LogWarning("Cannot save Development Authentication Token, as it has not been generated.");
```
jessicafalk: given that this should never happen, should that be an error?

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -33,18 +43,53 @@ private static void Generate()
...
+                var deserializedMessage = Json.Deserialize(receivedMessage);
+                if (deserializedMessage.TryGetValue(JsonDataKey, out var jsonData) &&
+                    ((Dictionary<string, object>) jsonData).TryGetValue(JsonTokenSecretKey, out var tokenSecret))
                 {
                     devAuthToken = (string) tokenSecret;
                 }
+                else
+                {
+                    throw new Exception(
+                        $@"{(deserializedMessage.TryGetValue(JsonErrorKey, out var errorMessage)
```
zeroZshadow: This looks like a bit overkill to all put in a string format, maybe pull the logic out?

workers/unity/Packages/com.improbable.gdk.tools/DevAuthTokenUtils.cs
```diff
@@ -33,18 +43,53 @@ private static void Generate()
...
+                }
             }
             catch (Exception e)
             {
                 Debug.LogError($"Unable to generate Dev Auth Token. {e.Message}");
-                return;
+                return false;
+            }
+
+            Debug.Log($"Saving token to Player Preferences.");
```
zeroZshadow: Is the log needed?

paulbalaji: given there's a couple ways of saving out a token (player prefs / .txt), I'd prefer to log where it's being saved

this way we'll be able to see if a user's actually created DAT.txt before complaining it doesn't exist

PR #962, opened 2019-05-31
Remove workaround for a schema component update bug (WRK-1031).

paulbalaji:
#### Description

Removed the workaround for a schema component update bug (WRK-1031).

consider reading 1st and 3rd commit (2nd commit is generated code)

includes drive-by deployment launcher fix for uploading assemblies

#### Tests

- [x] local launch
- [x] cloud launch
- [x] premerge
- [x] release qa

#### Documentation

internal changelog entry

#### Primary reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@
 ### Fixed
 
 - Fixed a bug where a worker's `World` could get disposed multiple times if you stopped the application inside the Editor while the worker is being created.
-- Fixed a bug when uploading assemblies with the deployment launcher window, by adding the `--enable_pre_upload_check=false` argument.
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@
 
 - Fixed a bug where a worker's `World` could get disposed multiple times if you stopped the application inside the Editor while the worker is being created.
 
+### Internal
+
+- Removed the workaround for a schema component update bug (WRK-1031).
+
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,11 @@
 ### Fixed
 
 - Fixed a bug where a worker's `World` could get disposed multiple times if you stopped the application inside the Editor while the worker is being created.
+- Fixed a bug when uploading assemblies with the deployment launcher window, by adding the `--enable_pre_upload_check=false` argument.
```
jamiebrynes7: What bug was this exactly?

jessicafalk: a bug that only exists in our develop branch afaik, so I don't think a changelog entry is necessary

workers/unity/Packages/com.improbable.gdk.deploymentlauncher/Commands/Assembly.cs
```diff
@@ -19,7 +19,8 @@ internal static class Assembly
                 config.AssemblyName,
                 "--project_name",
                 config.ProjectName,
-                "--json_output"
+                "--json_output",
+                "--enable_pre_upload_check=false"
```
jessicafalk: add a comma at the end so that you don't need to change this line if we ever need to update this

CHANGELOG.md
paulbalaji: will remove changelog entry - it's the issue where spatial `prepare for run` happens and breaks because our schema isn't all in one folder now

PR #976, opened 2019-06-11
API docs update for 0.2.3 (39867936)

paulbalaji:
Generated API docs taken from `3986793`, targeting `0.2.3` tag


paulbalaji:
will re-generate docs after a few in-code updates are made

PR #978, opened 2019-06-11
API docs update for 0.2.3 (generated from f029e1e8)

paulbalaji:


paulbalaji:
will regenerate once the RC branch actually gets merged into `develop`

PR #981, opened 2019-06-12
Worker abstraction rework

jamiebrynes7:
> Split up into commits for easier review. Not necessarily isolated commits however.

#### Description

Extending #844. This PR changes how workers are created and the objects which represent them.

- There is now a `WorkerInWorld` which, as the name suggests, represents a Worker which is tied to an ECS world.
- There is now a builder pattern for creating a `IConnectionHandler` (i.e. - a connection object).
- The `DefaultWorkerConnector` and `DefaultMobileWorkerConnector` have been removed as has many of the inheritable methods in `WorkerConnector`. Instead, a user composes how they want to connect with the builder pattern mentioned above.
- The `ReceptionistConfig`, `LocatorConfig`, and `AlphaLocatorConfig` have been removed. The data used to connect with a specific flow lives with the implementation of that connection flow (i.e. - `ReceptionistFlow`, `LocatorFlow`, and `AlphaLocatorFlow`, which all inherit from `IConnectionFlow`). 
- In order to allow reuse of common bits of functionality when configuring a connection, there exists a `IConnectionFlowInitializer<TConnectionFlow> where TConnectionFlow : IConnectionFlow` interface. The implementer of this interface describes how to populate the data required to connect for a particular flow.
- Joining the circle back up, the default implementation for flow configuration in `DefaultWorkerConnector` has been migrated to `CommandLineConnectionFlowInitializer` and the `DefaultMobileWorkerConnector` has been migrated to `MobileConnectionFlowInitializer` (plus some refactorings). 

The result of this PR is some changes to the internals (moving towards decoupling Workers & Worlds) as well as moving to composition over inheritance for creating a worker. 

This does result in a little bit more effort on a user's part to set up a `WorkerConnector` however I think this is okay for two reasons:

- Most users start from a starter project, which has preconfigured ones.
- There is much less magic (somewhat hidden inherited methods) which controlled behavior. 

##### Unresolved questions

- [x] How to handle connection parameters? Currently they are constructed and modified directly by the user. Do we want to provide a similar mechanism for reusable blobs of code to modify them. (Resolved in https://github.com/spatialos/gdk-for-unity/pull/981#discussion_r297219960) 

#### Left TODO

- [x] Changelog
- [x] Migration guide
- [x] Assorted TODOs in the code (these are really me asking you, the reader, these questions)

#### Tests

Will want to do thorough testing of each combinations of flows. Current semantics *should be preserved*.


workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlowImpls.cs
```diff
@@ -0,0 +1,269 @@
...
+                }
+
+                using (var connectionFuture = locator.ConnectAsync(deploymentName, parameters, _ => true))
+                {
+                    return await Utils.TryToConnectAsync(connectionFuture).ConfigureAwait(false);
+                }
+            }
+        }
+
+        protected virtual string SelectDeploymentName(DeploymentList deployments)
```
jessicafalk: let's call it `deploymentList` instead of `deployments`. looks at least a tiny bit better when calling `deploymentList.Deployments` instead of `deployments.Deployments` ;)

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlowImpls.cs
```diff
@@ -0,0 +1,269 @@
...
+
+    public class LocatorFlowImpl : IConnectionFlowImpl
+    {
+        public string LocatorHost = "locator.improbable.io";
+
+        public LocatorParameters LocatorParameters = new LocatorParameters
+        {
+            LoginToken = new LoginTokenCredentials
+            {
+                Token = string.Empty
```
jessicafalk: do we need to set all of those strings explicitly to `string.Empty`?

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlowImpls.cs
```diff
@@ -0,0 +1,269 @@
...
+            },
+            Steam = new SteamCredentials
+            {
+                Ticket = string.Empty,
+                DeploymentTag = string.Empty
+            },
+            Logging = new ProtocolLoggingParameters(),
+            ProjectName = string.Empty,
+            CredentialsType = LocatorCredentialsType.LoginToken,
+            EnableLogging = false
```
jessicafalk: add a comma at the end,
I also think we should expose the `EnableLogging` and protocol logs parameter to make debugging easier instead of having to change this file directly or inheriting from it

jessicafalk: same for alpha locator

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/Utility/CommandLineUtilityTest.cs
```diff
@@ -6,6 +6,7 @@ namespace Improbable.Gdk.Core.EditmodeTests
     [TestFixture]
     public class CommandLineUtilityTest
     {
+        // TODO: Rewrite this.
```
jessicafalk: I assume this will happen in a separate pr?

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
```diff
@@ -38,5 +65,20 @@ public override void Dispose()
...
         }
+
+        public Option<string> GetHostIp()
+        {
+            return ipAddress;
+        }
+
+        public Option<string> GetDevAuthToken()
+        {
+            return Option<string>.Empty;
```
jessicafalk: this should try to read the dev auth token from a txt asset

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
```diff
@@ -38,5 +65,20 @@ public override void Dispose()
...
+        }
+
+        public Option<ConnectionService> GetConnectionService()
+        {
+            return Option<ConnectionService>.Empty;
```
jessicafalk: this should handle it based on the `shouldConnectLocally` checkbox

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
```diff
@@ -1,20 +1,47 @@
+using System;
 using Improbable.Gdk.Core;
 using Improbable.Gdk.Mobile;
+using Improbable.Worker.CInterop;
 using UnityEngine;
 
 namespace Playground
 {
-    public class MobileClientWorkerConnector : DefaultMobileWorkerConnector
+    public class MobileClientWorkerConnector : WorkerConnector, IMobileSettingsProvider
```
jessicafalk: what will happen to the `DefaultMobileWorkerConnector`?

workers/unity/Assets/Playground/Scripts/Worker/ClientWorkerConnector.cs
```diff
@@ -15,13 +17,45 @@ public class ClientWorkerConnector : DefaultWorkerConnector
...
-        {
-            // This could be replaced with a splash screen asking to select a deployment or some other user-defined logic.
-            return deployments.Deployments[0].DeploymentName;
+            var connParams = new ConnectionParameters
+            {
+                WorkerType = WorkerUtils.UnityClient,
+                DefaultComponentVtable = new ComponentVtable(),
+                Network = new NetworkParameters
+                {
+                    UseExternalIp = UseExternalIp
```
jessicafalk: this will be overwritten when `resolve` is called, right?

workers/unity/Assets/Playground/Scripts/Worker/GameLogicWorkerConnector.cs
```diff
@@ -14,7 +16,27 @@ public class GameLogicWorkerConnector : DefaultWorkerConnector
...
             Application.targetFrameRate = 60;
-            await Connect(WorkerUtils.UnityGameLogic, new ForwardingDispatcher()).ConfigureAwait(false);
+
+            var connParams = new ConnectionParameters
+            {
+                WorkerType = WorkerUtils.UnityGameLogic,
+                DefaultComponentVtable = new ComponentVtable(),
+                Network = new NetworkParameters
+                {
+                    UseExternalIp = UseExternalIp
```
jessicafalk: this will be overwritten when `resolve` is called, right?

workers/unity/Assets/Playground/Config/DeploymentLauncherConfig.asset
```diff
@@ -14,11 +14,11 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   AssemblyConfig:
     ProjectName: unity_gdk
-    AssemblyName: playground
-    ShouldForceUpload: 0
+    AssemblyName: unity_gdk_0612_1204
```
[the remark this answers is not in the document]
jamiebrynes7: Yes - I put this up for a review of the structure rather than the details 😉

workers/unity/Packages/com.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -3,35 +3,50 @@
...
 
-        public static T GetCommandLineValue<T>(Dictionary<string, string> commandLineDictionary, string configKey,
-            T defaultValue)
+        public bool Contains(string key)
         {
-            if (TryGetConfigValue(commandLineDictionary, configKey, out T configValue))
+            return arguments.ContainsKey(key);
+        }
+
+        public void SetCommandLineValue<T>(string key, ref T value)
```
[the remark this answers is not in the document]
jamiebrynes7: Sure 👍

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlowImpls.cs
```diff
@@ -0,0 +1,269 @@
...
+                    return await Utils.TryToConnectAsync(connectionFuture).ConfigureAwait(false);
+                }
+            }
+        }
+
+        protected virtual string SelectDeploymentName(DeploymentList deployments)
+        {
+            if (deployments.Error != null)
+            {
+                throw new Exception($"Failed to list deployments with error: ${deployments.Error}");
```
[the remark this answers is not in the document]
jamiebrynes7: I think I've seen something similar

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlowImpls.cs
```diff
@@ -0,0 +1,269 @@
...
+        ///     Retrieves the login tokens for all active deployments that the player
+        ///     can connect to via the anonymous authentication flow.
+        /// </summary>
+        /// <param name="workerType">The type of the worker that wants to connect.</param>
+        /// <param name="playerIdentityToken">The player identity token of the player that wants to connect.</param>
+        /// <returns>A list of all available login tokens and their deployments.</returns>
+        protected virtual List<LoginTokenDetails> GetDevelopmentLoginTokens(string workerType, string playerIdentityToken)
+        {
+            var result = DevelopmentAuthentication.CreateDevelopmentLoginTokensAsync(
+                RuntimeConfigDefaults.LocatorHost,
```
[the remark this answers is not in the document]
jamiebrynes7: Yeah I know - its a mix of copy pasting / writing code

workers/unity/Packages/com.improbable.gdk.mobile/Utility/MobileConfigResolvers.cs
```diff
@@ -0,0 +1,183 @@
...
+            }
+
+            return Option<string>.Empty;
+        }
+
+        public Option<string> GetDevAuthToken()
+        {
+            var args = LaunchArguments.GetArguments();
+
+            if (args.TryGetCommandLineValue(RuntimeConfigNames.DevAuthTokenKey, out string devAuthToken))
```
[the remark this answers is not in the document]
jamiebrynes7: Which flow do you mean?

workers/unity/Packages/com.improbable.gdk.mobile/Utility/MobileConfigResolvers.cs
```diff
@@ -0,0 +1,183 @@
...
+                token = Resources.Load<TextAsset>("DevAuthToken")?.text.Trim();
+            }
+
+            alphaLocator.DevAuthToken = token ?? string.Empty;
+        }
+    }
+
+    public class CommandLineSettingsProvider : IMobileSettingsProvider
+    {
+        private const string HostIpPlayerPrefsKey = "SpatialOSHostIp";
```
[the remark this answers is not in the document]
jamiebrynes7: Yup agreed

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlowImpls.cs
jamiebrynes7: We _may_ do. I didn't test what happens if you call `args.SetCommandLineValue(key, ref something.dest);` where `something.dest is null` but `something` is valid

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlowImpls.cs
jamiebrynes7: Protocol logging params are in the `ConnectionParameters` as well. Which takes precedence? I'd prefer using the `ConnectionParameters` for configuring logging (as its irrelevant to connection flow)

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/Utility/CommandLineUtilityTest.cs
jamiebrynes7: It'll be done before anything gets merged - that's why all the tests auto-fail right now

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
jamiebrynes7: It will be removed (I may have not done that yet)

workers/unity/Assets/Playground/Scripts/Worker/ClientWorkerConnector.cs
jamiebrynes7: It can be

workers/unity/Assets/Playground/Scripts/Worker/GameLogicWorkerConnector.cs
jamiebrynes7: It can be

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
jamiebrynes7: I thought I added and pushed this 🤔

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
jamiebrynes7: What is the order of precedence? I.e. - who wins out if there are player prefs & command line args & this checkbox is set?

jamiebrynes7: Actually you kind of want three states for the checkbox as its like an override? 

- Not overridden
- Overidden and set
- Overidden and not set

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
jamiebrynes7: Ah I had it in a different place.

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
jamiebrynes7: Made obsolete by the mobile launcher 🎉

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
```diff
@@ -1,20 +1,49 @@
...
         public async void Start()
         {
-            await Connect(WorkerUtils.MobileClient, new ForwardingDispatcher()).ConfigureAwait(false);
+            var connParams = new ConnectionParameters
+            {
+                WorkerType = WorkerUtils.MobileClient,
+                DefaultComponentVtable = new ComponentVtable()
+            };
+
+            var flowSettingsProvider = new MobileConnectionFlowModifier(new PlayerPrefsSettingsProvider(), new CommandLineSettingsProvider(), this);
```
jessicafalk: I think the ordering needs to be:
1. first check if we got new command line args
2. then check if player prefs available
3. then do whatever fallback behaviour you have

cause otherwise command line args will be ignored after player prefs have been set

workers/unity/Packages/com.improbable.gdk.core/Tests/Editmode/Systems/SpatialOSSendSystemTests.cs
```diff
@@ -12,23 +12,22 @@ namespace Improbable.Gdk.Core.EditmodeTests.Systems
...
         [OneTimeSetUp]
         public void OneTimeSetup()
         {
-            world = new World("test-world");
-            logDispatcher = new TestLogDispatcher();
-            world.CreateSystem<WorkerSystem>(new MockConnectionHandler(), null, logDispatcher, TestWorkerType, Vector3.zero);
+            var mockConnectionBuilder = new MockConnectionHandlerBuilder();
+            worker = WorkerInWorld.CreateWorkerInWorldAsync(mockConnectionBuilder, "TestWorkerType",
+                new TestLogDispatcher(), Vector3.zero).Result;
+
```
jessicafalk: nit: nwe line

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlows.cs
```diff
@@ -0,0 +1,276 @@
...
+        {
+            var result = DevelopmentAuthentication.CreateDevelopmentLoginTokensAsync(
+                RuntimeConfigDefaults.LocatorHost,
+                RuntimeConfigDefaults.AnonymousAuthenticationPort,
+                new LoginTokensRequest
+                {
+                    WorkerType = workerType,
+                    PlayerIdentityToken = playerIdentityToken,
+                    UseInsecureConnection = false,
+                    DurationSeconds = 120,
```
jessicafalk: we may be able to remove this. Think I added it because of a bug in an older version of the worker sdk

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
```diff
@@ -1,20 +1,49 @@
...
+
+            var builder = new SpatialOSConnectionHandlerBuilder()
+                .SetConnectionParameters(connParams)
+                .SetConnectionFlowModifier(flowSettingsProvider)
+                .SetConnectionParametersModifier(new MobileConnectionParametersModifier());
+
+            switch (flowSettingsProvider.GetConnectionService())
```
jessicafalk: one thing that I'm still wondering: would it be possible to hide that from the user?
just let them set whichever flows they want and then handle the switch-case in the builder? or do we not want to do it?

jamiebrynes7: Hmm so I guess doing that would remove the simplest use case which is something like: 

```csharp
var builder = new SpatialOSConnectionHandlerBuilder()
	.SetConnectionParameters(connParams)
	.SetConnectionFlow(new ReceptionistFlow());

await Connect(builder, ...)
```

That is to say - there is no requirement currently that you provide anything beyond the two things above. 

---

Also where does the value from the switch-case come from? What do we do for the mobile case where the standard Locator flow is entirely unsupported? Does a user still provide the flow / just set it to null? 

I guess my main objection is that this PR greatly reduces the amount of magic that happens, at the cost of pushing a bit more logic up to the user code layer. Everything is opt in as opposed to opt out (status quo).

jessicafalk: quick idea, not completely thought trough: 
if the flow is not set and it will be selected for some reason, it should throw an exception,. e.g. a "default does-not-exist flow" object. you could instead have a 
`builder.SetConnectionServiceHandler(yourObject)` that implements the logic for the switch. that way you still expose it, but it looks more like a basic builder pattern and is easier to follow in user code.

jamiebrynes7: Discussed offline and reworked slightly

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
```diff
@@ -1,20 +1,57 @@
...
 #pragma warning restore 649
 
         private GameObject levelInstance;
 
         public async void Start()
         {
-            await Connect(WorkerUtils.MobileClient, new ForwardingDispatcher()).ConfigureAwait(false);
+            var connParams = CreateConnectionParameters(WorkerUtils.MobileClient);
+            connParams.Network.UseExternalIp = true;
+            connParams.Network.ConnectionType = NetworkConnectionType.Kcp;
```
zeroZshadow: Should we read this from commandline?

paulbalaji: if you did that, wouldn't you have to read in the parameters for it too?

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlowInitializers.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+        void Initialize(TConnectionFlow flow);
+    }
+
+    public class CommandLineConnectionFlowInitializer :
+        IConnectionFlowInitializer<ReceptionistFlow>,
+        IConnectionFlowInitializer<LocatorFlow>,
+        IConnectionFlowInitializer<AlphaLocatorFlow>
+    {
+        private CommandLineArgs commandLineArgs;
```
paulbalaji: readonly?

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlows.cs
```diff
@@ -0,0 +1,289 @@
...
+            },
+            Steam = new SteamCredentials
+            {
+                Ticket = string.Empty,
+                DeploymentTag = string.Empty
+            },
+            Logging = new ProtocolLoggingParameters(),
+            ProjectName = string.Empty,
+            CredentialsType = LocatorCredentialsType.LoginToken,
+            EnableLogging = false,
```
paulbalaji: do we need to explicitly set this to false? if we do here - should do the same when initialising `Alpha.LocatorParameters`

workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerSystem.cs
```diff
@@ -98,12 +97,6 @@ protected override void OnCreate()
             Enabled = false;
         }
 
-        protected override void OnDestroy()
-        {
-            ConnectionHandler.Dispose();
-            base.OnDestroy();
-        }
-
         protected override void OnUpdate()
```
paulbalaji: random question - why should this be empty instead of just not being implemented?

workers/unity/Packages/com.improbable.gdk.mobile/Utility/MobileConnectionFlowInitializer.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        public class CommandLineSettingsProvider : IMobileSettingsProvider
+        {
+            public Option<string> GetReceptionistHostIp()
+            {
+                var args = LaunchArguments.GetArguments();
+
+                var hostIp = string.Empty;
+
+                if (args.TryGetCommandLineValue(RuntimeConfigNames.ReceptionistHost, ref hostIp))
```
paulbalaji: invert ifs in `CommandLineSettingsProvider`

(seems you've done that with `GetConnectionService` in `PlayerPrefsSettingsProvider`)

workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs
```diff
@@ -21,47 +21,35 @@ public static class PlayerLifecycleHelper
         {
             template.AddComponent(new PlayerHeartbeatClient.Snapshot(),
                 EntityTemplate.GetWorkerAccessAttribute(clientWorkerId));
-
             template.AddComponent(new PlayerHeartbeatServer.Snapshot(), serverAccess);
             template.AddComponent(new OwningWorker.Snapshot(clientWorkerId), serverAccess);
         }
```
paulbalaji: we don't need line 2/3/5

(couldn't add a comment on the line itself)

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
jamiebrynes7: This relates to the "Unresolved Question" from the description: 

>  How to handle connection parameters? Currently they are constructed and modified directly by the user. Do we want to provide a similar mechanism for reusable blobs of code to modify them.

In any case - its left up to the user. I want to avoid magic behavior that isn't opt-in. We can certainly provide a `ConnectionParameters.FromCommandLine()` or similar that has a reasonable implementation.

workers/unity/Packages/com.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -34,263 +25,67 @@ public class Worker : IDisposable
...
-#if !DISABLE_REACTIVE_COMPONENTS
-            // Reactive components
-            ReactiveComponentsHelper.AddCommonSystems(World);
-#endif
-        }
-
-        public void Dispose()
+        public virtual void Dispose()
         {
+            ConnectionHandler?.Dispose();
```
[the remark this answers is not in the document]
jamiebrynes7: Awkwardly this means it cannot be `readonly` 😕

workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerSystem.cs
```diff
@@ -70,23 +70,22 @@ public bool HasEntity(EntityId entityId)
...
+            Worker.MessagesToSend.AddLogMessage(new LogMessageToSend(message, loggerName, logLevel, entityId?.Id));
         }
 
         public void SendMetrics(Metrics metrics)
         {
-            MessagesToSend.AddMetrics(metrics);
+            Worker.MessagesToSend.AddMetrics(metrics);
         }
 
         internal void GetMessages()
```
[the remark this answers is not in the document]
jamiebrynes7: Made it `Advance` to at least be consistent with the inner call.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -160,205 +133,6 @@ await Worker.CreateWorkerAsync(GetAlphaLocatorConfig(workerType), connectionPara
...
-            if (result.Value.Status.Code != ConnectionStatusCode.Success)
-            {
-                throw new AuthenticationFailedException("Failed to retrieve any login tokens.\n" +
-                    $"error code: {result.Value.Status.Code}\nerror message: {result.Value.Status.Detail}");
-            }
-
-            return result.Value.LoginTokens;
-        }
-
         protected virtual void HandleWorkerConnectionEstablished()
```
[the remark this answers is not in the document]
jamiebrynes7: Its not - although in practice we do generally. I'd prefer not to force users to implement this though

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandlerBuilder.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    public class SpatialOSConnectionHandlerBuilder : IConnectionHandlerBuilder
+    {
+        public enum ThreadingMode
+        {
+            SingleThreaded,
+            MultiThreaded
+        }
+
+        public string WorkerType => connectionParameters?.WorkerType ?? string.Empty;
```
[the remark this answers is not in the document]
jamiebrynes7:
```csharp       
public string WorkerType { get; private set; } = string.Empty;
```

This is valid syntax, TIL

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlows.cs
```diff
@@ -0,0 +1,289 @@
...
+    public interface IConnectionFlow
+    {
+        Task<Connection> CreateAsync(ConnectionParameters parameters, CancellationToken? token = null);
+    }
+
+    public class ReceptionistFlow : IConnectionFlow
+    {
+        public string ReceptionistHost = RuntimeConfigDefaults.ReceptionistHost;
+        public ushort ReceptionistPort = RuntimeConfigDefaults.ReceptionistPort;
+        public string WorkerId;
```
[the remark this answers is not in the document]
jamiebrynes7: The `initializer` is at liberty to change this

workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerSystem.cs
```diff
@@ -17,29 +17,29 @@ public class WorkerSystem : ComponentSystem
...
-            WorkerType = workerType;
-            Origin = origin;
-            ConnectionHandler = connectionHandler;
+            Worker = worker;
 
-            MessagesToSend = connectionHandler.GetMessagesToSendContainer();
+            LogDispatcher = worker.LogDispatcher;
+            WorkerType = worker.WorkerType;
+            WorkerId = worker.WorkerId;
+            Origin = worker.Origin;
```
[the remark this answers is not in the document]
jamiebrynes7: Nope, the `Origin` is readonly.

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -71,7 +68,7 @@ protected void OnDestroy()
         /// <param name="workerType">The type of the worker to connect as</param>
```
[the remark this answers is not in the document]
jamiebrynes7: Fixed!

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlows.cs
jamiebrynes7: Not strictly speaking - but we set everything to a safe default so it was just for consistency. Do you feel particularly strongly about this?

workers/unity/Packages/com.improbable.gdk.core/Systems/WorkerSystem.cs
jamiebrynes7: `ComponentSystem.OnUpdate()` is abstract

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlows.cs
jamiebrynes7: Leaving as is until we can confirm so

jamiebrynes7:
Tested local & cloud workflows for: https://github.com/spatialos/gdk-for-unity/pull/981/commits/78bbf7c45e0645a7715430cbef4d981ba6a4d447 on PC & mobile. All ✔️

workers/unity/Assets/Playground/Scripts/Worker/MobileClientWorkerConnector.cs
jamiebrynes7: Discussed offline with @zeroZshadow - going to implement a similar pattern to the connection flow initializers so you have something like: 

```csharp
var connParams = CreateConnectionParameters(WorkerUtils.MobileClient, new CommandLineConnectionParametersInitializer());
```

What a mouthful.

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
```diff
@@ -30,7 +30,7 @@ protected override void OnUpdate()
             try
             {
                 Profiler.BeginSample("GetMessages");
-                worker.GetMessages();
+                worker.Advance();
```
zeroZshadow: I disagree with this function name. It gives no indication of what it does.

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandlerBuilder.cs
```diff
@@ -47,7 +48,7 @@ public async Task<IConnectionHandler> CreateAsync(CancellationToken? token = nul
 
             if (flow == null)
             {
-                throw new InvalidOperationException("An executor implementation must be provided.");
+                throw new InvalidOperationException("An connection flow implementation must be provided.");
```
zeroZshadow: `An` -> `A`

workers/unity/Packages/com.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
jamiebrynes7: Changed to `Tick()`

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandlerBuilder.cs
jamiebrynes7: Nice catch

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlowInitializers.cs
```diff
@@ -0,0 +1,99 @@
+using Improbable.Worker.CInterop;
+
+namespace Improbable.Gdk.Core
+{
+    // TODO: Should this be contravarient?
```
paulbalaji: keeping the TODO in?

jamiebrynes7: Yes as I'm not entirely sure (need a use case) for doing so

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlows.cs
```diff
@@ -0,0 +1,372 @@
...
+        /// <summary>
+        ///     The development authentication token to use when connecting via with development authentication.
+        /// </summary>
+        public string DevAuthToken = string.Empty;
+
+        /// <summary>
+        ///    Denotes whether we should connect with development authentication.
+        /// </summary>
+        /// <remarks>
+        ///     If this is false, is it assumed that the <see cref="PlayerIdentityCredentials"/> element
```
paulbalaji:
```suggestion
        ///     If this is false, it is assumed that the <see cref="PlayerIdentityCredentials"/> element
```

workers/unity/Packages/com.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlows.cs
```diff
@@ -0,0 +1,372 @@
...
+            if (loginTokens.Count == 0)
+            {
+                throw new AuthenticationFailedException("Did not receive any login tokens. Do you have a valid deployment running?");
+            }
+
+            return loginTokens[0].LoginToken;
+        }
+
+        /// <summary>
+        ///     Gets the player id for the player trying to connect via the development authentication flow.
```
paulbalaji:
```suggestion
        ///     Gets the player ID for the player trying to connect via the development authentication flow.
```

id --> ID if you find any more

workers/unity/Packages/com.improbable.gdk.core/Worker/WorkerInWorld.cs
```diff
@@ -0,0 +1,104 @@
...
+using Improbable.Gdk.Subscriptions;
+using Unity.Entities;
+using UnityEngine;
+#if !DISABLE_REACTIVE_COMPONENTS
+using Improbable.Gdk.ReactiveComponents;
+#endif
+
+namespace Improbable.Gdk.Core
+{
+    public class WorkerInWorld : Worker
```
paulbalaji: add a summary for `WorkerInWorld`?

Issue #959, opened 2019-05-28
Invalid characters in your PATH cause exceptions at startup


jamiebrynes7:
**Affects:** Release v0.1.0 and up on Windows.
**Internal Ticket:** [UTY-2000](https://improbableio.atlassian.net/browse/UTY-2000)

---

#### Description

If your PATH environment variable has invalid characters in it, the GDK will throw exceptions at startup when trying to find the `spatial` and `dotnet` binaries.

#### Workaround

Inspect your PATH variable and remove elements with invalid characters. See the [MSDN naming conventions](https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#naming-conventions) for a list of invalid characters.

[Small guide on editing your PATH](https://www.howtogeek.com/118594/how-to-edit-your-system-path-for-easy-command-line-access/)

jamiebrynes7:
Fixed in 0.2.4

PR #986, opened 2019-06-13
Fix PATH errors on Windows

jamiebrynes7:
#### Description

When we encounter an invalid PATH element during the binary discovery process, we drop those elements instead of throwing an exception which would break the entire GDK.

The alternative proposed solution was to call `spatial`/`dotnet` and check the result, but apparently Mac is a little special and so that would not always work.

#### Tests

Will try to break my PATH and see what happens. 🤞 

#### Documentation

Change log and will close known issue #959

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


jamiebrynes7:
Added "oh;no" to my PATH - was able to generate code no problems (also verified that my PATH had actually updated)

PR #990, opened 2019-06-17
Transform sync bandwidth optimisations

paulbalaji:
#### Description

- Quaternions are now compressed from 4 floats --> uint32
- more granular change checking in the transform send system
- location/velocity are now fixed points, with x/y/z each an `sint32` (zig zag encoded)
- updated snapshot after schema changes were made
- drive-by changes to some playground systems

_Playground with 16 cubes + 2 spinners has gone from ~44 kBps on Transform updates down to ~22 kBps._

#### Tests

When running comparable deployments with the standard GDK playground snapshot (16 cubes, 2 spinners), we recorded the average bandwidth used by a UnityGameLogic worker sending TransformInternal component updates:

|Optimisation|Bandwidth (kBps)|
|-|-|
|None|43.9|
|Quaternion compression|36.5|
|More granular checking in transform send system|30.7|
|Fixed point location/velocity|21.4|
|Q15.16 --> Q21.10 for fixed points|20.2|

#### Documentation

TODO:
- [x] changelog
- [x] add comment to ticket
- [x] in-code docs where possible
- [x] update transform sync docs https://github.com/spatialos/gdk-for-unity/pull/998

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
```diff
@@ -11,51 +11,146 @@ public static bool HasChanged(Coordinates a, Coordinates b)
...
 
         // Checking for no change, so exact equality is fine
-        public static bool HasChanged(Quaternion a, Quaternion b)
+        public static bool HasChanged(FixedPointVector3 a, FixedPointVector3 b)
+        {
+            return a.X != b.X || a.Y != b.Y || a.Z != b.Z;
+        }
+
+        // Decompress from uint32 to Unity Quaternion
+        public static UnityEngine.Quaternion ToUnityQuaternion(this CompressedQuaternion quaternion)
```
jamiebrynes7: Please add comments for these compressions. Someone coming to this without context will have to slog through it 😉

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
```diff
@@ -11,51 +11,146 @@ public static bool HasChanged(Coordinates a, Coordinates b)
...
+            return new Coordinates
+            {
+                X = FixedToFloat(fixedPointVector3.X),
+                Y = FixedToFloat(fixedPointVector3.Y),
+                Z = FixedToFloat(fixedPointVector3.Z)
+            };
         }
 
-        public static Velocity ToImprobableVelocity(this Vector3 velocity)
+        private const int FixedPointOne = 0x00010000;
```
samiwh: Is this accuracy deliberate? I would have expected you to want 10 bits (mm accuracy) after the binary point and 21 bits before + a sign bit.
This looks like you are going down to about half a micrometer accuracy, which would give you about 16km to play with if you are feeding this from a double. But from a float you can only represent ~ 2^24 as an integer without loss of precision. Which means that you can go up to about +-256m before it loses precision. Arguably this loss in prevision isn't important as it will be in the bit that probably already has too much precision, but it still seems a bit odd. It might be worth working out what precision loss is expected at expected values.

I'd also not do this is a single hex string. Make it something like `1u << 10` to make the precision clear. Also probably put a comment on what rough decimal range this corresponds too.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
```diff
@@ -11,51 +11,146 @@ public static bool HasChanged(Coordinates a, Coordinates b)
...
         }
 
         // Checking for no change, so exact equality is fine
-        public static bool HasChanged(Quaternion a, Quaternion b)
+        public static bool HasChanged(FixedPointVector3 a, FixedPointVector3 b)
+        {
+            return a.X != b.X || a.Y != b.Y || a.Z != b.Z;
+        }
+
+        // Decompress from uint32 to Unity Quaternion
```
samiwh: These functions need comments that either describe the algorithm exactly or link to something that does.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
```diff
@@ -11,51 +11,146 @@ public static bool HasChanged(Coordinates a, Coordinates b)
...
+        public static UnityEngine.Quaternion ToUnityQuaternion(this CompressedQuaternion quaternion)
+        {
+            var compressedValue = quaternion.Data;
+            var q = new float[4];
+
+            const uint mask = (1u << 9) - 1u;
```
samiwh: Why 9

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/DefaultUpdateLatestTransformSystem.cs
```diff
@@ -44,25 +43,19 @@ private void UpdateRigidbodyData()
         {
             Entities.With(rigidbodyGroup).ForEach((ref TransformToSend transformToSend, Rigidbody rigidbody) =>
             {
-                transformToSend = new TransformToSend
-                {
-                    Position = rigidbody.position,
-                    Velocity = rigidbody.velocity,
-                    Orientation = rigidbody.rotation
-                };
+                transformToSend.Position = rigidbody.position;
```
zeroZshadow: The previous method should be faster due to less dereferencing

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
```diff
@@ -11,51 +11,146 @@ public static bool HasChanged(Coordinates a, Coordinates b)
...
+        public static bool HasChanged(FixedPointVector3 a, FixedPointVector3 b)
+        {
+            return a.X != b.X || a.Y != b.Y || a.Z != b.Z;
+        }
+
+        // Decompress from uint32 to Unity Quaternion
+        public static UnityEngine.Quaternion ToUnityQuaternion(this CompressedQuaternion quaternion)
+        {
+            var compressedValue = quaternion.Data;
+            var q = new float[4];
```
zeroZshadow: Can this be a NativeArray or use stackalloc to avoid allocation

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
```diff
@@ -11,51 +11,146 @@ public static bool HasChanged(Coordinates a, Coordinates b)
...
+
+        private const float SqrtHalf = 0.70710678118f;
+
+        // Compress from Unity Quaternion to uint32
+        public static CompressedQuaternion ToCompressedQuaternion(this UnityEngine.Quaternion quaternion)
         {
-            return a.W != b.W || a.X != b.X || a.Y != b.Y || a.Z != b.Z;
+            quaternion = quaternion.normalized;
+
+            var q = new float[4] { quaternion.x, quaternion.y, quaternion.z, quaternion.w };
```
zeroZshadow: avoid allocation here

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
```diff
@@ -11,51 +11,146 @@ public static bool HasChanged(Coordinates a, Coordinates b)
...
+                    uint magnitude = (uint) (((1u << 9) - 1u) * (Mathf.Abs(q[i]) / SqrtHalf) + 0.5f);
+                    compressedQuaternion = (compressedQuaternion << 10) | (signBit << 9) | magnitude;
+                }
+            }
+
+            return new CompressedQuaternion(compressedQuaternion);
         }
 
-        public static UnityEngine.Quaternion ToUnityQuaternion(this Quaternion quaternion)
+        public static UnityEngine.Vector3 ToUnityVector3(this FixedPointVector3 fixedPointVector3)
```
zeroZshadow: All of these helper utilities should be normal functions. To avoid users seeing them when using the unity types.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
```diff
@@ -11,51 +11,146 @@ public static bool HasChanged(Coordinates a, Coordinates b)
...
+            const uint mask = (1u << 9) - 1u;
+
+            int largestIndex = (int) (compressedValue >> 30);
+            float sumSquares = 0;
+            for (var i = 3; i >= 0; --i)
+            {
+                if (i != largestIndex)
+                {
+                    uint magnitude = compressedValue & mask;
+                    uint signBit = (compressedValue >> 9) & 0x1;
```
samiwh: Is the reason the mask is 9 bits to add room for a sign bit then? Can't the mapping from float to fixed just encode this in the main 10 bits?

samiwh: This seems like it just adds floating point comparisons and multiplications.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
samiwh: If you want to represent 0 exactly generally you can map from the quaternion float range to the fixed int range -1. You don't need to lose an entire bit.

samiwh: Actually now that I think about it that's the only way of doing it. 2^n is always even :P

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
paulbalaji: It's not deliberate - you're right that we probably we don't need more than mm accuracy

workers/unity/Assets/Playground/Editor/SnapshotGenerator/SnapshotGenerator.cs
```diff
@@ -84,14 +83,15 @@ private static void AddCubeGrid(Snapshot snapshot, int cubeCount)
                         return;
                     }
 
+                    var location = new Vector3(x, 1, z);
                     var positionSnapshot = new Position.Snapshot
                     {
-                        Coords = new Coordinates(x, 1, z)
+                        Coords = TransformUtils.ToCoordinates(location)
```
zeroZshadow: I think this can be reverted to an extension method.

workers/unity/Assets/Playground/Editor/SnapshotGenerator/SnapshotGenerator.cs
```diff
@@ -84,14 +83,15 @@ private static void AddCubeGrid(Snapshot snapshot, int cubeCount)
...
                     };
                     var transformSnapshot = new TransformInternal.Snapshot
                     {
-                        Location = new Location(x, 1, z),
-                        Rotation = new Quaternion(1, 0, 0, 0),
+                        Location = TransformUtils.ToFixedPointVector3(location),
```
zeroZshadow: Creation of an TransformInternal Snapshot needs a helper utility.
The user should not have to deal with compression.

workers/unity/Assets/Playground/Scripts/MonoBehaviours/SpawnCubeCommandReceiver.cs
```diff
@@ -49,16 +48,17 @@ private void OnEntityIdsReserved(WorldCommands.ReserveEntityIds.ReceivedResponse
...
             cubeEntityTemplate.SetComponent(new Position.Snapshot
             {
-                Coords = new Coordinates(location.X, location.Y + 2, location.Z)
+                Coords = TransformUtils.ToCoordinates(location)
             });
             cubeEntityTemplate.SetComponent(new TransformInternal.Snapshot
             {
-                Location = new Location(location.X, location.Y + 2, location.Z),
-                Rotation = new Quaternion(1, 0, 0, 0)
+                Location = TransformUtils.ToFixedPointVector3(location),
```
zeroZshadow: User should not have to deal with compression here

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/DirectTransformUpdateSystem.cs
```diff
@@ -46,9 +46,9 @@ protected override void OnUpdate()
 
                     transformToSet = new TransformToSet
                     {
-                        Position = transformInternal.Location.ToUnityVector3() + worker.Origin,
-                        Velocity = transformInternal.Velocity.ToUnityVector3(),
-                        Orientation = transformInternal.Rotation.ToUnityQuaternion()
+                        Position = TransformUtils.ToUnityVector3(transformInternal.Location) - worker.Origin,
```
zeroZshadow: Why did this change from `+ origin` to `- origin`?

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
paulbalaji: atm the identity quaternion (0,0,0,1) gets packed to [3 0 0 0] (index 3, value 0, value 0, value 0)

wouldn't it be easier to make 0 the identity by switching the order in compression to wxyz, so the packed structure would become [0 0 0 0]?

samiwh: I don't understand. If you want 0 to be the identity then special case it.

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
paulbalaji: I'll leave this as-is for now to get the changes in, and raise a new ticket for further optimisations

samiwh: It's not for optimisation so much as it seems odd.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 ### Breaking Changes
 
 - The constructor for the `ForwardingDispatcher` now accepts a `UnityEngine.LogType` instead of `Improbable.Worker.CInterop.LogLevel` as its parameter. [#987](https://github.com/spatialos/gdk-for-unity/pull/987)
+- The schema for the Transform Synchronization feature module has been optimised to reduce bandwidth. If you use this module, please make use of the updated helper methods and regenerate your snapshot. [#990](https://github.com/spatialos/gdk-for-unity/pull/990)
+    - Quaternions are now compressed from 4 floats to a single uint32.
+    - Location and velocity are now `Q21.10` fixed point values.
+    - Regenerated `default.snapshot` after schema changes were made.
```
jamiebrynes7: This isn't a breaking change.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 ### Breaking Changes
 
 - The constructor for the `ForwardingDispatcher` now accepts a `UnityEngine.LogType` instead of `Improbable.Worker.CInterop.LogLevel` as its parameter. [#987](https://github.com/spatialos/gdk-for-unity/pull/987)
+- The schema for the Transform Synchronization feature module has been optimised to reduce bandwidth. If you use this module, please make use of the updated helper methods and regenerate your snapshot. [#990](https://github.com/spatialos/gdk-for-unity/pull/990)
+    - Quaternions are now compressed from 4 floats to a single uint32.
+    - Location and velocity are now `Q21.10` fixed point values.
```
jamiebrynes7: What is `Q21.10` ?

workers/unity/Assets/Playground/Scripts/MonoBehaviours/RotationBehaviour.cs
```diff
@@ -14,6 +14,6 @@ public class RotationBehaviour : MonoBehaviour
     private void Update()
     {
         transform.rotation *=
-            Quaternion.Euler((RotatingClockWise ? Vector3.up : Vector3.down) * Time.deltaTime * 20.0f);
+            Quaternion.Euler(Time.deltaTime * 20.0f * (RotatingClockWise ? Vector3.up : Vector3.down));
```
jamiebrynes7: Nit: Prefer constants as first element in the chain. Not a big deal tho

workers/unity/Packages/com.improbable.gdk.transformsynchronization/Systems/ResetForAuthorityGainedSystem.cs
```diff
@@ -22,6 +21,9 @@ protected override void OnCreate()
             worker = World.GetExistingSystem<WorkerSystem>();
             updateSystem = World.GetExistingSystem<ComponentUpdateSystem>();
 
+            // TODO: we need "auth loss" exposed to set the component group filters below correctly.
```
jamiebrynes7: Do we have a ticket tracking this?

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
```diff
@@ -2,60 +2,324 @@
 
 namespace Improbable.Gdk.TransformSynchronization
 {
+    /// <summary>
+    ///     A collection of utility functions for use with the Transform Synchronization Feature Module.
+    /// </summary>
     public static class TransformUtils
```
jamiebrynes7: Should everything in here be `public`? Or can some be internal?

CHANGELOG.md
paulbalaji: https://en.wikipedia.org/wiki/Q_(number_format)

workers/unity/Packages/com.improbable.gdk.transformsynchronization/TransformUtils.cs
paulbalaji: the only things left public now are:
- `CreateTransformSnapshot` using `Vector3` for location
- `CreateTransformSnapshot` using `Coordinates` for location
- `ToCoordinates` from `Vector3`

UPGRADE_GUIDE.md
```diff
@@ -0,0 +1,34 @@
...
+
+## From `0.2.3` to `Unreleased`
+
+### General changes
+
+When constructing a `ForwardingDispatcher`, you must now provide a `UnityEngine.LogType` instead of a `Improbable.Worker.CInterop.LogLevel`.
+
+### Transform Synchronization changes
+
+Several utility methods in the `TransformUtils` class have been made `internal` to the Transform Synchronization package. We highly recommend that **if you were using the `TransformInternal` component in your logic, please use `UnityEngine.Transform` instead.**
```
jamiebrynes7:
```suggestion
Several utility methods in the `TransformUtils` class have been made `internal` to the Transform Synchronization package. We highly recommend that **if you were using the `TransformInternal` component in your logic, to use `UnityEngine.Transform` instead.**
```

Also the utility methods being made internal shouldn't affect a user?

UPGRADE_GUIDE.md
```diff
@@ -0,0 +1,34 @@
...
+// Sample of INVALID code - the `Location` and `Quaternion` schema types have been removed.
+var transform = new TransformInternal.Snapshot
+{
+    Location = new Location((float) coords.X, (float) coords.Y, (float) coords.Z),
+    Rotation = new Quaternion(1, 0, 0, 0),
+    TicksPerSecond = 1f / Time.fixedDeltaTime
+};
+```
+
+Instead, you must use the `CreateTransformSnapshot` method provided by the `TransformUtils` class. This is to ensure that the contents of an entity's `TransformInternal` component are compressed using the Transform Synchronization module's new compression scheme.
```
jamiebrynes7:
```suggestion
Instead, you must use the `TransformUtils.CreateTransformSnapshot` static method. This is to ensure that the contents of an entity's `TransformInternal` component are compressed using the Transform Synchronization module's new compression scheme.
```

PR #991, opened 2019-06-17
Fix code generation path escaping

jamiebrynes7:
#### Description
See title. There were a few un-escaped paths.

Changelog incoming


jamiebrynes7:
This resolves #992

PR #995, opened 2019-06-20
enabling scaler v2

jessicafalk:


efarem:
For reference this PR adds the :

- `scaler_version=2` (required) - Use BuildKite scaler version 2
- `agent_count=1` (required) - The number of BuildKite agents that will spawn on each instance
- `minimum_instances=1` (optional) - Means that the BuildKite scaler will never scale below 1 instance

Scaling downward is currently restricted to out of the London office's working hours (7pm-9am).

The machines created, by default will use a machine type of `n1-standard-4`, should you require a different machine size you can use the keys from this map https://github.com/improbable/platform/blob/master/go/src/improbable.io/lib/nodesizes/nodesizes.go#L24-L68 e.g. `machine_type="octa"` will tell the scaler to use `n1-standard-32`

PR #1012, opened 2019-07-02
Split build target from build environment

paulbalaji:
#### Description

- relies on https://github.com/spatialos/gdk-for-unity-shared-ci/pull/39
- separate build environment from build target when building workers
- renamed `BUILD_TARGET` to `BUILD_ENVIRONMENT` to accurately capture what this variable was actually doing ("local"/"cloud")
- introduced a new `BUILD_TARGETS` which specifies what platform to build for (`ios`/`android` etc)
	- if unspecified, shared-ci will default this to "all" platforms for an environment (e.g. UnityClient cloud will build for mac and windows)
	- can specify multiple targets using comma separated values (e.g. `win,winx86,macos`)
note:
- includes temp changes to bootstrap whilst the shared-ci PR is not yet merged

#### Tests

- ran CI

#### Documentation

- [x] changelog (breaking changes entry)
- [x] upgrade guide
- [ ] update build system module docs

#### Primary reviewers


workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -32,17 +32,46 @@ public static void Build()
             try
             {
                 var args = CommandLineArgs.FromCommandLine();
-                var buildTargetArg = args.GetCommandLineValue("buildTarget", "local");
+                var buildEnvironmentArg = args.GetCommandLineValue("buildEnvironment", "local");
+                var buildTargetArg = args.GetCommandLineValue("buildTarget", "All");
```
jamiebrynes7: This is a breaking change as this is a public API - make sure to note in the changelog + upgrade guide. 

Also need to update https://docs.improbable.io/unity/alpha/modules/build-system/cli

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -32,17 +32,46 @@ public static void Build()
...
                         break;
+                    default:
+                        throw new BuildFailedException("Unknown build environment value: " + buildEnvironmentArg);
+                }
+
+                BuildTarget buildTarget;
+                switch (buildTargetArg.ToLower())
+                {
+                    case "all":
+                        buildTarget = BuildTarget.NoTarget;
```
jamiebrynes7: `all` should not map to NoTarget...

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
jamiebrynes7: Also why not support multiple build targets with comma delineated arguments given we are supporting this. I.e. - `win,linux`

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -165,12 +194,14 @@ private static bool BuildWorkers(string[] workerTypes, BuildEnvironment buildEnv
...
         }
 
-        private static bool BuildWorkerForEnvironment(string workerType, BuildEnvironment targetEnvironment, ScriptingImplementation? scriptingBackend = null)
+        private static bool BuildWorkerForEnvironment(string workerType, BuildEnvironment targetEnvironment, BuildTarget buildTarget, ScriptingImplementation? scriptingBackend = null)
         {
             var spatialOSBuildConfiguration = BuildConfig.GetInstance();
             var environmentConfig = spatialOSBuildConfiguration.GetEnvironmentConfigForWorker(workerType, targetEnvironment);
 
-            var enabledTargets = environmentConfig?.BuildTargets.Where(t => t.Enabled).ToList();
+            var enabledTargets = buildTarget == BuildTarget.NoTarget
```
jamiebrynes7: This feels like a good time to refactor into:

```csharp
bool BuildWorkerForEnvironment(string workerType, BuildEnvironment targetEnvironment, ScriptingImplementation? scriptingBackend = null);

bool BuildWorkerForEnvironmentAndTarget(string workerType, BuildEnvironment targetEnvironment, BuildTarget target, ScriptingImplementation? scriptingBackend = null);
```

then call the correct one according to if we are looking for a particular target over a all targets 

(If you end up adding comma-delineated support, you'll want to make it a `IEnumerable<BuildTarget>` but a similar argument applies)

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
paulbalaji: will update the PR description to do these^

paulbalaji: wish granted 🎉

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -44,9 +47,33 @@ public static void Build()
                         buildEnvironment = BuildEnvironment.Local;
                         break;
                     default:
-                        throw new BuildFailedException("Unknown build target value: " + buildTargetArg);
+                        throw new BuildFailedException("Unknown build environment value: " + buildEnvironmentArg);
                 }
 
+                var buildTargets = buildTargetsArg
+                    .Split(',')
+                    .SkipWhile(target => target == "all")
```
jamiebrynes7: If you have something like `all,ios,all` this will throw the unknown build target value. We should enforce that its `all` or one or more specific targets, not a mix (with a more descriptive error message)

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -165,12 +196,18 @@ private static bool BuildWorkers(string[] workerTypes, BuildEnvironment buildEnv
...
+            BuildEnvironment buildEnvironment,
+            IEnumerable<BuildTarget> buildTargets,
+            ScriptingImplementation? scriptingBackend = null)
         {
             var spatialOSBuildConfiguration = BuildConfig.GetInstance();
-            var environmentConfig = spatialOSBuildConfiguration.GetEnvironmentConfigForWorker(workerType, targetEnvironment);
+            var environmentConfig = spatialOSBuildConfiguration.GetEnvironmentConfigForWorker(workerType, buildEnvironment);
 
-            var enabledTargets = environmentConfig?.BuildTargets.Where(t => t.Enabled).ToList();
+            var enabledTargets = buildTargets.Any()
```
jamiebrynes7: This still feels wrong - an empty build targets array = build them all?

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -165,12 +196,18 @@ private static bool BuildWorkers(string[] workerTypes, BuildEnvironment buildEnv
...
+                ? environmentConfig?.BuildTargets.Where(t => buildTargets.Contains(t.Target)).ToList()
```
jamiebrynes7: What about the `Enabled` property?

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
jamiebrynes7: I'd rather have a more comprehensive refactor to enable this behaviour

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
paulbalaji: I think if we pass in a set of targets, the expectation is that the target gets built regardless of whether the build config sets it as enabled

workers/unity/Packages/com.improbable.gdk.buildsystem/WorkerBuilder.cs
paulbalaji: Given the only real difference between passing in targets and inferring them is the contents of `enabledTargets`, I'm not really sure what the added value of splitting it out into different functions is

PR #1013, opened 2019-07-03
Add `LinkedGameObjectMap` API

jamiebrynes7:
#### Description
Added an API to expose the `EntityId -> [GameObject]` relationship. This is wrapped in an object to ensure that we expose a list which is readonly (we really don't want users to mutate the underlying data). 

Also added a subscription manager for this (only in the GameObjectCreation module as the linker actually lives in a system here!) and added tests for this. 

**Commits are ordered and logical!**

#### Tests
Added some tests for the subscription fulfillment. 

#### Documentation
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Added the `LinkedGameObjectMap` class for finding the `GameObject`(s) linked with a specified `EntityId`.
+- Added the `LinkedGameObjectMap` class for finding the `GameObject`(s) linked with a specified `EntityId`. [#1013](https://github.com/spatialos/gdk-for-unity/pull/1013)
     - This can be used with the `[Require]` annotation to inject it into your `MonoBehaviours` provided you are using the `GameObjectCreation` feature module. For example: `[Require] private LinkedGameObjectMap gameObjectMap;`
 
 ### Internal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Added
 
 - Added the `LinkedGameObjectMap` class for finding the `GameObject`(s) linked with a specified `EntityId`.
-    - This can be used with the `[Require]` annotation to inject it into your `MonoBehaviours` provided you are using the `GameObjectCreation` feature module.
+    - This can be used with the `[Require]` annotation to inject it into your `MonoBehaviours` provided you are using the `GameObjectCreation` feature module. For example: `[Require] private LinkedGameObjectMap gameObjectMap;`
 
 ### Internal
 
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Added
+
+- Added the `LinkedGameObjectMap` class for finding the `GameObject`(s) linked with a specified `EntityId`.
+    - This can be used with the `[Require]` annotation to inject it into your `MonoBehaviours` provided you are using the `GameObjectCreation` feature module.
```
zeroZshadow: I would rephrase this line a bit to make it clearer what you put require on to have this work.

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/LinkedGameObjectMap.cs
```diff
@@ -0,0 +1,51 @@
+using System.Collections.Generic;
+using Improbable.Gdk.Core;
+using UnityEngine;
+
+namespace Improbable.Gdk.Subscriptions
+{
+    /// <summary>
+    ///     Represents the mapping between SpatiaLOS entity IDs and linked GameObjects.
```
paulbalaji:
```suggestion
    ///     Represents the mapping between SpatialOS entity IDs and linked GameObjects.
```

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/LinkedGameObjectMap.cs
```diff
@@ -0,0 +1,51 @@
...
+        ///     if any are linked; otherwise, null. This parameter is passed uninitialized.
+        /// </param>
+        /// <returns>True, if there are any GameObjects linked to the <see cref="EntityId"/>; otherwise false</returns>
+        public bool TryGetLinkedGameObjects(EntityId entityId, out IReadOnlyList<GameObject> linkedGameObjects)
+        {
+            linkedGameObjects = GetLinkedGameObjects(entityId);
+            return linkedGameObjects != null;
+        }
+    }
+
```
paulbalaji: nit: empty lines

workers/unity/Packages/com.improbable.gdk.core/Subscriptions/LinkedGameObjectMap.cs
jamiebrynes7: Its actual spatia line-of-sight

CHANGELOG.md
jamiebrynes7: Added an example at the end?

zeroZshadow: 👍

PR #1017, opened 2019-07-05
Update ScoreCollision.cs

KazikBozia:
This code lines is unnecessary


zeroZshadow:
@KazikBozia Sorry but we do not currently accept community PR's. The scoring system is an old test we disabled, and will be removing entirely soon.

Issue #1026, opened 2019-07-11
Missing exclamation mark on if statement in PlayerLifecycleHelper.cs


TagXy:
**Affects**: Release v0.2.4

-------

#### Description
Hi! Sorry if I'm wrong on this, however, in PlayerLifecycleHelper.cs , I saw what should be an error in PlayerLifecycleHelper.IsOwningWorker() (line:36) :
```
if (entitySystem.GetEntitiesInView().Contains(entityId))
{
                throw new InvalidOperationException(
                    $"Entity with SpatialOS Entity ID {entityId.Id} is not in this worker's view");
}
```
An entityId found will throw the error since `HashSet<T>.Contains()` return true if something is found : 

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.contains?view=netframework-4.8#System_Collections_Generic_HashSet_1_Contains__0_

Link of the line : 
https://github.com/spatialos/gdk-for-unity/blob/36cf14b1a230b37a1603b2a034a92f13e01b6086/workers/unity/Packages/com.improbable.gdk.playerlifecycle/PlayerLifecycleHelper.cs#L36

#### Workaround
shouldn't it be : 
`if (!entitySystem.GetEntitiesInView().Contains(entityId))`
In other words, adding the missing exclamation mark would simply fix it.

PS: I'm not used to GitHub, and sorry for my english.

jamiebrynes7:
Hey @TagXy thanks for the report! I've submitted a fix for this in #1027 and added some test cases to cover this sort of thing.

This fix should be released in `v0.2.5`

jamiebrynes7:
This bug was released yesterday in [0.2.5](https://github.com/spatialos/gdk-for-unity/releases/tag/0.2.5)!

PR #1027, opened 2019-07-11
Fix `PlayerLifecycleHelper.IsOwningWorker`

jamiebrynes7:
#### Description
Fixed the bug described by #1026 and added test coverage for this method.


zeroZshadow:
Thinking about it, should it maybe return false instead of throw?

PR #1042, opened 2019-07-18
Fix mac symlinks

zeroZshadow:
#### Description
Somehow `dotnet run -p <project>` breaks if the filepath contains a symlinked folder, and the has local dependencies.
This fix makes sure to resolve the symlink for the code generator to resolve this issue.


workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -82,6 +82,25 @@ private static void Generate()
                 var projectPath = Path.GetFullPath(Path.Combine(Common.GetThisPackagePath(),
                     CsProjectFile));
 
+
```
paulbalaji: nit: extra line

PR #1052, opened 2019-07-19
Worker SDK 13.8.2 Upgrade

jamiebrynes7:
#### Description
Upgraded the Worker SDK to 13.8.2:

- Bumped package versions
- Took Jared's schema bundle implementation 
- Updated our generation for the new schema bundle format.

Follow up PRs will: 
- [x] Add implementation for schema `entity` type. #1053 
- [x] Pull exhaustive generated code package instead of our one. #1056 

#### Tests
All tests are green! Fixed old ones.

#### Documentation
Changelog

#### Primary reviewers


jamiebrynes7:
> As mentioned, I think some of these classes can be cleaned and made readonly to show their intent as immutable data more clearly.

I agree entirely. I think we should defer this until we start on the modular codegen work since we will heavily refactoring these files then

PR #1069, opened 2019-07-30
Fix code snippet

jamiebrynes7:
Fix code sample


paulbalaji:
cherry-pick into live docs as well pls :)

PR #1071, opened 2019-07-30
Fix subscriptions coupling

jamiebrynes7:
#### Description
@zeroZshadow and I found a bug where subscriptions could get coupled together through the following series of events:

- Two (or more) different MonoBehaviours both require either a: `World`, `WorkerId,` or `LogDispatcher`. (These were cached incorrectly)
- These are enabled and then disabled.
- At this point, we have multiple references to the same handlers in the `SubscriptionAggregate.Handler.Pool`.
- As these handlers get popped off the pool, the new subscriptions set the `aggregate` field. However, since everything is a reference type, there is contention between the `aggregate` fields. Only one of the subscriptions now has the correct `handler.aggregate` field. 
- Different subscriptions on different `MonoBehaviour`s/`GameObject`s are now coupled. One being disabled could affect the other causing very strange behaviour.

This PR fixes this bug by removing the caching for the offending subscription types and adds tests for all hand-implemented subscription managers. This will increase the amount of GC slightly, but a more comprehensive fix will require a larger rework of the subscriptions code.

#### Tests
Yes there are tests for this bug.

#### Documentation
- [x] Changelog

#### Primary reviewers


jamiebrynes7:
So we can try to enforce that any new subscription managers get test entries added by finding all subscription managers (that aren't in generated code) and doing some reflection magic to check that we have a test for that case.

Sadly, we can't auto-generate the test cases since we can't use `Reflection.Emit` or similar in IL2CPP builds (notably iOS). 

Thoughts?

test-project/Assets/PlaymodeTests/SubscriptionAggregateCouplingTests.cs
```diff
@@ -0,0 +1,171 @@
...
+            connectionHandler.RemoveComponent(entityId, Position.ComponentId);
+            connectionHandler.RemoveEntity(entityId);
+            receiveSystem.Update();
+            requireLifecycleSystem.Update();
+
+            Object.DestroyImmediate(gameObject);
+        }
+
+        // Drop all logs, we don't care about them for this test.
+        private class IgnoreLogsDispatcher : ILogDispatcher
```
zeroZshadow: Is this still something that's true? It feels easy to erase errors with this.

jamiebrynes7: I can check to see if it fires off any log messages now. Sadly we'd need to "expect" all of them which is slightly annoying 😕

jamiebrynes7: Didn't need this in the end 💯

ci/test.sh
```diff
@@ -18,6 +18,8 @@ CODE_GENERATOR_TEST_RESULTS_FILE="${PROJECT_DIR}/logs/code-generator-test-result
 EDITMODE_TEST_RESULTS_FILE="${PROJECT_DIR}/logs/editmode-test-results.xml"
 PLAYMODE_TEST_RESULTS_FILE="${PROJECT_DIR}/logs/playmode-test-results.xml"
 TEST_PROJECT_EDITMODE_TEST_RESULTS_FILE="${PROJECT_DIR}/logs/test-project-editmode-test-results.xml"
+TEST_PROJECT_PLAYMODE_TEST_RESULTS_FILE="${PROJECT_DIR}/logs/test-project-playmode-test-results.xml"
+
```
paulbalaji: newline

ci/test.sh
```diff
@@ -66,3 +68,15 @@ pushd "test-project"
         -logfile "${PROJECT_DIR}/logs/test-project-editmode-test-run.log" \
         -testResults "${TEST_PROJECT_EDITMODE_TEST_RESULTS_FILE}"
 popd
+
+echo "--- Testing Unity: Test Project Playmode :joystick:"
```
paulbalaji: 💯 emoji game strong

test-project/Assets/PlaymodeTests/Subscriptions/SubscriptionAggregateCouplingTests.cs
```diff
@@ -0,0 +1,201 @@
...
+    public class SubscriptionAggregateCouplingTests
+    {
+        private MockConnectionHandler connectionHandler;
+        private WorkerInWorld workerInWorld;
+        private EntityGameObjectLinker linker;
+
+        private SpatialOSReceiveSystem receiveSystem;
+        private RequireLifecycleSystem requireLifecycleSystem;
+
+        private const string WorkerType = "TestWorkerType";
```
paulbalaji: we have "TestWorkerType" sprinkled throughout our tests - should there be one place where we define this?

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/LinkedGameObjectMapSubscriptionManager.cs
```diff
@@ -17,15 +16,12 @@ public LinkedGameObjectMapSubscriptionManager(World world)
 
         public override Subscription<LinkedGameObjectMap> Subscribe(EntityId entityId)
         {
-            if (linkedGameObjectMapSubscription == null)
-            {
-                linkedGameObjectMapSubscription = new Subscription<LinkedGameObjectMap>(this, new EntityId(0));
+            var linkedGameObjectMapSubscription = new Subscription<LinkedGameObjectMap>(this, new EntityId(0));
```
paulbalaji: random point, but considering we use `new EntityId(0)` in a few places, should we include a built-in `EntityId.zero`?

jamiebrynes7: Tbh I don't think so - the utility to users is probably pretty low.

test-project/Assets/PlaymodeTests/Subscriptions/SubscriptionAggregateCouplingTests.cs
jamiebrynes7: It actually doesn't matter at all what you put here 😬

PR #1085, opened 2019-08-01
Tidy up test utils package

jamiebrynes7:
#### Description
Was going to add some stuff to this, but first noticed that its rotted a fair bit. Ended up being able to delete most of it. (Will add some more stuff back in soon ™️).

- Removed `WorkerOpFactory` as its not used. Use `MockConnectionHandler` instead (afaict you can't even use this anymore!)
- Removed `TestMonoBehaviour` as its not used and not something that needs to be provided in a package..
- Removed `HybridGdkSystemTestBaseTests.cs` as it was a relic of a previous era and is no longer testing anything useful. 
- Moved `DisableAutoCreationSystemTests.cs` to the test-project as it feels like a better fit there. 

#### Tests

Everything still runs fine 😄 

#### Documentation
- [x] Changelog

Maybe upgrade guide? But no one could have been really been using `WorkerOpFactory` since ops have been abstracted away for a while and `TestMonoBehaviour` doesn't really need an upgrade guide entry imo. Thoughts?


workers/unity/Assets/Plugins.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: 34c0720497daf334290b4cb0ac5f7db2
+guid: a2bb714765233254eb4d9463c199df91
```
zeroZshadow: ? Why is this file changed?

jamiebrynes7: Discussed offline - this is created due to Rider plugins every time we open the project. However, since nothing else is in the folder and hence not checked in, we have a `.meta` file and no folder.

Solution - add a dummy file to the Plugins directory so it can be committed and the meta file won't change 😄

PR #1091, opened 2019-08-05
Bump `current_version` variable

jamiebrynes7:
In preparation for the `0.2.6` release. Bump this.


jamiebrynes7:
Well, we can re-run CI after FPS has been tagged and then merge this in

PR #1097, opened 2019-08-06
Proof of Concept: Platform SDK as a package

jamiebrynes7:
#### Description
This PR adds an `io.improbable.worker.sdk.platform` package which contains the Platform SDK. Note that we are using a specific commit of the `Platform SDK` and building it from source since we _really really_ want the ILPacked version to get out of dependency hell.

When `14.1` releases, we should just be able to grab the `NuGet` package as normal.


workers/unity/Packages/io.improbable.worker.sdk.platform/.gitignore
```diff
@@ -0,0 +1,7 @@
+# Ignore hidden folders
+/.*/
+
+# Ignore everything besides .meta in Plugins
+Plugins/**/*.*
+!Plugins/**/*.meta
+!.npmignore
```
zeroZshadow: This needs an npmignore file, otherwise the package wont include the meta files.

zeroZshadow: You can check for correctness by using `npm pack` in the package folder, and see if the resulting tar includes the right files.

jamiebrynes7: Yeah I know 😉 I just threw this up as a proof of concept so didn't do the entire process.

I don't think this is strictly merge-able in its current state given that we are compiling a pre-release version of the Platform SDK.

PR #1099, opened 2019-08-08
Add scaler v2 to docs pipeline

jamiebrynes7:
See title, need to do this by Aug 14th


jamiebrynes7:
Forgot I needed a `docs/` prefix

PR #1110, opened 2019-08-13
Make Options readonly, fix tests

zeroZshadow:
#### Description
As per sami's old PR
Make Options readonly, fix tests

#### Tests
Updated tests

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.core/Tests/Editmode/Components/OptionTests.cs
+++ b/workers/unity/Packages/io.improbable.gdk.core/Tests/Editmode/Components/OptionTests.cs
@@ -6,17 +6,10 @@ namespace Improbable.Gdk.Core.EditmodeTests
     public class OptionTests
     {
         [Test]
-        public void Parameterless_constructor_creates_empty_option_for_value_types()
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/io.improbable.gdk.core/Tests/Editmode/Components/OptionTests.cs
```diff
@@ -179,9 +146,9 @@ public void Implicit_conversion_to_payload_type_works()
         public void Implicit_conversion_to_payload_type_using_empty_option_throws()
         {
             var option = new Option<bool>();
-            Assert.Throws<CalledValueOnEmptyOptionException>(() =>
+            Assert.Throws<EmptyOptionException>(() =>
             {
-                var _ = option.Value;
+                bool _ = option.Value;
```
jamiebrynes7: The test name and content don't really match up.

I'd expect something like: 

```
bool _ = option;
```

to actually test the `Option<T> -> T` implicit conversion

zeroZshadow: Knew I forgot something, thanks!

PR #1118, opened 2019-08-15
Fix local launch on mac

paulbalaji:
#### Description

local launches through editor should work again on macos

#### Tests

tried local launch through editor on macos

#### Documentation

n/a because bug caught before release

#### Primary reviewers


paulbalaji:
tested on Windows - doesn't break 🎉

paulbalaji:
Validated the fix works on both MacOS and Windows

Issue #1057, opened 2019-07-22
Mono crash when using recursive options in schema


jamiebrynes7:
**Affects:** Release v0.2.3 and up (observed in Unity 2019.1.3)
**Internal Ticket:** [UTY-2091](https://improbableio.atlassian.net/browse/UTY-2091)

---

#### Description

Schema that attempts to use an option recursively:

```text
type TypeA {
	option<TypeA> test = 1;
}
```

Will cause a Mono crash like the following:

```
* Assertion at ..\mono\metadata\class.c:1998, condition `image_is_dynamic (klass->image) || field_offsets [i] > 0' not met
```

#### Workaround

There is no workaround to allow recursive options, you can emulate this with a `list` with 0 or 1 elements. Support for reference-semantic options will enable this.

---

To get yourself back to a good state:

1. Remove the problematic schema entry.
2. Delete the `Assets/Generated/Source` directory.
3. Re-open Unity.

jamiebrynes7:
Fixed in release v0.2.8

PR #1131, opened 2019-08-20
UTY-2091: Fix crash on recursive options

jamiebrynes7:
#### Description
Historically, we have not supported recursive options due to reference-semantic options imposing serious performance penalties. However, this is supported at the schema level. Previously, this would simply cause a compile error, but recently it started causing a Mono hard crash.

This PR fixes this bug by explicitly skipping recursive optional fields in schema types. Additionally, this PR contains updates to the `exhaustive_bundle.json` used for testing and some extra tests for the recursive schema case.

The first 5 commits are mostly housekeeping, the commits after those will be the more interesting ones.

#### Tests
- There are some nice little test cases in here.

#### Documentation
- [x] Changelog
- [x] Add a note to the docs that recursive options are skipped. #1132

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Workers will now log their `PlayerConnection` ports to SpatialOS after connecting. This port can be used for connecting the Unity profiler to workers running in the cloud. [#1128](https://github.com/spatialos/gdk-for-unity/pull/1128)
     - Note that this will only happen if the worker was built as a "Development Build".
 
+### Fixed
+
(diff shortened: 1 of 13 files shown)


jamiebrynes7:
Fixes #1057

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- Fixed a bug where recursive options in schema types would cause a Mono hard crash. [#1131](https://github.com/spatialos/gdk-for-unity/pull/1131)
```
zeroZshadow: This should mention, we skip them now.

PR #1132, opened 2019-08-20
Add warning about recursive schema options

jamiebrynes7:
#### Description
Add a warning that we will skip these fields. Also noted that we do actually support `double`s 😉

### DIFF
--- a/docs/reference/concepts/code-generation.md
+++ b/docs/reference/concepts/code-generation.md
@@ -24,7 +24,7 @@ The code generator maps from the [schema primitive types](https://docs.improbabl
 | `uint64`                       |               `ulong`                |
 | `float`                        |               `float`                |
 | `double`                       |               `double`               |
-| `bool`                         |           `BlittableBool`            |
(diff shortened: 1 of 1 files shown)


docs/reference/concepts/code-generation.md
```diff
@@ -23,6 +23,7 @@ The code generator maps from the [schema primitive types](https://docs.improbabl
 | `int64` / `sint64`/ `fixed64`  |                `long`                |
 | `uint64`                       |               `ulong`                |
 | `float`                        |               `float`                |
+| `double`                       |               `double`               |
```
paulbalaji: woops

docs/reference/concepts/code-generation.md
```diff
@@ -23,6 +23,7 @@ The code generator maps from the [schema primitive types](https://docs.improbabl
...
 | `bool`                         |           `BlittableBool`            |
```
zeroZshadow: Much Blittable

PR #1133, opened 2019-08-21
Port Forwarding UI

jamiebrynes7:
#### Description
Adds a UI window to complement the feature implemented in #1128. This allows users to port forward without touching the CLI.

- Validation against user input
- Little responsive UI telling you what state the connection is in.
- Assembly reloading locked while the window is up.

~**Question about the lifetime of the port forwarding: While the window is open? While Unity is open?**~

~If the latter, we will need to serialize the process `Task` _somehow_.~

Decided to tie the lifetime to the window lifetime.

Screenshots:

image
_entering data_

image
_connected successfully_

image
_failed attempt_

#### Tests
Tested the various states for connecting and the validation logic.

#### Documentation
- [x] Changelog
- [x] Update #1130 to use the GUI and only mention the CLI as an option in an expandable.

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
+++ b/workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
@@ -93,6 +93,11 @@ namespace Improbable.Gdk.Tools
             }
         }
 
+        private void OnDisable()
+        {
+            Cleanup();
+        }
+
         private void OnDestroy()
         {
             Cleanup();
@@ -168,10 +173,13 @@ namespace Improbable.Gdk.Tools
 
         private void Cleanup()
         {
-            activePortForwarding.Dispose();
-            activePortForwarding = null;
-            tokenSrc.Dispose();
+            tokenSrc?.Cancel();
+            tokenSrc?.Dispose();
             tokenSrc = null;
+
+            activePortForwarding?.Wait();
+            activePortForwarding?.Dispose();
+            activePortForwarding = null;
 
             EditorApplication.UnlockReloadAssemblies();
         }
--- a/workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
+++ b/workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
@@ -10,9 +10,11 @@ namespace Improbable.Gdk.Tools
 {
     internal class PortForwardingWindow : EditorWindow
     {
-        private string deploymentName = "";
-        private string workerId = "";
-        private int port;
+        [SerializeField] private string deploymentName = "";
+
+        [SerializeField] private string workerId = "";
+
+        [SerializeField] private int port;
 
         private Task<RedirectedProcessResult> activePortForwarding;
         private CancellationTokenSource tokenSrc;
--- a/workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
+++ b/workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
@@ -1,0 +1,179 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+using UnityEditor;
+using UnityEngine;
+
+namespace Improbable.Gdk.Tools
+{
+    internal class PortForwardingWindow : EditorWindow
+    {
+        private string deploymentName = "";
+        private string workerId = "";
+        private int port;
+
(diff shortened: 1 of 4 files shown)


workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
```diff
@@ -0,0 +1,179 @@
...
+                messageType = MessageType.Error;
+                message = "Failed to establish connection. Check the Unity Console for more information.";
+            }
+
+            if (message != null)
+            {
+                EditorGUILayout.HelpBox(message, messageType);
+            }
+
+            if (activePortForwarding != null)
```
jared-improbable: Can combine the two if statements

jamiebrynes7:
> One option could be to serialize the PID out to a file.

I guess the problem is we can't restore all the information that is associated with the `Task<RedirectedProcessOutput>` with just the PID 😕 

> What are the implications to user workflow to having this window open and assembly reloading locked?

Given the expected usage while the port forwarding is active is that you are profiling (and maybe debugging) - I think this is okay. 
Its a safety measure that we kind of have to implement since the hot reload will lose all information about the state of the window. 

I can't think of a use case where I'm profiling and want to make changes and see the hot reload take place.

workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
jamiebrynes7: True, although I feel like the intent is clearer as written.

I read this as: 

- If the task isn't running, don't render the button.
- If the button is clicked, cancel the task.

Whereas if its combined:

- If the task is running and the button is clicked cancel the task

Thoughts?

jared-improbable: I don't feel particularly strongly either way!

workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using UnityEditor;
+using UnityEngine;
+
+namespace Improbable.Gdk.Tools
+{
+    internal class PortForwardingWindow : EditorWindow
+    {
+        private string deploymentName = "";
```
zeroZshadow: mark these fields as serialized

workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
```diff
@@ -0,0 +1,187 @@
...
+                return;
+            }
+
+            // Cancelling or killing the process will result in a non-zero exit code, so ensure that we weren't
+            // previously connected successfully.
+            portForwardingHasFailed = activePortForwarding.Result.ExitCode != 0 && !isConnected;
+
+            if (portForwardingHasFailed)
+            {
+                foreach (var error in activePortForwarding.Result.Stdout)
```
zeroZshadow: Do we not want to concatenate this into 1 error?

workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        private IEnumerable<string> GetErrors()
+        {
+            if (port > ushort.MaxValue || port < 1)
+            {
+                yield return "Port must be a number in the range of 1 to 65535.";
+            }
+
+            if (string.IsNullOrEmpty(deploymentName))
```
zeroZshadow: Do note that this technically means you can iterate over the `GetErrors()` while changing the deploymentName during the iteration. That's a bit flaky. Might be better just to return a list of errors?

workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
jamiebrynes7: Done!

workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
jamiebrynes7: Discussed offline - increases noise from the error messaging

workers/unity/Packages/io.improbable.gdk.tools/PortForwardingWindow.cs
jamiebrynes7: I kinda prefer this pattern since it makes bubbling up errors much nicer. The caller has to be aware of `IEnumerable` semantics (which I would expect)

PR #1140, opened 2019-09-02
2019.2.0 docs changes

jamiebrynes7:
#### Description

`/s/2019.1.3/2019.2.0/g`


jamiebrynes7:
Remembered why we didn't use a variable previously - the variable macros are not expanded within an `Include` macro (which the setup page is exclusively used in)

PR #1151, opened 2019-09-09
Modular codegen first pass

jamiebrynes7:
This PR implements a first pass at modular codegen which utilizes [dotnet templates](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-new?tabs=netcore22) to assist in generating a skeleton project.

The process goes as follows:

1. User hits code generation button.
2. Install the dotnet template by shelling out `cd $UNITY_PACKAGES_DIR/io.improbable.gdk.tools/.CodeGenTemplate/ && dotnet new -i ./`
3. Wipe `$UNITY_BUILD_SCRATCH/codegen` directory.
4. Create a copy of the dotnet template: `cd $UNITY_BUILD_SCRATCH/codegen && dotnet new gdk-for-unity-codegen`
5. Find all `.codegen` directories in packages.
6. Edit the generated `.csproj` and link in any source, templates, and partials in each of the `.codegen` directories.
7. Run the generated csproj.

For this first implementation, any time a user runs codegen, all the steps above happen. This has have an impact on performance of operation:

image

What particularly hurts is we lose all build artifact caching built into `dotnet/msbuild` since we wipe the `codegen` directory _every time_. For future optimizations, we should focus on only doing this when necessary. 

<details>
<summary><b>Possible optimizations</b></summary>

* We only need to install the dotnet template if the template has changed. For any user consuming the GDK as packages, this will _only_ be on upgrade. For those iterating on the template semi-regularly (mostly us) this is a little harder to detect. 
* We don't need to wipe the `codegen` directory unless the template has changed.
* We don't need to create a copy of the dotnet template unless the template has changed. If we can replace the blob of XML that links the packages, we can update the _same_ `.csproj` file.

All of these optimizations rely on us detecting the `template` has changed. If we can do this, we can adjust the solution above to: 

1. User hits code generation button.
2. Check if the template has changed. 
	a. If yes, do steps 2-4 from above.
3. Find all `.codegen` directories in packages.
4. Edit the generated `.csproj` and link in any source, templates, and partials in each of the `.codegen` directories.
5. Run the generated csproj.

</details>

---

This PR  also contains a driveby refactor on `RedirectedProcess` such that the synchronous `Run()` now returns the `RedirectedProcessResult` type to make it consistent with the asynchronous one.

**Remaining work**
- [ ] Changelog
- [x] Investigate optimizations - https://github.com/spatialos/gdk-for-unity/pull/1151#issuecomment-529475056


jamiebrynes7:
### Optimization Results

* Only `Generate Code (Force)` will rebuild the project from scratch.
* We now support incrementally updating the `ItemGroup` containing each of the module's codegen bits.

This brought down the time from ~5000ms - ~3000ms after the first run.

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -253,11 +389,11 @@ private static IEnumerable<string> GetSchemaDirectories()
             }
 
             var packagePathsWithSchema = request.Result
```
zeroZshadow: Rename these to not reference schema

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -224,26 +253,133 @@ private static string[] ConstructArgs(string projectPath, string schemaCompilerP
...
+                var idEle = new XElement("GdkPackageSource");
+                idEle.SetAttributeValue("Include", dir);
+                itemGroup.Add(idEle);
+
+                var sourceDir = Path.Combine(dir, "Source");
+                if (Directory.Exists(sourceDir))
+                {
+                    var ele = new XElement("Compile");
+                    ele.SetAttributeValue("Include", Path.Combine(sourceDir, "**"));
+                    itemGroup.Add(ele);
```
jessicafalk: that whole if statement (and the ones below it) could be factored out into a method. that might also makes this a bit easier to read :)

jamiebrynes7: Discussed offline - this may not help legibility since its only generalizable in two places (the `Partial` one is different) without going down an `params (string nodeName, string attributeName)[]` approach on the helper method which seems a little bit overkill

PR #1153, opened 2019-09-10
Find codegen jobs via reflection

jamiebrynes7:
#### Description

This PR removes the hard dependency between the `EntryPoint` and the code generation job by finding all types that are non-abstract, inherit from `CodegenJob`, and don't have the `IgnoreCodegenJob` attribute on them (used for tests). 

- Also makes the `CodeGenerationOptions` accessible as a singleton instance.
- Also includes a few tweaks/fixes in how we generate the `csproj`.  

#### Tests

Ran codegen locally, everything worked as expected.

#### Documentation
N/A

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/EntryPoint.cs
+++ b/workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/EntryPoint.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using Improbable.Gdk.CodeGeneration.FileHandling;
 using Improbable.Gdk.CodeGeneration.Jobs;
 using Improbable.Gdk.CodeGeneration.Model;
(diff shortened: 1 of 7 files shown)


workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -337,7 +338,7 @@ private static void UpdateModules()
             var gdkItemGroups = projectNode
                 .Elements("ItemGroup")
                 .Where(ele => ele.Element("GdkPackageSource") != null)
-                .ToDictionary(ele => ele.Element("GdkPackageSource").Attribute("Include").Value, ele => ele);
+                .ToDictionary(ele => ele.Element("GdkPackageSource").Attribute("Remove").Value, ele => ele);
```
jamiebrynes7: This stops a phantom folder appearing in the `csproj` if you open it in VS/Rider.

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/EntryPoint.cs
```diff
@@ -60,12 +61,17 @@ public int Run()
             var bundlePath = GenerateBundle();
             var schemaBundle = SchemaBundle.LoadBundle(File.ReadAllText(bundlePath));
             var store = new DetailsStore(schemaBundle, options.SerializationOverrides);
-            var workerGenerationJob = new WorkerGenerationJob(options.NativeOutputDirectory, options, fileSystem);
-            var singleJob = new SingleGenerationJob(options.NativeOutputDirectory, store, fileSystem);
 
-            var runner = new JobRunner(fileSystem);
+            var jobs = AppDomain.CurrentDomain
```
zeroZshadow: This should probably guard against the `ReflectionTypeLoadException` and others.
Since the assembly might have loaded third-party dll's that mess with this.

PR #1154, opened 2019-09-10
Add test failure reporting in Buildkite

jamiebrynes7:
#### Description

This PR introduces a Buildkite plugin which will report test failures as annotations on a build:

image

This allows you to understand which tests are broken without diving into the logs. This feature is using the new `dag` feature from Buildkite which allows us to declare dependencies between jobs. The reporting step will only run after the tests has completed, but doesn't care about any of the other steps.


jamiebrynes7:
cc @jared-improbable you can use this as a reference implementation for doing this on the OS side. All you need is `docker` on the agent assigned to the plugin step.

PR #1155, opened 2019-09-10
UTY-2292: modular codegen tests

jamiebrynes7:
#### Description
Add a basic set of test cases for each of the type of objects we support in modular codegen: raw source code, templates, and partials. Also a small driveby fix to another test.

#### Tests
Yes I am test.

### DIFF
--- a/test-project/Assets/EditmodeTests/ModularCodegenTests.cs
+++ b/test-project/Assets/EditmodeTests/ModularCodegenTests.cs
@@ -1,0 +1,21 @@
+using System;
+using System.Linq;
+using NUnit.Framework;
+
+namespace Improbable.Gdk.EditmodeTests
+{
+    [TestFixture]
+    public class ModularCodegenTests
+    {
+        // This checks that the codegen module was correctly picked up.
+        [Test]
+        public void Improbable_Gdk_ModularCodegenTests_Test_exists()
+        {
+            // Assembly.FullName == "<name> <version> <culture> <publickeytoken>", so we just .Contains() it.
+            var generatedCodeAssembly = AppDomain.CurrentDomain.GetAssemblies()
+                .First(assembly => assembly.FullName.Contains("Improbable.Gdk.Generated"));
+
+            Assert.IsTrue(generatedCodeAssembly.GetTypes().Any(type => type.FullName == "Improbable.Gdk.ModularCodegenTests.Test"));
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


test-project/Assets/EditmodeTests/ModularCodegenTests.cs
```diff
@@ -0,0 +1,45 @@
...
+    [TestFixture]
+    public class ModularCodegenTests
+    {
+        private Assembly generatedCodeAssembly;
+
+        [OneTimeSetUp]
+        public void Setup()
+        {
+            // Assembly.FullName == "<name> <version> <culture> <publickeytoken>", so we just .Contains() it.
+            generatedCodeAssembly = AppDomain.CurrentDomain.GetAssemblies()
```
zeroZshadow: It might be safer to use `typeof` on a known type in the Generated assembly, and get the `.Assembly` through that.

test-project/Assets/EditmodeTests/ModularCodegenTests.cs
```diff
@@ -0,0 +1,45 @@
...
+                .First(assembly => assembly.FullName.Contains("Improbable.Gdk.Generated"));
+        }
+
+        // This checks that the codegen module was correctly picked up.
+        [Test]
+        public void Improbable_Gdk_ModularCodegenTests_Test_exists()
+        {
+            Assert.IsTrue(generatedCodeAssembly.GetTypes().Any(type => type.FullName == "Improbable.Gdk.ModularCodegenTests.Test"));
```
zeroZshadow: Why not use `generatedCodeAssembly.GetType("Improbable.Gdk.ModularCodegenTests.Test")` ? It will return null if the type was not found.

test-project/Assets/EditmodeTests/ModularCodegenTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+
+        [Test]
+        public void Improbable_Tests_ModularTarget_has_partial_injected()
+        {
+            var modularTargetType = generatedCodeAssembly.GetTypes()
+                .First(type => type.FullName == "Improbable.Tests.ModularTarget");
+
+            modularTargetType.GetNestedTypes().Any(type => type.Name == "InteriorClass");
```
zeroZshadow: `GetNestedType`?

PR #1158, opened 2019-09-11
update publish packages script

jamiebrynes7:
#### Description

Update package publishing script to prepare it for running it in CI.

This will:

- Grab the Worker SDK/`spot`/schema std lib
- Pack each package
- Push it to cloudsmith, overwriting any package that exists with that version (this allows us to overwrite the packages in the case where we detect a defect later in QA). 

There will be work that needs to be done in https://github.com/spatialos/gdk-for-unity-shared-ci to setup and run this script.

### DIFF
--- a/ci/publish_packages.sh
+++ b/ci/publish_packages.sh
@@ -27,6 +27,7 @@ pushd "${PKG_ROOT}" > /dev/null
             # Created package is the last line of output.
             PACKAGE_NAME=$(npm pack | tail -n 1)
             cloudsmith push npm spatialos/gdk-for-unity "${PACKAGE_NAME}" --republish "${EXTRA_CLOUDSMITH_ARGS}"
+            rm "${PACKGE_NAME}"
         popd > /dev/null
     done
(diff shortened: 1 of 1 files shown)


ci/publish_packages.sh
```diff
@@ -2,18 +2,26 @@
...
     do
-        echo $dir
-        pushd $dir
-            npm publish --registry="${REGISTRY}"
-        popd
+        echo "${package_dir}"
+        pushd "${package_dir}" > /dev/null
+            # Created package is the last line of output.
+            PACKAGE_NAME=$(npm pack | tail -n 1)
+            cloudsmith push npm spatialos/gdk-for-unity "${PACKAGE_NAME}" --republish
```
paulbalaji: Is `--republish` something cloudsmith-specific?

jamiebrynes7: Yes - it allows packages to overwrite existing versions. Otherwise if we find a regression we must manually pull them all.

PR #1173, opened 2019-09-23
Add component metaclass / command metaclasses

jamiebrynes7:
#### Description

In the quest to expose (component ID, command index) -> command name pairs in preparation for command network stats GUI work, we found that we could simplify and reduce the amount of reflection required at startup by introducing metadata classes.

Whereas previously, for every type we wanted to reflect over related to a component (i.e. - serializer, diff storage, etc) we would do a full search through all the assemblies for these types.

Now we statically generate this information in the metaclass and do reflection _once_ to find those metaclasses. This also has the benefit of allowing us to expose the information ((component ID, command index) -> command name pairs) that we originally wanted in a clean way.

Along the way we made a small breaking change to improve the coherency of the API.

#### Tests

- Ran playground 
- Ran tests

#### Documentation
- [x] Changelog
- [x] Upgrade guide
- ~❓ Docs?~ Turns out there are none

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.core/Dynamic/ComponentDatabase.cs
+++ b/workers/unity/Packages/io.improbable.gdk.core/Dynamic/ComponentDatabase.cs
@@ -1,43 +1,46 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using Improbable.Gdk.Core.Commands;
 
 namespace Improbable.Gdk.Core
 {
-    internal static class ComponentDatabase
+    public static class ComponentDatabase
     {
-        public static Dictionary<uint, IDynamicInvokable> IdsToDynamicInvokers { get; }
+        internal static Dictionary<uint, IComponentMetaclass> Metaclasses { get; }
 
-        public static Dictionary<Type, uint> ComponentsToIds { get; }
+        private static Dictionary<Type, uint> ComponentsToIds { get; }
 
-        public static Dictionary<Type, uint> SnapshotsToIds { get; }
+        private static Dictionary<Type, uint> SnapshotsToIds { get; }
 
         static ComponentDatabase()
         {
-            IdsToDynamicInvokers = new Dictionary<uint, IDynamicInvokable>();
-            ComponentsToIds = new Dictionary<Type, uint>();
-            SnapshotsToIds = new Dictionary<Type, uint>();
+            Metaclasses = ReflectionUtility.GetNonAbstractTypes(typeof(IComponentMetaclass))
+                .Select(type => (IComponentMetaclass) Activator.CreateInstance(type))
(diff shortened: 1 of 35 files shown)


workers/unity/Packages/io.improbable.gdk.core/Worker/MessagesToSend.cs
```diff
@@ -83,6 +90,8 @@ public MessagesToSend()
             typeToCommandStorage.Add(typeof(WorldCommands.DeleteEntity.Request), worldCommandStorage);
             typeToCommandStorage.Add(typeof(WorldCommands.ReserveEntityIds.Request), worldCommandStorage);
             typeToCommandStorage.Add(typeof(WorldCommands.EntityQuery.Request), worldCommandStorage);
+
+            Debug.Log("Y");
```
zeroZshadow: But Y?

workers/unity/Packages/io.improbable.gdk.core/Dynamic/ComponentDatabase.cs
```diff
@@ -1,43 +1,46 @@
...
-        public static Dictionary<uint, IDynamicInvokable> IdsToDynamicInvokers { get; }
+        internal static Dictionary<uint, IComponentMetaclass> Metaclasses { get; }
 
-        public static Dictionary<Type, uint> ComponentsToIds { get; }
+        private static Dictionary<Type, uint> ComponentsToIds { get; }
 
-        public static Dictionary<Type, uint> SnapshotsToIds { get; }
+        private static Dictionary<Type, uint> SnapshotsToIds { get; }
 
         static ComponentDatabase()
```
jessicafalk: do we not want an access modifier here

workers/unity/Packages/io.improbable.gdk.core/Dynamic/Dynamic.cs
```diff
@@ -21,6 +23,15 @@ namespace Improbable.Gdk.Core
 
     public static class Dynamic
     {
+        private static readonly Dictionary<uint, IDynamicInvokable> DynamicInvokers;
+
+        static Dynamic()
```
jessicafalk: access modifier

workers/unity/Packages/io.improbable.gdk.core/Dynamic/ComponentDatabase.cs
jamiebrynes7: Static constructors can't have access modifiers 😉

workers/unity/Packages/io.improbable.gdk.core/Dynamic/Dynamic.cs
jamiebrynes7: Same as above :)

PR #1183, opened 2019-10-09
Mobile SDK import settings fix + CodeGen casing

zeroZshadow:
#### Description
Mobile SDK Plugins had wrong import settings, which allowed them to clash in the editor.
This fixes the settings, and re-exports them in the new 2019.2 format.
Additionally fixes the codegenerator path to use the right casing for case-sensitive platforms.


zeroZshadow:
This will need a change in our versioning scheme for the worker sdk's. So don't merge this yet until that is done.

PR #1194, opened 2019-10-15
Choose emulator/device to launch android app on + refactor mobile launcher

paulbalaji:
#### Description

- added ability to choose which emulator/device to launch apk on
- heavily refactored `MobileLauncherWindow` to minimise duplicated code
- added some comments here and there

image
image

#### Tests

I've tested all possible emulator/device, local/cloud and android/ios combinations possible on both windows and mac (enumerated below).

Will re-validate this at the end

- Android
	- Windows
		- [x] emulator local
		- [x] emulator cloud
		- [x] device local
		- [x] device cloud
	- MacOS
		- [x] emulator local
		- [x] emulator cloud
		- [x] device local
		- [x] device cloud
- iOS
	- [ ] Simulator local
	- [ ] Simulator cloud
	- [x] device local
	- [x] device cloud

#### Documentation

- [ ] update docs on how to launch mobile clients
- [x] changelog

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.mobile/Editor/AndroidLaunchUtils.cs
```diff
@@ -1,29 +1,92 @@
...
 using UnityEngine;
 using Debug = UnityEngine.Debug;
 
 namespace Improbable.Gdk.Mobile
 {
-    // TODO UTY-2095: Improve Android workflow and refactor the code
     public static class AndroidLaunchUtils
     {
-        public static void Launch(bool shouldConnectLocally, string runtimeIp)
+        private static readonly Regex DeviceMatchRegex = new Regex("(?:(?<id>[\\w_\\d-]+)\\s*device).*"
```
jamiebrynes7: Pls comment for what this matches against

workers/unity/Packages/io.improbable.gdk.mobile/Editor/AndroidLaunchUtils.cs
```diff
@@ -1,29 +1,92 @@
...
+            + "(?:product:(?<product>[\\w_\\d]+))\\s*"
+            + "(?:model:(?<model>[\\w_\\d]+))\\s*"
+            + "(?:device:(?<device>[\\w_\\d]+)).*");
+
+        public static (Dictionary<string, string>, Dictionary<string, string>) RetrieveAvailableEmulatorsAndDevices()
```
jamiebrynes7: You can name the tuples here 😄

workers/unity/Packages/io.improbable.gdk.mobile/Editor/AndroidLaunchUtils.cs
```diff
@@ -1,29 +1,92 @@
...
+                    {
+                        availableDevices[$"{match.Groups["model"].Value} ({deviceId})"] = deviceId;
+                    }
+                })
+                .RedirectOutputOptions(OutputRedirectBehaviour.None)
+                .Run();
+
+            if (result.ExitCode == 0)
+            {
+                if (!available)
```
jamiebrynes7: `available` is basically equivalent to: `!availableEmulators.IsEmpty() || !availableDevices.IsEmpty()` right? 

If so, we could drop it entirely?

workers/unity/Packages/io.improbable.gdk.mobile/Editor/AndroidLaunchUtils.cs
```diff
@@ -1,29 +1,92 @@
...
+                {
+                    Debug.Log("No Android emulators or devices found.");
+                }
+
+                return (availableEmulators, availableDevices);
+            }
+
+            Debug.LogError("Failed to find Android emulators or devices.");
```
jamiebrynes7: Probably want to log the `stderr` in this case

workers/unity/Packages/io.improbable.gdk.mobile/Editor/AndroidLaunchUtils.cs
```diff
@@ -1,29 +1,92 @@
...
+            availableEmulators.Clear();
+            availableDevices.Clear();
+            return (availableEmulators, availableDevices);
+        }
+
+        public static void Launch(bool shouldConnectLocally, string deviceId, string runtimeIp, bool useEmulator)
```
jamiebrynes7: Not a huge fan of the bools here. Could do a little refactoring and make some intermediate data structures:

```csharp
internal struct MobileLaunchOptions 
{
	public bool ShouldConnectLocally;
	public string RuntimeIp;
	
	// If you move this in here, you can drop the `MobileLaunchUtils` class entirely.
	public string ToArgs()
	{
		...
	}
}

internal struct MobileDeviceTarget
{
	public string DeviceId;
	public bool IsEmulator;
	
	// Maybe rename this..? 
	public string DeviceOrEmulator => IsEmulator ? "emulator" : "device";
}

public static void Launch(MobileDeviceTarget target, MobileLaunchOptions options)
{
	...
}
```

workers/unity/Packages/io.improbable.gdk.mobile/Editor/MobileLauncherWindow.cs
```diff
@@ -52,15 +67,54 @@ public void OnGUI()
...
+            DisplayiOSMenu();
+#endif
+        }
+
+        private void RefreshAndroidEmulatorsAndDevices()
+        {
+            (androidEmulators, androidDevices) = AndroidLaunchUtils.RetrieveAvailableEmulatorsAndDevices();
+
+            androidEmulatorNames = androidEmulators.Keys.ToArray();
+            androidDeviceNames = androidDevices.Keys.ToArray();
```
jamiebrynes7: Why do we do this? Couldn't we iterate over the keys as needed?

workers/unity/Packages/io.improbable.gdk.mobile/Editor/MobileLauncherWindow.cs
```diff
@@ -76,84 +130,107 @@ public void OnGUI()
...
-                    {
-                        simulatorNames = iOSLaunchUtils.RetrieveAvailableiOSSimulators().Keys.ToArray();
-                        simulatorNameIndex = 0;
-                    }
+                DisplayDevicesAndLaunchButton(isAndroid: false, isEmulator: false);
+            }
+        }
+#endif
+
+        private void DisplayDevicesAndLaunchButton(bool isAndroid, bool isEmulator)
```
jamiebrynes7: Can collapse these bools into an enum: 

```csharp
public enum DeviceType
{
	AndroidDevice = 0,
	AndroidEmulator = 1,
	iOSDevice = 2,
	iOSEmulator = 3
}
```

Then the below are switch-case statements rather than nested ternaries.

workers/unity/Packages/io.improbable.gdk.mobile/Editor/MobileLauncherWindow.cs
```diff
@@ -76,84 +130,107 @@ public void OnGUI()
...
+            // Draw a launch button that will launch the app on the correct platform and chosen device
+            using (new EditorGUI.DisabledScope(names.Length == 0))
+            {
+                if (!GUILayout.Button($"Launch {androidOriOs} app on {deviceOrEmulator}"))
+                {
+                    return;
+                }
 
-                using (new GUILayout.HorizontalScope())
+                if (devices.TryGetValue(names[index], out var deviceId))
```
jamiebrynes7: If we stored an object here instead of an ID, we might be able to do something like: 

```csharp
if (devices.TryGetValue(names[index], out var device)
{
	device.Launch(options);
}
```

and delegate implementation to the object.

workers/unity/Packages/io.improbable.gdk.mobile/Editor/iOSLaunchUtils.cs
```diff
@@ -18,62 +18,61 @@ public static class iOSLaunchUtils
...
                     }
                 })
                 .RedirectOutputOptions(OutputRedirectBehaviour.None)
                 .Run();
 
-            if (result.ExitCode != 0)
+            if (result.ExitCode == 0)
             {
-                Debug.LogError("Failed to find connected iOS devices. Make sure you have the Command line tools for XCode (https://developer.apple.com/download/more/) installed and check the logs.");
+                if (!available)
```
jamiebrynes7: Similar comment about `available` here as above

workers/unity/Packages/io.improbable.gdk.mobile/Editor/AndroidLaunchUtils.cs
paulbalaji: for the record: I haven't forgotten this comment

workers/unity/Packages/io.improbable.gdk.mobile/Editor/MobileLauncherWindow.cs
```diff
@@ -48,112 +53,181 @@ public void OnGUI()
...
-                    if (GUILayout.Button(buttonIcon, EditorStyles.miniButton, GUILayout.ExpandWidth(false)))
-                    {
-                        deviceNames = iOSLaunchUtils.RetrieveAvailableiOSDevices().Keys.ToArray();
-                        deviceNameIndex = 0;
-                    }
+                    return;
                 }
 
-                using (new EditorGUI.DisabledScope(deviceNames.Length == 0))
+                if (!devices[index].deviceName.Equals(names[index]))
```
paulbalaji: I think this check is moot and we can call Launch directly but the issue is that the device list might be out of data

we could either:
- refresh the list before calling `launch`, and match on the device of the same name
- let it try to launch and if it fails in either adb/instruments then user can resolve that 

depends on how much we want to verify for ourselves here

workers/unity/Packages/io.improbable.gdk.mobile/Editor/MobileLauncherWindow.cs
```diff
@@ -48,112 +53,181 @@ public void OnGUI()
...
+                    (index, devices) = androidEmulators;
+                    break;
+                case DeviceType.iOSDevice:
+                    (index, devices) = iOSDevices;
+                    break;
+                case DeviceType.iOSSimulator:
+                    (index, devices) = iOSSimulators;
+                    break;
+                default:
+                    throw new ArgumentException($"Invalid Device Type: {deviceType}");
```
jamiebrynes7:
```suggestion
                    throw new ArgumentException($"Unknown device type: {deviceType}");
```

workers/unity/Packages/io.improbable.gdk.mobile/Editor/MobileLauncherWindow.cs
```diff
@@ -48,112 +53,181 @@ public void OnGUI()
...
-
-                    if (GUILayout.Button(buttonIcon, EditorStyles.miniButton, GUILayout.ExpandWidth(false)))
-                    {
-                        simulatorNames = iOSLaunchUtils.RetrieveAvailableiOSSimulators().Keys.ToArray();
-                        simulatorNameIndex = 0;
-                    }
+                    SetDeviceIndex(deviceType,
+                        noDevicesFound
+                            ? EditorGUILayout.Popup(prettyDeviceType, 0, emptyDeviceNameList)
+                            : EditorGUILayout.Popup(prettyDeviceType, index, names));
```
jamiebrynes7: Instead of this, we could not have a popup when there are no devices and display simple text instead.

Not sure if that will play nice with the layout engine though

workers/unity/Packages/io.improbable.gdk.mobile/Editor/MobileLauncherWindow.cs
jamiebrynes7: Under what conditions will this happen?

paulbalaji: 1. connect device
2. refresh list
3. select device
4. disconnect device
5. hit launch button on device that is no longer connected

workers/unity/Packages/io.improbable.gdk.mobile/Editor/MobileLauncherWindow.cs
paulbalaji: I think I prefer having the popup still there with this text because it then makes it clear why it's disabled

workers/unity/Packages/io.improbable.gdk.mobile/Editor/MobileLauncherWindow.cs
jamiebrynes7: I think we just do:

> let it try to launch and if it fails in either adb/instruments then user can resolve that

as long as we handle and log the errors properly.

paulbalaji: yeah that's basically what I ended up doing after discussing with Martijn yesterday evening

PR #1196, opened 2019-10-17
Codegen Split: GameObject Creation

zeroZshadow:
#### Description
Moved code generation files required for GameObject Creation support to the respective package.
Fixed missing dependency in TransformSync package.

#### Documentation
* [x] Changelog


workers/unity/Packages/io.improbable.gdk.gameobjectcreation/.codegen/Source/GameObjectGenerationJob.cs
```diff
@@ -0,0 +1,85 @@
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Improbable.Gdk.CodeGeneration.FileHandling;
+using Improbable.Gdk.CodeGeneration.Jobs;
+using Improbable.Gdk.CodeGeneration.Model.Details;
+using Improbable.Gdk.CodeGeneration.Utils;
+
+namespace Improbable.Gdk.CodeGenerator.GameObjectCreation
```
paulbalaji: I think we should rename this module since it encompasses everything to do with game objects (not just creation)

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/.codegen/Source/GameObjectGenerationJob.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    public class GameObjectGenerationJob : CodegenJob
```
paulbalaji: `GameObjectCodegenJob` might be nicer?

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/.codegen/Source/GameObjectGenerationJob.cs
zeroZshadow: Agreed, though out of scope for this

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/.codegen/Source/GameObjectCodegenJob.cs
```diff
@@ -0,0 +1,85 @@
...
+
+                var componentReaderWriterFileName =
+                    Path.ChangeExtension($"{componentName}ComponentReaderWriter", FileExtension);
+                var componentReaderWriterCode =
+                    componentReaderWriterGenerator.Generate(componentTarget.Content, package);
+                Content.Add(Path.Combine(relativeOutputPath, componentReaderWriterFileName), componentReaderWriterCode);
+            }
+        }
+
+        private struct GenerationTarget<T>
```
jamiebrynes7: Maybe move this into the base lib? Since its used in both core and here

PR #1205, opened 2019-10-31
Fix command response key leak

zeroZshadow:
#### Description
Command response callbacks were registering a key with the system, but these were never cleared.
The key itself isn't actually used, since response callbacks will always clean up after sending.

This PR removes the key registration for response callbacks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 ### Fixed
 
 - Fixed a bug where the Deployment Launcher window would accept tags with 33 characters. [#1202](https://github.com/spatialos/gdk-for-unity/pull/1202)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 ### Fixed
 
 - Fixed a bug where the Deployment Launcher window would accept tags with 33 characters. [#1202](https://github.com/spatialos/gdk-for-unity/pull/1202)
+- Fixed small memory leak with command response callbacks using Monobehaviours. [#1205](https://github.com/spatialos/gdk-for-unity/pull/1205)
```
paulbalaji:
```suggestion
- Fixed a small memory leak with command response callbacks using MonoBehaviours. [#1205](https://github.com/spatialos/gdk-for-unity/pull/1205)
```

PR #1206, opened 2019-10-31
Quick code snippets fixes

paulbalaji:
#### Description

- turn off colliders in HealthPickupServerBehaviour (UTY-1977)
- fix missing schema import for `Empty` (UTY-2099)

#### Tests

- not regaining health anymore when pickup is disabled
- codegen broke without import, works fine now

#### Documentation

- yes this is docs

#### Primary reviewers


jamiebrynes7:
This could equally be pointed at `docs-current` right?

paulbalaji:
> This could equally be pointed at `docs-current` right?

yes, it can

PR #1207, opened 2019-10-31
Inherited required test

zeroZshadow:
#### Description
Remove an unneeded binding flag + adds a test to ensure this behaviour keeps working


test-project/Assets/EditmodeTests/Subscriptions/InjectionCriteriaTests.cs
```diff
@@ -114,7 +114,30 @@ public void Injection_happens_if_all_constraints_are_satisfied()
             Assert.IsNotNull(compositeBehaviour.Writer);
         }
 
+        [Test]
+        public void Injection_happens_if_inherited_constraints_are_satisfied()
+        {
+            ConnectionHandler.ChangeAuthority(EntityId, Position.ComponentId, Authority.Authoritative);
```
jamiebrynes7:
```suggestion
```

Don't think we need this line since the requireable is a `Reader`

PR #1208, opened 2019-10-31
Worker sdk upgrade

zeroZshadow:
#### Description
Upgrade the Worker SDK to 14.2.1

* [x] Merge shared-ci and update pin

#### Documentation
* [x] Changelog
* [x] Open PR for docs worker sdk version change


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 ### Changed
 
 - Generated Worker ID's for local development are now smaller and easier to read for debugging. [#1197](https://github.com/spatialos/gdk-for-unity/pull/1197)
+- Upgraded to Worker SDK 14.2.1. [#1208](https://github.com/spatialos/gdk-for-unity/pull/1208)
```
jessicafalk: we should mention somewhere that we changed the version numbers of the worker sdk packages from worker sdk version to gdk release version

zeroZshadow: Would this fall under a breaking change? The sdk's are dependencies of our core packages, so our users shouldn't have to change anything.

jessicafalk: don't think it's a breaking change as we do update the version numbers of all packages with each release. But it should be mentioned in the release notes even if they don't have to change anything

PR #1210, opened 2019-10-31
Docs - worker sdk upgrade 14.2.1

zeroZshadow:
Upping the version number for docs

### DIFF
--- a/docs/variable.properties
+++ b/docs/variable.properties
@@ -1,2 +1,2 @@
 current_version=0.2.10
-worker_sdk_version=14.0
+worker_sdk_version=14.2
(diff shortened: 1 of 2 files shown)


docs/variable.properties
```diff
@@ -1,2 +1,2 @@
 current_version=0.2.10
-worker_sdk_version=14.0
\ No newline at end of file
+worker_sdk_version=14.2
```
jessicafalk: should it be 14.2.1?

paulbalaji: there's no docs versions for patch releases

PR #1219, opened 2019-11-18
Deployment Launcher editor performance improvement

paulbalaji:
#### Description

- stop saving to disk on every change
	- only write unsaved changes if it's been more than 1 second since the last change
	- also write to disk if the window is being closed

#### Tests

- feels a lot nicer
- profile before/after

#### Documentation

probably worth adding a note in the deployment launcher docs about the save interval

- [x] changelog

#### Primary reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@
 
 ### Fixed
 
-- The Deployment Launcher window no longer saves changes after every input. [#1219](https://github.com/spatialos/gdk-for-unity/pull/1219)
-    - It will now wait for at least 1 second to elapse after the last change before writing the config back to disk.
+- Fixed an issue where the Deployment Launcher window would feel unresponsive due to saving changes after every input. [#1219](https://github.com/spatialos/gdk-for-unity/pull/1219)
+    - It will now wait for at least 1 second to elapse after the last change before writing the configuration back to disk.
 
 ## `0.3.0` - 2019-11-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 - A `WorkerSystem` now exposes the underlying Worker's `IsConnected` property. [#1217](https://github.com/spatialos/gdk-for-unity/pull/1217)
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


workers/unity/Packages/io.improbable.gdk.deploymentlauncher/DeploymentLauncherWindow.cs
```diff
@@ -882,6 +888,23 @@ private string GetStatusMessage()
             return sb.ToString();
         }
 
+        private void MarkConfigAsDirty()
+        {
+            EditorUtility.SetDirty(launcherConfig);
+            lastEditTime = DateTime.Now;
+        }
+
+        private void SaveChanges()
```
jamiebrynes7:
```suggestion
        private void TrySaveChanges()
```

Nit - clearer naming, doesn't always save.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 - A `WorkerSystem` now exposes the underlying Worker's `IsConnected` property. [#1217](https://github.com/spatialos/gdk-for-unity/pull/1217)
 
+### Fixed
+
+- The Deployment Launcher window no longer saves changes after every input. [#1219](https://github.com/spatialos/gdk-for-unity/pull/1219)
```
jamiebrynes7: Its not clear why this is under "Fixed" from this description. Why does a user care about this?

paulbalaji: I was debating whether to put this in changed or fixed

should I:
a - move to changed
b - change this to something like "Fixed a performance issue where the Deployment Launcher window would feel slow because xyz.."

workers/unity/Packages/io.improbable.gdk.deploymentlauncher/DeploymentLauncherWindow.cs
```diff
@@ -75,11 +78,16 @@ private void OnDestroy()
 
         private void OnExit()
         {
+            Debug.Log("Saving Deployment Launcher configuration before closing.");
```
zeroZshadow: This log isn't needed

Issue #958, opened 2019-05-24
Potential name clash between nested types and fields in schema components


jamiebrynes7:
**Affects:** Release v0.1.0 and up
**Internal Ticket:** [UTY-1416](https://improbableio.atlassian.net/browse/UTY-1416)

---

#### Description

Given a schema file like the following: 

```
type Foo {
    enum Bar {
        YES= 0;
        NO = 1;
        MAYBE = 2;
    }

    Bar bar = 1;
}
```

The code generator will generate invalid C# code that does not compile.

> Note this occurs when the name of a nested type and the name of an field inside a component are _the same_.

#### Workaround

Rename either the nested type or the field.

paulbalaji:
A resolution has been merged and will be part of our 0.3.1 release next week.

jamiebrynes7:
Released in https://github.com/spatialos/gdk-for-unity/releases/tag/0.3.1

Issue #957, opened 2019-05-24
Potential name clash between components and events/commands in generated code


jamiebrynes7:
**Affects:** Release v0.1.0 and up 
**Internal Ticket:** [UTY-1962](https://improbableio.atlassian.net/browse/UTY-1962)

---

#### Description

Given a schema file like one of the following:

```
package name;

type Foo {
    float test_field = 1;
}

component Bar
{
    id = 200;
    event Foo bar;
}
```

```
package name;

type Foo {
    float test_field = 1;
}

component Bar
{
    id = 200;
    command Foo bar(Foo);
}
```

The code generator will generate invalid C# code that does not compile.

> Note this occurs when the name of a component and the name of an event/command inside that component are _the same_

#### Workaround

Rename either the component or the event/command.

paulbalaji:
A resolution has been merged and will be part of our 0.3.1 release next week.

jamiebrynes7:
Released in https://github.com/spatialos/gdk-for-unity/releases/tag/0.3.1

PR #1222, opened 2019-11-19
Resolving type/command/event name clashes

paulbalaji:
#### Description

- #958 (UTY-1416)
	- codegen will now catch this, skip the field, and log an error
- #957 (UTY-1962)
	- codegen will now catch this, skip the event/command, and log an error

It currently just does a basic log to stderr, and forwarding that to the Editor:
image

TODO:

- [x] changelog
- [x] move valid sample schema to test project

~~add tests for the illegal schema sample~~

> UTY-2311 will implement logging in the codegen, which should provider a nicer way of outputting the information/errors that we've added in this PR. For now, how it currently logs can be left as an implementation detail.

#### Tests

Schema linked in #957 and #958 will no longer cause compilation errors, but codegen will log an error that gets picked up in the Editor

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/UnityComponentDetails.cs
```diff
@@ -51,5 +52,33 @@ public void PopulateFields(DetailsStore store)
...
+            var componentName = ComponentName;
+
+            var clashingCommands = CommandDetails
+                .Where(commandDetail => commandDetail.CommandName.Equals(componentName));
+            foreach (var clashingCommand in clashingCommands)
+            {
+                isValid = false;
+                Console.Error.WriteLine(
+                    $"Error in component \"{componentName}\". " +
+                    $"Command \"{clashingCommand.RawCommandName}\" clashes with component name.");
```
jamiebrynes7: Nit: Prefer having a long line over concatentating strings. Also you can use single quotes and get rid of the escaping which aids readability

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/UnityComponentDetails.cs
```diff
@@ -51,5 +52,33 @@ public void PopulateFields(DetailsStore store)
                     .AsReadOnly();
             }
         }
+
+        public bool IsValid()
+        {
+            var isValid = true;
+            var componentName = ComponentName;
```
jamiebrynes7: Why do we have a local variable here?

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/UnityComponentDetails.cs
```diff
@@ -51,5 +52,33 @@ public void PopulateFields(DetailsStore store)
...
+                .Where(eventDetail => eventDetail.EventName.Equals(componentName));
+            foreach (var clashingEvent in clashingEvents)
+            {
+                isValid = false;
+                Console.Error.WriteLine(
+                    $"Error in component \"{componentName}\". " +
+                    $"Event \"{clashingEvent.RawEventName}\" clashes with component name.");
+            }
+
+            return isValid;
```
jamiebrynes7: This could be something like `clashingEvents.Any() || clashingCommands.Any()` and we can drop tracking `isValid` through the function? 

Also helps understand what this condition is imo

jamiebrynes7: (Obviously be careful with double iteration over the IEnumerables)

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/UnityTypeDetails.cs
```diff
@@ -61,6 +62,37 @@ public void PopulateFields(DetailsStore store)
...
+                foreach (var childType in clashingChildTypes)
+                {
+                    isValid = false;
+                    Console.Error.WriteLine(
+                        $"Error in type \"{typeName}\". " +
+                        $"Field \"{fieldDetail.Raw.Name}\" clashes with child type \"{childType.CamelCaseName}\".");
+                }
+            }
+
+            return isValid;
```
jamiebrynes7: Same comments apply to this one too :smile:

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/DetailsStore.cs
```diff
@@ -81,16 +82,22 @@ public DetailsStore(SchemaBundle bundle, List<string> serializationOverrides, IF
...
+            Types = new Dictionary<string, UnityTypeDetails>(types
+                .Select(kv =>
+                {
+                    kv.Value.PopulateFields(this);
+                    kv.Value.PopulateChildren(this);
+                    return kv;
+                })
+                .Where(kv => kv.Value.IsValid()));
+
+            Components = new Dictionary<string, UnityComponentDetails>(components
```
jamiebrynes7: Do we need the assignments on line 78 and 79 now?

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/DetailsStore.cs
```diff
@@ -3,6 +3,7 @@
 using System.Collections.Immutable;
 using System.Collections.ObjectModel;
 using System.Linq;
+using System.Threading;
```
jamiebrynes7: Nit: Unnecessary import

test-project/Assets/.schema/illegal.schema
```diff
@@ -0,0 +1,27 @@
+// Wait, that's illegal!
```
jamiebrynes7: :clap:

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/DetailsStore.cs
```diff
@@ -81,16 +82,22 @@ public DetailsStore(SchemaBundle bundle, List<string> serializationOverrides, IF
...
-                kv.Value.PopulateFields(this);
-            }
+            Types = new Dictionary<string, UnityTypeDetails>(types
+                .Select(kv =>
+                {
+                    kv.Value.PopulateFields(this);
+                    kv.Value.PopulateChildren(this);
+                    return kv;
+                })
+                .Where(kv => kv.Value.IsValid()));
```
jamiebrynes7: Shouldn't we be filtering out the offending field in the type rather than the whole type? 

Imagine I had the following schema: 

```
type MyType {
	int32 data = 1;
}
```

And this type was used all over my codebase. Then at some point in the future someone adds an inner type and field (that clashes):

```
type MyType {
	type Inner {}
	
	int32 data = 1;
	Inner inner = 2;
}
```

Now I'll get code compilation errors since `MyType` has disappeared entirely rather than the new field being omitted? 

It also aligns with us dropping the event/command when it clashes instead of the entire component

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/UnityComponentDetails.cs
paulbalaji: I was using single quotes before and I think it's not really that readable in the Editor

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/UnityComponentDetails.cs
```diff
@@ -12,8 +13,8 @@ public class UnityComponentDetails
         public bool IsBlittable { get; }
 
         public IReadOnlyList<UnityFieldDetails> FieldDetails { get; private set; }
-        public IReadOnlyList<UnityCommandDetails> CommandDetails { get; }
-        public IReadOnlyList<UnityEventDetails> EventDetails { get; }
+        public IReadOnlyList<UnityCommandDetails> CommandDetails { get; private set; }
```
zeroZshadow: Why are these 2 private set now?

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/UnityComponentDetails.cs
```diff
@@ -26,11 +27,33 @@ public UnityComponentDetails(string package, ComponentDefinition componentDefini
...
                 .Select(command => new UnityCommandDetails(command))
+                .Where(commandDetail =>
+                {
+                    // Return true to keep commands that do not have a name clash with the component
+                    if (!commandDetail.CommandName.Equals(ComponentName))
+                    {
+                        return true;
+                    }
+
+                    Console.Error.WriteLine($"Error in component \"{ComponentName}\". Command \"{commandDetail.RawCommandName}\" clashes with component name.");
```
zeroZshadow: This should FAIL codegen. It shouldn't be possible to write schema that throws errors but still generates code.

jamiebrynes7:
> I think codegen needs to fail when running into clashes, not skip them.

The problem is until we have a way of surfacing these errors reliably, failing code generation (i.e. - outputting nothing) obscures the problem due to the avalanche of compile errors you will hit.

paulbalaji:
> > I think codegen needs to fail when running into clashes, not skip them.
> 
> The problem is until we have a way of surfacing these errors reliably, failing code generation (i.e. - outputting nothing) obscures the problem due to the avalanche of compile errors you will hit.

Discussed offline

Having heard both sides, I've decided to keep the messages as errors without failing the entire codegen step for now, and we can refine the idea further when we have better codegen logging. Once that is in place, we are likely to be able to surface these issues better to users and (maybe) cause the codegen step to fail in a way that doesn't stop users from A) knowing that the error was in codegen and B) knowing where to look to figure out where their error comes from.

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/UnityComponentDetails.cs
paulbalaji: resolved offline

workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/Model/Details/UnityComponentDetails.cs
paulbalaji: nice spot - these are relics from when the filtering was done in a separate method

PR #1211, opened 2019-11-01
Events no longer trigger component updates

zeroZshadow:
#### Description
Received component ops that don't contain any field updates, no longer trigger the OnUpdate callback.

#### Documentation
* [x] Changelog

PR #1229, opened 2019-12-03
Cleared field bugfix of #1211

paulbalaji:
#### Description

- UTY-2382
- Check cleared field count as well when applying update to diff
- Fixes bug introduced in #1211 

#### Tests

- removed 1 map element every second on server, received updated on client
- logged map count on server/client
- observed client never made it to 0 before bugfix
- correct behaviour observed after bugfix

#### Documentation

- [x] changelog

#### Primary reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
     - It will now wait for at least 1 second to elapse after the last change before writing the configuration back to disk.
 - Fixed issues ([#957](https://github.com/spatialos/gdk-for-unity/issues/957), [#958](https://github.com/spatialos/gdk-for-unity/issues/958)) where valid schema would generate invalid code due to name clashes. [#1222](https://github.com/spatialos/gdk-for-unity/pull/1222)
     - The offending schema properties will no longer be generated and are now logged in the Unity Editor.
+- Fixed a bug where cleared fields in a received update would not be applied to the diff. [#1229](https://github.com/spatialos/gdk-for-unity/pull/1229)
 
 ### Internal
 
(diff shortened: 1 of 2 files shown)


jamiebrynes7:
For future reference - we could use the [JSON representation of schema](https://docs.improbable.io/reference/14.2/shared/schema/json#json-representation-of-schema-data) to test that we are serializing data correctly: 

- Serialize data/update into `Schema_*`
- Convert into JSON
- Validate the JSON against the expected structure

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
     - It will now wait for at least 1 second to elapse after the last change before writing the configuration back to disk.
 - Fixed issues ([#957](https://github.com/spatialos/gdk-for-unity/issues/957), [#958](https://github.com/spatialos/gdk-for-unity/issues/958)) where valid schema would generate invalid code due to name clashes. [#1222](https://github.com/spatialos/gdk-for-unity/pull/1222)
     - The offending schema properties will no longer be generated and are now logged in the Unity Editor.
+- Fixed a bug where cleared fields in a received update would not be applied to the diff. [#1229](https://github.com/spatialos/gdk-for-unity/pull/1229)
```
jamiebrynes7: Wrong section 😛

paulbalaji: woops

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- Fixed a bug where cleared fields in a received update would not be applied to the diff. [#1229](https://github.com/spatialos/gdk-for-unity/pull/1229)
```
```suggestion
- Fixed a bug where empty list, maps, or options in a component update would not be applied correctly. [#1229](https://github.com/spatialos/gdk-for-unity/pull/1229)
```

Cleared fields + diff is an implementation detail

paulbalaji: fine

PR #1243, opened 2020-01-16
Tidied CodeWriter and tweaked API for annotations/methods

paulbalaji:
#### Description

Before:
- annotation defined within the lambda for methods/types/enums
- can only populate a method using `Action<MethodBlock>`

Now:
- do `.Annotate` on the parent and then call `.Method`/`.Type`/`.Enum`
	- e.g. `ns.Annotate("Serializable").Type("public struct Foobar", t => {});`
- .Method has a separate constructor taking Func<IEnumerable<string>> instead of Action<MethodBlock>
	- e.g.
	```csharp
	t.Method("public static void Test()", () => new[]
	{
		"var x = 42;",
		"var y = 360;",
		"return x + y;"
	});
	```
	```csharp
	t.Method("public static void Test()", () => someList.some(a => wesome).linq());
	```

Also enforced that you can only define namespace/using directives at the top-level `CodeWriter`

#### Tests

- will use for build system codegen port

#### Documentation

- [ ] changelog

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/CodeWriter/CodeWriter.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using CodeGenerationLib.CodeWriter.Scopes;
```
zeroZshadow: Why is this added?

PR #1248, opened 2020-01-20
Port enum/type content and generators

paulbalaji:
#### Description

ported:

- resolve minor lint issues in CoreCodegenJob
- enum content
- enum generator
- type content
- type generator

#### Tests

generated code

#### Documentation

- will add changelog in final core port PR

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.core/.codegen/Source/Generators/Core/UnityTypeGenerator.cs
```diff
@@ -0,0 +1,30 @@
...
+            {
+                cgw.UsingDirectives(
+                    "System.Linq",
+                    "Improbable.Gdk.Core",
+                    "UnityEngine"
+                );
+
+                cgw.Namespace(qualifiedNamespace, ns =>
+                {
+                    ns.Type(UnityTypeContent.Generate(details, qualifiedNamespace));
```
zeroZshadow: Can this not be inlined now?

paulbalaji: worth coming back to on the post-tt-purge codegen pass 

not so easy because the generator has the `using` and `namespace`, whereas the content is just the type class/struct

PR #1256, opened 2020-01-24
Generate code generator run configurations

paulbalaji:
#### Description

- update csproj with program arguments
- also add rider-specific run configuration
- generate the configs on force generate + every save of gdk tools config
- log warning if user-defined schema dir is not found when generating run configs
- log error in code generator if any input schema dir does not exist

#### Tests

- tried running code generator without unity in VS/Rider
- tried modifying codegen settings and regen code (not force) to see if changes are propagated
- tried generating code with schema input dir that doesn't exist - nice warnings and errors

#### Documentation

- [x] changelog

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +450,91 @@ private static IEnumerable<string> FindDirInPackages(string directory)
...
+            if (toolsConfig.VerboseLogging)
+            {
+                codegenArgs.Add($"--verbose");
+            }
+
+            var codegenOutputPath = Path.GetFullPath(Path.Combine(Application.dataPath, "..", toolsConfig.CodegenOutputDir));
+            codegenArgs.Add($"--native-output-dir=\"{codegenOutputPath}\"");
+
+            // Add user defined schema directories
+            codegenArgs.AddRange(toolsConfig.SchemaSourceDirs
```
jared-improbable: If an input directory doesn't exist, does that imply the user has something mis-configured? It may be better to at least warn that a directory doesn't exist

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +450,91 @@ private static IEnumerable<string> FindDirInPackages(string directory)
...
+  </configuration>
+</component>
+");
+                }
+            }
+            catch (Exception e)
+            {
+                Debug.LogError("Unable to generate Rider run configuration.");
+                Debug.LogException(e);
+                throw;
```
jared-improbable: Will this exception effectively be logged twice because of the re-throw?

paulbalaji: yes - though I'll now change it to just throw an Exception with this context and catch/log it at a higher level

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
paulbalaji: mis-configuration is captured when we validate the tools config, but we should still warn here in case a directory a user expects is not present for whatever reason

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -245,6 +230,36 @@ private static void Generate()
             }
         }
 
+        private static string GetDefaultLoggerOutputPath(GdkToolsConfiguration toolsConfig = null)
+        {
+            toolsConfig = toolsConfig ?? GdkToolsConfiguration.GetOrCreateInstance();
+            return Path.GetFullPath(Path.Combine(toolsConfig.CodegenLogOutputDir, "codegen-output.log"));
+        }
```
jamiebrynes7: Should this be a method on the `GdkToolsConfiguration` class?

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +450,94 @@ private static IEnumerable<string> FindDirInPackages(string directory)
 
             return packagePaths.Union(cachedPackagePaths).Distinct();
         }
+
+        internal static void GenerateCodegenRunConfigs(GdkToolsConfiguration toolsConfig = null)
```
jamiebrynes7: Why does this have a default argument of `null`?

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +450,94 @@ private static IEnumerable<string> FindDirInPackages(string directory)
...
+
+                codegenArgs.Add($"--schema-path=\"{Path.GetFullPath(schemaDir)}\"");
+            }
+
+            // Add package schema directories
+            codegenArgs.AddRange(FindDirInPackages(SchemaPackageDir)
+                .Select(directory => $"--schema-path=\"{directory}\""));
+
+            // Schema Descriptor
+            codegenArgs.Add($"--descriptor-dir=\"{toolsConfig.DescriptorOutputDir}\"");
```
jamiebrynes7: This could be in the initial list right?

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +450,94 @@ private static IEnumerable<string> FindDirInPackages(string directory)
...
+
+                var args = propertyGroup.Element("StartArguments");
+                args?.Remove();
+
+                propertyGroup.Add(XElement.Parse($"<StartArguments>{codegenArgsString}</StartArguments>"));
+                csprojXml.Save(CodegenExe);
+            }
+            catch
+            {
+                throw new Exception("Unable to update csproj with run configuration.");
```
jamiebrynes7: This should be something like:

```
catch (Exception e)
{
	throw new Exception($"Unable to add run configurations to '{CodegenExe}'", e);
}
```

such that we don't swallow the underlying exception and instead wrap it.

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +450,94 @@ private static IEnumerable<string> FindDirInPackages(string directory)
...
+    <option name=""PROJECT_PATH"" value=""$PROJECT_DIR$/CodeGen/CodeGen.csproj"" />
+    <option name=""PROGRAM_PARAMETERS"" value=""{codegenArgsString.Replace("\"", "&quot;")}"" />
+  </configuration>
+</component>
+");
+                }
+            }
+            catch
+            {
+                throw new Exception("Unable to generate Rider run configuration.");
```
jamiebrynes7: Same comment as above

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +450,94 @@ private static IEnumerable<string> FindDirInPackages(string directory)
...
+            {
+                throw new Exception("Unable to update csproj with run configuration.");
+            }
+
+            // For jetbrains rider
+            try
+            {
+                var runConfigPath = Path.Combine(CodegenExeDirectory, ".idea", ".idea.CodeGen", ".idea",
+                    "runConfigurations");
+                Directory.CreateDirectory(runConfigPath);
```
jamiebrynes7: What is this directory already exists with files in it?

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +450,94 @@ private static IEnumerable<string> FindDirInPackages(string directory)
...
+                using (var w = new StreamWriter(Path.Combine(runConfigPath, "CodeGen.xml"), false))
```
jamiebrynes7: Use the `name: value` syntax for the bool argument please :)

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -245,6 +230,36 @@ private static void Generate()
...
+            {
+                case RuntimePlatform.WindowsEditor:
+                    schemaCompilerPath = Path.ChangeExtension(schemaCompilerPath, ".exe");
+                    break;
+                case RuntimePlatform.LinuxEditor:
+                case RuntimePlatform.OSXEditor:
+                    RedirectedProcess.Command("chmod")
+                        .WithArgs("+x", $"\"{schemaCompilerPath}\"")
+                        .InDirectory(Path.GetFullPath(Path.Combine(Application.dataPath, "..")))
+                        .Run();
```
jamiebrynes7: Pretty big side-effect for a method called `GetSchemaCompilerPath()` 😛

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -245,6 +230,36 @@ private static void Generate()
...
+        private static string GetSchemaCompilerPath()
+        {
+            var schemaCompilerPath = SchemaCompilerPath;
+
+            switch (Application.platform)
+            {
+                case RuntimePlatform.WindowsEditor:
+                    schemaCompilerPath = Path.ChangeExtension(schemaCompilerPath, ".exe");
+                    break;
+                case RuntimePlatform.LinuxEditor:
```
jamiebrynes7: Pretty sure that Linux is unsupported too? We don't download the `schema_compiler` for that platform

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
paulbalaji: remnant of an old idea 😅

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
paulbalaji: err - yes it could

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
paulbalaji: moved this + getting the codegen output path to properties on tools config class

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
paulbalaji: this'll do nothing

Q: but do we not need to wipe the contents of the dir?
A: we only care about the run config xml and we're overwriting that anyway

Q: but we remove the args in the csproj?
A: sadly there's no nice way of easily overwriting the csproj node

Q: ok

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 ### Added
 
 - Added public toolkit for writing code generators. [#1240](https://github.com/spatialos/gdk-for-unity/pull/1240) [#1243](https://github.com/spatialos/gdk-for-unity/pull/1243) [#1244](https://github.com/spatialos/gdk-for-unity/pull/1244) [#1245](https://github.com/spatialos/gdk-for-unity/pull/1245) [#1250](https://github.com/spatialos/gdk-for-unity/pull/1250)
+- Added generation of code generator run configurations. [#1256](https://github.com/spatialos/gdk-for-unity/pull/1256)
```
```suggestion
- Added generation of code generator run configurations for Jetbrains Rider, Visual Studio, and the `dotnet` CLI. [#1256](https://github.com/spatialos/gdk-for-unity/pull/1256)
```

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
...
+    - This means you no longer need to keep the Unity Editor constantly open to iterate on generators.
+- Added error logging in the code generator to log when input schema source directories do not exist. [#1256](https://github.com/spatialos/gdk-for-unity/pull/1256)
```
```suggestion
- The code generator now logs an error when an input schema source directory does not exist. [#1256](https://github.com/spatialos/gdk-for-unity/pull/1256)
```

workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfiguration.cs
```diff
@@ -22,6 +22,9 @@ public class GdkToolsConfiguration
         public string DevAuthTokenFilepath => Path.Combine(DevAuthTokenFullDir, "DevAuthToken.txt");
         public int DevAuthTokenLifetimeHours => TimeSpan.FromDays(DevAuthTokenLifetimeDays).Hours;
 
+        public string FullCodegenOutputPath => Path.GetFullPath(Path.Combine(Application.dataPath, "..", CodegenOutputDir));
```
jamiebrynes7: Can we make `CodegenOutputDir` private given we expect to use this path instead? Or nah?

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -245,6 +230,30 @@ private static void Generate()
...
+                case RuntimePlatform.OSXEditor:
+                    RedirectedProcess.Command("chmod")
+                        .WithArgs("+x", $"\"{schemaCompilerPath}\"")
+                        .InDirectory(Path.GetFullPath(Path.Combine(Application.dataPath, "..")))
+                        .Run();
```
jamiebrynes7: We should probably check the result of `chmod +x`-ing the file

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +444,89 @@ private static IEnumerable<string> FindDirInPackages(string directory)
...
+            {
+                codegenArgs.Add("--verbose");
+            }
+
+            // Add user defined schema directories, warn if directory does not exist
+            foreach (var schemaDir in toolsConfig.SchemaSourceDirs)
+            {
+                if (!Directory.Exists(schemaDir))
+                {
+                    Debug.LogWarning($"Schema directory defined in GDK tools configuration does not exist: {schemaDir}");
```
jamiebrynes7: Currently if the directory doesn't exist we will get a warning then an error. Do we need this one if we check it inside the code generator? 

The latter should always catch any missing directories?

paulbalaji: one'd expect they'd be different errors: a mismatch with toolsconfig (which may not be an error) VS a mismatch with a previously valid run config (which imo definitely should be an error)

This check catches cases where a toolsconfig defines something but you can't find it at runconfig generation time - which might not actually be an error e.g. if you never actually made use of the schema in that extra directory.

The check in the code generator would catch everything else e.g. dir exists at run-config generation time but not when the code is actually being generated

workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfiguration.cs
paulbalaji: nah because we currently set it in toolsconfigwindow

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
paulbalaji: 🙊

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -245,6 +230,33 @@ private static void Generate()
...
+                    schemaCompilerPath = Path.ChangeExtension(schemaCompilerPath, ".exe");
+                    break;
+                case RuntimePlatform.OSXEditor:
+                    var result = RedirectedProcess.Command("chmod")
+                        .WithArgs("+x", $"\"{schemaCompilerPath}\"")
+                        .InDirectory(Path.GetFullPath(Path.Combine(Application.dataPath, "..")))
+                        .Run();
+                    if (result.ExitCode != 0)
+                    {
+                        throw new IOException($"Failed to chmod schema compiler:\n{string.Join("\n", result.Stderr)}");
```
jamiebrynes7: `\n` -> `Environment.Newline` ?

paulbalaji: we use `\n` everywhere else atm so... nah

workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -476,5 +448,89 @@ private static IEnumerable<string> FindDirInPackages(string directory)
...
+                $"--native-output-dir=\"{toolsConfig.FullCodegenOutputPath}\""
+            };
+
+            if (toolsConfig.VerboseLogging)
+            {
+                codegenArgs.Add("--verbose");
+            }
+
+            // Add user defined schema directories, warn if directory does not exist
+            foreach (var schemaDir in toolsConfig.SchemaSourceDirs)
```
zeroZshadow: Could this be replaced with a `toolsConfig.Validate` or similar?
At this point the config should be valid.

paulbalaji: good catch - should enforce the tools config is valid at this point

PR #1262, opened 2020-01-30
Fix code generation failures relating to `dotnet template` calls

jamiebrynes7:
#### Description

The `dotnet new` commands rely on all the templates having stable paths. This isn't always the case and can result in `dotnet new` always failing until you manually clear the cache.

We no longer use any of the templating functionality in `dotnet new` so we have just swapped over to a simple directory copy.

#### Tests
- [x] Force generate locally.
- [x] Open project from a `git clean -xdf` state
- [x] Test from FPS/Blank

#### Documentation

- [x] Changelog


workers/unity/Packages/io.improbable.gdk.tools/GenerateCode.cs
```diff
@@ -103,12 +103,8 @@ private static void SetupProject()
...
             try
             {
-                Profiler.BeginSample("Install dotnet template");
-                InstallDotnetTemplate();
-                Profiler.EndSample();
-
-                Profiler.BeginSample("Create dotnet template");
-                CreateTemplate();
+                Profiler.BeginSample("Copy codegen template");
+                CopyTemplate();
```
paulbalaji: :chefkiss:

PR #1265, opened 2020-02-03
Copy `libc++_shared.so` for Android builds

jamiebrynes7:
#### Description

The Worker SDK no longer links this library. Gradle doesn't link it properly. So we manually copy it across from the NDK. 

#### Tests

- [x] Windows w/ built in NDK
- [x] MacOS w/ built in NDK

#### Documentation

This is a breaking change since the NDK is required:

- [x] Changelog
- [x] Upgrade Guide

### DIFF
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -2,6 +2,12 @@
 
 ## From `0.3.2` to `0.3.3`
 
+### Building for Android now requires the NDK
+
+You can install the NDK from within the Unity Hub. Select "Add Modules" for your Unity install and ensure that the "Unity SDK & NDK" is ticked under the "Android Build Support".
+
+See [Unity's documentation](https://docs.unity3d.com/Manual/android-sdksetup.html) for more information.
+
 ### Generated code now requires unsafe
 
 Code generated from schema now contains code marked as `unsafe`, which needs to be specifically allowed in the relevant [Assembly Definition file](https://docs.unity3d.com/Manual/class-AssemblyDefinitionImporter.html).
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -4,7 +4,7 @@
 
 ### Building for Android now requires the NDK
 
-You can install the NDK from within the Unity Hub. Select "Add Modules" for your Unity install and ensure that the "Unity SDK & NDK" is ticked under the "Android Build Support".
+You can install the NDK from within the Unity Hub. Select "Add Modules" for your Unity install and ensure that the "Unity SDK & NDK" option is ticked underneath "Android Build Support".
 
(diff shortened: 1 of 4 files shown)


workers/unity/Packages/io.improbable.worker.sdk.mobile/CopyLibcpp.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            var jniLibsPath = Path.Combine(gradleRootPath, "src", "main", "jniLibs");
+            var dirInfo = new DirectoryInfo(jniLibsPath);
+            return dirInfo.EnumerateDirectories();
+        }
+
+        private static void CopyLibcppSo(DirectoryInfo architectureFolder)
+        {
+            var architecture = architectureFolder.Name;
+            File.Copy(Path.Combine(NdkLibRoot, architecture, "libc++_shared.so"), Path.Combine(architectureFolder.FullName, "libc++_shared.so"));
```
jessicafalk: should we store `libc++_shared.so` in a field as we use it twice here

UPGRADE_GUIDE.md
```diff
@@ -2,6 +2,12 @@
 
 ## From `0.3.2` to `0.3.3`
 
+### Building for Android now requires the NDK
+
+You can install the NDK from within the Unity Hub. Select "Add Modules" for your Unity install and ensure that the "Unity SDK & NDK" is ticked under the "Android Build Support".
```
paulbalaji:
```suggestion
You can install the NDK from within the Unity Hub. Select "Add Modules" for your Unity install and ensure that the "Unity SDK & NDK" option is ticked underneath "Android Build Support".
```

workers/unity/Packages/io.improbable.worker.sdk.mobile/CopyLibcpp.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.IO;
+using UnityEditor;
+using UnityEditor.Android;
+
+namespace Improbable.Gdk.Mobile
+{
+    // Since 14.4.0, the Worker SDK no longer links 'libc++_shared.so' and Gradle doesn't seem to
+    // properly link it in either.
+    // So we just manually copy it across.
+    // ¯\_(ツ)_/¯
```
paulbalaji: 💯

workers/unity/Packages/io.improbable.worker.sdk.mobile/CopyLibcpp.cs
paulbalaji: +1

jamiebrynes7: Could do - although given that we won't touch this code much I'm okay with it being like this. Also the locality is nice

PR #1267, opened 2020-02-03
Build deployment launcher in CI

jamiebrynes7:
#### Description
We realised that we _weren't_ doing this and it was only getting run on our local machines & at release time. This PR adds an extra build step to build the `DeploymentLauncher.csproj` to catch any regressions.

#### Tests

Yes I am tests.

#### Documentation

N/A

### DIFF
--- a/.buildkite/premerge.steps.yaml
+++ b/.buildkite/premerge.steps.yaml
@@ -73,6 +73,12 @@ steps:
             type: details
     <<: *linux
 
+  - label: "build deployment launcher :rocket:"
+    command: bash -c ci/build-deployment-launcher.sh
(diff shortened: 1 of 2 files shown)


.buildkite/premerge.steps.yaml
```diff
@@ -73,6 +73,12 @@ steps:
             type: details
     <<: *linux
 
+  - label: "build deployment launcher :rocket:"
```
paulbalaji: you know what to do

jamiebrynes7: I'm gonna pretend like I didn't see this.

PR #1289, opened 2020-02-17
Add runtime version override

jamiebrynes7:
#### Description

Added an optional override for the runtime version in the GDK Tools Configuration JSON & window. This is passed through to both `spatial local launch` and the deployment launcher. 

Also a small driveby in the `GdkToolsConfiguration` to reorganise some fields and ensure that defaults are properly set.

Commits are ordered for your reviewing pleasure 

#### Tests

- Ran a local deployment & cloud deployment with an override set.
- Ran a local deployment & cloud deployment w/o an override set.

#### Documentation

- [x] Changelog
- [ ] Docs? (Noticed we don't actually have a page documenting the configuration stuff..)

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.deploymentlauncher/.DeploymentLauncher/Commands/Create.cs
+++ b/workers/unity/Packages/io.improbable.gdk.deploymentlauncher/.DeploymentLauncher/Commands/Create.cs
@@ -63,6 +63,11 @@ namespace Improbable.Gdk.DeploymentLauncher.Commands
                     }
                 }
 
+                if (!string.IsNullOrEmpty(options.RuntimeVersion))
+                {
+                    deployment.RuntimeVersion = options.RuntimeVersion;
+                }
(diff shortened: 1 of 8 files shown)


workers/unity/Packages/io.improbable.gdk.deploymentlauncher/.DeploymentLauncher/Commands/Create.cs
```diff
@@ -63,6 +63,11 @@ private static int CreateDeploymentInternal<TOptions>(TOptions options, Func<TOp
                     }
                 }
 
+                if (options.RuntimeVersion != null)
```
zeroZshadow: isNullOrEmpty?

jamiebrynes7: Errr - yes!

PR #1299, opened 2020-02-21
Golden pinned runtime version

jamiebrynes7:
#### Description

The GDK now has a concept of a golden runtime version which is a pinned version for each GDK version. This is used if the override is not set for local & cloud deployments.

I've also edit the launch script to read from the pinned version or an override via environment variable.

#### Tests

- [x] Ran local deployment w/o override
- [x] Ran local deployment w/ override
- [x] Ran cloud deployment w/o override
- [x] Ran cloud deployment w/override
- [x] Release QA pipeline w/o override
- [x] Release QA pipeline w/ override

#### Documentation

- [x] Changelog


jamiebrynes7:
> Is this technically a breaking change for the deployment launcher? (In case anyone uses it manually)

Yes it is! Good catch

PR #1301, opened 2020-02-24
Enable SpatialD tests on MacOS

zeroZshadow:
Magically the tests have started working consistently.

It seems the machine needed the reboot that happened by moving the machine from EV's desk to the server rack?

### DIFF
--- a/.buildkite/premerge.steps.yaml
+++ b/.buildkite/premerge.steps.yaml
@@ -94,7 +94,7 @@ steps:
             type: details
     <<: *linux
 
-    - label: ":debian: ~ annotate :darwin: test results :pencil2:"
+  - label: ":debian: ~ annotate :darwin: test results :pencil2:"
     depends_on:
       - step: "test-darwin"
         allow_failure: true
--- a/.buildkite/premerge.steps.yaml
+++ b/.buildkite/premerge.steps.yaml
@@ -92,6 +92,7 @@ steps:
               type: nunit
           formatter:
             type: details
+          context: test-summary-windows
     <<: *linux
 
   - label: ":debian: ~ annotate :darwin: test results :pencil2:"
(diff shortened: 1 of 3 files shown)


.buildkite/premerge.steps.yaml
```diff
@@ -73,16 +73,28 @@ steps:
...
+        allow_failure: true
     plugins:
       - improbable/test-summary#d7289cac8297018fd1b452dcf828979307b3ebc6:
           inputs:
             - label: ":octagonal_sign: :windows: Test failures "
               artifact_path: "logs\\nunit\\*.xml"
               type: nunit
+            - label: ":octagonal_sign: :darwin: Test failures "
+              artifact_path: "logs/nunit/*.xml"
+              type: nunit
```
jamiebrynes7: Its funny that this oddity in BuildKite actually helps us out here

ci/test.sh
```diff
@@ -20,6 +20,7 @@ dotnet test \
     --logger:"nunit;LogFilePath=${TEST_RESULTS_DIR}/code-gen-lib-test-results.xml" \
     workers/unity/Packages/io.improbable.gdk.tools/.CodeGenTemplate/CodeGenerationLib/CodeGenerationLib.csproj
 
+echo ""
```
jamiebrynes7: Add a comment why this exists

.buildkite/premerge.steps.yaml
zeroZshadow: I'm not 100% sure this actually works as intended.

jamiebrynes7: Force a test failure (in different ways for each platform) and see what happens?

zeroZshadow: Just going to enable the test first. will test do another pass for the annotation afterwards

ci/test.sh
zeroZshadow: Done

PR #1302, opened 2020-02-24
Display runtime version in deployment launcher

jamiebrynes7:
#### Description
Added a label which diplays the current runtime version & a button which opens the GDK Tools Configuration window to the deployment launcher. 

image

#### Tests

- [x] Changed the runtime override and observed the change in the deployment launcher.

#### Documentation

- [x] Changelog
- [x] Probably some docs updates --> #1303

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.deploymentlauncher/DeploymentLauncherWindow.cs
+++ b/workers/unity/Packages/io.improbable.gdk.deploymentlauncher/DeploymentLauncherWindow.cs
@@ -7,6 +7,7 @@ using System.Text;
 using Improbable.Gdk.Core.Editor;
 using Improbable.Gdk.Tools;
 using Improbable.Gdk.Tools.MiniJSON;
+using Improbable.Worker.CInterop;
 using UnityEditor;
 using UnityEngine;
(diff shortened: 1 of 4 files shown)


workers/unity/Packages/io.improbable.gdk.deploymentlauncher/DeploymentLauncherWindow.cs
```diff
@@ -247,6 +248,26 @@ private void OnGUI()
                     EditorGUILayout.LabelField("Project Name", projectName);
                 }
 
+                using (new EditorGUILayout.HorizontalScope())
+                {
+                    using (new EditorGUI.DisabledScope(manager.IsActive))
+                    {
+                        var buttonIcon = new GUIContent(EditorGUIUtility.IconContent(BuiltInEditIcon))
```
zeroZshadow: Is it making this every frame?

jamiebrynes7: Hmm good question, to be honest - I cargo-culted this from the other buttons in this window. I can try making them readonly/static and seeing what happens

PR #1305, opened 2020-02-25
The beginnings of a useful testing framework

jamiebrynes7:
#### Description

This PR condenses some of the patterns we've used in testing into a `MockWorld` class which helps you write integration tests in a fluent and easy to understand way. The goal of this was to make it very easy to spin up the full stack with a mock connection and tick through frames. This is visible in the API which looks something like:

```csharp
MockWorld.Create(new MockWorld.Options())
	.Step(world => /* first frame */)
	.Step(world => /* second frame */)
	...
```

The `MockWorld` classes has a minimal feature set, and can be extended in the future:

- Handling the creation and management of the world & player loop. 
- Create and link GameObjects (with a given component), a future extension could be adding a `<T1, T2>` & `<T1, T2, T3>` version.
- Passing arbitrary context through 'frames' simply by returning an object in the lambda. This context will be passed through any number of frames until it is replaced:
	```csharp
	MockWorld.Create(new MockWorld.Options())
		.Step(world => 10)
		.Step((world, number) => /* do something with number */)
		.Step(world => /* do something without number */)
		.Step((world, number) => { /* do something with number */ return "foo"; })
		.Step((world, str) => /* do something with string */);
	```

This is all wrapped up in the `MockBase` abstract class which test fixtures can inherit from. This has a simple `SetUp` and `TearDown` where the World is created and then disposed so each test can have a fresh world to start with. 

I've ported the existing tests in the Test Project to this new paradigm, there are tests in `Core` that would benefit from this API but we may want to move around the location of the `Mock*` objects in order to use them there. 

**Commits are ordered!**

#### Tests

Old tests ported & still passing.

#### Documentation

- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@
 - De-duplicated code for generated `ComponentDiffStorage` instances. [#1290](https://github.com/spatialos/gdk-for-unity/pull/1290)
 - `Improbable.Gdk.Core.EntityId` is now a readonly struct. [#1290](https://github.com/spatialos/gdk-for-unity/pull/1290)
 - The Playground project now uses QBI instead of CBI. [#1370](https://github.com/spatialos/gdk-for-unity/pull/1307)
+- Added `MockWorld` and `MockBase` classes to the `Improbable.Gdk.TestUtils` package. These are designed as a framework for testing Core code. [#1305](https://github.com/spatialos/gdk-for-unity/pull/1305)
 
 ## `0.3.3` - 2020-02-14
 
--- a/test-project/Assets/EditmodeTests/Subscriptions/InjectionCriteriaTests.cs
+++ b/test-project/Assets/EditmodeTests/Subscriptions/InjectionCriteriaTests.cs
@@ -1,6 +1,7 @@
+using System;
 using Improbable.Gdk.Core;
 using Improbable.Gdk.Subscriptions;
-using Improbable.Gdk.TestBases;
+using Improbable.Gdk.TestUtils;
 using Improbable.TestSchema;
 using Improbable.Worker.CInterop;
 using NUnit.Framework;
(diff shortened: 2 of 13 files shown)


test-project/Assets/EditmodeTests/Subscriptions/InjectionCriteriaTests.cs
```diff
@@ -12,118 +13,161 @@ namespace Improbable.Gdk.EditmodeTests.Subscriptions
...
-            Assert.IsFalse(readerBehaviour.enabled);
-            Assert.IsNull(readerBehaviour.Reader);
+            World
+                .Step(world => { world.Connection.CreateEntity(EntityId, BasicEntity()); })
+                .Step(world =>
+                {
+                    var (_, readerBehaviour) = world.CreateGameObject<ExhaustiveSingularReaderBehaviour>(EntityId);
+                    return readerBehaviour;
+                })
+                .Step((world, reader) =>
```
zeroZshadow: This does change the test? As there is now a step between creation and assertion.

test-project/Assets/EditmodeTests/Subscriptions/InjectionCriteriaTests.cs
```diff
@@ -12,118 +13,161 @@ namespace Improbable.Gdk.EditmodeTests.Subscriptions
...
-            createdGameObject = CreateAndLinkGameObjectWithComponent<ExhaustiveSingularReaderBehaviour>(EntityId);
-            var readerBehaviour = createdGameObject.GetComponent<ExhaustiveSingularReaderBehaviour>();
-
-            Assert.IsFalse(readerBehaviour.enabled);
-            Assert.IsNull(readerBehaviour.Reader);
+            World
+                .Step(world => { world.Connection.CreateEntity(EntityId, BasicEntity()); })
+                .Step(world =>
+                {
+                    var (_, readerBehaviour) = world.CreateGameObject<ExhaustiveSingularReaderBehaviour>(EntityId);
```
zeroZshadow: Its a little unclear why this test works the way it does. Can we somehow make it clearer that this works because the `BasicEntity` does not contain the component that the monobehaviour here has a reader for?.

test-project/Assets/EditmodeTests/Subscriptions/InjectionCriteriaTests.cs
```diff
@@ -12,118 +13,161 @@ namespace Improbable.Gdk.EditmodeTests.Subscriptions
...
-            createdGameObject = CreateAndLinkGameObjectWithComponent<PositionWriterBehaviour>(EntityId);
-            var writerBehaviour = createdGameObject.GetComponent<PositionWriterBehaviour>();
-
-            Assert.IsTrue(writerBehaviour.enabled);
-            Assert.IsNotNull(writerBehaviour.Writer);
+            World
+                .Step(world =>
+                {
+                    world.Connection.CreateEntity(EntityId, BasicEntity());
+                    world.Connection.ChangeAuthority(EntityId, Position.ComponentId, Authority.Authoritative);
```
zeroZshadow: Do we want to make the component type safe here by using a generic type and the database instead of a uint?

test-project/Assets/EditmodeTests/Subscriptions/RequireablesDisableTests.cs
```diff
@@ -1,117 +1,158 @@
...
+                    return (receiverBehaviour: receiverBehaviour, receiver: receiverBehaviour.Receiver);
+                })
+                .Step(world =>
+                {
+                    world.Connection.ChangeAuthority(EntityId, TestCommands.ComponentId,
+                        Authority.NotAuthoritative);
+                })
+                .Step((world, context) =>
+                {
+                    var (receiverBehaviour, testCommandsCommandReceiver) = context;
```
zeroZshadow: nit: some tests have a new line after the context deconstruction, and some dont
pick one ;)

workers/unity/Packages/io.improbable.gdk.testutils/MockWorld.cs
```diff
@@ -0,0 +1,137 @@
...
+            Worker.World.Update();
+        }
+
+        public (GameObject, T) CreateGameObject<T>(long entityId) where T : MonoBehaviour
+        {
+            var gameObject = new GameObject("TestGameObject");
+            gameObjects.Add(gameObject);
+
+            var component = gameObject.AddComponent<T>();
+            component.enabled = false;
```
zeroZshadow: I think the component will have already called `OnEnabled` at this point.

If it has, you might want to instantiate the test object as a child to a disabled root game object, to avoid it waking up until setup is complete.

workers/unity/Packages/io.improbable.gdk.testutils/MockWorld.cs
```diff
@@ -0,0 +1,137 @@
...
+                UnityEngine.Object.DestroyImmediate(go);
+            }
+
+            Worker?.Dispose();
+            Connection?.Dispose();
+        }
+
+        public struct MockWorldWithContext<T>
+        {
+            internal T Context;
```
zeroZshadow: Can these be private?

test-project/Assets/EditmodeTests/Subscriptions/InjectionCriteriaTests.cs
jamiebrynes7: There was an `Update` in the old `CreateAndLinkGameObject<T>` - https://github.com/spatialos/gdk-for-unity/blob/master/test-project/Assets/TestBases/SubscriptionsTestBase.cs#L45

test-project/Assets/EditmodeTests/Subscriptions/InjectionCriteriaTests.cs
jamiebrynes7: I'd rather not change the tests themselves in this PR

test-project/Assets/EditmodeTests/Subscriptions/InjectionCriteriaTests.cs
jamiebrynes7: Why is this not type-safe currently?

workers/unity/Packages/io.improbable.gdk.testutils/MockWorld.cs
jamiebrynes7: Quite possibly

test-project/Assets/EditmodeTests/Subscriptions/RequireablesDisableTests.cs
jamiebrynes7: I could only find one w/ a new line and its because it has 4 asserts rather than 2 after it so its harder to read when all jumbled together.

Leaving this as is

workers/unity/Packages/io.improbable.gdk.testutils/MockWorld.cs
jamiebrynes7: Had a look at this and paired offline - it seems that in Editmode tests `OnEnable` isn't called

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@
 - Reduce complexity in `ViewDiff` and `MessagesToSend` classes. [#1290](https://github.com/spatialos/gdk-for-unity/pull/1290)
 - De-duplicated code for generated `ComponentDiffStorage` instances. [#1290](https://github.com/spatialos/gdk-for-unity/pull/1290)
 - `Improbable.Gdk.Core.EntityId` is now a readonly struct. [#1290](https://github.com/spatialos/gdk-for-unity/pull/1290)
+- Added `MockWorld` and `MockBase` classes to the `Improbable.Gdk.TestUtils` package. These are designed to help us write tests for Core code. [#1305](https://github.com/spatialos/gdk-for-unity/pull/1305)
```
zeroZshadow:
```suggestion
- Added `MockWorld` and `MockBase` classes to the `Improbable.Gdk.TestUtils` package. These are designed to help write tests for Core code. [#1305](https://github.com/spatialos/gdk-for-unity/pull/1305)
```

workers/unity/Packages/io.improbable.gdk.core/Worker/ConnectionHandlers/MockConnectionHandler.cs
```diff
@@ -114,7 +114,6 @@ public MessagesToSend GetMessagesToSendContainer()
 
         public void PushMessagesToSend(MessagesToSend messages, NetFrameStats netFrameStats)
         {
-            throw new System.NotImplementedException();
```
jessicafalk: why did we remove this?

jamiebrynes7: The full stack runs (including the send system), which means this is being called. I'm not sure what we want to do with this right now, so just removing the error

PR #1313, opened 2020-03-09
Create EntityRangeCollection for upcoming EntityReservationSystem

zeroZshadow:
#### Description
PRing this separately from the reservation system so that it can be reviewed while I work on it.

#### Tests
There are tests

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.core/Collections/EntityRangeCollection.cs
+++ b/workers/unity/Packages/io.improbable.gdk.core/Collections/EntityRangeCollection.cs
@@ -32,10 +32,9 @@ namespace Improbable.Gdk.Core
                 }
                 else
                 {
-                    var splitRange = new EntityIdRange(firstElement.FirstEntityId, count);
+                    var (splitRange, remainder) = firstElement.Split(count);
                     collection.Add(splitRange);
-                    firstElement.FirstEntityId = new EntityId(firstElement.FirstEntityId.Id + count);
-                    firstElement.Count -= count;
+                    firstElement = remainder;
                     break;
                 }
             }
@@ -79,15 +78,28 @@ namespace Improbable.Gdk.Core
             version++;
         }
 
-        internal struct EntityIdRange
+        internal readonly struct EntityIdRange
         {
-            public EntityId FirstEntityId;
-            public uint Count;
+            public readonly EntityId FirstEntityId;
+            public readonly uint Count;
 
             public EntityIdRange(EntityId firstEntityId, uint count)
             {
                 Count = count;
                 FirstEntityId = firstEntityId;
+            }
+
+            public (EntityIdRange part, EntityIdRange remainder) Split(uint count)
+            {
+                if (count > Count)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(count), count, "Split count out of range");
+                }
+
+                var part = new EntityIdRange(FirstEntityId, count);
+                var remainder = new EntityIdRange(new EntityId(FirstEntityId.Id + count), Count - count);
+
+                return (part, remainder);
             }
         }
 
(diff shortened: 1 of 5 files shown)


workers/unity/Packages/io.improbable.gdk.core/Collections/EntityRangeCollection.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        public void Clear()
+        {
+            firstElement = new EntityIdRange();
+            queue.Clear();
+            Count = 0;
+            version++;
+        }
+
+        internal struct EntityIdRange
```
jamiebrynes7: Consider making this immutable and making all the operations methods which return new `EntityIdRange` objects:

```csharp
public (EntityIdRange part, EntityIdRange remainder) Split(uint count)
{
	...
}
```

workers/unity/Packages/io.improbable.gdk.core/Collections/EntityRangeCollection.cs
```diff
@@ -0,0 +1,184 @@
...
+        {
+            public EntityId FirstEntityId;
```
jamiebrynes7: Maybe deal with `long`s until the public API layer for simplicity?

workers/unity/Packages/io.improbable.gdk.core/Collections/EntityRangeCollection.cs
```diff
@@ -0,0 +1,184 @@
...
+                index = 0;
+                version = collection.version;
+                current = default;
+            }
+
+            public bool MoveNext()
+            {
+                if (version != collection.version)
+                {
+                    throw new InvalidOperationException("Cannot iterate over collection which has been altered.");
```
jamiebrynes7: Suggest coping the wording of the standard exceptions for these things if this isn't it already

workers/unity/Packages/io.improbable.gdk.core/Collections/EntityRangeCollection.cs
zeroZshadow: I'd rather leave it like this in case we get more restrictions on EntityId's.
Currently this lets me use `IsValid` to check for <= 0 id's, layer we might check for the <= 62^2-1 range as well.

workers/unity/Packages/io.improbable.gdk.core/Collections/EntityRangeCollection.cs
zeroZshadow: `Collection was modified; enumeration operation may not execute`

zeroZshadow:
EntityRangeCollection is no longer public, so it wont need a changelog

workers/unity/Packages/io.improbable.gdk.core/Tests/Editmode/Collections/EntityRangeCollectionTests.cs
```diff
@@ -0,0 +1,152 @@
...
+            Assert.Throws<ArgumentOutOfRangeException>(() => collection.Take(ValidRangeCount * 2));
+        }
+
+        [Test]
+        public void EntityRangeCollection_clear_resets_collection()
+        {
+            var collection = new EntityRangeCollection();
+            collection.Add(validRange);
+            collection.Clear();
+            Assert.AreEqual(0, collection.Count);
```
jamiebrynes7: Also check that trying to dequeue/take throws

workers/unity/Packages/io.improbable.gdk.core/Tests/Editmode/Collections/EntityRangeCollectionTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        [Test]
+        public void EntityRangeCollection_enumerator_single_range()
```
jamiebrynes7: The name of this test and the one below don't really describe it well

PR #1318, opened 2020-03-13
Fastlane

paulbalaji:
#### Description

Fastlane

#### Tests

- [x] can build and deploy to ipad

#### Documentation

n/a

#### Primary reviewers

### DIFF
--- a/fastlane/Matchfile
+++ b/fastlane/Matchfile
@@ -1,0 +1,9 @@
+git_url("git@github.com:improbable/apple-codesigning.git")
+storage_mode("git")
+type("development")
+app_identifier("io.improbable.*")
+keychain_name("login.keychain")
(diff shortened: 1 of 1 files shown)


fastlane/Matchfile
```diff
@@ -0,0 +1,9 @@
+git_url("git@github.com:improbable/apple-codesigning.git")
+storage_mode("git")
+type("development")
+app_identifier("io.improbable.*")
```
paulbalaji: wasn't sure if I should keep this or make it `io.improbable.gdk.*`

jessicafalk: io.improbable.* is the correct one to use here.

paulbalaji: 👍

paulbalaji:
Validated without logging into master apple account 👍

PR #1326, opened 2020-03-19
Fix Readers not re-validating with AggregateSubscription

zeroZshadow:
#### Description
This fixes the mysterious exceptions with injected, but invalid, readers when using QBI.
Some additional safety improvements.

#### Tests
Added 2 tests, 1 fails without the fix.

#### Documentation
 - [x] Changelog

### DIFF
--- a/test-project/Assets/EditmodeTests/Subscriptions/RequireablesDisableTests.cs
+++ b/test-project/Assets/EditmodeTests/Subscriptions/RequireablesDisableTests.cs
@@ -2,6 +2,7 @@ using Improbable.Gdk.Core;
 using Improbable.Gdk.Subscriptions;
 using Improbable.Gdk.Test;
 using Improbable.Gdk.TestUtils;
+using Improbable.Tests;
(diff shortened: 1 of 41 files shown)


test-project/Assets/EditmodeTests/Subscriptions/RequireablesDisableTests.cs
```diff
@@ -179,6 +241,12 @@ private class MultipleReaderBehaviour : MonoBehaviour
             [Require] public PositionReader PositionReader;
             [Require] public TestCommandsReader TestCommandsReader;
         }
+        
+        private class MultipleComponentBehaviour : MonoBehaviour
```
paulbalaji: how is this behaviour different to the one above? both have component readers?

zeroZshadow: The above one has commands, I wanted to ensure its just components. with nothing on them

PR #1310, opened 2020-03-03
Generated Docs Pipeline

jamiebrynes7:
#### Description
This PR introduces a BuildKite pipeline for producing generated documentation and publishing it to Readme. This PR implements the first part of this. (See the TODO for the second part).

You can see an example build here - https://buildkite.com/improbable/gdk-for-unity-generate-docs/builds/9#_ where the generated docs are present in the artifacts.

PR #1328, opened 2020-03-23
Generated docs pipeline part two

jamiebrynes7:
#### Description

A follow on from #1310, this uses both a pinned docker image reference and triggers the platform pipeline to actually publish the documentation

#### Tests

Changed a source link in one ReadMe doc. Ran the pipeline & checked that it was reverted back. 

#### Documentation

Will update the Wiki for the release process.

### DIFF
--- a/ci/generate-docs.sh
+++ b/ci/generate-docs.sh
@@ -22,7 +22,6 @@ function cleanUp() {
 trap cleanUp EXIT
 
 TAG=$(buildkite-agent meta-data get release-version)
-CATEGORY_ID=$(buildkite-agent meta-data get category-id)
 
 # Check if this tag is valid.
 git rev-parse "${TAG}"
@@ -53,7 +52,7 @@ function generate_step() {
     echo "    build:"
     echo "      env:"
     echo "        bk_readme_project_name: gdk-for-unity"
-    echo "        bk_readme_catergory_id: ${CATEGORY_ID}"
+    echo "        bk_readme_category_slug: api-reference"
     echo "        bk_readme_version: ${TAG}"
     echo "        bk_readme_artifact_path: \"**/*\""
 }
--- a/ci/generate-docs.sh
+++ b/ci/generate-docs.sh
@@ -6,7 +6,7 @@ cd "$(dirname "$0")/../"
 
 DOCGEN_PIN="${DOCGEN_OVERRIDE:-v1.0}"
 CURRENT_DIR=$(pwd)
-TMP_DIR=$(mktemp -d)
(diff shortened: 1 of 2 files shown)


ci/generate-docs.sh
```diff
@@ -30,23 +31,30 @@ CLONE_URL="git@github.com:spatialos/gdk-for-unity.git"
...
         --namespace-filter=".*DeploymentLauncher" \
         --namespace-filter=".*PlaymodeTests" \
         --git-tag="${TAG}"
 
-# TODO: Trigger Filip's pipeline when that's live
+pushd "${OUTPUT_DIR}"
+    buildkite-agent artifact upload "**/*.md"
+popd
+
+function generate_step() {
```
jessicafalk: very small nit: I personally prefer to have the functions grouped together at the start

ci/generate-docs.sh
```diff
@@ -30,23 +31,30 @@ CLONE_URL="git@github.com:spatialos/gdk-for-unity.git"
...
+    echo "steps:"
+    echo "  - trigger: platform-release-docs-readme"
+    echo "    label: Upload documentation"
+    echo "    build:"
+    echo "      env:"
+    echo "        bk_readme_project_name: gdk-for-unity"
```
jessicafalk: nit: think we generally use all uppercase for env variables

ci/generate-docs.sh
```diff
@@ -4,8 +4,9 @@ set -e -u -x -o pipefail
 
 cd "$(dirname "$0")/../"
 
+DOCGEN_PIN="${DOCGEN_OVERRIDE:-v1.0}"
 CURRENT_DIR=$(pwd)
-TMP_DIR=$(mktemp -d)
+TMP_DIR=$(mktemp -d "${TMPDIR:-/tmp}/XXXXXXXXX")
```
jessicafalk: why so many X? what does that do?

zeroZshadow: xxxx?

ci/generate-docs.sh
jamiebrynes7: I don't control this. Blame @filipVisko 😛

ci/generate-docs.sh
jamiebrynes7: Controls how many random characters to generate in the path. 8 characters for sufficiently small chances of collision

jamiebrynes7: See https://github.com/spatialos/gdk-for-unity/pull/1328#discussion_r396390594

ci/generate-docs.sh
jamiebrynes7: Sure, can move it

PR #1333, opened 2020-03-24
Better gameobject initialization

paulbalaji:
#### Description

- `EntityTypeExpectations` keeps track of entitytype -> expectedtype[] map
	- ability to define a default type[] as a fallback
- gameobject initialization system will check if entity has the expected types
	- if it does -> call `OnEntityCreated`
    - if not, return 

#### Tests

- [x] playground
- [x] existing unit tests
- [x] fps

#### Documentation

- [x] changelog
- [x] upgrade guide
- [ ] readme

#### Primary reviewers

### DIFF
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -37,8 +37,8 @@ If you have written custom GameObject creators implementing `IEntityGameObjectCr
 
 The `EntityTypeExpectations` class provides two public methods for defining a set of components expected on an entity to be able create GameObjects for a given entity type:
 
-- `void RegisterDefault(Type[] defaultComponentTypes = null)`
-- `void RegisterEntityType(string entityType, Type[] expectedComponentTypes = null)`
+- `void RegisterDefault(IEnumerable<Type> defaultComponentTypes = null)`
+- `void RegisterEntityType(string entityType, IEnumerable<Type> expectedComponentTypes = null)`
 
 For example, the `GameObjectCreatorFromMetadata` class implements the method like so:
 
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -33,11 +33,14 @@ public static EntityTemplate Player(EntityId entityId, string workerId, byte[] a
 
 #### Implement `PopulateEntityTypeExpectations` method
 
-If you have written custom GameObject creators implementing `IEntityGameObjectCreator`, you will now have to implement a `PopulateEntityTypeExpectations` method.
+If you have written custom GameObject creators implementing `IEntityGameObjectCreator`, you will now have to implement a `void PopulateEntityTypeExpectations(EntityTypeExpectations entityTypeExpectations)` method.
 
-This method is used to define a set of components expected on an entity to be able create GameObjects for a given entity type.
+The `EntityTypeExpectations` class provides two public methods for defining a set of components expected on an entity to be able create GameObjects for a given entity type:
 
-For example, the `GameObjectCreatorFromMetadata` class implements it like so:
+- `void RegisterDefault(Type[] defaultComponentTypes = null)`
+- `void RegisterEntityType(string entityType, Type[] expectedComponentTypes = null)`
+
+For example, the `GameObjectCreatorFromMetadata` class implements the method like so:
 
 ```csharp
 public void PopulateEntityTypeExpectations(EntityTypeExpectations entityTypeExpectations)
@@ -48,6 +51,20 @@ public void PopulateEntityTypeExpectations(EntityTypeExpectations entityTypeExpe
     });
 }
 ```
+
+The `AdvancedEntityPipeline` in the FPS Starter Project makes use of both public methods on the `EntityTypeExpectations` class. This is done in order to use a specifc method for creating "Player" GameObjects, and a fallback for any other entity type:
+
+```csharp
+public void PopulateEntityTypeExpectations(EntityTypeExpectations entityTypeExpectations)
+{
+    entityTypeExpectations.RegisterEntityType(PlayerEntityType, new[]
+    {
+        typeof(OwningWorker.Component), typeof(ServerMovement.Component)
+    });
+
+    fallback.PopulateEntityTypeExpectations(entityTypeExpectations);
+}
+``
 
 #### Add `string entityType` as argument to `OnEntityCreated`
 
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -29,21 +29,48 @@ public static EntityTemplate Player(EntityId entityId, string workerId, byte[] a
 }
 ```
 
-### The `IEntityGameObjectCreator` now requires the `ComponentType[] MiniumComponentTypes { get; }` property
+### `IEntityGameObjectCreator` changes
 
-If you have written custom GameObject creators implementing `IEntityGameObjectCreator`, you will have to define the minimum set of components required on an entity to trigger the `OnEntityCreated` method.
+#### Implement `PopulateEntityTypeExpectations` method
 
-For example, the following has been added to the `GameObjectCreatorFromMetadata` class:
+If you have written custom GameObject creators implementing `IEntityGameObjectCreator`, you will now have to implement a `PopulateEntityTypeExpectations` method.
+
+This method is used to define a set of components expected on an entity to be able create GameObjects for a given entity type.
+
+For example, the `GameObjectCreatorFromMetadata` class implements it like so:
 
 ```csharp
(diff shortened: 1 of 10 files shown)


workers/unity/Packages/io.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -7,7 +6,7 @@
 namespace Improbable.Gdk.GameObjectCreation
 {
     /// <summary>
-    ///     For each newly added SpatialOS entity, calls IEntityGameObjectCreator to get an associated GameObject
+    ///     For each newly added SpatialOS entity, calls IEntityGagit meObjectCreator to get an associated GameObject
```
paulbalaji:
```suggestion
    ///     For each newly added SpatialOS entity, calls IEntityGameObjectCreator to get an associated GameObject
```

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/EntityTypeExpectations.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public void RegisterDefault(Type[] defaultComponentTypes)
+        {
+            defaultExpectation = defaultComponentTypes;
+        }
+
+        public void RegisterEntityType(string entityType, Type[] expectedComponentTypes)
+        {
+            entityExpectations.Add(entityType, expectedComponentTypes ?? new Type[] { });
+        }
```
jamiebrynes7: If the second parameter is optional, make it default `null`. If its not optional, it should be an exception

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/GameObjectInitializationSystem.cs
```diff
@@ -106,5 +105,32 @@ protected override void OnUpdate()
...
+
+        private void ProcessNewEntities()
+        {
+            Entities.With(newEntitiesQuery).ForEach(
+                (Entity entity, ref SpatialEntityId spatialEntityId, ref Metadata.Component metadata) =>
+                {
+                    var entityType = metadata.EntityType;
+                    var expectedTypes = entityTypeExpectations.GetExpectedTypes(entityType);
+
+                    if (expectedTypes != null)
```
jamiebrynes7: If we guarantee in `ExpectedEntityTypes` that this cannot be null, we can get rid of this check. 

(I think there's any reason why this should ever be null, since an empty array is semantically equivalent to `null` here)

UPGRADE_GUIDE.md
```diff
@@ -29,21 +29,48 @@ public static EntityTemplate Player(EntityId entityId, string workerId, byte[] a
...
+For example, the `GameObjectCreatorFromMetadata` class implements it like so:
 
 ```csharp
-public ComponentType[] MinimumComponentTypes { get; } =
+public void PopulateEntityTypeExpectations(EntityTypeExpectations entityTypeExpectations)
 {
-    ComponentType.ReadOnly<Metadata.Component>(),
-    ComponentType.ReadOnly<Position.Component>()
-};
+    entityTypeExpectations.RegisterDefault(new[]
```
zeroZshadow: I would like a non-default registration here as well, with some bogus component names. Just to show the API

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/EntityTypeExpectations.cs
```diff
@@ -8,29 +10,34 @@ namespace Improbable.Gdk.GameObjectCreation
...
-        internal Type[] GetExpectedTypes(string entityType)
+        internal ComponentType[] GetExpectedTypes(string entityType)
         {
             if (!entityExpectations.TryGetValue(entityType, out var types))
             {
                 return defaultExpectation;
             }
 
-            return types;
+            return types ?? new ComponentType[] { };
```
jamiebrynes7: Why not handle the null when we add things to the dictionary rather than here?

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/EntityTypeExpectations.cs
```diff
@@ -8,29 +10,34 @@ namespace Improbable.Gdk.GameObjectCreation
...
+        private readonly Dictionary<string, ComponentType[]> entityExpectations
+            = new Dictionary<string, ComponentType[]>();
 
-        public void RegisterDefault(Type[] defaultComponentTypes)
+        public void RegisterDefault(Type[] defaultComponentTypes = null)
         {
-            defaultExpectation = defaultComponentTypes;
+            defaultExpectation = defaultComponentTypes?
+                .Select(type => new ComponentType(type, ComponentType.AccessMode.ReadOnly))
+                .ToArray();
```
jamiebrynes7: This should never be null. If the user gives `null` it should be an empty array.

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/EntityTypeExpectations.cs
jamiebrynes7: Especially as this is called in a hot-loop

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/EntityTypeExpectations.cs
jamiebrynes7: (Note on L37 we just return `defaultExpectation`)

jamiebrynes7: Also its odd to have a single optional param. Would it make more sense for `defaultExpectation` to be an empty array by default and then you call with method (always with a valid type array)  to override it?

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/EntityTypeExpectations.cs
paulbalaji: fixed in yet-to-be-pushed commit

workers/unity/Packages/io.improbable.gdk.gameobjectcreation/EntityTypeExpectations.cs
paulbalaji: > Would it make more sense for defaultExpectation to be an empty array by default and then you call with method (always with a valid type array) to override it?

That is the behaviour we have now (latest commit)

defaultlist: empty
RegisterDefault(): resets defaultlist to empty
RegisterDefault(null): resets defaultlist to empty
RegisterDefault(yourlist): sets defaultlist to yourlist

UPGRADE_GUIDE.md
```diff
@@ -37,8 +37,8 @@ If you have written custom GameObject creators implementing `IEntityGameObjectCr
 
 The `EntityTypeExpectations` class provides two public methods for defining a set of components expected on an entity to be able create GameObjects for a given entity type:
 
-- `void RegisterDefault(Type[] defaultComponentTypes = null)`
-- `void RegisterEntityType(string entityType, Type[] expectedComponentTypes = null)`
+- `void RegisterDefault(IEnumerable<Type> defaultComponentTypes = null)`
+- `void RegisterEntityType(string entityType, IEnumerable<Type> expectedComponentTypes = null)`
```
jamiebrynes7: The explanation on L55 is a little out of date (it mentions a specific method which ins't true anymore).

UPGRADE_GUIDE.md
```diff
@@ -29,21 +29,65 @@ public static EntityTemplate Player(EntityId entityId, string workerId, byte[] a
...
-};
+    entityTypeExpectations.RegisterDefault(new[]
+    {
+        typeof(Position.Component)
+    });
+}
 ```
 
-> You will need to add `using Unity.Entities;` to the top of the file and reference it in the assembly that your custom GameObject creator is in.
+The `AdvancedEntityPipeline` in the FPS Starter Project makes use of both public methods on the `EntityTypeExpectations` class. This is done in order to use a specifc method for creating "Player" GameObjects, and a fallback for any other entity type:
```
paulbalaji:
```suggestion
The `AdvancedEntityPipeline` in the FPS Starter Project makes use of both public methods on the `EntityTypeExpectations` class. This is done in order to wait for specific components when creating "Player" GameObjects, and a different set of components for any other entity type:
```

UPGRADE_GUIDE.md
paulbalaji: fixed

PR #1341, opened 2020-04-06
Mention RegisterDomainUnload in upgrade guide

zeroZshadow:

### DIFF
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -12,6 +12,12 @@ We've upgraded our dependency on Unity's entities package from 0.1.0 to 0.7.0. A
 
 You'll have to add the built-in module `com.unity.modules.assetbundle` to your project dependencies. This can be done through the package manager window.
 
+If you are using the Blank Project, you will have to remove the following line from `OneTimeInitialisation.cs`.
+
+```csharp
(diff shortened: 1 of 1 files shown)


UPGRADE_GUIDE.md
```diff
@@ -12,6 +12,12 @@ We've upgraded our dependency on Unity's entities package from 0.1.0 to 0.7.0. A
 
 You'll have to add the built-in module `com.unity.modules.assetbundle` to your project dependencies. This can be done through the package manager window.
 
+If you are using the Blank Project, you will have to remove the following line from `OneTimeInitialisation.cs`.
```
jamiebrynes7, line 15: Does this not apply to the FPS?

zeroZshadow: The FPS is an example project, not a project base.

zeroZshadow:
Turns out, we never did this on the Blank Project, only in the playground. Oops

UPGRADE_GUIDE.md
jamiebrynes7: Right, but if you forked the FPS and you were bumping the packages you'd need to do this? I assume people don't rebase their entire project on the new FPS tag

PR #1354, opened 2020-04-24
Hide Android in mobile launcher if build support not installed

paulbalaji:
#### Description

Only show android section of mobile launcher if android build support is installed.

#### Tests

- [x] windows with android
- [x] windows without android
- [x] mac with android
- [x] mac without android

#### Documentation

- [x] changelog

#### Primary reviewers


paulbalaji:
Should we be doing this for iOS as well?

PR #1356, opened 2020-05-01
GDK Tools dialog auto-saves

austinkelmore:
#### Description
Made the GDK Tools dialog auto-save. Also moved the reset button to the bottom instead of the top to make it a bit cleaner.

#### Documentation
* [x] Changelog
* [x] Documentation

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfigurationWindow.cs
+++ b/workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfigurationWindow.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Collections.Generic;
+using System.IO;
 using UnityEditor;
 using UnityEngine;
 
@@ -22,8 +24,7 @@ namespace Improbable.Gdk.Tools
         private const string CodeGeneratorLabel = "Code generator";
         private const string CustomSnapshotPathLabel = "Selected snapshot path";
         private const string GeneralSectionLabel = "General";
-        private const string ResetConfigurationButtonText = "Reset to default";
-        private const string SaveConfigurationButtonText = "Save";
+        private const string ResetConfigurationButtonText = "Reset all to default";
 
         private static GUIContent AddSchemaDirButton;
         private static GUIContent RemoveSchemaDirButton;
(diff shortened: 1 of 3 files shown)


austinkelmore:
> LGTM, though I'm not entirely sure about that timer. Not sure what could replace it though.

I'm not a huge fan of the timer, either. I wish Unity had a concept of "finished entering in text, go ahead and update whatever you need to". I've copied the timer from other sections of code in the GDK, so it's at least consistent.

workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfigurationWindow.cs
```diff
@@ -224,5 +248,22 @@ private void DrawCustomSnapshotDir()
...
+        {
+            var timeSinceLastSave = DateTime.Now - lastSaveTime;
+            if (!hasUnsavedData || timeSinceLastSave < FileSavingInterval)
+            {
+                return;
+            }
+
+            toolsConfig.Save();
+            lastSaveTime = DateTime.Now;
+        }
```
jamiebrynes7, line 267: I think there's a race here where we could save invalid data.

1. User inputs some valid value, `hasUnsavedData` is set to true.
2. User changes a value to invalid. 
3. The interval passes and we are able to save.
4. Invalid data is saved

There should be a check against any config errors before we are able to save: 

```csharp
if (!hasUnsavedData || timeSinceLastSave < FileSavingInterval || configErrors.Any())
```

workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfigurationWindow.cs
```diff
@@ -224,5 +248,22 @@ private void DrawCustomSnapshotDir()
...
+        private void TrySaveChanges()
+        {
+            var timeSinceLastSave = DateTime.Now - lastSaveTime;
+            if (!hasUnsavedData || timeSinceLastSave < FileSavingInterval)
+            {
+                return;
+            }
+
+            toolsConfig.Save();
+            lastSaveTime = DateTime.Now;
```
jamiebrynes7, line 266: Shouldn't we set `hasUnsavedData` to `false` after saving?

austinkelmore: Oh goodness. Good catch!

workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfigurationWindow.cs
austinkelmore: Sounds good, making this change.

PR #1357, opened 2020-05-05
Expose cluster in deployment launcher

jamiebrynes7:
#### Description

As usual, commits are ordered and self contained.

This PR adds support for choosing a specific cluster for a deployment. This is exposed to the user as a exclusive choice between a region _or_ cluster. 

image

image

The same change is applied to the `DeploymentLauncher` csproj. I chose to make the `--cluster` and `--region` flags both optional, but it will throw if neither are provided. In the case where both are provided, `--cluster` takes precedence since it is the more specific option.

This change was made in a way such that it should be backward compatible, there should be no action required upon upgrade.

Also if anyone knows a better way of aligning `GUILayout` controls with `EditorGUILayout` scopes, please let me know 😅 

#### Tests
- [x] Ran a deployment against a region
- [x] Ran a deployment in a specific cluster

#### Documentation

- [x] Changelog
- [x] Readme.io

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.deploymentlauncher/DeploymentLauncherWindow.cs
+++ b/workers/unity/Packages/io.improbable.gdk.deploymentlauncher/DeploymentLauncherWindow.cs
@@ -579,7 +579,37 @@ namespace Improbable.Gdk.DeploymentLauncher
 
             dest.SnapshotPath = EditorGUILayout.TextField("Snapshot Path", source.SnapshotPath);
             dest.LaunchJson = EditorGUILayout.TextField("Launch Config", source.LaunchJson);
-            dest.Region = (DeploymentRegionCode) EditorGUILayout.EnumPopup("Region", source.Region);
+
+            var foldoutState = stateManager.GetStateObjectOrDefault<bool>((source.Name + "region").GetHashCode());
+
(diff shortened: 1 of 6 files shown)


paulbalaji:
Piggybacking off Austin's suggestion, how about something like this:
image

Disabling the half that isn't selected

workers/unity/Packages/io.improbable.gdk.deploymentlauncher/DeploymentLauncherWindow.cs
```diff
@@ -579,7 +579,37 @@ private void RenderBaseDeploymentConfig(BaseDeploymentConfig source, BaseDeploym
 
             dest.SnapshotPath = EditorGUILayout.TextField("Snapshot Path", source.SnapshotPath);
             dest.LaunchJson = EditorGUILayout.TextField("Launch Config", source.LaunchJson);
-            dest.Region = (DeploymentRegionCode) EditorGUILayout.EnumPopup("Region", source.Region);
+
+            var foldoutState = stateManager.GetStateObjectOrDefault<bool>((source.Name + "region").GetHashCode());
```
zeroZshadow, line 583: string interpolation here please, and 2 lines down

jamiebrynes7:
@austinkelmore changed the UI to feel a bit more natural (see updated screenshots above). Shoutout to @paulbalaji for code snippets for that

PR #1359, opened 2020-05-11
Produce test coverage reports

jamiebrynes7:
#### Description

When running our tests, we will now output an `.xml` file with test coverage information. This will later be consumed by a service. I've also removed the Playmode tests for the `playground` project as we have none of those. I'm pretty sure that the existing Playmode tests can be ported to Editmode in the `test-project` too. This should cut down on runtime :) 

#### Tests

Observed artifacts in CI. 

#### Documentation

- [ ] Changelog

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -14,6 +14,9 @@ PROJECT_DIR="$(pwd)"
 TEST_RESULTS_DIR="${PROJECT_DIR}/logs/nunit"
 mkdir -p "${TEST_RESULTS_DIR}"
 
+COVERAGE_OPTIONS="generateHtmlReport\;assemblyFilters:+Improbable.Gdk.*,-Improbable.Gdk.Generated,-Improbable.Gdk.Generated.BuildSystem"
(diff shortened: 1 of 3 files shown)


jamiebrynes7:
> This should cut down on runtime.

Cuts approximately 2 minutes off the test step 🎉

ci/test.sh
```diff
@@ -63,6 +60,9 @@ echo "--- Testing Unity: Test Project Playmode :joystick:"
 pushd "test-project"
     dotnet run -p "${PROJECT_DIR}/.shared-ci/tools/RunUnity/RunUnity.csproj" -- \
         -batchmode \
+        -enableCodeCoverage \
+        -coverageResultsPath "${COVERAGE_RESULTS_PATH}/test-project" \
```
zeroZshadow, line 64: Will it be oke that this and the editmode tests go to the same location?

jamiebrynes7: It creates an `Editmode/Playmode` folder for each 😄

PR #1360, opened 2020-05-12
Add non-generic overloads to `EntityTemplate`

jamiebrynes7:
#### Description

If you had an `ISpatialComponentSnapshot`, you couldn't interact with the `EntityTemplate` object w/o reflection or using the Dynamic API. This PR completes the set of APIs for type-erased usage of the `EntityTemplate` class.

Also driveby fixed a bug where if you added `EntityAcl.Snapshot` with write access, the snapshot wouldn't be added, but the write access would!

#### Tests

Yes there are tests.

#### Documentation

- [x] Class level docs
- [x] Changelog

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.core/Utility/EntityTemplate.cs
+++ b/workers/unity/Packages/io.improbable.gdk.core/Utility/EntityTemplate.cs
@@ -42,6 +42,41 @@ namespace Improbable.Gdk.Core
         public void AddComponent<TSnapshot>(TSnapshot snapshot)
             where TSnapshot : struct, ISpatialComponentSnapshot
         {
+            AddComponent((ISpatialComponentSnapshot) snapshot);
+        }
+
+        /// <summary>
+        ///     Adds a SpatialOS component to the EntityTemplate with write permissions specified.
+        /// </summary>
+        /// <param name="snapshot">The component snapshot to add.</param>
+        /// <param name="writeAccess">
+        ///     The worker attribute that should be granted write access over the <see cref="TSnapshot" /> component.
+        /// </param>
+        /// <typeparam name="TSnapshot">The type of the component snapshot.</typeparam>
+        /// <exception cref="InvalidOperationException">
+        ///     Thrown if the EntityTemplate already contains a component snapshot of type <see cref="TSnapshot" />.
+        /// </exception>
+        /// <remarks>
+        ///     EntityACL is handled automatically by the EntityTemplate, so a EntityACL snapshot will be ignored.
+        /// </remarks>
+        public void AddComponent<TSnapshot>(TSnapshot snapshot, string writeAccess)
+            where TSnapshot : struct, ISpatialComponentSnapshot
+        {
+            AddComponent((ISpatialComponentSnapshot) snapshot, writeAccess);
+        }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


workers/unity/Packages/io.improbable.gdk.core/Utility/EntityTemplate.cs
```diff
@@ -41,6 +41,41 @@ public static string GetWorkerAccessAttribute(string workerId)
         /// </remarks>
         public void AddComponent<TSnapshot>(TSnapshot snapshot)
             where TSnapshot : struct, ISpatialComponentSnapshot
+        {
+            AddComponent((ISpatialComponentSnapshot) snapshot);
```
zeroZshadow, line 45: Do we still need the generic functions?

workers/unity/Packages/io.improbable.gdk.core/Utility/EntityTemplate.cs
```diff
@@ -94,6 +131,17 @@ public void AddComponent<TSnapshot>(TSnapshot snapshot, string writeAccess)
             return null;
         }
 
+        /// <summary>
+        ///     Attempts to get a component snapshot stored in the EntityTemplate.
+        /// </summary>
+        /// <param name="componentId">The ID of the component to fetch.</param>
+        /// <returns>The component snapshot, if the component snapshot exists, null otherwise.</returns>
+        public ISpatialComponentSnapshot GetComponent(uint componentId)
```
zeroZshadow, line 139: This is a little confusing, I'd rather make this a `TryGet` method.
`ISpatialComponentSnapshot` is always a struct otherwise, so having this return null feels odd

zeroZshadow:
This seems to have a changelog from another PR?

jamiebrynes7:
> This seems to have a changelog from another PR?

Yeah I forgot it in the previous one 😬

workers/unity/Packages/io.improbable.gdk.core/Utility/EntityTemplate.cs
jamiebrynes7: Hmm possibly not, it should be backwards compatible to remove the generic ones. Let me have a think

workers/unity/Packages/io.improbable.gdk.core/Utility/EntityTemplate.cs
jamiebrynes7: I'm fine to add a `TryGet` as well. 

`ISpatialComponentSnapshot` could not be a struct, this function doesn't know that and we can't constrain it as such.

workers/unity/Packages/io.improbable.gdk.core/Utility/EntityTemplate.cs
jamiebrynes7: Removed the generic `Add` and `Set` variants. The rest have both a generic and non-generic.

workers/unity/Packages/io.improbable.gdk.core/Utility/EntityTemplate.cs
jamiebrynes7: Added `TryGetComponent` methods for both typed and type-erased variants

PR #1361, opened 2020-05-12
Disable flakey tests on MacOS

jamiebrynes7:
#### Description

These are flaking enough that we should disable them and potentially investigate why they flake. I considered test categories, but there doesn't seem to be support for _excluding_ a category when running tests, so every other test would need to have a category that says 'not flakey'. Would also require extra bash logic which is always undesirable. 

#### Tests

Going to check the test results on the MacOS agent to ensure these aren't run.

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.testutils/Editor/Tests/SpatialDeploymentManagerTests.cs
+++ b/workers/unity/Packages/io.improbable.gdk.testutils/Editor/Tests/SpatialDeploymentManagerTests.cs
@@ -1,3 +1,4 @@
+#if !UNITY_EDITOR_OSX
 using System;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


workers/unity/Packages/io.improbable.gdk.testutils/Editor/Tests/SpatialDeploymentManagerTests.cs
```diff
@@ -1,3 +1,4 @@
+#if !UNITY_EDITOR_OSX
```
paulbalaji, line 1: image

jamiebrynes7:
> Going to check the test results on the MacOS agent to ensure these aren't run.

Can confirm, they were not run.

PR #1362, opened 2020-05-12
Move test to editmode

jamiebrynes7:
#### Description

There was a playmode test that was in the wrong place, moved it to the correct place + ported to the new mocking framework we added a little while ago. I was hoping to get rid of the Playmode tests, but we do actually have one :( 

#### Tests

Yes these are tests.


test-project/Assets/EditmodeTests/Subscriptions/SubscriptionAggregateCouplingTests.cs
```diff
@@ -157,45 +140,49 @@ public void DifferentAggregateSubscriptions_should_not_couple_together(Type requ
...
-            RemoveAndUnlinkGameObject(1, gameObject1);
-            RemoveAndUnlinkGameObject(2, gameObject2);
-
-            var gameObject3 = CreateAndLinkGameObject<T>(3);
-            var gameObject4 = CreateAndLinkGameObject<T>(4);
-
-            RemoveAndUnlinkGameObject(3, gameObject3);
-
-            Assert.IsTrue(gameObject4.GetComponent<T>().enabled, "gameObject4.GetComponent<T>().enabled");
+            World
```
paulbalaji, line 143: noice.

test-project/Assets/EditmodeTests/Subscriptions/SubscriptionAggregateCouplingTests.cs
```diff
@@ -157,45 +140,49 @@ public void DifferentAggregateSubscriptions_should_not_couple_together(Type requ
...
+                    world.Connection.CreateEntity(2, GetEntityTemplate());
+                })
+                .Step(world =>
+                {
+                    world.CreateGameObject<T>(1);
+                    world.CreateGameObject<T>(2);
+                })
+                .Step(world =>
+                {
+                    world.Connection.RemoveComponent(1, Position.ComponentId);
```
zeroZshadow, line 156: This is a workaround for `RemoveEntity` not removing components right?

jamiebrynes7: I don't believe its required/matters in this case, just porting the test as close as possible.

PR #1364, opened 2020-05-15
Remove the View

zeroZshadow:
#### Description
Experimental change to remove the View and ViewStorage classes.
This also removes our support for AuthorityImminentLoss.

Might move the WorkerFlags from the `WorkerSystem` to the `Worker` class. Opinions welcome.

#### Documentation
This will need a rather heavy upgrade guide on:
* Reading worker flags.
* Deprecating AuthorityImminentLoss.
* The view is gone, its just gone.


workers/unity/Packages/io.improbable.gdk.core/Subscriptions/CommandSenderReceiverSubscriptionManagerBase.cs
```diff
@@ -124,19 +124,22 @@ public override void ResetValue(ISubscription subscription)
...
-        private readonly uint componentId;
+        private readonly ComponentType componentType;
+        private readonly ComponentType componentAuthType;
 
         protected CommandReceiverSubscriptionManagerBase(World world, uint componentId) : base(world)
         {
-            this.componentId = componentId;
-            componentUpdateSystem = world.GetExistingSystem<ComponentUpdateSystem>();
+            componentType = ComponentDatabase.Metaclasses[componentId].Data;
+            componentAuthType = ComponentDatabase.Metaclasses[componentId].Authority;
```
jamiebrynes7, line 140: Use `ComponentDatabase.GetMetaclass(componentId)` instead? It gives a more helpful error if this ever was to throw an error

workers/unity/Packages/io.improbable.gdk.core/Subscriptions/CommandSenderReceiverSubscriptionManagerBase.cs
```diff
@@ -196,8 +199,8 @@ public override Subscription<T> Subscribe(EntityId entityId)
             }
 
             if (WorkerSystem.TryGetEntity(entityId, out var entity)
-                && componentUpdateSystem.HasComponent(componentId, entityId)
-                && componentUpdateSystem.GetAuthority(entityId, componentId) != Authority.NotAuthoritative)
+                && entityManager.HasComponent(entity, componentType)
```
jamiebrynes7, line 202: Is this actually required now? The presence of the Auth component is sufficient right?

workers/unity/Packages/io.improbable.gdk.core/Subscriptions/Reader.cs
```diff
@@ -35,7 +37,7 @@ public TComponent Data
             }
         }
 
-        public Authority Authority
+        public bool Authority
```
jamiebrynes7, line 40: If we are changing this to `bool` we should rename it to `IsAuthoritative`

workers/unity/Packages/io.improbable.gdk.core/Subscriptions/WriterSubscriptionManager.cs
```diff
@@ -9,17 +9,19 @@ namespace Improbable.Gdk.Subscriptions
...
-        private readonly ComponentUpdateSystem componentUpdateSystem;
+        private readonly EntityManager entityManager;
         private Dictionary<EntityId, HashSet<Subscription<TWriter>>> entityIdToWriterSubscriptions;
 
         private readonly HashSet<EntityId> entitiesMatchingRequirements = new HashSet<EntityId>();
         private readonly HashSet<EntityId> entitiesNotMatchingRequirements = new HashSet<EntityId>();
 
         private static readonly uint ComponentId = ComponentDatabase.GetComponentId<TComponent>();
+        private static readonly ComponentType ComponentType = ComponentDatabase.Metaclasses[ComponentId].Data;
+        private static readonly ComponentType ComponentAuthType = ComponentDatabase.Metaclasses[ComponentId].Authority;
```
jamiebrynes7, line 20: Use `ComponentDatabase.GetMetaclass(componentId)` instead? It gives a more helpful error if this ever was to throw an error

workers/unity/Packages/io.improbable.gdk.core/Subscriptions/Reader.cs
```diff
@@ -17,7 +17,9 @@ public abstract class Reader<TComponent, TUpdate> : IRequireable
         protected readonly EntityManager EntityManager;
         protected readonly Entity Entity;
         protected readonly EntityId EntityId;
+
         private static readonly uint ComponentId = ComponentDatabase.GetComponentId<TComponent>();
+        private static readonly ComponentType ComponentAuthType = ComponentDatabase.Metaclasses[ComponentId].Authority;
```
jamiebrynes7, line 22: Use `ComponentDatabase.GetMetaclass(componentId)` instead? It gives a more helpful error if this ever was to throw an error

workers/unity/Packages/io.improbable.gdk.core/Subscriptions/WriterSubscriptionManager.cs
```diff
@@ -84,8 +86,8 @@ public override Subscription<TWriter> Subscribe(EntityId entityId)
             }
 
             if (WorkerSystem.TryGetEntity(entityId, out var entity)
-                && componentUpdateSystem.HasComponent(ComponentId, entityId)
-                && componentUpdateSystem.GetAuthority(entityId, ComponentId) != Authority.NotAuthoritative)
+                && entityManager.HasComponent(entity, ComponentType)
```
jamiebrynes7, line 89: Same comment here, the auth component should be sufficient?

workers/unity/Packages/io.improbable.gdk.core/Systems/ComponentUpdateSystem.cs
```diff
@@ -72,26 +70,11 @@ public List<EntityId> GetComponentsRemoved(uint componentId)
...
-
-        public T GetComponent<T>(EntityId entityId) where T : struct, ISpatialComponentSnapshot
-        {
-            return worker.View.GetComponent<T>(entityId);
-        }
-
         public void AcknowledgeAuthorityLoss(EntityId entityId, uint componentId)
         {
             worker.MessagesToSend.AcknowledgeAuthorityLoss(entityId.Id, componentId);
         }
```
jamiebrynes7, line 76: I thought this PR was removing support for this?

workers/unity/Packages/io.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
```diff
@@ -31,17 +31,17 @@ protected override void OnUpdate()
         {
             try
             {
-                worker.Tick();
+                workerSystem.Tick();
 
-                var diff = worker.Diff;
-                worker.View.ApplyDiff(diff);
+                var diff = workerSystem.Diff;
+                workerSystem.ApplyDiff(diff);
```
jamiebrynes7, line 37: This feels a little weird. Shouldn't `Tick` just do `ApplyDiff` internally?

workers/unity/Packages/io.improbable.gdk.core/Worker/Worker.cs
```diff
@@ -111,16 +109,6 @@ public void SendLogMessage(LogLevel logLevel, string message, string loggerName,
...
 
-        /// <summary>
-        ///     Gets the value for a given worker flag.
-        /// </summary>
-        /// <param name="key">The key of the worker flag.</param>
-        /// <returns>The value of the flag, if it exists, null otherwise.</returns>
-        public string GetWorkerFlag(string key)
-        {
-            return View.GetWorkerFlag(key);
-        }
```
jamiebrynes7, line 122: Given this was exposed here before.. can we keep the worker flags in here and not add another breaking change?

workers/unity/Packages/io.improbable.gdk.core/Subscriptions/Reader.cs
zeroZshadow: `HasAuthority` ?

workers/unity/Packages/io.improbable.gdk.core/Systems/SpatialOSReceiveSystem.cs
zeroZshadow: I explicitly didn't do this to keep all places where the diff is applied in 1 place.
This could later change to some pubsub system where the workerSystem publishes the new diff and others system subscribe and apply it

jamiebrynes7: If we do https://github.com/spatialos/gdk-for-unity/pull/1364#discussion_r464977855 instead it may fix itself?

workers/unity/Packages/io.improbable.gdk.core/Subscriptions/Reader.cs
```diff
@@ -35,16 +37,16 @@ public TComponent Data
...
             get
             {
                 if (!IsValid)
                 {
-                    throw new InvalidOperationException("Cannot read authority when Reader is not valid");
+                    throw new InvalidOperationException($"Cannot read {nameof(HasAuthority)} when Reader is not valid");
                 }
 
-                return ComponentUpdateSystem.GetAuthority(EntityId, ComponentId);
+                return EntityManager.HasComponent(Entity, ComponentAuthType);
```
jamiebrynes7, line 49: On balance, should we make this breaking change?

workers/unity/Packages/io.improbable.gdk.core/Dynamic/ComponentDatabase.cs
```diff
@@ -43,6 +43,16 @@ public static IComponentMetaclass GetMetaclass(uint componentId)
             return metaclass;
         }
 
+        public static IComponentMetaclass GetMetaclass<T>() where T : ISpatialComponentData
```
zeroZshadow, line 46: This was added to avoid having to get the component id first.
However mostly to make sonar stop complaining about the generic statics in Reader.cs

workers/unity/Packages/io.improbable.gdk.core/Dynamic/ComponentDatabase.cs
```diff
@@ -43,6 +43,16 @@ public static IComponentMetaclass GetMetaclass(uint componentId)
...
+        {
+            if (!ComponentsToIds.TryGetValue(typeof(T), out var id))
+            {
+                throw new ArgumentException($"Can not find Metaclass for unregistered SpatialOS component {nameof(T)}.");
+            }
+
+            return Metaclasses[id];
+        }
```
jamiebrynes7, line 54: Not sure you needed this overload, but doesn't matter that much 😛

workers/unity/Packages/io.improbable.gdk.core/Dynamic/ComponentDatabase.cs
jamiebrynes7: Ah that explains https://github.com/spatialos/gdk-for-unity/pull/1364#discussion_r465687105 then 😂

workers/unity/Packages/io.improbable.gdk.core/Subscriptions/Reader.cs
zeroZshadow: What is the breaking change here?

jamiebrynes7: Both the type and name of the property are changing. 

We could keep this backwards compatible by changing the return to:

```
return EntityManager.HasComponent(Entity, ComponentAuthType) ? Authority.Authoritative : Authority.NotAuthoritative;
```

jamiebrynes7: Right, but removing auth loss imminent doesn't mandate or require this change

paulbalaji: I guess I just don't see the value in keeping backwards compatibility given we'll be imposing that `Authority` is essentially nothing but true or false

Sure it's breaking but it does simplify logic for users (no more ==/!= comparisons)

jamiebrynes7: My opinion is that the bar for making a breaking change should be relatively high and be done for good reason. With this case I don't feel the rationale is strong enough, but if you two both feel otherwise, I'm not gonna die on this hill.

UPGRADE_GUIDE.md
```diff
@@ -1,5 +1,23 @@
...
+* `componentUpdateSystem.GetAuthority(entityId, componentId)` to `EntityManager.HasComponent<ComponentType.Authority>(entity)`.
+* `componentUpdateSystem.GetComponent<ComponentType.Snapshot>(entityId)` to `EntityManager.GetComponentData<ComponentType.Component>(entity)`.
+* `componentUpdateSystem.HasComponent(componentId, entityId)` to `EntityManager.HasComponent<ComponentType.Component>(entity)`
+
+The `WorkerSystem` and `ComponentDatabase` can be used to convert from `entityId` to `entity`, and from `componentId` to a component type.
+
+### Readers and Writers Authority
+
+The `Reader` and `Writer` classes have had their `Authority` property changed to `HasAuthority`.
+Instead of returning an Enum, they now return a boolean indicating whether you have authority over the given component or not.
```
jamiebrynes7, line 19:
```suggestion
The `Reader` and `Writer` classes have had their `public Authority Authority` property changed to `public bool HasAuthority`. `HasAuthority` now indicates whether you have authority over the given component or not.
```

PR #1365, opened 2020-05-19
Change up worker connector semantics slightly

jamiebrynes7:
#### Description

`WorkerConnecter.Connect` now throws when it fails to connect. This makes the `HandleWorkerConnectionFailed` redundant and has been removed. I've also made a couple of behavioural tweaks:

- The `WorkerConnector` no longer destroys itself when it fails to connect. Only concern I have here is embedded in the WorkerConnector is this odd disposable MonoBehaviour, I _think_ the reason it destroyed itself was that it was considered "consumed" after a disconnect, but there's no reason why you can't use the same object to connect again! This feels like its tied up in a confused abstraction (where the `WorkerConnector` is both a way to connect a worker and something to represent the worker after connection).
- The WorkerConnector won't log the failure (since an unhandled exception would log, and a handled exception.. is well.. handled).

I expect we will want further work to refine this abstraction (probably related to some of the thoughts that @zeroZshadow had). 

#### Tests

Tested successful & failed worker connections to observe the behaviour.

#### Documentation

- [x] Changelog
- [x] Upgrade guide
- [x] Public docs changes (https://documentation.improbable.io/gdk-for-unity/docs/worker-connectors & https://documentation.improbable.io/gdk-for-unity/docs/blank-project-tutorial-3a-instantiate-a-level)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Breaking Changes
 
 - `WorkerConnector.HandleWorkerConnectionFailure` has been removed and `WorkerConnector.Connect` now throws exceptions for connection errors instead. [#1365](https://github.com/spatialos/gdk-for-unity/pull/1365)
+- `WorkerConnector` no longer destroys itself in `Dispose`. [#1365](https://github.com/spatialos/gdk-for-unity/pull/1365)
 
 ### Added
 
@@ -15,7 +16,6 @@
 ### Changed
 
 - GDK Tools Configuration window now autosaves. [#1356](https://github.com/spatialos/gdk-for-unity/pull/1356)
-- `WorkerConnector` no longer destroys itself in `Dispose`. [#1365](https://github.com/spatialos/gdk-for-unity/pull/1365)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Breaking Changes
+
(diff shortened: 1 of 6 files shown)


UPGRADE_GUIDE.md
```diff
@@ -1,5 +1,53 @@
 # Upgrade Guide
 
+## From `0.3.5` to `0.3.6`
+
+### WorkerConnector changes
+
+There are two related changes that you will need to adjust for. Previously, the `WorkerConnector.Connect` class would trigger the `HandleWorkerConnectionFailure` callback if the connection failed. 
+
+We've removed this callback, and `WorkerConnector.Connect` will now throw an exception for a failed connection. For example:
```
paulbalaji, line 9:
```suggestion
There are two related changes that you will need to adjust for.

The `WorkerConnector.Connect` class would previously trigger the `HandleWorkerConnectionFailure` callback if the connection failed. We've now removed this callback, and `WorkerConnector.Connect` will now throw an exception for a failed connection.

For example:
```

CHANGELOG.md
```diff
@@ -11,6 +15,7 @@
 ### Changed
 
 - GDK Tools Configuration window now autosaves. [#1356](https://github.com/spatialos/gdk-for-unity/pull/1356)
+- `WorkerConnector` no longer destroys itself in `Dispose`. [#1365](https://github.com/spatialos/gdk-for-unity/pull/1365)
```
paulbalaji, line 18: is this not a breaking change too?

jamiebrynes7: Ah yes you're correct

PR #1367, opened 2020-05-20
Remove the MultiThreadedSpatialOSConnectionHandler

zeroZshadow:
#### Description
Remove the MultiThreadedSpatialOSConnectionHandler

Also remove the CommandMetaDataManager and CommandMetaDataAggregate, these were only there to make multithreading work.

#### Tests
Made sure everything still works in the playground.

#### Documentation
* [x] Changelog
* [x] Update Readme


paulbalaji:
Don't forget to remove the related API docs pages as well

workers/unity/Packages/io.improbable.gdk.core/Worker/SerializedMessagesToSend.cs
```diff
@@ -230,7 +226,7 @@ public void AddRequest(CommandRequest request, uint commandId, long entityId, ui
         public void AddResponse(CommandResponse response, uint requestId)
         {
             responses.Add(new ResponseToSend(response, requestId));
-            netFrameStats.AddCommandResponse(response, message: null);
+            netFrameStats.AddCommandResponse(response, null);
```
austinkelmore, line 229: Any reason to remove this? It seems a bit more clear what's being passed null with the message signifier.

UPGRADE_GUIDE.md
```diff
@@ -2,6 +2,11 @@
 
 ## From `0.3.5` to `0.3.6`
 
+### Multithreaded connection removed
+
+The old `MultiThreadedSpatialOSConnectionHandler` has been removed and the method `SpatialOSConnectionHandlerBuilder.SetThreadingMode` has been removed to reflect this change.
+In a future release, threading will be implemented into the `SpatialOSConnectionHandler`.
```
austinkelmore, line 8: Does this mean threading is no longer handled? That's currently how I'm reading it - or is there some other underlying reason to remove this that users would understand?

zeroZshadow:
Deleted the API docs for deleted classes

workers/unity/Packages/io.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandler/SpatialOSConnectionHandler.cs
```diff
@@ -36,12 +36,14 @@ public List<string> GetWorkerAttributes()
 
         public void GetMessagesReceived(ref ViewDiff viewDiff)
         {
+            commandMetaData.FlushRemovedIds();
```
austinkelmore, line 39: Does this only get called once a frame or can this have bad side effects if it's called before everything is handled? Is there any way to guard or error against this? Or could that be intended behaviour?

workers/unity/Packages/io.improbable.gdk.core/Worker/SerializedMessagesToSend.cs
zeroZshadow: Seems the formatter got rid of this, will put it back.

UPGRADE_GUIDE.md
zeroZshadow: This means that currently, the mode to choose between a multi and single threaded connection handler has been removed. In the future, THE connection handler will be threaded through jobs as planned.

workers/unity/Packages/io.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandler/SpatialOSConnectionHandler.cs
zeroZshadow: This is just cleaning up garbage data

UPGRADE_GUIDE.md
austinkelmore: I guess what I'm asking is what happens if someone was using the multi threaded handler? Are they no longer able to do that? Are there impacts onto users who were doing that before? We say it's going to be handled in the future, but what about the mean time?

zeroZshadow: If someone was using this, they can no longer use it.
However, we've never publicly documented this feature outside of the api existing in the generated api reference docs.

workers/unity/Packages/io.improbable.gdk.core/Worker/ConnectionHandlers/SpatialOSConnectionHandler/SpatialOSConnectionHandler.cs
austinkelmore: When does it become garbage data? I'm wondering if re-entering into this function and flushing it before it becomes garbage is a possibility.

zeroZshadow: `CommandMetaData` is a storage provider for linking a send Request (data) to the request id we got back when the request was send over the network.
Whenever the response comes back, we use the request id that comes with it to lookup the initial Request through `CommandMetaData` to include in the command `Response` structure for further processing. At this point the request ID is marked for removal as it has been processed.

This could likely be refactored to simply remove the request id straight away, instead of a mark and delete.

zeroZshadow: Updated the PR delete immediately

workers/unity/Packages/io.improbable.gdk.core/Worker/CommandMetaData.cs
```diff
@@ -46,96 +44,46 @@ public CommandMetaData()
...
-            if (!componentIdToCommandIdToStorage.TryGetValue(componentId, out var commandIdToStorage))
-            {
-                throw new ArgumentException($"Can not find command meta data. Unknown component ID {componentId}");
-            }
-
-            if (!commandIdToStorage.TryGetValue(commandId, out var storage))
+            if (!componentCommandToStorage.TryGetValue((componentId, commandId), out var storage))
             {
-                throw new ArgumentException($"Can not find command meta data storage. Unknown command ID {commandId}");
+                throw new ArgumentException($"Can not find command meta data. Unknown component:command ID {componentId}:{commandId}");
```
paulbalaji, line 83:
```suggestion
                throw new ArgumentException($"Can not find command metadata. Unknown command with ID {commandId} on component {componentId}.");
```

UPGRADE_GUIDE.md
austinkelmore: I think the only thing I'd suggest is removing "in a future release" unless we have a plan for it happening sometime soon. If it's part of the backlog and we can prioritize it, then that works, or if not let's get a ticket in for it if it's important. Otherwise maybe just remove that line.

zeroZshadow: Removed the line

UPGRADE_GUIDE.md
```diff
@@ -2,6 +2,10 @@
 
 ## From `0.3.5` to `0.3.6`
 
+### Multithreaded connection removed
+
+The old `MultiThreadedSpatialOSConnectionHandler` has been removed and the method `SpatialOSConnectionHandlerBuilder.SetThreadingMode` has been removed to reflect this change.
```
paulbalaji, line 7: Not sure if this is needed in the upgrade guide? Perhaps mention in the changelog itself that the `SetThreadingMode` method has been removed

jamiebrynes7, line 7:
```suggestion
The `MultiThreadedSpatialOSConnectionHandler` has been removed and the method `SpatialOSConnectionHandlerBuilder.SetThreadingMode` has been removed to reflect this change.
```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - `WorkerConnector.HandleWorkerConnectionFailure` has been removed and `WorkerConnector.Connect` now throws exceptions for connection errors instead. [#1365](https://github.com/spatialos/gdk-for-unity/pull/1365)
 - `WorkerConnector` no longer destroys itself in `Dispose`. [#1365](https://github.com/spatialos/gdk-for-unity/pull/1365)
+- `MultiThreadedSpatialOSConnectionHandler` has been removed. [#1367](https://github.com/spatialos/gdk-for-unity/pull/1367)
```
jamiebrynes7, line 9: And the `SetThreadingMode` on the builder right?

zeroZshadow: Yes, thats why this is mentioned in the upgrade guide. Didn't think it belonged in the changelog

PR #1369, opened 2020-05-21
Add worker sdk override env variables

jamiebrynes7:
#### Description

Added an override for the Worker SDK version in the `init.sh` and `init.ps1`. This allows for someone to run a Buildkite build against the GDK with a given Worker SDK version by setting the `WORKER_SDK_OVERRIDE` environment variable in the build options.

#### Tests

- Checked the value of `SDK_VERSION`/`$SdkVersion` with and without the environment variable set.

#### Documentation

Not sure where to put this info... wiki maybe?

### DIFF
--- a/init.ps1
+++ b/init.ps1
@@ -15,7 +15,12 @@ $SdkPath = $PkgRoot + "/io.improbable.worker.sdk"
 $SdkMobilePath = $PkgRoot + "/io.improbable.worker.sdk.mobile"
 $TestSdkPath="test-project/Packages/io.improbable.worker.sdk.testschema"
 
-$SdkVersion = Get-Content ($SdkPath + "/.sdk.version")
+if (Test-Path env:WORKER_SDK_OVERRIDE) {
+    $SdkVersion = $env:WORKER_SDK_OVERRIDE;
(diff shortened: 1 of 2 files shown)


init.ps1
```diff
@@ -15,7 +15,12 @@ $SdkPath = $PkgRoot + "/io.improbable.worker.sdk"
 $SdkMobilePath = $PkgRoot + "/io.improbable.worker.sdk.mobile"
 $TestSdkPath="test-project/Packages/io.improbable.worker.sdk.testschema"
 
-$SdkVersion = Get-Content ($SdkPath + "/.sdk.version")
+if (Test-Path env:WORKER_SDK_OVERRIDE) {
```
paulbalaji, line 18: why do we still have the ps1 script?

jamiebrynes7: No idea 🤷

jamiebrynes7:
@RNabel @tbimprob when this rolls in, you'd be able to test Worker SDK bundles against the GDK builds. This won't catch _all_ regressions (dynamic linking errors at runtime for example), but would give a reasonable first pass.

PR #1370, opened 2020-05-21
Add convenience method to the `MockConnection` 

jamiebrynes7:
#### Description

See branch name for motivation, you can now remove components & an entity the same way you add them.

#### Tests

- [x] Updated a test to double check!

#### Documentation

- [x] Changelog?

### DIFF
--- a/test-project/Assets/EditmodeTests/Collections/ReferenceProviderTests.cs
+++ b/test-project/Assets/EditmodeTests/Collections/ReferenceProviderTests.cs
@@ -51,7 +51,7 @@ namespace Improbable.Gdk.EditmodeTests.Collections
         {
             var template = new EntityTemplate();
             template.AddComponent(new Position.Snapshot(), "some-worker");
-            template.AddComponent(new Metadata.Snapshot(), "some-worker");
+            template.AddComponent(new Metadata.Snapshot("AReferenceType"), "some-worker");
             return template;
         }
     }
--- a/test-project/Assets/EditmodeTests/Collections/ReferenceProviderTests.cs
+++ b/test-project/Assets/EditmodeTests/Collections/ReferenceProviderTests.cs
@@ -17,13 +17,7 @@ namespace Improbable.Gdk.EditmodeTests.Collections
             {
                 mockWorld.Step(world =>
                 {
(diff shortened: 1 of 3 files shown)


test-project/Assets/EditmodeTests/Collections/ReferenceProviderTests.cs
```diff
@@ -35,25 +29,30 @@ public void Disposed_world_cleans_provider()
...
 
                 Assert.AreEqual(0, ReferenceProvider<string>.Count);
             }
         }
+
+        private EntityTemplate GetTemplate()
+        {
+            var template = new EntityTemplate();
+            template.AddComponent(new Position.Snapshot(), "some-worker");
+            template.AddComponent(new Metadata.Snapshot(), "some-worker");
```
zeroZshadow, line 54: This should have a value to keep the tests the same

test-project/Assets/EditmodeTests/Collections/ReferenceProviderTests.cs
```diff
@@ -35,25 +29,30 @@ public void Disposed_world_cleans_provider()
...
-                });
+                mockWorld
+                    .Step(world =>
+                    {
+                        world.Connection.CreateEntity(entityId, template);
+                    })
+                    .Step(world =>
+                    {
+                        Assert.AreEqual(1, ReferenceProvider<string>.Count);
+                        world.Connection.RemoveEntityAndComponents(entityId, template);
```
zeroZshadow, line 43: The test previously only removed the component, not the whole entity?

test-project/Assets/EditmodeTests/Collections/ReferenceProviderTests.cs
jamiebrynes7: Hahaha this actually means we pushed a `null` into the provider

test-project/Assets/EditmodeTests/Collections/ReferenceProviderTests.cs
jamiebrynes7: Discussed offline - this is what created the unhappiness in the first place (the fact that `RemoveEntity` didn't work originally). So we are good to swap this over

PR #1372, opened 2020-05-22
Typed command ids

zeroZshadow:
#### Description
Make command request ID's typed.
Split interfaces for commands into their own files.

Small drive-by on making comparer classes sealed.

#### Documentation
* [x] Changelog
* [x] Upgrade Guide


workers/unity/Packages/io.improbable.gdk.core/Commands/CommandRequestId.cs
```diff
@@ -0,0 +1,124 @@
...
+            return Raw.GetHashCode();
+        }
+
+        public int CompareTo(CommandRequestId other)
+        {
+            return Raw.CompareTo(other.Raw);
+        }
+    }
+
+    public readonly struct InternalCommandRequestId : IEquatable<InternalCommandRequestId>, IComparable<InternalCommandRequestId>
```
austinkelmore, line 65: What is the difference between `InternalCommandRequestId` and `CommandRequestId`? Is it that they're two different wrappers for two different types of commands and could theoretically deviate from each other? Or are you trying to create type safety?

workers/unity/Packages/io.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
```diff
@@ -65,11 +65,16 @@ private void SendPlayerCreatorEntityQuery()
 
         private void HandleEntityQueryResponses()
         {
+            if (!playerCreatorEntityQueryId.HasValue)
```
austinkelmore, line 68: What does it not having a value mean has happened? How has that changed with the rest of the code, or is this something you found because of the change?

UPGRADE_GUIDE.md
```diff
@@ -2,6 +2,13 @@
 
 ## From `0.3.5` to `0.3.6`
 
+### Command Request ID's are now typed
+
+The return value of `CommandSystem.SendCommand` has changed from `long` to `CommandRequestID`.
+A similar change has been made to all API's that take request ID's as a parameter.
```
jamiebrynes7, line 8:
```suggestion
A similar change has been made to all APIs that take request IDs as a parameter.
```

UPGRADE_GUIDE.md
```diff
@@ -2,6 +2,13 @@
...
+
+Anywhere you store request ids, you need to change the type from `long` to `CommandRequestId`.
```
jamiebrynes7, line 10:
```suggestion
Anywhere you store request IDs, you need to change the type from `long` to `CommandRequestId`.
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - `WorkerConnector.HandleWorkerConnectionFailure` has been removed and `WorkerConnector.Connect` now throws exceptions for connection errors instead. [#1365](https://github.com/spatialos/gdk-for-unity/pull/1365)
 - `WorkerConnector` no longer destroys itself in `Dispose`. [#1365](https://github.com/spatialos/gdk-for-unity/pull/1365)
 - `MultiThreadedSpatialOSConnectionHandler` and `SpatialOSConnectionHandlerBuilder.SetThreadingMode` have been removed. [#1367](https://github.com/spatialos/gdk-for-unity/pull/1367)
+- Command request ID's are now typed as `CommandRequestID` instead of `long`. [#1372](https://github.com/spatialos/gdk-for-unity/pull/1372)
```
jamiebrynes7, line 10:
```suggestion
- Command request IDs are now typed as `CommandRequestID` instead of `long`. [#1372](https://github.com/spatialos/gdk-for-unity/pull/1372)
```

workers/unity/Packages/io.improbable.gdk.playerlifecycle/Systems/PlayerCreation/SendCreatePlayerRequestSystem.cs
zeroZshadow: Previously a check was done `if(long? == long)` which worked because there is an implicit conversion there. This implicit conversion is gone.

If there is no Value in that nullable nothing is supposed to happen, so I changed it to an early out.

workers/unity/Packages/io.improbable.gdk.core/Commands/CommandRequestId.cs
zeroZshadow: They are the same data wise, but 1 is an id the GDK generates when you schedule a command to be send. And the internal ID is the ID the worker SDK returns when the command is actually send over the network.
Using 2 different types here ensures they aren't accidentally mixed up,

workers/unity/Packages/io.improbable.gdk.core/Commands/CommandRequestId.cs
```diff
@@ -0,0 +1,84 @@
+using System;
+
+namespace Improbable.Gdk.Core.Commands
+{
+    public readonly struct CommandRequestId : IEquatable<CommandRequestId>, IComparable<CommandRequestId>
+    {
+        public readonly long Raw;
```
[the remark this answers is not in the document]
zeroZshadow: For us no, but I wanted to keep this public in case someone used the command ID's for something.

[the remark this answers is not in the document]
jamiebrynes7: But then they can use this right? It has equality and comparison operators.

PR #1375, opened 2020-05-27
Worker Inspector window & Entity List

jamiebrynes7:
#### Description

This PR is the first in a series for the Worker Inspector window. In it, we find:

- The window definition itself
- A UI element for listing and selecting entities. 
- Searching over the list of entities by entity ID and entity type (metadata)
- Relatively efficient list rebuilding
	- Only rebuilds when entities have been added/removed.
	- With 10k SpatialOS entities in the Editor => ~2-3ms for a full rebuild
	- With 100k SpatialOS entities in the Editor => ~20-30ms for a full rebuild
	- (On a i7 7700k)
	- Possible optimizations in the future would be to use the [`EntityManagerDiffer`](https://docs.unity3d.com/Packages/com.unity.entities@0.11/api/Unity.Entities.EntityManagerDiffer.html) to track changes, but this would require the underlying data storage to be equipped for fast additions/removals. For now, I feel what we have is acceptable performance
	- Or alternatively jobify the list building.
- Some other misc changes (sortable `EntityId` structs, hacky `ViewVersion` for diff tracking)

A quick screenshot, the right hand panel will be used for displaying the detail for a given entity.

image

#### Tests

- [x] Some quick perf tests on my local machine, and for searching.
- [x] Unit tests for the searching.

#### Documentation

- I'll add a changelog once the rest is implemented

PR #1376, opened 2020-05-28
Add first class support for 'Editor-only' generated code

jamiebrynes7:
#### Description

In order to continue with #1375, I'll need to generate some code for rendering a given component. This code will be Editor only code. We currently have one 'Editor-only' generated assembly for the 'Build worker' menu items.

However, this was just sort of inserted into the existing generated code. As most generated code uses the namespace of the schema to derive its file path, this could cause collisions.

E.g.

```
package improbable.buildsystem;

...
```

Would generate code _inside_ the Build System 'Editor-only' assembly. 

This PR rectifies this and lays the groundwork for future 'Editor-only' generated code modules by:

- Adding a entirely separate directory for `Editor-only` generated code. (This can't go into source since otherwise this causes collisions). 
- Refactoring the `CodegenJob` classes slightly to allow for easy declaration as editor or non-editor. (It defaults to non-editor).

This is done when passing down arguments from your derived job class:

```csharp
public class MyCodegenJob : CodegenJob
{
	// Note the AsEditor() call.
	public MyCodegenJob(CodegenJobOptions opts, IFileSystem fs, DetailsStore store) : base (opts.AsEditor(), fs, store) {}
}
```

which will correctly set the base output directory path.

#### Tests

Verified that the build system generated code is placed in the correct place.

#### Documentation

- [x] Changelog
- [x] Upgrade guide (extra `.gitignore` entries)


test-project/.gitignore
```diff
@@ -12,4 +12,6 @@ Assets/Generated/Source.meta
 Assets/Generated/Source/**/*.*
 !Assets/Generated/Source/improbable/tests/**/*.cs
 !Assets/Generated/Source/improbable/testschema/**/*.cs
-!Assets/Generated/Source/improbable/dependentschema/**/*.cs
\ No newline at end of file
+!Assets/Generated/Source/improbable/dependentschema/**/*.cs
+Assets/Generated/Editor.meta
+Assets/Generated/Editor/**/*.*
```
paulbalaji, line 17: newline

workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfigurationWindow.cs
```diff
@@ -16,6 +16,7 @@ public class GdkToolsConfigurationWindow : EditorWindow
         internal const string VerboseLoggingLabel = "Verbose logging";
         internal const string CodegenLogOutputDirLabel = "Log output directory";
         internal const string CodegenOutputDirLabel = "C# output directory";
+        internal const string CodegenEditorOutputDirLabel = "C# Editor output directory";
```
paulbalaji, line 19: As a drive-by, could update the label to make it clearer that this is for generated code

workers/unity/Packages/io.improbable.gdk.buildsystem/.codegen/Source/WorkerGenerationJob.cs
```diff
@@ -18,8 +18,8 @@ public class WorkerGenerationJob : CodegenJob
         private readonly string relativeOutputPath = Path.Combine("improbable", "buildsystem");
         private readonly string relativeEditorPath = Path.Combine("improbable", "buildsystem", "Editor");
 
-        public WorkerGenerationJob(string baseOutputDir, IFileSystem fileSystem, DetailsStore detailsStore, bool force)
-            : base(baseOutputDir, fileSystem, detailsStore, force)
+        public WorkerGenerationJob(CodegenJobOptions options, IFileSystem fileSystem, DetailsStore detailsStore)
+            : base(options.AsEditor(), fileSystem, detailsStore)
```
paulbalaji, line 22: I quite like this way of using `.AsEditor` to signify it's for editor code

workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfigurationWindow.cs
jamiebrynes7: These are rendered under the "Code Generation" header in the settings, what did you have in mind? 

image

workers/unity/Packages/io.improbable.gdk.buildsystem/.codegen/Source/WorkerGenerationJob.cs
jamiebrynes7: It was either that or an attribute, and I didn't feel like writing more reflection code 😂

workers/unity/Packages/io.improbable.gdk.tools/GdkToolsConfigurationWindow.cs
paulbalaji: ah yes I didn't expand to see the heading 

ignore this

PR #1377, opened 2020-05-28
Enable Accelerator

paulbalaji:
#### Description

Provides `ACCELERATOR_ENDPOINT` argument to agents and provides the relevant arguments to use either the local/cloud Accelerators for mac/windows agents.

Depends on https://github.com/spatialos/gdk-for-unity-shared-ci/pull/72

#### Tests

Ran the pipeline. Multiple times.

#### Documentation

n/a

### DIFF
--- a/.buildkite/premerge.steps.yaml
+++ b/.buildkite/premerge.steps.yaml
@@ -30,7 +30,7 @@ windows: &windows
         # Workaround for flaky Git clones, likely due to - https://github.com/PowerShell/Win32-OpenSSH/issues/1322
       - exit_status: 128
         limit: 3
-  env: *windows-env
+  env: &windows-env
     ACCELERATOR_ENDPOINT: "unity-accelerator.production-intinf-eu1.i8e.io"
 
 linux: &linux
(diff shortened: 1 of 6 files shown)


.buildkite/premerge.steps.yaml
```diff
@@ -76,19 +76,25 @@ steps:
     <<: *windows
     artifact_paths:
       - logs/**/*
+    env:
+      ACCELERATOR_ENDPOINT: "unity-accelerator.production-intinf-eu1.i8e.io"
```
zeroZshadow, line 80: Can these be set on queue level? So we set it once per windows/mac ?

paulbalaji: We could probably set this in the `pre-command` hook, make use of `BUILDKITE_AGENT_META_DATA_PLATFORM` that the agent will know

jamiebrynes7: I feel like you can set the env for a given agent queue.

cc @petemounce ?

paulbalaji: Moved it up, going to test and see if this works

paulbalaji: You can set it, but it'll be wiped if you define env per step:

see the environment tab on the `build` steps here: https://buildkite.com/improbable/gdk-for-unity-premerge/builds/4019

PR #1382, opened 2020-06-03
Add worker details to worker inspector

jamiebrynes7:
#### Description

A small section for general worker information: 

- Worker Type
- Worker ID
- Worker flags

image

Sadly Unity doesn't have `inline-flex` as an option, so we can't have the worker flags section expand up to a given size with flex. For expediency's sake, this is just set to a fixed height percentage. The worker flags section will have a scrollbar if the contents are too large.

Unity also doesn't have the ability to define what happens for text overflows in an element, so very very long worker flags can overflow the container.

#### Tests

Ran it on a local deployment w/ some worker flags, and some without. Swapping worlds, etc.

#### Documentation

- Big changelog for this and the previous PRs


workers/unity/Packages/io.improbable.gdk.debug/WorkerInspector/WorkerDetail.cs
```diff
@@ -0,0 +1,112 @@
...
+            workerFlags.Refresh();
+        }
+
+        private static VisualElement MakeItem()
+        {
+            var container = new VisualElement();
+            container.AddToClassList("flag-element-container");
+
+            var key = new Label { name = FlagKeyName, };
+
```
zeroZshadow, line 85: Why all the whitespace here?

jamiebrynes7: The initializer was spanning multiple lines, but then I linted and it collapsed it. 🤷‍♂️

PR #1392, opened 2020-06-11
Add Option<T> rendering to the Worker Inspector

jamiebrynes7:
#### Description

This PR adds rendering support for `Option<T>` to the Worker Inspector. 

- Add `OptionVisualElement` and `NullableVisualElement` to render `Option<T>` and `T?` respectively. (We use `T?` except for when `T` is a reference type, i.e. - string and bytes).
- Refactor `FieldTypeHandler` a bit to support the above VisualElements. 

#### Tests

Eyeball test of the `ExhaustiveOptional` component. 

image

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.debug/.codegen/Source/Generators/DebugExtensions/FieldTypeHandler.cs
+++ b/workers/unity/Packages/io.improbable.gdk.debug/.codegen/Source/Generators/DebugExtensions/FieldTypeHandler.cs
@@ -31,8 +31,20 @@ namespace Improbable.Gdk.CodeGenerator
                     }
 
                     return $"private readonly {uiType} {fieldDetails.CamelCaseName}Field;";
+                case OptionFieldType optionFieldType:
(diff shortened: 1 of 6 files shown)


workers/unity/Packages/io.improbable.gdk.debug/.codegen/Source/Generators/DebugExtensions/FieldTypeHandler.cs
```diff
@@ -66,8 +78,35 @@ public IEnumerable<string> ToFieldInitialisation(UnityFieldDetails fieldDetails,
...
+                    }
+
+                    if (optionFieldType.ContainedType.Category == ValueType.Enum)
+                    {
+                        yield return $"{fieldDetails.CamelCaseName}InnerField.Init(default({optionFieldType.ContainedType.FqnType}));";
+                    }
+
+                    var element = optionFieldType.IsNullable ? "NullableVisualElement" : "OptionVisualElement";
+
+                    yield return
```
paulbalaji, line 104: beautiful

PR #1396, opened 2020-06-15
Add list support to Worker Inspector

jamiebrynes7:
#### Description

This PR introduces list support for the worker inspector by adding a `PaginatedListView<TElement, TData>` visual element for use in the worker inspector. This maintains a pool of `TElement` objects that it will reuse. By default, the number of elements per page is `5`, we can tweak this pretty easily. The `PaginatedListView` accepts two closures in its constructor:

- `Func<TElement> makeElement`. This is what the pool calls when it needs to create a new element.
- `Action<int, TData, TElement> bindElement`. This is what is called when the list view is updating an element. The `int` parameter is the absolute position of the element in the list (i.e. - not the position in the page). 

The pagination handles elements being removed from the list underneath you as well as additions. 

image

(See `CubeSpawner` component for the list)

#### Tests

- [x] Cube spawner in playground
- [x] Some unit tests for the list view around pagination and viewing a list slice.
- [x] Exhaustive list renderer.

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.debug/WorkerInspector/Codegen/PaginatedListView.cs
+++ b/workers/unity/Packages/io.improbable.gdk.debug/WorkerInspector/Codegen/PaginatedListView.cs
@@ -16,7 +16,7 @@ namespace Improbable.Gdk.Debug.WorkerInspector.Codegen
 
         private readonly Action<int, TData, TElement> bindElement;
         private readonly VisualElement container;
-        private readonly Pool elementPool;
+        private readonly ElementPool<TElement> elementPool;
         private readonly int elementsPerPage;
 
         private readonly VisualElement controlsContainer;
@@ -42,14 +42,12 @@ namespace Improbable.Gdk.Debug.WorkerInspector.Codegen
             pageCounter = this.Q<Label>(name: "page-counter");
 
             backButton = this.Q<Button>(name: "back-button");
-            backButton.text = "<";
             backButton.clickable.clicked += () => ChangePageCount(-1);
 
             forwardButton = this.Q<Button>(name: "forward-button");
-            forwardButton.text = ">";
             forwardButton.clickable.clicked += () => ChangePageCount(1);
 
-            elementPool = new Pool(makeElement);
+            elementPool = new ElementPool<TElement>(makeElement);
         }
 
         public void Update(List<TData> newData)
@@ -79,7 +77,7 @@ namespace Improbable.Gdk.Debug.WorkerInspector.Codegen
 
         private void CalculatePages()
         {
-            numPages = (int) Math.Ceiling((double) data.Count / elementsPerPage);
+            numPages = Mathf.CeilToInt((float) data.Count / elementsPerPage);
             numPages = Mathf.Clamp(numPages, 1, numPages);
             currentPage = Mathf.Clamp(currentPage, 0, numPages - 1);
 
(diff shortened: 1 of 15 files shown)


jamiebrynes7:
Also as a note, the elements in the list view are lazily created (rather than creating the max-per-page up front) as certain schema will cause a stack overflow. I.e. - 

```schema
type RecursiveList {
	list<RecursiveList> list = 1;
}
```

(This could happen with options, but we prevent recursive options for other reasons!)

workers/unity/Packages/io.improbable.gdk.debug/WorkerInspector/Codegen/PaginatedListView.cs
```diff
@@ -0,0 +1,151 @@
...
+            {
+                bindElement(firstIndex + j, data[firstIndex + j], (TElement) child);
+                j++;
+            }
+
+            backButton.SetEnabled(currentPage != 0);
+            forwardButton.SetEnabled(currentPage != numPages - 1);
+        }
+
+        private class Pool
```
zeroZshadow, line 130: This class should be moved out so that we generate less types

workers/unity/Packages/io.improbable.gdk.debug/WorkerInspector/Codegen/PaginatedListView.cs
```diff
@@ -0,0 +1,151 @@
...
+            template.CloneTree(this);
+
+            this.Q<Label>(name: "list-name").text = label;
+            container = this.Q<VisualElement>(className: "user-defined-type-container-data");
+
+            controlsContainer = this.Q<VisualElement>(className: "paginated-list-controls");
+            pageCounter = this.Q<Label>(name: "page-counter");
+
+            backButton = this.Q<Button>(name: "back-button");
+            backButton.text = "<";
```
zeroZshadow, line 45: This should probably be set in the UXML file

workers/unity/Packages/io.improbable.gdk.debug/WorkerInspector/Codegen/PaginatedListView.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            currentPage += diff;
+            currentPage = Mathf.Clamp(currentPage, 0, numPages - 1);
+            CalculatePages();
+            RefreshView();
+        }
+
+        private void CalculatePages()
+        {
+            numPages = (int) Math.Ceiling((double) data.Count / elementsPerPage);
```
zeroZshadow, line 82: Could be replaced with `Mathf.CeilToInt` usage, to not use both math libraries

workers/unity/Packages/io.improbable.gdk.debug/WorkerInspector/Codegen/PaginatedListView.cs
```diff
@@ -0,0 +1,151 @@
...
+            else if (diff < 0)
+            {
+                for (var i = diff; i < 0; i++)
+                {
+                    container.Add(elementPool.GetOrCreate());
+                }
+            }
+
+            // At this point, container.Children() has the same length as the slice.
+            var j = 0;
```
zeroZshadow, line 119: `var elementIndex= firstIndex;` and then increase that index, instead of using `j`

jamiebrynes7: Ah yeah this got missed in a refactor, used to be a `for` rather than `foreach`

PR #1409, opened 2020-06-29
UTY 2088: Set external ip from cli

BryanJY-Wong:
#### Description
Allow the setting of `UseExternalIp` flag from the cli. 
- Added new bool field for `UseExternalIp` in `ReceptionistFlow` and `LocatorFlow`
- Added appropriate name and defaults in the RuntimeConfig classes `RuntimeConfigNames` and `RuntimeConfigDefaults`
- Added unit test to check if the `UseExternalIp` is set if the argument exists and set to default if it does not

#### Tests
- Added and ran unit test in `CommandLineConnectionFlowInitializerTests`

#### Documentation
https://documentation.improbable.io/gdk-for-unity/v0.3.8/suggested-edits

#### Primary reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Added `map<k,v>` support to the Worker Inspector window. [#1403](https://github.com/spatialos/gdk-for-unity/pull/1403)
 - Added `Open inspector V2` menu item that opens the new inspector in the browser. [#1407](https://github.com/spatialos/gdk-for-unity/pull/1407)
 - Added the option to set `UseExternalIp` using the command line argument `+useExternalIp` with the `CommandLineConnectionParameterInitializer`. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
+- Added support for flag arguments in `CommanLineParser`. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Added `map<k,v>` support to the Worker Inspector window. [#1403](https://github.com/spatialos/gdk-for-unity/pull/1403)
 - Added `Open inspector V2` menu item that opens the new inspector in the browser. [#1407](https://github.com/spatialos/gdk-for-unity/pull/1407)
 - Added the option to set `UseExternalIp` using the command line argument `+useExternalIp` with the `CommandLineConnectionParameterInitializer`. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
-- Added support for flag arguments in `CommanLineParser`. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
+- Added support for flag arguments in `CommandLineParser`. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Added `Open inspector V2` menu item that opens the new inspector in the browser. [#1407](https://github.com/spatialos/gdk-for-unity/pull/1407)
 - Added the option to set `UseExternalIp` using the command line argument `+useExternalIp` with the `CommandLineConnectionParameterInitializer`. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
 - Added support for flag arguments in `CommandLineParser`. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
+- Added `Scripting Backend` option dropdown to the Build Configuration UI. [#1411](https://github.com/spatialos/gdk-for-unity/pull/1411)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Added `map<k,v>` support to the Worker Inspector window. [#1403](https://github.com/spatialos/gdk-for-unity/pull/1403)
 - Added `Open inspector V2` menu item that opens the new inspector in the browser. [#1407](https://github.com/spatialos/gdk-for-unity/pull/1407)
-- Added the option to set `UseExternalIp` using the command and argument. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
+- Added the option to set `UseExternalIp` using the command line argument `+useExternalIp` with the `CommandLineConnectionParameterInitializer`. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
 
 ### Changed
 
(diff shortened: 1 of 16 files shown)


workers/unity/Packages/io.improbable.gdk.core/Worker/ConnectionHandlers/ConnectionFlows.cs
```diff
@@ -55,6 +60,8 @@ public ReceptionistFlow(string workerId, IConnectionFlowInitializer<Receptionist
 
         public async Task<Connection> CreateAsync(ConnectionParameters parameters, CancellationToken? token = null)
         {
+            parameters.Network.UseExternalIp = UseExternalIp;
```
paulbalaji, line 63: Just noticed that if we set `UseExternalIp` outside of the connection flow, it gets overwritten by whatever we set as the default

e.g. in the editor if we set the `UseExternalIp` field on the `ClientConnector`, it gets overwritten when we call `CreateAsync`

BryanJY-Wong: Should I handle this only on the `Receptionist` or on the `Locator` as well? I could set the cli arg to take precedence; i.e.  if we set `UseExternalIp` on in the connection flow we'll use this value, if not just use whatever's supplied through the `ConnectionParameters`.

paulbalaji: We won't have to worry about what flow gets used if we move this into `CommandLineConnectionParameterInitializer`

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Added `map<k,v>` support to the Worker Inspector window. [#1403](https://github.com/spatialos/gdk-for-unity/pull/1403)
 - Added `Open inspector V2` menu item that opens the new inspector in the browser. [#1407](https://github.com/spatialos/gdk-for-unity/pull/1407)
+- Added the option to set `UseExternalIp` using the command and argument. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
```
jamiebrynes7, line 9:
```suggestion
- Added the option to set `UseExternalIp` using the command line argument `+useExternalIp` with the `CommandLineConnectionParameterInitializer`. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
```

paulbalaji, line 9:
```suggestion
- Added the option to set `UseExternalIp` using the `useExternalIp` command line flag. [#1409](https://github.com/spatialos/gdk-for-unity/pull/1409)
```

workers/unity/Packages/io.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -59,6 +59,11 @@ public bool TryGetCommandLineValue<T>(string key, ref T value)
             var desiredType = typeof(T);
             if (arguments.TryGetValue(key, out var strValue))
             {
+                if (strValue == string.Empty)
+                {
+                    throw new ArgumentException($"Cannot convert flag only argument, did you mean to call Contains?");
```
BryanJY-Wong, line 64: Not sure if this is the right exception 🤔

workers/unity/Packages/io.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -80,25 +85,28 @@ public bool TryGetCommandLineValue<T>(string key, ref T value)
...
+                if (isFlag(flag))
                 {
-                    if (i + 1 >= args.Count)
-                    {
-                        throw new ArgumentException(
-                            $"Flag \"{flag}\" requires an argument\nArguments: \"{string.Join(", ", args)}\"");
-                    }
-
-                    var flagArg = args[i + 1];
+                    var isNextArgFlagOrEnd = i + 1 >= args.Count || isFlag(args[i + 1]);
```
zeroZshadow, line 101: I think this can be simplified

```csharp
var strippedOfPlus = flag.Substring(1, flag.Length - 1);
if (i + 1 >= args.Count || isFlag(args[i + 1]))
{
    config[strippedOfPlus] = string.Empty;
}
else
{
    config[strippedOfPlus] = args[i + 1];
    i++;
}
```

Something like this

workers/unity/Packages/io.improbable.gdk.core/Tests/Editmode/Connection/CommandLineConnectionParametersInitializerTests.cs
```diff
@@ -30,7 +30,6 @@ public void Initialize_should_set_network_protocol(string protocolStr, NetworkCo
 
         [TestCase("ranket")]
         [TestCase("raknet")]
-        [TestCase("")]
```
BryanJY-Wong, line 33: Had to remove this. The initializer now assumes empty string is a result of parsing a flag argument and will throw `InvalidOperationException` when you try to use `TryGetCommandLineValue` on it.

workers/unity/Packages/io.improbable.gdk.core/Utility/CommandLineUtility.cs
```diff
@@ -59,6 +59,11 @@ public bool TryGetCommandLineValue<T>(string key, ref T value)
             var desiredType = typeof(T);
             if (arguments.TryGetValue(key, out var strValue))
             {
+                if (string.IsNullOrEmpty(strValue))
```
zeroZshadow, line 62: I think this will be hit when you do the following as well `+someArgument ""`.
I don't think there is a difference currently between "no" argument and an "empty" argument?

BryanJY-Wong: @zeroZshadow Should I allow empty argument to pass or throw a different exception for empty argument?

zeroZshadow: I think an empty string as a value is valid, so we'd need a different way to differentiate between a flag and an empty value. @jamiebrynes7 @paulbalaji @seanjparker Opinions?

BryanJY-Wong: I can just set it to `null` for flags.

jamiebrynes7: `null` for flags and `string.Empty` for truly empty arguments seems reasonable to me

workers/unity/Packages/io.improbable.gdk.core/Tests/Editmode/Utility/CommandLineUtilityTest.cs
```diff
@@ -66,5 +67,71 @@ public void ParseCommandLineArgs_should_process_args_when_string_array_is_provid
...
+            var stringVal = string.Empty;
+            Assert.Throws<InvalidOperationException>(() => args.TryGetCommandLineValue("flag", ref stringVal));
+        }
+
+        [Test]
+        public void TryGetCommandLineValue_is_valid_for_empty_arg()
+        {
+            var args = CommandLineArgs.From(new List<string>
+            {
+                "+key1",
```
zeroZshadow, line 126: This would be a better test by changing the keys around. An empty one first, and then one with a value.
This way you test that the parser doesn't skip the empty key

PR #1415, opened 2020-07-02
Merge testing/sonarcloud CI steps

paulbalaji:
#### Description

Merge testing/sonarcloud CI steps

#### Tests

Ran it through buildkite

#### Documentation

n/a

#### Primary reviewers


paulbalaji:
> Could we reduce the stdout spam from the scanner?

Already included in the PR, msbuild and sonar were both spamming stdout, so that's been redirected to a couple of new logfiles

jamiebrynes7:
> Already included in the PR since msbuild and sonar were both spamming stdout. So that's been redirected to a couple of new logfiles

Can we `tee` them instead? So we get both the log file and live stdout? 

@zeroZshadow this is the opposite side of the issue you were complaining about with Unity 😂

PR #1421, opened 2020-07-07
Remove Windows x86 support in build configuration

seanjparker:
For [UTY-2589](https://improbableio.atlassian.net/browse/UTY-2589)

#### Description

- Removed support for Windows x86 in the build configuration

#### Tests
Tested that all CI builds passed

#### Documentation

- Checked for references to x86/32-bit windows in docs
- Added Changelog entry


workers/unity/Packages/io.improbable.gdk.buildsystem/Util/CommandlineParser.cs
```diff
@@ -34,8 +34,6 @@ internal static List<BuildTarget> GetBuildTargetFilter(CommandLineArgs args)
                             return BuildTarget.Android;
                         case "ios":
                             return BuildTarget.iOS;
-                        case "winx86":
-                            return BuildTarget.StandaloneWindows;
```
austinkelmore, line 38: Thoughts on throwing an exception saying this is no longer supported and to use the "win" target instead of falling through to the general case saying it's an unknown build target?

paulbalaji: In this scenario, would we treat x86 specially forever or would it drop down to a standard "unknown target" after a few releases?

seanjparker: Yep, I think that would be better than the `unknown` error message

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - All generated C# enums will now start from 0, being shifted to schema values on serialization and shifted back to C# values on deserialization. [#1412](https://github.com/spatialos/gdk-for-unity/pull/1412)
     - A warning will be generated when enums defined in schema do not start from 0.
+- Removed Windows x86 support in the Cloud build configuration. [#1421](https://github.com/spatialos/gdk-for-unity/pull/1421)
```
jamiebrynes7, line 9:
```suggestion
- Removed Windows x86 build support. [#1421](https://github.com/spatialos/gdk-for-unity/pull/1421)
```

workers/unity/Packages/io.improbable.gdk.buildsystem/Util/CommandlineParser.cs
jamiebrynes7: > In this scenario, would we treat x86 specially forever or would it drop down to a standard "unknown target" after a few releases?

I imagine sometime in the future it can drop down to the 'unknown target' option. Maybe the next time we touch this code 😉

paulbalaji: SGTM

workers/unity/Packages/io.improbable.gdk.buildsystem/Configuration/BuildConfigEditor.cs
```diff
@@ -632,28 +633,56 @@ private GUIContent GetBuildTargetGuiContents(BuildTargetConfig c)
...
         private void DrawBuildTargets(BuildEnvironmentConfig env, int hash)
         {
+            // Build a mapping the build target position in the toolbar and it's position in the BuildTargets array.
+            // Sometimes there is a different of targets in the BuildTargets array and those in the toolbar as
+            // some are deprecated
+            if (toolbarMapping.Count == 0)
+            {
+                var index = 0;
+                var skipped = 0;
+                env.BuildTargets.ForEach(t =>
```
zeroZshadow, line 643: I think you can make this a normal for loop to make it a little clearer. `skipped` will become the for index.

workers/unity/Packages/io.improbable.gdk.buildsystem/Configuration/BuildConfigEditor.cs
```diff
@@ -632,28 +633,56 @@ private GUIContent GetBuildTargetGuiContents(BuildTargetConfig c)
...
+
             // Draw selected build target.
-            var buildTarget = env.BuildTargets[selectedBuildTarget.Index];
+            var buildTarget = env.BuildTargets[selectedTargetIndex];
             var canBuildTarget = BuildSupportChecker.CanBuildTarget(buildTarget.Target);
 
             var options = buildTarget.Options;
             var enabled = buildTarget.Enabled;
             var required = buildTarget.Required;
+            var deprecated = buildTarget.Deprecated;
```
zeroZshadow, line 685: Isn't this impossible to be `true` here?

seanjparker: Yes, I think this was a remnant of a previous attempt to get it working, so it can be removed safely

workers/unity/Packages/io.improbable.gdk.buildsystem/Configuration/BuildConfigEditor.cs
```diff
@@ -633,24 +633,19 @@ private GUIContent GetBuildTargetGuiContents(BuildTargetConfig c)
...
-                var skipped = 0;
-                env.BuildTargets.ForEach(t =>
+                var toolbarIndex = 0;
+                for (var buildTargetIndex = 0; buildTargetIndex < env.BuildTargets.Count; buildTargetIndex++)
                 {
-                    if (t.Deprecated)
+                    if (!env.BuildTargets[buildTargetIndex].Deprecated)
                     {
-                        skipped++;
+                        toolbarMapping.Add(toolbarIndex++, buildTargetIndex++);
```
zeroZshadow, line 646: I don't think you should be incrementing `buildTargetIndex` here?

seanjparker: Oops! Very true, I'll get that fixed

seanjparker:
Unfortunately, the deprecated build targets don't get removed from the `BuildConfiguration.asset` file. Instead, each target gets marked with the deprecated field (either 0 or 1). I tried a few ways where the it removes the deprecated targets in the asset but we couldn't get it to work, it would always seem to cause an error in the UI saying something about missing support for building the target.

In the end, the a solution that worked was to do the following:

- Add the `deprecated` field to each build target
- Filter out these targets from the `BuildContext` list, so the final builds aren't created
- Hide these targets in the UI, so we only see the non-deprecated targets

Everything else remains the same, all the targets are serialised and de-serialised by Unity into the `.asset` file, we just hide the deprecated targets where necessary.

jamiebrynes7:
Given there's not much churn in supported build targets, that seems like a good solution to me 👍

PR #1425, opened 2020-07-13
Prevent worker builds with compiler errors in Unity Editor

seanjparker:
For [UTY-2478](https://improbableio.atlassian.net/browse/UTY-2478)

#### Description
Prevents building the workers if there are compiler errors in the editor

#### Tests
Tested manually that dialog appeared with script errors that prevented building without confirmation

#### Documentation
Added entry into Changelog


workers/unity/Packages/io.improbable.gdk.buildsystem/Util/CompilerValidator.cs
```diff
@@ -0,0 +1,29 @@
+using System.Linq;
+using UnityEditor;
+using UnityEditor.Compilation;
+
+namespace Packages.io.improbable.gdk.buildsystem.Util
```
jamiebrynes7, line 5: This namespace should be `Improbable.Gdk.BuildSystem.Util`. Rider/VS will typically generate it based off the file path (which is why it looks as it does).

workers/unity/Packages/io.improbable.gdk.buildsystem/Util/CompilerValidator.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    [InitializeOnLoad]
+    public static class CompilerValidator
+    {
+        private static bool hasCompileErrors;
+        private static int compileErrorCount;
+
+        public static bool HasCompileErrors => hasCompileErrors;
+
+        public static int CompileErrorCount => compileErrorCount;
```
jamiebrynes7, line 15: You can use the auto-property syntax to simplify this: 

```csharp
public static bool HasCompileErrors { get; private set; }
public static int CompileErrorCount { get; private set; } 
```

workers/unity/Packages/io.improbable.gdk.buildsystem/Util/CompilerValidator.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        static CompilerValidator()
+        {
+            CompilationPipeline.assemblyCompilationFinished += ProcessCompileFinished;
```
jamiebrynes7, line 19: As this is a callback for a single assembly finishing compilation, could we have ordering problems that cause logic errors here? That is to say: 

1. Assemblies A, B, and C kick off compilation at the same time.
2. Assembly A finishes w/o errors
3. Assembly B finishes w/ errors.
4. Assembly C finishes w/o errors.

I think right now, `HasCompileErrors` would be false for this case. 

You may need to listen for [`compilationStarted`](https://docs.unity3d.com/ScriptReference/Compilation.CompilationPipeline-compilationStarted.html) and [`compilationFinished`](https://docs.unity3d.com/ScriptReference/Compilation.CompilationPipeline-compilationFinished.html) to merge assembly compile results and reset state.

workers/unity/Packages/io.improbable.gdk.buildsystem/WorkerBuilder.cs
```diff
@@ -68,8 +69,20 @@ internal static void MenuBuild(BuildEnvironment environment, params string[] wor
...
+
+                    if (!CompilerValidator.HasCompileErrors || overrideCompilerError)
+                    {
+                        var buildContexts = BuildContext.GetBuildContexts(workerTypes, environment);
+                        BuildWorkers(buildContexts);
+                    }
+                    else
+                    {
+                        throw new BuildFailedException($"Compiling failed with {CompilerValidator.CompileErrorCount} errors.");
+                    }
```
jamiebrynes7, line 85: I'd possibly invert this if-else block: 

```csharp
if (CompilerValidtor.HasCompileErrors && !overrideCompilerError)
{
	throw new BuildFailedException($"Cannot build worker as compile failed with {CompilerValidator.CompileErrorCount} errors.");
}

var buildContexts = BuildContext.GetBuildContexts(workerTypes, environment);
BuildWorkers(buildContexts);
```

There's also an argument that rather than throwing an exception, it should just return as the user has already been alerted. It may be a bit odd to get the first dialog, hit "No", then get a second dialog (in the `catch` block) telling you the build failed.

workers/unity/Packages/io.improbable.gdk.buildsystem/Util/CompilerValidator.cs
seanjparker: Oh nice, I didn't know you could use `private` in the auto-prop!

workers/unity/Packages/io.improbable.gdk.buildsystem/Util/CompilerValidator.cs
seanjparker: That's true. I'll need to rework how I check for the errors in the assemblies. I'll look into those two events 👍

workers/unity/Packages/io.improbable.gdk.buildsystem/WorkerBuilder.cs
seanjparker: Yeah, I would agree that it is a nicer UX to cancel the build when they Click "No" instead of seeing the standard "Build Failed" dialog

workers/unity/Packages/io.improbable.gdk.buildsystem/Util/BuildSupportChecker.cs
```diff
@@ -39,6 +39,11 @@ public static bool CanBuildHeadless(BuildTarget target)
             || target == BuildTarget.StandaloneLinux64;
     }
 
+    public static bool EditorHasCompileErrors()
+    {
+        return EditorUtility.scriptCompilationFailed;
```
jamiebrynes7, line 44: I'd possibly just inline this into `UnityWorkerMenuGenerator`

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Added
 
 - Added support for multiple annotations in Code Writer API. [#1427](https://github.com/spatialos/gdk-for-unity/pull/1427)
+- Added dialog to prevent building workers with Unity Editor compile errors. [#1425](https://github.com/spatialos/gdk-for-unity/pull/1425)
```
jamiebrynes7, line 8: Its no longer a dialog, but just disabled right?

seanjparker: Yes, I still need to update the Changelog

workers/unity/Packages/io.improbable.gdk.buildsystem/.codegen/Source/Generators/BuildSystem/UnityWorkerMenuGenerator.cs
```diff
@@ -85,6 +99,11 @@ public static CodeWriter Generate(List<string> workerTypes)
                             "WorkerBuilder.Clean();",
                             "Debug.Log(\"Clean completed\");"
                         });
+
+                        buildWorkerMenu.Method("private static bool CanMenuBuild()", () => new[]
+                        {
+                            "return !BuildSupportChecker.EditorHasCompileErrors();"
+                        });
```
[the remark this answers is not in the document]
seanjparker: Good idea! I've pushed a new commit with this method

workers/unity/Packages/io.improbable.gdk.buildsystem/.codegen/Source/Generators/BuildSystem/UnityWorkerMenuGenerator.cs
```diff
@@ -64,6 +75,7 @@ public static CodeWriter Generate(List<string> workerTypes)
                                 "MenuBuildCloud(AllWorkers);"
                             });
 
+
```
paulbalaji, line 78: extraneous newline

seanjparker: Fixed!

PR #1428, opened 2020-07-16
UTY-2504: Nullable command response

BryanJY-Wong:
#### Description
The `GetResponse<T>(CommandRequestId)` method in the `IDiffCommandResponseStorage` and `CommandSystem` now returns a `T?` instead of `MessageSpan<T>`

#### Tests
Simple validation in the playground's `PlayerCommandSystem`
1. log sent request 
2. log the checking for response until response is received
3. log received response
- More tests in [UTY-2508](https://improbableio.atlassian.net/browse/UTY-2508)

#### Documentation

- [x] **Did you remember a changelog entry?**

#### Primary reviewers


workers/unity/Assets/Playground/Scripts/Player/PlayerCommandsSystem.cs
```diff
@@ -39,6 +41,20 @@ protected override void OnCreate()
 
         protected override void OnUpdate()
         {
+            if (lastId.HasValue)
```
zeroZshadow, line 44: Do we want to add this to playground?
This might be rather spammy in a cloud build

BryanJY-Wong: It only logs when you left/right-click, but if its still too spammy I can remove it. @jamiebrynes7 @paulbalaji what do you think?

paulbalaji: It might be useful locally, but don't think it's worth keeping for cloud builds

workers/unity/Assets/Playground/Scripts/Player/PlayerCommandsSystem.cs
```diff
@@ -102,9 +106,10 @@ protected override void OnUpdate()
                         command == PlayerCommand.LaunchLarge ? LargeEnergy : SmallEnergy,
                         playerId
                     ));
-
+#if UNITY_EDITOR
                 lastId = commandSystem.SendCommand(request, entities[0]);
```
paulbalaji, line 110: we still want to send the command, it's only the logging we don't really care about

workers/unity/Assets/Playground/Scripts/Player/PlayerCommandsSystem.cs
```diff
@@ -86,8 +106,12 @@ protected override void OnUpdate()
                         command == PlayerCommand.LaunchLarge ? LargeEnergy : SmallEnergy,
                         playerId
                     ));
-
+#if UNITY_EDITOR
+                lastId = commandSystem.SendCommand(request, entities[0]);
+                Debug.Log($"Launching {lastId.Value.Raw}");
+#else
```
paulbalaji, line 112: Just checking if you tested this works? I _think_ this now means that it won't send commands in the editor

BryanJY-Wong: 😕 if editor send command, store id and log else send command

PR #1429, opened 2020-07-17
Fix bug where CommandSender would not reenable after being disabled

jamiebrynes7:
#### Description

If you have a behaviour that looks like:

```csharp
public class MyBehaviour : MonoBehaviour 
{
	[Require] private ComponentCommandSender sender;
	[Require] private OtherComponentCommandReceiver receiver;
}
```

If this behaviour is disabled due to `OtherComponent` losing authority, `receiver` will be set to `null` and the underlying object has `IsValid` set to `false`. 

If then the behaviour is re-enabled due to `OtherComponent` gaining authority, `receiver` will be re-injected, but `IsValid` will not be set to true. 

This is the same bug as https://github.com/spatialos/gdk-for-unity/pull/1326, but for the sender rather than the readers. The fix is to have everything implement `IRequireable`. This is a consequence of https://github.com/spatialos/gdk-for-unity/pull/1297 where we changed the behaviour of behaviours with multiple requireables. 

#### Tests

Test cases for reader/writer/commandsender/commandreceiver

#### Documentation

- [x] Changelog

### DIFF
--- a/workers/unity/Assets/Tests/EditmodeTests/Correctness/Subscriptions/RequireablesReenableTests.cs
+++ b/workers/unity/Assets/Tests/EditmodeTests/Correctness/Subscriptions/RequireablesReenableTests.cs
@@ -1,0 +1,117 @@
+using System;
+using System.Collections.Generic;
+using Improbable.Gdk.Core;
+using Improbable.Gdk.Core.Commands;
+using Improbable.Gdk.PlayerLifecycle;
+using Improbable.Gdk.Subscriptions;
+using Improbable.Gdk.Test;
+using Improbable.Gdk.TestUtils;
+using Improbable.Worker.CInterop;
+using NUnit.Framework;
+using UnityEngine;
+using UnityEngine.TestTools;
+using Empty = Improbable.Gdk.Core.Empty;
+using Entity = Unity.Entities.Entity;
+
+namespace Improbable.Gdk.EditmodeTests
+{
+    [TestFixture]
+    public class RequireablesReenableTests : MockBase
+    {
+        private const long EntityId = 100;
+
+        [Test]
+        public void CommandSenders_and_Receivers_are_valid_after_reinjection()
+        {
+            World
+                .Step(world => world.Connection.CreateEntity(EntityId, GetTemplate()))
+                .Step(world =>
+                    world.Connection.ChangeAuthority(EntityId, TestCommands.ComponentId, Authority.Authoritative))
+                .Step(world =>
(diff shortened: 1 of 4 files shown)


workers/unity/Assets/Tests/EditmodeTests/Correctness/Subscriptions/RequireablesReenableTests.cs
```diff
@@ -0,0 +1,117 @@
...
+            return template;
+        }
+
+        private class CommandSenderTestBehaviour : MonoBehaviour
+        {
+            [Require] public TestCommandsCommandReceiver CommandReceiver;
+            [Require] public PlayerHeartbeatClientCommandSender CommandSender;
+        }
+
+        private class ReaderTestBehaviour : MonoBehaviour
```
paulbalaji, line 111: I got a bit confused by the name of this, perhaps `ReaderWriterTestBehaviour` and similarly `CommandSenderReceiverTestBehaviour`?

jamiebrynes7: Sure thing

workers/unity/Assets/Tests/EditmodeTests/Correctness/Subscriptions/RequireablesReenableTests.cs
```diff
@@ -0,0 +1,121 @@
...
+                    Assert.IsNotNull(behaviour.CommandSender);
+                    Assert.IsTrue(behaviour.CommandReceiver.IsValid);
+                    Assert.IsTrue(behaviour.CommandSender.IsValid);
+                })
+                .Step(world =>
+                    world.Connection.ChangeAuthority(EntityId, TestCommands.ComponentId, Authority.NotAuthoritative))
+                .Step((world, behaviour) =>
+                {
+                    Assert.IsNull(behaviour.CommandReceiver);
+                    Assert.IsNull(behaviour.CommandSender);
```
zeroZshadow, line 42: Should this check if the captured command sender/receiver is invalidated?

jamiebrynes7: Those are captured in https://github.com/spatialos/gdk-for-unity/blob/develop/workers/unity/Assets/Tests/EditmodeTests/Correctness/Subscriptions/RequireablesDisableTests.cs :)

PR #1432, opened 2020-07-20
UTY-2659: set different entt list version

BryanJY-Wong:
#### Description
- Fixed a bug in the Worker Inspector where switching to a worker with no entities checked out does not refresh the entity list correctly

#### Tests
Set up client worker to not checkout anything and switch from game logic to client view in worker inspector.

#### Documentation
- [x] Changelog

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.debug/WorkerInspector/EntityList.cs
```diff
@@ -70,7 +70,7 @@ public void Update()
         public void SetWorld(World world)
         {
             entitySystem = world?.GetExistingSystem<EntitySystem>();
-            lastViewVersion = 0;
+            lastViewVersion = entitySystem?.ViewVersion - 1 ?? 0;
```
jamiebrynes7, line 73:
```suggestion
            lastViewVersion = -1;
```

Even simpler? 😂

jamiebrynes7:
Ah don't forget a changelog here 😄

PR #1437, opened 2020-07-22
UTY-2508: enable command tests

BryanJY-Wong:
#### Description
Created a `MockCommandSender` to inject into `CommandSystem` during construction of the `MockWorld`. This allows generating stub responses either singularly through a known `CommandRequestId` or generating all responses for all "sent" requests.

#### Tests
Created example use case using the `Launcher` component.

#### Documentation

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
```diff
@@ -0,0 +1,75 @@
...
+        private readonly Dictionary<Type, HashSet<long>> requestIds = new Dictionary<Type, HashSet<long>>();
+
+        private readonly Dictionary<long, ICommandRequest> outboundRequests = new Dictionary<long, ICommandRequest>();
+
+        internal MockCommandSender(MockWorld world)
+        {
+            this.world = world;
+        }
+
+        public void Setup<TRequest>(uint componentId)
```
BryanJY-Wong, line 26: I could probably get rid of this setup function using 
`componentId = (uint?) declaringType?.GetField("ComponentId", BindingFlags.Public | BindingFlags.Static)?.GetValue(null);`

workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
```diff
@@ -0,0 +1,80 @@
...
+        public void GenerateResponses<TRequest, TResponse>(Func<CommandRequestId, TRequest, TResponse> creator)
+            where TRequest : ICommandRequest
+            where TResponse : struct, IReceivedCommandResponse
+        {
+            var ids = requestIds[typeof(TRequest)];
+            var componentId = componentIds[typeof(TRequest)];
+            foreach (var id in ids)
+            {
+                var request = outboundRequests[id];
+                world.Connection.AddCommandResponse(creator(new CommandRequestId(id), (TRequest) request), componentId, (uint) id);
```
jamiebrynes7, line 62: I think you are passing in the wrong `id` here for the last parameter to `AddCommandResponse`. It should be `commandId` (which I believe is the schema command index of the command) rather than `requestId`.

workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            ids.Clear();
+        }
+
+        public void GenerateResponse<TRequest, TResponse>(CommandRequestId id, Func<CommandRequestId, TRequest, TResponse> creator)
+            where TRequest : ICommandRequest
+            where TResponse : struct, IReceivedCommandResponse
+        {
+            var componentId = componentIds[typeof(TRequest)];
+            var request = outboundRequests[id.Raw];
```
jamiebrynes7, line 74: This should be a `TryGetValue` with some error handling

workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
```diff
@@ -0,0 +1,80 @@
+using System;
+using System.Collections.Generic;
+using Improbable.Gdk.Core;
+using Improbable.Gdk.Core.Commands;
+using Improbable.Gdk.TestUtils;
+using Unity.Entities;
+
+namespace Packages.io.improbable.gdk.testutils
```
jamiebrynes7, line 8:
```suggestion
namespace Improbable.Gdk.TestUtils
```

workers/unity/Packages/io.improbable.gdk.testutils/MockWorld.cs
```diff
@@ -45,11 +44,38 @@ public static MockWorld Create(Options options)
...
+            CommandSender.GenerateResponses(creator);
+        }
+
+        public void GenerateResponse<TRequest, TResponse>(CommandRequestId id,
+            Func<CommandRequestId, TRequest, TResponse> creator)
+            where TRequest : ICommandRequest
+            where TResponse : struct, IReceivedCommandResponse
+        {
+            CommandSender.GenerateResponse(id, creator);
+        }
```
jamiebrynes7, line 77: I don't think these methods belong here. Similarly to the `MockConnectionHandler` where we don't add passthroughs here we should make the `MockCommandSender` public and allow users to access it as: `World.CommandSender.GenerateResponses(...);`.

workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
jamiebrynes7: (Similarly below)

workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
```diff
@@ -0,0 +1,87 @@
...
+        internal MockCommandSender(MockWorld world)
+        {
+            this.world = world;
+        }
+
+        public IEnumerable<(long, TRequest)> GetOutboundCommandRequests<TRequest>() where TRequest : ICommandRequest
+        {
+            return requestIds[typeof(TRequest)].Select(id => (id, (TRequest) outboundRequests[id]));
+        }
+
```
BryanJY-Wong, line 29: 🤔 Do you guys think this function is needed?

workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
```diff
@@ -0,0 +1,87 @@
...
+        internal MockCommandSender(MockWorld world)
+        {
+            this.world = world;
+        }
+
+        public IEnumerable<(CommandRequestId, TRequest)> GetOutboundCommandRequests<TRequest>() where TRequest : ICommandRequest
+        {
+            return requestIds[typeof(TRequest)].Select(id => (new CommandRequestId(id), (TRequest) outboundRequests[id]));
+        }
```
BryanJY-Wong, line 28: 🤔 Do you guys think this is needed?

workers/unity/Packages/io.improbable.gdk.testutils/MockWorld.cs
```diff
@@ -22,18 +19,18 @@ public struct Options
 
         public WorkerInWorld Worker { get; private set; }
         public MockConnectionHandler Connection { get; private set; }
-
         public EntityGameObjectLinker Linker { get; private set; }
 
         private readonly HashSet<GameObject> gameObjects = new HashSet<GameObject>();
+        public MockCommandSender CommandSender { get; private set; }
```
paulbalaji, line 25: Does this need to be public? Is there a way to only interact with the `CommandSystem`?

paulbalaji: I guess the point is why should we have to GenerateResponses in the test, instead of it being done for us in between steps?

BryanJY-Wong: My thinking was this allows you to simulate the responses arriving at non consecutive frames or even not arriving at all. At the same time I feel like this makes it _too_ different to the way you set up other things like component updates, events, etc.
Since everything else is done through `world.Connection.Addxxx()` whereas commands need to be called through `world.CommandSender.SendCommand(xxx)` or `world.GetSystem<CommandSystem>().SendCommand(xxx)`

paulbalaji: > allows you to simulate the responses arriving at non consecutive frames or even not arriving at all

Ah yes. That makes sense 👍

workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
paulbalaji: So this is all requests that haven't had a response yet? I think it could be useful 🤷‍♂️ 

but I'm trying to think a test case that would want it right now

jamiebrynes7: It may be that `IEnumerable<CommandRequestId>` would be sufficient for the use case of 'all requests that haven't had a response yet'. We can also choose to omit this and add it later.

workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public CommandRequestId SendCommand<TRequest>(TRequest request, Entity sendingEntity = default) where TRequest : ICommandRequest
```
jamiebrynes7, line 30: It may make sense to break this class into two as currently it does two different things. 

```csharp
public MockCommandSender : ICommandSender
{
	private OutgoingCommands outgoingCommands;
	public CommandRequestId SendCommand<TRequest>(...) { outgoingCommands.Add(...); }
}

public OutgoingCommands 
{
	public void GenerateResponses<TRequest, TResponse(...) {}
	public void GenerateResponse<TRequest, TResponse>(...) {}
}
```

And only giving public access to `OutgoingCommands` in the `MockWorld`. Currently this does two things and a user shouldn't have access to the `ICommandSender` directly as its implementation detail.

BryanJY-Wong: How about just putting the dictionaries and `GenerateResponse[s]` functions in the `MockConnectionHandler` so that it conforms to the existing API where you call `mockWorld.Connection.(do spatial stuff)` instead of having an additional object in `MockWorld`?

workers/unity/Assets/Tests/EditmodeTests/Correctness/Core/NullableResponseTests.cs
```diff
@@ -0,0 +1,73 @@
+using Improbable.Gdk.Core;
+using Improbable.Gdk.Core.Commands;
+using Improbable.Gdk.TestUtils;
+using Improbable.Worker.CInterop;
+using NUnit.Framework;
+using Playground;
+
+namespace Improbable.Gdk.EditmodeTests.Core
+{
+    public class NullableResponseTests : MockBase
```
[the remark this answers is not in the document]
BryanJY-Wong: Any ideas? All I got is `ReceivedResponseTests`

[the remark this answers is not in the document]
jamiebrynes7: `CommandReceivedResponseTests`?

workers/unity/Packages/io.improbable.gdk.testutils/MockCommandSender.cs
jamiebrynes7: Yeah that could work!

workers/unity/Assets/Tests/EditmodeTests/Correctness/Core/CommandReceivedResponseTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                world.Connection.CreateEntity(EntityId, GetTemplate());
+            }).Step(world =>
+            {
+                var id = world.GetSystem<CommandSystem>().SendCommand(GetRequest());
+                world.Connection.GenerateResponse<Launcher.LaunchEntity.Request, Launcher.LaunchEntity.ReceivedResponse>(id, ResponseGenerator);
+                return id;
+            }).Step((world, id) =>
+            {
+                Assert.IsTrue(world.GetSystem<CommandSystem>().GetResponse<Launcher.LaunchEntity.ReceivedResponse>(id).HasValue);
```
jamiebrynes7, line 40: It may be worth adding an `Assert` against the contents of the response to ensure its the one that we responded with rather than the API returning something else.

workers/unity/Assets/Tests/EditmodeTests/Correctness/Core/CommandReceivedResponseTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        [Test]
+        public void ReceivedResponse_isNotNull_when_Response_is_received()
+        {
+            World.Step(world =>
+            {
+                world.Connection.CreateEntity(EntityId, GetTemplate());
+            }).Step(world =>
+            {
+                var id = world.GetSystem<CommandSystem>().SendCommand(GetRequest());
+                world.Connection.GenerateResponse<Launcher.LaunchEntity.Request, Launcher.LaunchEntity.ReceivedResponse>(id, ResponseGenerator);
```
jamiebrynes7, line 36: I might move the `GenerateResponse` to its own 'Step()` as it closer matches what happens in reality. Right now we both send the command and receive the response in the same frame which can't happen in reality.

jamiebrynes7: A similar comment applies to the other test fixture

workers/unity/Packages/io.improbable.gdk.core/Dynamic/ComponentDatabase.cs
```diff
@@ -21,6 +25,12 @@ static ComponentDatabase()
 
             ComponentsToIds = Metaclasses.ToDictionary(pair => pair.Value.Data, pair => pair.Key);
             SnapshotsToIds = Metaclasses.ToDictionary(pair => pair.Value.Snapshot, pair => pair.Key);
+            RequestsToCommandMetaclass = Metaclasses.Where(pair => pair.Value.Commands.Length > 0)
+                .SelectMany(pair => pair.Value.Commands.Select(cmd => (cmd.Request, cmd)))
+                .ToDictionary(pair => pair.Request, pair => pair.cmd);
+            ReceivedRequestsToCommandMetaclass = Metaclasses.Where(pair => pair.Value.Commands.Length > 0)
+                .SelectMany(pair => pair.Value.Commands.Select(cmd => (cmd.ReceivedRequest, cmd)))
+                .ToDictionary(pair => pair.ReceivedRequest, pair => pair.cmd);
```
jamiebrynes7, line 33:
```suggestion
            RequestsToCommandMetaclass = Metaclasses
                .SelectMany(pair => pair.Value.Commands.Select(cmd => (cmd.Request, cmd)))
                .ToDictionary(pair => pair.Request, pair => pair.cmd);
            ReceivedRequestsToCommandMetaclass = Metaclasses
                .SelectMany(pair => pair.Value.Commands.Select(cmd => (cmd.ReceivedRequest, cmd)))
                .ToDictionary(pair => pair.ReceivedRequest, pair => pair.cmd);
```

The `Where` is redundant as the inner closure of `SelectMany` will return an empty enumerable for the case where `Commands.Length == 0`

workers/unity/Packages/io.improbable.gdk.core/Systems/CommandSystem.cs
```diff
@@ -49,13 +46,37 @@ public class CommandSystem : ComponentSystem
...
+        private class OutgoingCommandHandler : IOutgoingCommandHandler
+        {
+            public OutgoingCommandHandler(WorkerSystem worker)
+            {
+                Worker = worker;
+            }
+
+            private long nextRequestId = 1;
+
+            private WorkerSystem Worker { get; }
```
jamiebrynes7, line 67: Can we put these fields above the constructor?

workers/unity/Packages/io.improbable.gdk.core/Worker/ConnectionHandlers/MockConnectionHandler.cs
```diff
@@ -211,6 +290,29 @@ public RemoveEntityTemplateDynamicHandler(EntityTemplate template, long entityId
             }
         }
 
+        private class MockOutgoingCommandHandler : IOutgoingCommandHandler
+        {
+            private long nextRequestId = 1;
+
+            private readonly Action<long, Type, ICommandRequest> onSendCommand;
+
+            internal MockOutgoingCommandHandler(Action<long, Type, ICommandRequest> onSendCommand)
```
jamiebrynes7, line 299: Why does this need a closure rather than accepting the `MockConnectionHandler` concretely? There's now a level of indirection between them that isn't necessarily required.

workers/unity/Packages/io.improbable.gdk.core/Worker/ConnectionHandlers/MockConnectionHandler.cs
```diff
@@ -211,6 +290,29 @@ public RemoveEntityTemplateDynamicHandler(EntityTemplate template, long entityId
...
+
+            public CommandRequestId SendCommand<TRequest>(TRequest request, Entity sendingEntity = default) where TRequest : ICommandRequest
+            {
+                onSendCommand(nextRequestId, typeof(TRequest), request);
+                return new CommandRequestId(nextRequestId++);
+            }
+
+            public void SendResponse<T>(T response) where T : ICommandResponse
+            {
+                // Do nothing
```
jamiebrynes7, line 312: We should either implement this or have this throw a [`NotImplementedException`](https://docs.microsoft.com/en-us/dotnet/api/system.notimplementedexception?view=netcore-3.1)

BryanJY-Wong: If I throw the exception, it would require tests to have try-catch when they call `CommandSystem.SendResponse`

workers/unity/Assets/Tests/EditmodeTests/Correctness/Core/CommandReceivedResponseTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        private const long EntityId = 101;
+        private const long LaunchedId = 102;
+
+        [Test]
+        public void ReceivedResponse_isNull_when_Response_is_not_received()
+        {
+            World.Step(world => world.Connection.CreateEntity(EntityId, GetTemplate()))
+                .Step(world => world.GetSystem<CommandSystem>().SendCommand(GetRequest()))
+                .Step((world, id) => Assert.IsFalse(world.GetSystem<CommandSystem>().GetResponse<Launcher.LaunchEntity.ReceivedResponse>(id).HasValue));
```
paulbalaji, line 20: does this have to be all on one line?

(same with a few other tests)

paulbalaji: Generally why are some things one line and other things split across lines?

BryanJY-Wong: 😅 I just do it in one line if its just a single function call

PR #1393, opened 2020-06-11
Entity representation

zeroZshadow:
#### Description
Change entity representation to be defined through assets
image

#### Documentation

* [x] Upgrade Guide
* [x] Changelog

PR #1438, opened 2020-07-24
Fix the `PrefabPreprocessor`

jamiebrynes7:
#### Description

We broke this in by #1393 as we forgot to change the string when we re-named it. I've changed it to a `typeof()` to avoid this in the future.

#### Tests

- Removed a behaviour and re-added it. Hit play and checked the prefab to make sure it was disabled.

#### Documentation

- [x] Changelog

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.core/Representation/Editor/PrefabPreprocessor.cs
+++ b/workers/unity/Packages/io.improbable.gdk.core/Representation/Editor/PrefabPreprocessor.cs
@@ -44,7 +44,7 @@ namespace Improbable.Gdk.Core.Representation.Editor
 
         private static void PreprocessPrefabs()
         {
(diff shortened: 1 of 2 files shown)


workers/unity/Packages/io.improbable.gdk.core/Representation/Editor/PrefabPreprocessor.cs
```diff
@@ -44,7 +44,7 @@ private static void OnPlaymodeStateChanged(PlayModeStateChange playModeStateChan
 
         private static void PreprocessPrefabs()
         {
-            var resolvers = AssetDatabase.FindAssets("t:EntityLinkerDatabase")
+            var resolvers = AssetDatabase.FindAssets($"t:{typeof(EntityRepresentationMapping)}")
```
paulbalaji, line 47: nice

PR #1441, opened 2020-07-27
Fix Android support check

jamiebrynes7:
#### Description

This PR is two bug fixes combined in one! 

The first commit is a driveby where selecting the 'Cloud Build Options' and then the 'Local Build Options' and then selecting the `iOS` target would throw errors. TL;DR - We built a cache and forget to clear it when its invalidated.

The second is the fix for the associated ticket, if the editor pref isn't set, we look to ensure that the default location for the Android SDK installed by Unity is present. 

#### Tests

Put my build configuration in a configuration that would give me the missing Android SDK error (even though it was installed), then applied the code changes and checked that it correctly identifies it as installed. 

#### Documentation

- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


jamiebrynes7:
Woops realised the ticket reference in the branch name is wrong! Its actually UTY-2628

Issue #1225, opened 2019-11-25
Local launch errors on MacOS when project path contains spaces


paulbalaji:
**Affects:** Release v0.1.0 and up
**Internal Ticket:** [UTY-2380](https://improbableio.atlassian.net/browse/UTY-2380)

---

#### Description

If your spatial project path contains spaces, attempting a local launch will result in errors in the Unity Editor like below:
```
Errors occured - For more information, check the spatial local launch logfile: /Users/username/project path with spaces/logs/spatial_20191125T162432.log
UnityEngine.Debug:LogError(Object)
Improbable.Gdk.Tools.<>c__DisplayClass13_0:<LaunchLocalDeployment>b__0(Object, EventArgs) (at Library/PackageCache/io.improbable.gdk.tools@0.3.0/LocalLaunch.cs:280)
System.Threading._ThreadPoolWaitCallback:PerformWaitCallback()
```

#### Workaround

Remove all spaces from your project path.

jamiebrynes7:
Released a fix in [0.3.10](https://github.com/spatialos/gdk-for-unity/releases/tag/0.3.10)

PR #1442, opened 2020-07-29
Fix MacOS local launch errors when there are spaces in paths

jamiebrynes7:
#### Description
We use `osascript` for launching a local deployment with `spatial`. This has some particular rules around quotation, especially as we are running the script inline from a shell. 

This resulted in errors when paths had spaces in them. To handle these, I've used the built-in quoting abilities in `osascript` rather than trying to escape them manually twice (once for shell and once for osascript!). 

#### Tests

- Renamed `gdk-for-unity` to `gdk for unity` on my local machine to test launching a deployment & standalone client.

#### Documentation

- [x] Changelog
- [ ] Known issue - #1225

### DIFF
--- a/workers/unity/Packages/io.improbable.gdk.tools/LocalLaunch.cs
+++ b/workers/unity/Packages/io.improbable.gdk.tools/LocalLaunch.cs
@@ -221,7 +221,7 @@ namespace Improbable.Gdk.Tools
             var command = Common.SpatialBinary;
 
 #if UNITY_EDITOR_OSX
-            var commandArgs = $"local launch --enable_pre_run_check=false --snapshot '{toolsConfig.CustomSnapshotPath}' --experimental_runtime={toolsConfig.RuntimeVersion}";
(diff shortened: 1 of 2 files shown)


workers/unity/Packages/io.improbable.gdk.tools/LocalLaunch.cs
```diff
@@ -91,8 +91,9 @@ public static void LaunchClient()
             {
                 command = "osascript";
                 commandArgs = $@"-e 'tell application ""Terminal""
+                                     set projectDir to ""{Common.SpatialProjectRootDir}""
                                      activate
-                                     do script ""cd {Common.SpatialProjectRootDir} && {Common.SpatialBinary} {commandArgs}""
+                                     do script ""cd "" & quoted form of projectDir & "" && {Common.SpatialBinary} {commandArgs}""
```
paulbalaji, line 96: I find it almost magic how "quoted form of X" is valid osascript

jamiebrynes7: Yeah its pretty wild

PR #1443, opened 2020-07-30
UTY-2646: new dotnet version

BryanJY-Wong:
#### Description
Update queue with .NET Core SDK 3.1.302
- [x] Merge shared-ci and update ci/shared-ci.pinned

#### Tests
- [x] Run premerge

#### Documentation
- [x] If test pass, update docs

#### Primary reviewers


ci/sonar-scanner.sh
```diff
@@ -66,9 +66,11 @@ pushd "workers/unity"
     # Wildcards don't appear to play nice with this.
     args+=("-d:sonar.cs.opencover.reportsPaths=$(find ./logs -name "*.xml" | paste -sd "," -)")
 
+    dotnet new globaljson --sdk-version 2.2.402 --force
     traceStart "Execute sonar-scanner :sonarqube:"
         dotnet-sonarscanner begin "${args[@]}"
         dotnet msbuild ./unity.sln -t:Rebuild -nr:false | tee "${PROJECT_DIR}/logs/sonar-msbuild.log"
         dotnet-sonarscanner end "-d:sonar.login=${TOKEN}" | tee "${PROJECT_DIR}/logs/sonar-postprocess.log"
     traceEnd
+    rm -f global.json
```
zeroZshadow, line 75: This won't happen if something breaks in the scanner. This should be replaced with an exit trap to ensure the global.json is ALWAYS removed

workers/unity/.gitignore
```diff
@@ -23,3 +23,6 @@
 # Jetbrains
 /Assets/Plugins/Editor/Jetbrains
 /Assets/Plugins/Editor/Jetbrains.meta
+
+# Sonarcube dependency
+global.json
```
zeroZshadow, line 28: This shouldn't be in the gitignore, and the file won't ever survive a `test.sh` run

ci/sonar-scanner.sh
```diff
@@ -66,9 +66,11 @@ pushd "workers/unity"
     # Wildcards don't appear to play nice with this.
     args+=("-d:sonar.cs.opencover.reportsPaths=$(find ./logs -name "*.xml" | paste -sd "," -)")
 
+    dotnet new globaljson --sdk-version 2.2.402 --force
```
zeroZshadow, line 69: Which of the 3 dotnet runs below require 2.2?
If it's only the middle one, could the override not be added to that execution specifically?

ci/sonar-scanner.sh
jamiebrynes7: That's not a problem since if the scanner breaks, the job ends and the build dir gets wiped (including this file)

ci/sonar-scanner.sh
BryanJY-Wong: What do you mean? Put the `new globaljson` command after the first `dotnet-sonarscanner begin`?

zeroZshadow: I mean running `dotnet msbuild --runtime 2.2.402 ...` instead of the global override

BryanJY-Wong: I tried but couldn't find a way to do that. Everywhere I looked just referred to that json file :(

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
   - This no longer supports AuthorityImminentLoss.
 - ComponentUpdateSystem no longer has the API `GetAuthority`, `GetComponent`, and `HasComponent`. [#1364](https://github.com/spatialos/gdk-for-unity/pull/1364)
   - Use the Unity Entities `EntityManager` instead.
+- Tooling (codegen & deployment launcher) now depends on .NET Core v3.1.302 instead of v2.2.402. [#1443](https://github.com/spatialos/gdk-for-unity/pull/1443)
```
jamiebrynes7, line 11:
```suggestion
- The GDK now depends on .NET Core v3.1.3xx instead of v2.2.2xx. [#1443](https://github.com/spatialos/gdk-for-unity/pull/1443)
```

To align with the existing and incoming docs :)

UPGRADE_GUIDE.md
```diff
@@ -17,6 +17,10 @@ The `WorkerSystem` and `ComponentDatabase` can be used to convert from `entityId
 
 The `Reader` and `Writer` classes have had their `public Authority Authority` property changed to `public bool HasAuthority`. `HasAuthority` now indicates whether you have authority over the given component or not.
 
+### .NET Core Update
+
+Tooling for the GDK, in particular the code generator and deployment launcher, has been updated to use [.NET Core SDK v3.1.302](https://dotnet.microsoft.com/download/dotnet-core/3.1). To update, download and install the new version from the link provided.
```
jamiebrynes7, line 22:
```suggestion
The GDK, in particular the code generator and deployment launcher, has been updated to use [.NET Core SDK v3.1.3xx](https://dotnet.microsoft.com/download/dotnet-core/3.1). 

To update, download and install the new version from the link provided above.
```

workers/unity/.gitignore
zeroZshadow: Why was this resolved?

BryanJY-Wong: I think I clicked it by mistake when trying to resolve one of the other 2, anyway, I fixed it now

PR #1444, opened 2020-07-30
Native Event Tracing IO Wrappers

seanjparker:
For [UTY-2685](https://improbableio.atlassian.net/browse/UTY-2685)

#### Description
 - Added native C# wrappers for IO operations with the Worker SDK

#### Tests
 - Manually tested that native event wrappers worked with the current C-C# bindings

#### Documentation

- [x] Changelog


workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,120 @@
...
+                    // Check that the number of bytes we tried to write was the number of bytes written
+                    throw new IOException("Failed to write all data to stream.");
+                }
+            }
+        }
+
+        public byte[] Read(uint bytesToRead)
+        {
+            var streamData = new byte[bytesToRead];
+            var dataHandle = GCHandle.Alloc(streamData, GCHandleType.Pinned);
```
zeroZshadow, line 67: This could use `fixed` instead of a GCHandle. This also avoid the memory leak this currently has

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,120 @@
...
+                throw new IOException($"Failed to read {bytesToRead} bytes, only read {bytesRead}.");
+            }
+
+            return dataHandle.Target as byte[];
+        }
+
+        public byte[] Peek(uint bytes)
+        {
+            var streamData = new byte[bytes];
+            var dataHandle = GCHandle.Alloc(streamData, GCHandleType.Pinned);
```
zeroZshadow, line 87: This could use `fixed` instead of a GCHandle. This also avoid the memory leak this currently has

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    // Check that the number of bytes we tried to write was the number of bytes written
+                    throw new IOException("Failed to write all data to stream.");
+                }
+            }
+        }
+
+        public byte[] Read(uint bytesToRead)
+        {
+            var streamData = new byte[bytesToRead];
```
zeroZshadow, line 66: Perhaps make an additional version of this function that lets you provide an array to re-use as well?

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,120 @@
...
+                var rawError = CIO.StreamGetLastError(stream);
+                throw new IOException(ApiInterop.FromUtf8Cstr(rawError));
+            }
+
+            if (bytesRead != bytesToRead)
+            {
+                throw new IOException($"Failed to read {bytesToRead} bytes, only read {bytesRead}.");
+            }
+
+            return dataHandle.Target as byte[];
```
zeroZshadow, line 81: This can just return `streamData`

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        public void Write(byte[] data)
+        {
+            var remainingCapacity = CIO.StreamGetRemainingWriteCapacityBytes(stream);
+            if (remainingCapacity < data.Length)
+            {
+                throw new IOException("Not enough stream capacity to write data.");
+            }
+
+            fixed (byte* dataToWrite = data)
```
zeroZshadow, line 45: This fixed scope only needs to wrap the `StreamWrite` call.

workers/unity/Packages/io.improbable.worker.sdk/IOStorage.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            storage = CIO.StorageCreate();
+        }
+
+        /// <inheritdoc cref="IDisposable"/>
+        public void Dispose()
+        {
+            storage.Dispose();
+            GC.SuppressFinalize(this);
+        }
```
jamiebrynes7, line 20: Couple comments on the implementation of `Dispose()` here and in `IOStream`:

- We are calling `GC.SuppressFinalize(this);`, but we don't have a destructor defined (`~IOStorage()`). I think you may be getting a compiler warning about this.. if you mark this class as `sealed` it will fix that warning as well. 
- If a user called `Dispose()` twice, we would get double-free errors. We should track if this object has already been disposed and no-op the second (or subsequent) times after that.

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,120 @@
...
+            stream.Dispose();
+            GC.SuppressFinalize(this);
+        }
+
+        public static IOStream CreateRingBufferStream(uint capacity)
+        {
+            return new IOStream(CIO.CreateRingBufferStream(capacity));
+        }
+
+        public static IOStream CreateFileStream(string fileName)
```
jamiebrynes7, line 29: This should probably take the 'OpenMode' as a parameter to avoid future breaking changes. 

You may need to redefine the enum here to avoid leaking the internal namespace. (I'd see what the Worker SDK currently does in these kinds of cases, like `Authority`)

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,120 @@
...
+                    // The Stream write failed, check the error message
+                    var rawError = CIO.StreamGetLastError(stream);
+                    throw new IOException(ApiInterop.FromUtf8Cstr(rawError));
+                }
+
+                if (bytesWritten != data.Length)
+                {
+                    // Check that the number of bytes we tried to write was the number of bytes written
+                    throw new IOException("Failed to write all data to stream.");
+                }
```
jamiebrynes7, line 60: I think we should return the `bytesWritten` to the caller so they can decide what to do in this case. Right now we throw an exception, but they can't necessarily recover from it since they don't know how many bytes were written.

workers/unity/Packages/io.improbable.worker.sdk/IOStorage.cs
seanjparker: > We should track if this object has already been disposed and no-op the second (or subsequent) times after that.

The worker SDK doesn't seem to do this as far as I can see. If they have already called dispose on the object, won't that mean any state in the class could have been freed already?

jamiebrynes7: Hmmm maybe its handled inside the `CPtrHandle` class which the `CIO.StreamHandle` inherits from.

seanjparker: It doesn't look like it's handled in `CptrHandle`. The only place I see something like this is when the `WrappedGcHandle` is used and the `IsAllocated` flag is checked before the handle is disposed.

seanjparker: Maybe they just trust that anyone who inherits from this class won't try and double-free the memory?

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
seanjparker: Yep, very true. I've switch to using a `fixed` block for this case and the other places in this class

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
seanjparker: Yeah, that would be a good addition. I'm which way will be more useful yet though. Once I start with the Event tracing wrappers I'll find out which one is more useful 😄

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
seanjparker: Yep, fixed 👍

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
seanjparker: Yeah, good point. I've updated some of the other functions to have a similar behaviour as this too

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -26,95 +32,104 @@ public static IOStream CreateRingBufferStream(uint capacity)
             return new IOStream(CIO.CreateRingBufferStream(capacity));
         }
 
-        public static IOStream CreateFileStream(string fileName)
+        public static IOStream CreateFileStream(string fileName, OpenMode openMode)
         {
             fixed (byte* fileNameBytes = ApiInterop.ToUtf8Cstr(fileName))
             {
-                return new IOStream(CIO.CreateFileStream(fileNameBytes, CIO.OpenMode.OpenModeDefault));
+                return new IOStream(CIO.CreateFileStream(fileNameBytes, (CIO.OpenMode) openMode));
```
zeroZshadow, line 39: I'm not sure if this kind of cast is actually valid. the C# Interop has some fancy conversion class for this reason.
Might want to double check

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -26,95 +32,104 @@ public static IOStream CreateRingBufferStream(uint capacity)
...
+            {
+                return bytesWritten;
+            }
+
+            var rawError = CIO.StreamGetLastError(stream);
+            throw new IOException(ApiInterop.FromUtf8Cstr(rawError));
         }
 
-        public byte[] Read(uint bytesToRead)
+        public (byte[], long) Read(uint bytesToRead)
```
zeroZshadow, line 66: Not sure if we want to use a tuple, or an `out` var here.
@jamiebrynes7 ?

jamiebrynes7: Probably an `out` parameter like: `public long Read(uint bytesToRead, out bytes[] bytes)` or similar

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
jamiebrynes7: It would be a compiler error if it was an invalid cast I think (the `as` casts fail at runtime)

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,133 @@
...
+                return new IOStream(CIO.CreateFileStream(fileNameBytes, (CIO.OpenMode) openMode));
+            }
+        }
+
+        public long Write(byte[] data)
+        {
+            var remainingCapacity = CIO.StreamGetRemainingWriteCapacityBytes(stream);
+            if (remainingCapacity < data.Length)
+            {
+                throw new IOException("Not enough stream capacity to write data.");
```
jamiebrynes7, line 46: It may be worth have a different exception for this case as opposed to the general error case. It means users can write logic to specifically handle it.

It seems that [`NotSupportedException`](https://docs.microsoft.com/en-us/dotnet/api/system.notsupportedexception?view=netcore-3.1) may be a good candidate as its used elsewhere in the C# standard library for 'stream full' errors (e.g. - in [`Memory Stream`](https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream.write?view=netcore-3.1#System_IO_MemoryStream_Write_System_Byte___System_Int32_System_Int32_))

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            if (bytesPeeked != -1)
+            {
+                return streamData;
+            }
+
+            var rawError = CIO.StreamGetLastError(stream);
+            throw new IOException(ApiInterop.FromUtf8Cstr(rawError));
+        }
```
jamiebrynes7, line 118: Should we make this `public uint Peek(uint bytes, out byte[] bytes)` to match the `Read` method?

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        public void Ignore(uint bytesToIgnore)
```
jamiebrynes7, line 120: This should probably return `bytesIgnored`

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
seanjparker: The `Read` method has two implementations, one where the caller provides the `bytes` array which is written to, and the other takes a the number of bytes to read and creates the array for them.

Should I add another implementation of `Peek` so that it matches `Read`?

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
seanjparker: Yep, good point

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -0,0 +1,156 @@
...
+            var rawError = CIO.StreamGetLastError(stream);
+            throw new IOException(ApiInterop.FromUtf8Cstr(rawError));
+        }
+
+        public uint GetRemainingCapacity()
+        {
+            return CIO.StreamGetRemainingWriteCapacityBytes(stream);
+        }
+
+        private void CheckIfStreamClosed()
```
jamiebrynes7, line 148: Nit: Prefer `ThrowIfStreamClosed` over `CheckIfStreamClosed`, I would assume `CheckIfStreamClosed` would return me a result rather than throwing an exception.

PR #1448, opened 2020-08-04
UTY-2533: add commands tests for subscription

BryanJY-Wong:
#### Description
Add some more command tests for subscription system. This was part of UTY-2533 (closed ticket).

#### Tests
This is a test.

#### Documentation
Do I need a Changelog entry?

#### Primary reviewers


workers/unity/Assets/Tests/EditmodeTests/Correctness/Subscriptions/CommandTests.cs
```diff
@@ -57,8 +109,8 @@ private static TestCommands.Test.ReceivedResponse ResponseGenerator(CommandReque
         private static EntityTemplate GetTemplate()
         {
             var template = new EntityTemplate();
-            template.AddComponent(new Position.Snapshot(), "worker");
-            template.AddComponent(new TestCommands.Snapshot(), "worker");
+            template.AddComponent(new Position.Snapshot(), WorkerID);
+            template.AddComponent(new TestCommands.Snapshot(), WorkerID);
```
jamiebrynes7, line 113: There should be an overload w/o the string that you can use (since the write access isn't used at all in the tests).

workers/unity/Assets/Tests/EditmodeTests/Correctness/Subscriptions/CommandTests.cs
```diff
@@ -40,6 +42,56 @@ public void SubscriptionSystem_invokes_callback_on_receiving_response()
...
+                    return world.CreateGameObject<CommandStub>(EntityId).Item2;
+                })
+                .Step((world, mono) =>
+                {
+                    mono.Sender.SendTestCommand(GetRequest(), response => throw new AssertionException("Don't call"));
+                })
+                .Step((world, mono) =>
+                {
+                    world.Connection.GenerateResponses<TestCommands.Test.Request, TestCommands.Test.ReceivedResponse>(ResponseGenerator);
+                    world.Linker.UnlinkGameObjectFromEntity(new EntityId(EntityId), mono.gameObject);
```
jamiebrynes7, line 87: I'd possibly split these into two `Step`s with the unlink happening first.

workers/unity/Assets/Tests/EditmodeTests/Correctness/Subscriptions/CommandTests.cs
```diff
@@ -40,6 +42,56 @@ public void SubscriptionSystem_invokes_callback_on_receiving_response()
...
+                })
+                .Step((world, sender) =>
+                {
+                    sender.SendTestCommand(GetRequest(), response => throw new AssertionException("Don't call"));
+                })
+                .Step((world, sender) =>
+                {
+                    world.Connection.GenerateResponses<TestCommands.Test.Request, TestCommands.Test.ReceivedResponse>(
+                        ResponseGenerator);
+                    sender.IsValid = false;
```
jamiebrynes7, line 64: Hmm this shouldn't be allowed from the user API, cc @zeroZshadow @paulbalaji thoughts?

paulbalaji: I think I'd agree. Should there be a way to directly invalidate command senders, or should this test be indirect by having another reader/writer on the script that we invalidate

BryanJY-Wong: Actually this may not be needed right? Since the `IsValid` bool is only when the user captures the command sender.

jamiebrynes7: My point was more that a user shouldn't be able to set `IsValid` since we should be controlling the lifecycle of those objects.

zeroZshadow: Probably cause its internal and internals are exposed to the test assemblies

jamiebrynes7: Its public set! https://github.com/spatialos/gdk-for-unity/blob/master/workers/unity/Packages/io.improbable.gdk.core/Subscriptions/Reader.cs#L14

jamiebrynes7: Raised UTY-2712 to fix this

PR #1451, opened 2020-08-10
Removed all usages and support for Authority.AuthorityLossImminent

zeroZshadow:
#### Description
Remove AuthorityLossImminent API's

#### Documentation
* [x] Changelog
* [x] Upgrade guide


UPGRADE_GUIDE.md
```diff
@@ -2,6 +2,13 @@
 
 ## From `0.3.9` to `0.3.10`
 
+### AuthorityLossImminent support removed
+
+The GDK for Unity no longer supports the AuthorityLossImminent authority state. If this feature is currently enabled, it will not trigger the authority state for any reader or writer. The methods to `Ack` the authority state has also been removed.
```
jamiebrynes7, line 7:
```suggestion
The GDK for Unity no longer supports the `AuthorityLossImminent` authority state. If this feature is currently enabled, it will not trigger the authority state for any reader or writer. The methods to `Ack` the authority state has also been removed.
```

UPGRADE_GUIDE.md
```diff
@@ -2,6 +2,13 @@
...
+
+If you currently make use of this feature, we recommend using the `MonoBehaviour.OnDisabled` method instead.
```
jamiebrynes7, line 9: This probably isn't great advice for some use cases.

For example, if you have something which tries to sync some local state when loss is imminent, by the time this is called the authority is already gone.

zeroZshadow: Any suggestions? Because I can't really think of any

zeroZshadow: As discussed offline, removing the "advice"

PR #1440, opened 2020-07-24
Event Tracing API C# Bindings

seanjparker:
For [UTY-2671](https://improbableio.atlassian.net/browse/UTY-2671)

#### Description
Initial work for the implementation of the Event Tracing API, introduces some internal C# bindings for the C API

#### Tests
How did you test these changes prior to submitting this pull request?
What automated tests are included in this PR?
WIP

#### Documentation
How is this documented (for example: release note, upgrade guide, feature page, in-code documentation)?
WIP

- [x] Changelog

PR #1452, opened 2020-08-10
Event Tracing API wrapper

seanjparker:
For [UTY-2686](https://improbableio.atlassian.net/browse/UTY-2686)

#### Description
- The PR adds wrappers around the C-C# binding layer that was introduced in #1440 

#### Tests
- Manual testing
  - [x] Span
  - [x] Event
  - [x] Event Data
  - [x] Event Tracer
  - [x] Item

#### Documentation
Nothing yet

- [x] Changelog


workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        public static unsafe void ConvertEventTracer(EventTracerParameters[] parameters, out CEventTrace.EventTracerParameters[] internalParameters)
+        {
+            List<WrappedGcHandle> handles = new List<WrappedGcHandle>(
+                parameters.Length);
+
+            // ConvertTracerParameters(handles, ...)
+            internalParameters = new CEventTrace.EventTracerParameters[parameters.Length];
+        }
```
seanjparker, line 76: Still need to figure out the conversion for Event-tracers

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,349 @@
...
+
+        public void AddCollectionOfFields(Dictionary<string, string> fields)
+        {
+            foreach (var entry in fields)
+            {
+                AddField(entry.Key, entry.Value);
+            }
+        }
+
+        public void AddField(string key, string value)
```
jamiebrynes7: Discussed a way to simplify this (and `AddCollectionOfFields` above) offline

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,349 @@
...
+    public unsafe class TraceEventData : IDisposable
+    {
+        internal GCHandle eventData;
+
+        private readonly GcHandlePool fieldHandles;
+
+        public TraceEventData()
+        {
+            var internalEventData = CEventTrace.EventDataCreate();
+            eventData = GCHandle.Alloc(internalEventData, GCHandleType.Pinned);
```
jamiebrynes7, line 144: Why do we need to pin the `internalEventData` field? The location of the wrapper class in the heap doesn't matter as we care about the pointer to the native object inside of it

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,349 @@
...
+            fixed (byte* nativeKey = ApiInterop.ToUtf8Cstr(key))
+            {
+                value = CEventTrace.EventDataGetFieldValue(eventData.AddrOfPinnedObject(), nativeKey);
+            }
+
+            return ApiInterop.FromUtf8Cstr(value);
+        }
+    }
+
+    public unsafe class TraceItem
```
jamiebrynes7, line 200: It seems that most of these methods belong on the `Item` struct that then take in `IOStorage/IOStream` as required.

For example:

```csharp
public struct Item 
{
	// Field declarations
	
	public void AddToStorage(IOStorage storage);
	public void SerializeToStream(IOStream stream);
	public long GetSerializedSize(); // Does this need to be public? 

	public static Item DeserializeFromStream(IOStream stream);
}
```

This feels more natural than having a secondary class which takes in both a storage and stream and maybe will use them depending on what you do. Unless I'm misunderstanding what this class is for?

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    public unsafe delegate void TraceCallback(
+        UIntPtr userData, IntPtr item);
+
+    public class EventTracerParameters
+    {
+        public TraceCallback TraceCallback;
+        public UIntPtr UserData;
+    }
```
jamiebrynes7, line 338: These should use non-pointer types: 

```csharp
public delegate void TraceCallback(object userData, Item item);

public class EventTracerParameters 
{
	public TraceCallback Callback;
	public object UserData
}
```

and we should handle the interop in the library itself (so users don't need to worry about it).

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,140 @@
...
+            ApiInterop.Memcpy(internalSpanId.Data, spanId.Data, (UIntPtr) 16);
+            return internalSpanId;
+        }
+
+        public static unsafe CEventTrace.Event ConvertEvent(Event eventToConvert)
+        {
+            var internalEvent = new CEventTrace.Event();
+            internalEvent.UnixTimestampMillis = eventToConvert.UnixTimestampMillis;
+            internalEvent.Id = ConvertSpanId(eventToConvert.Id);
+            internalEvent.Data = eventToConvert.Data?.eventData.AddrOfPinnedObject() ?? IntPtr.Zero;
```
jamiebrynes7, line 57: Same comment here, don't we want the handle underlying `eventData` rather than a pointer to the wrapper class?

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            fixed (byte* eventType = ApiInterop.ToUtf8Cstr(eventToConvert.Type))
+            fixed (byte* eventMessage = ApiInterop.ToUtf8Cstr(eventToConvert.Message))
+            {
+                internalEvent.Type = eventType;
+                internalEvent.Message = eventMessage;
+            }
```
jamiebrynes7, line 64: This isn't safe as once we leave the `fixed` block, the pointers `eventType` and `eventMessage` may or may not be valid anymore. 

We will need to pin these using the handles and return the handles to the caller. 

Alternatively, have a look at `ConvertPlayerIdentityTokenRequest` or `ConvertLoginTokensRequest` in the Worker SDK for an different pattern where the caller provides a closure which will use the `internalEvent` while these pointers are still valid.

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,140 @@
...
+        public static void FreeTracerParameterHandles(IEnumerable<WrappedGcHandle> tracerParameterHandles)
+        {
+            foreach (var tracerParameterHandle in tracerParameterHandles)
+            {
+                tracerParameterHandle.Dispose();
+            }
+        }
+
+        private static void ConvertTracerParameters(List<WrappedGcHandle> handles,
+            Dictionary<uint, EventTracerParameters> parameterDict, EventTracerParameters defaultParameters,
```
jamiebrynes7, line 99: What is the significance of the `uint` as the key?

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        public static bool SpanIdEqual(SpanId spanId1, SpanId spanId2)
+        {
+            return CEventTrace.SpanIdEqual(
+                ParameterConversion.ConvertSpanId(spanId1),
+                ParameterConversion.ConvertSpanId(spanId2)) > 0;
+        }
+
+        public static unsafe SpanId GetNullSpanId()
```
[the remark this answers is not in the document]
seanjparker: The function needed to be marked as `unsafe` since it uses the `Data` property which is a pointer to the array. Once I moved it inside the `SpanId` struct that is marked as `unsafe`, I could remove the `unsafe` modifier from the function.

If I wanted to remove `unsafe` from the SpanId struct, would I need to introduce a wrapper around the struct?

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,349 @@
...
+            var nativeSpanId = CEventTrace.EventTracerGetActiveSpanId(eventTracer);
+            var spanId = new SpanId();
+            ApiInterop.Memcpy(spanId.Data, nativeSpanId.Data, (UIntPtr) 16);
+            return spanId;
+        }
+
+        public void AddSpanToCauses(int causeCount, SpanId[] causes)
+        {
+            var createdSpan = CEventTrace.EventTracerAddSpan(eventTracer, null, (uint) causeCount);
+        }
```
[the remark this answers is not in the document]
seanjparker: Oops, you're right. I still need to finish that one off 🤦‍♂️

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
seanjparker: That function was taken from the Worker SDK, I don't think I need the dictionary, they included this since they used the ComponentID to get the relevant object from the dictionary iirc. I think I can change this to just use a List instead 🙂

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,349 @@
...
+    public enum ItemType
+    {
+        Span = 1,
+        Event = 2
+    }
+
+    public struct Item
+    {
+        public ItemType ItemType;
+        public ITraceItem TraceItem;
```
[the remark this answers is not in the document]
seanjparker: Oh right, I think you mentioned this previously when we talked about the design. I think I followed how the Worker SDK implemented the idea of different types of `IConstraint`

I've switched to the nullable properties and it ended up looking cleaner 👍

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,349 @@
...
+        {
+            return CEventTrace.SpanIdHash(ParameterConversion.ConvertSpanId(spanId));
+        }
+
+        public static bool SpanIdEqual(SpanId spanId1, SpanId spanId2)
+        {
+            return CEventTrace.SpanIdEqual(
+                ParameterConversion.ConvertSpanId(spanId1),
+                ParameterConversion.ConvertSpanId(spanId2)) > 0;
+        }
```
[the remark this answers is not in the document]
seanjparker: Good point!

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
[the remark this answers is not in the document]
jamiebrynes7: You can also have a standalone `unsafe { }` block inside a safe function like: 

```csharp
public static SpanId GetNullSpanId()
{
	unsafe 
	{
		// Do unsafe things in here. By doing so, we are saying that we've ensured the invariants
		// of the unsafe code are met and calling this function is now 'safe'. 
	} 
}
```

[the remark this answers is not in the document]
seanjparker: Ohh right, I forgot you could do that! Ty!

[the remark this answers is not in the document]
seanjparker: Oh also, since the `Data` field is fixed, is it okay to do:

```csharp
public unsafe fixed byte Data[16];
```

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
seanjparker: I've move the method previously in `TraceItem` into the `Item` struct. Because of the way the Item conversion from user-facing to internal works, I needed to add the IOStorage parameter to some methods where it doesn't seem like you would need it. i.e. `SerializeItemToStream`

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
seanjparker: So I pinned the `internalEventData` since I needed to get the `IntPtr` for the `EventData` class. The other way I could see to do it is by using the `DangerousGetHandle()` method from the `SafeHandle` class.

seanjparker: Based on what Sonar cloud [complained about](https://sonarcloud.io/code?id=spatialos_gdk-for-unity&pullRequest=1452&selected=spatialos_gdk-for-unity%3APackages%2Fio.improbable.worker.sdk%2FParameterConversion.cs&line=59) it doesn't seem like I should use `DangerousGetHandle()`

jamiebrynes7: It looks like in the cases where you need access to the underlying pointer, the Worker SDK adds a method to expose it: 

https://code.i8e.io/github.com/improbable/everything@5eeb0768de55361b1977ce65b406921389f094a2/-/blob/worker_sdk/csharp_cinterop/sdk/worker/internal/Handles.cs#L105

Of course this only serves here to silence SonarCloud about `DangerousGetHandle()` 😂 

Imo - the warnings its complaining about are valid, but we can safe guard against the safety issues by ensuring that the `EventDataHandle` hasn't been disposed before calling `ConvertEvent`. 

Its also important to understand the context of `SafeHandle` which was originally designed for [Windows OS handles](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle?view=netcore-3.1#why-safehandle).

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
seanjparker: I've had a go at using the same pattern as the `ConvertPlayerIdentityTokenRequest`

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
seanjparker: Oh right! I did wonder why that method was included in the Worker SDK 😆 I've added it to the EventData class, I had to modify it a little compared to the Worker SDK, but it does the same thing I think.

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,403 @@
...
+                        newItem->ItemUnion.Span.Causes[i] = ParameterConversion.ConvertSpanId(spanItem.Causes[i]);
+                    }
+
+                    break;
+                case ItemType.Event:
+                    var eventItem = item.Event.GetValueOrDefault();
+                    ParameterConversion.ConvertEvent(eventItem, internalEvent =>
+                    {
+                        newItem->ItemUnion.Event = internalEvent;
+                    });
```
jamiebrynes7, line 200: This is still a little unsafe. The closure passed to `ConvertEvent` needs to encapsulate the entire process until the raw event is consumed. 

As it stands, the pointers inside `newItem->ItemUnion.Event` could be invalid after we exit this closure. We then return the item to the call who presumably does something with it.

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,403 @@
...
+                return newSpanId;
+            }
+        }
+
+        public SpanId AddSpan()
+        {
+            var newSpanId = new SpanId();
+            unsafe
+            {
+                var createdSpanId = CEventTrace.EventTracerAddSpan(eventTracer, null, 1);
```
jamiebrynes7, line 306: The cause count on this should be `0` rather than `1` no?

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,403 @@
...
+                var fields = new Dictionary<string, string>();
+                fixed (byte** keys = nativeKeys)
+                fixed (byte** values = nativeValues)
+                {
+                    CEventTrace.EventDataGetStringFields(eventData, keys, values);
+
+                    for (var i = 0; i < numberOfFields; i++)
+                    {
+                        fields.Add(ApiInterop.FromUtf8Cstr(nativeKeys[i]), ApiInterop.FromUtf8Cstr(nativeValues[i]));
+                    }
```
jamiebrynes7, line 374: Nit: This doesn't strictly need to be in the `fixed` block

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,403 @@
...
+            ParameterConversion.ConvertEvent(@event, internalEvent =>
+            {
+                shouldSampleEvent = CEventTrace.EventTracerShouldSampleEvent(eventTracer, internalEvent) > 0;
+            });
+
+            return shouldSampleEvent;
+        }
+    }
+
+    public class TraceEventData
```
jamiebrynes7, line 333: Nice work on eliminating the handles in here!

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,136 @@
+using System;
+using System.Collections.Generic;
+using System.Runtime.ConstrainedExecution;
+using System.Runtime.InteropServices;
+
+namespace Improbable.Worker.CInterop.Internal
+{
+    // Merge with /worker_sdk/csharp_cinterop/sdk/worker/internal/ParameterConversion.cs
+    // when changes upstreamed to Worker SDK
+    internal static class ParameterConversion
```
jamiebrynes7, line 10: Nice work on these - they can be a little tricky!

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,403 @@
...
+            return CEventTrace.SpanIdHash(ParameterConversion.ConvertSpanId(this));
+        }
+
+        public static SpanId GetNullSpanId()
+        {
+            var nativeSpanId = new SpanId();
+            var nullSpanId = CEventTrace.SpanIdNull();
+            unsafe
+            {
+                ApiInterop.Memcpy(nativeSpanId.Data, nullSpanId.Data, SpanIdSize);
```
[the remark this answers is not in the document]
seanjparker: Yeah, I thought so too, but that's how they defined in the ApiInterop 🤷‍♂️

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,403 @@
...
+
+
+        public void AddItemToStorage(IOStorage storage, Item item)
+        {
+            unsafe
+            {
+                var traceItem = ConvertItem(storage, item);
+                CEventTrace.ItemCreate(storage.Storage, traceItem);
+            }
+        }
```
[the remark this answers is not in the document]
seanjparker: Good point 👍 I'll change how these methods work + make the naming more clear

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
seanjparker: Yes, that should be `0`

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,403 @@
...
+            }
+
+            return nativeSpanId;
+        }
+    }
+
+    public struct Span
+    {
+        public SpanId Id;
+        public int CauseCount;
```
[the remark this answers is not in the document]
seanjparker: Ohhh, that's right! I'll get rid of it 🙂

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,136 @@
...
+            var handles = new List<WrappedGcHandle>(parameters.Length);
+            ConvertTracerParameters(handles, parameters, out var internalParameters);
+
+            fixed (CEventTrace.EventTracerParameters* parameterBuffer = internalParameters)
+            {
+                callback(parameterBuffer, handles);
+            }
+        }
+
+        public static unsafe WrappedGcHandle ConvertTracerParameter(EventTracerParameters parameter, ref CEventTrace.EventTracerParameters internalParameters)
```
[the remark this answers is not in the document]
seanjparker: Nope, it can be made private 👍

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,136 @@
...
+            return internalSpanId;
+        }
+
+        public static unsafe void ConvertEvent(Event eventToConvert, EventParametersCallback callback)
+        {
+            CEventTrace.Event internalEvent = new CEventTrace.Event();
+            internalEvent.UnixTimestampMillis = eventToConvert.UnixTimestampMillis;
+            internalEvent.Id = ConvertSpanId(eventToConvert.Id);
+
+            if (!eventToConvert.Data.eventData.IsClosed && eventToConvert.Data.eventData != null)
```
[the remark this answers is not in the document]
seanjparker: Fixed, good point!

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,403 @@
...
+                    newEvent.Data = Marshal.PtrToStructure<TraceEventData>(itemContainer->ItemUnion.Event.Data);
+                    break;
+                default:
+                    throw new NotSupportedException("Invalid Item Type provided.");
+            }
+
+            return newItem;
+        }
+
+        internal static unsafe CEventTrace.Item* ConvertItem(IOStorage storage, Item item)
```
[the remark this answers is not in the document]
seanjparker: Hmm, I think that should work. I've added a new method that gets the Thread-local item and nulls/zeros any data before returning the pointer to the Item

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
seanjparker: Discussed offline, refactored this approach to control the lifetime of the `internalEvent` until it's consumed by using a closure.

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,403 @@
...
+
+                    var newEvent = newItem.Event.GetValueOrDefault();
+                    newEvent.Id = new SpanId();
+                    ApiInterop.Memcpy(newEvent.Id.Data, itemContainer->ItemUnion.Event.Id.Data, SpanId.SpanIdSize);
+
+                    newEvent.UnixTimestampMillis = itemContainer->ItemUnion.Event.UnixTimestampMillis;
+                    newEvent.Type = ApiInterop.FromUtf8Cstr(itemContainer->ItemUnion.Event.Type);
+                    newEvent.Message = ApiInterop.FromUtf8Cstr(itemContainer->ItemUnion.Event.Message);
+
+                    newEvent.Data = Marshal.PtrToStructure<TraceEventData>(itemContainer->ItemUnion.Event.Data);
```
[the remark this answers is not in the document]
seanjparker: I've added a method which can convert the EventData between the internal + external structs, it feels a bit hacky imo but I've tested it and it does work, but I'll still look into a better way to do this.

workers/unity/Packages/io.improbable.worker.sdk/CEventTrace.cs
```diff
@@ -18,13 +18,23 @@ internal unsafe class CEventTrace
...
         {
             protected override bool ReleaseHandle()
             {
                 EventDataDestroy(handle);
                 return true;
             }
+
+            internal void SetUnderlying(IntPtr newHandle)
+            {
+                SetHandle(newHandle);
```
zeroZshadow, line 31: This allows you to override the previous handle, this could lead to a leak?

workers/unity/Packages/io.improbable.worker.sdk/CIO.cs
```diff
@@ -30,8 +30,28 @@ protected override bool ReleaseHandle()
 
         public enum OpenMode
```
zeroZshadow: As discussed offline, this could be changed to a `[Flags]` enum, typed at Uint32.

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -4,45 +4,67 @@
 
 namespace Improbable.Worker.CInterop
 {
+    [Flags]
     public enum OpenMode
```
zeroZshadow: Isn't this a duplicate from `OpenMode` in `workers/unity/Packages/io.improbable.worker.sdk/CIO.cs` ?

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -4,45 +4,67 @@
...
         }
 
         public static IOStream CreateRingBufferStream(uint capacity)
         {
             return new IOStream(CIO.CreateRingBufferStream(capacity));
         }
 
-        public static IOStream CreateFileStream(string fileName, OpenMode openMode)
+        public static IOStream CreateFileStream(string fileName,
+            OpenMode? openMode = OpenMode.OpenModeRead | OpenMode.OpenModeWrite | OpenMode.OpenModeTruncate)
```
zeroZshadow, line 55: Why is `openMode` a nullable with a default?

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
```diff
@@ -4,45 +4,67 @@
...
         {
             fixed (byte* fileNameBytes = ApiInterop.ToUtf8Cstr(fileName))
             {
-                return new IOStream(CIO.CreateFileStream(fileNameBytes, (CIO.OpenMode) openMode));
+                return new IOStream(CIO.CreateFileStream(fileNameBytes, (uint) openMode.GetValueOrDefault()));
```
zeroZshadow: The `default` here will return `0`, is that valid?

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,233 @@
...
+            {
+                case ItemType.Span:
+                    var spanItem = item.Span.GetValueOrDefault();
+                    newItem->ItemUnion.Span = new CEventTrace.Span();
+                    newItem->ItemUnion.Span.Id = ConvertSpanId(spanItem.Id);
+                    newItem->ItemUnion.Span.CauseCount = (uint) spanItem.Causes.Length;
+                    for (var i = 0; i < newItem->ItemUnion.Span.CauseCount; i++)
+                    {
+                        newItem->ItemUnion.Span.Causes[i] = ConvertSpanId(spanItem.Causes[i]);
+                        callback(newItem);
```
zeroZshadow, line 156: Is it correct that this will call multiple times?

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,233 @@
...
+                    break;
+                default:
+                    throw new NotSupportedException("Invalid Item Type provided.");
+            }
+        }
+
+        public static unsafe void ConvertEventTracerParameters(EventTracerParameters[] parameters, EventTracerParametersCallback callback)
+        {
+            var handles = new List<WrappedGcHandle>(parameters.Length);
+            ConvertTracerParameters(handles, parameters, out var internalParameters);
```
zeroZshadow, line 177: Shouldn't this do something with the return value?

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,233 @@
...
+
+                    break;
+                default:
+                    throw new NotSupportedException("Invalid Item Type provided.");
+            }
+        }
+
+        public static unsafe void ConvertEventTracerParameters(EventTracerParameters[] parameters, EventTracerParametersCallback callback)
+        {
+            var handles = new List<WrappedGcHandle>(parameters.Length);
```
zeroZshadow, line 176: If this `List` is of a static size, might be better to use an array instead.

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            internalParameters.UserData = (void*) wrappedParameterObject.Get();
+            internalParameters.TraceCallback = parameter.TraceCallback == null
+                ? IntPtr.Zero
+                : Marshal.GetFunctionPointerForDelegate(CallbackThunkDelegates.TraceCallbackThunkDelegate);
+
+            return wrappedParameterObject;
+        }
+
+        public static void FreeTracerParameterHandles(IEnumerable<WrappedGcHandle> tracerParameterHandles)
```
zeroZshadow, line 197: Is this worth a function?

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,233 @@
...
+        {
+            foreach (var tracerParameterHandle in tracerParameterHandles)
+            {
+                tracerParameterHandle.Dispose();
+            }
+        }
+
+        private static void ConvertTracerParameters(List<WrappedGcHandle> handles, EventTracerParameters[] parameters,
```
zeroZshadow, line 205: should this not be allocating the handles array itself, and return it?

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        private unsafe class CallbackThunkDelegates
+        {
+            public static readonly CEventTrace.TraceCallback TraceCallbackThunkDelegate = TraceCallbackThunk;
+
+            [MonoPInvokeCallback(typeof(CEventTrace.TraceCallback))]
+            private static void TraceCallbackThunk(void* callbackHandlePtr, CEventTrace.Item* item)
+            {
+                var callbackHandle = GCHandle.FromIntPtr((IntPtr) callbackHandlePtr);
+                var callback = (Action<Item>) callbackHandle.Target;
```
zeroZshadow, line 227: Thats a scary cast, is that correct?

workers/unity/Packages/io.improbable.worker.sdk/CEventTrace.cs
seanjparker: Yes, this could definitely lead to a leak, however, it's only used in one place:
https://github.com/spatialos/gdk-for-unity/pull/1452/files#diff-f0724dab9c5e9627ca1d44f44414bd32R120

It was the only way I could figure out to convert the EventData from the internal to external representations. In the `TraceEventData` internal constructor, we take a copy of the old handle and call `EventDataDestroy` to make sure any allocated memory is released

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
seanjparker: It is, I think I discussed this with Jamie that we shouldn't provide access to the internal Enum in `CIO.cs` and just have the same enum and just convert between the inner and outer enums

workers/unity/Packages/io.improbable.worker.sdk/CEventTrace.cs
zeroZshadow: What about making the function private, and add an additional constructor?

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
seanjparker: Oops, I can remove the `?`

workers/unity/Packages/io.improbable.worker.sdk/IOStream.cs
seanjparker: I changed the openMode parameter to no longer be nullable, so the default value would now be `OpenMode.OpenModeRead | OpenMode.OpenModeWrite | OpenMode.OpenModeTruncate`

seanjparker: > The `default` here will return `0`, is that valid?

A value of `0` would not be valid in Worker SDK v14.8

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
seanjparker: Whoops, I don't think that should be inside the loop!

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
seanjparker: It uses the `internalParameters` in the fixed statement below in order to call the callback

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
seanjparker: Good point 👍

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
seanjparker: Probably not, I think I can move the logic to inside the Dispose method of the EventTracer instead

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
seanjparker: That's the bit that isn't working at the moment, so I don't think the cast is correct

workers/unity/Packages/io.improbable.worker.sdk/CEventTrace.cs
seanjparker: Yep, that seems to work, it made it a bit more clear I think 👍

seanjparker:
cc @improbable-jonphillips, @tbimprob I was wondering if you could take a look at these C# wrappers for the C API, in case you had any comments/suggestions about them before creating a PR into `everything`.

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
seanjparker: Discussed offline, now fixed 😄

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
seanjparker: Yes, that would be better 👍

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,222 @@
...
+        public static unsafe void ConvertItem(Item item, ItemCallback callback)
+        {
+            var newItem = Item.GetThreadLocalItem();
+
+            newItem->ItemUnion = new CEventTrace.Item.Union();
+            newItem->ItemType = (CEventTrace.ItemType) item.ItemType;
+            switch (item.ItemType)
+            {
+                case ItemType.Span:
+                    var spanItem = item.Span.GetValueOrDefault();
```
jamiebrynes7, line 146:
```suggestion
                    var spanItem = item.Span.Value;
```

We want this to throw if the item type is `Span`, but there is no span data (rather than silently pass)

workers/unity/Packages/io.improbable.worker.sdk/ParameterConversion.cs
```diff
@@ -0,0 +1,222 @@
...
+                    for (var i = 0; i < newItem->ItemUnion.Span.CauseCount; i++)
+                    {
+                        newItem->ItemUnion.Span.Causes[i] = ConvertSpanId(spanItem.Causes[i]);
+                    }
+
+                    callback(newItem);
+
+                    break;
+                case ItemType.Event:
+                    var eventItem = item.Event.GetValueOrDefault();
```
jamiebrynes7, line 161:
```suggestion
                    var eventItem = item.Event.Value;
```

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,425 @@
...
+using Improbable.Worker.CInterop.Internal;
+
+namespace Improbable.Worker.CInterop
+{
+    public struct SpanId : IEquatable<SpanId>
+    {
+        internal static UIntPtr SpanIdSize = new UIntPtr(16);
+        public unsafe fixed byte Data[16];
+
+        public static SpanId NullSpanId => GetNullSpanId();
```
jamiebrynes7, line 13:
```suggestion
        public static SpanId Null => GetNullSpanId();
```

`SpanId.Null` is just as descriptive as `SpanId.NullSpanId`

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
```diff
@@ -0,0 +1,425 @@
...
+        public TraceEventData()
+        {
+            EventData = CEventTrace.EventDataCreate();
+        }
+
+        public string this[string key]
+        {
+            get => GetValue(key);
+            set => AddField(key, value);
+        }
```
jamiebrynes7, line 339: Nice!

jamiebrynes7 (approved):
Couple nits, but otherwise this LGTM. Don't forget a CHANGELOG entry too!

workers/unity/Packages/io.improbable.worker.sdk/CIO.cs
```diff
@@ -10,7 +11,7 @@ namespace Improbable.Worker.CInterop.Internal
 {
     internal unsafe class CIO
     {
-        public class StorageHandle : CptrHandle
+        internal class StorageHandle : CptrHandle
```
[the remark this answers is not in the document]
seanjparker: Good point 👍

workers/unity/Packages/io.improbable.worker.sdk/EventTracing.cs
seanjparker: Updated in latest commit

BryanJY-Wong (approved):

paulbalaji (approved):

PR #1465, opened 2020-08-26
Fix authority changes ordering problem

jamiebrynes7:
#### Description

Authority changes were stored in reverse order (newest to oldest, rather than oldest to newest). This meant that if you iterated through the `MessageSpan<AuthorityChangeReceived>` changes assuming it would be oldest to newest, the state could end up in correct.

This PR rectifies this issue by ensuring they are ordered in the way we receive them.

#### Tests

- [x] Unit tests

#### Documentation

- [x] Changelog

### DIFF
--- a/workers/unity/Assets/Tests/EditmodeTests/Correctness/Core/OpOrderingTests.cs
+++ b/workers/unity/Assets/Tests/EditmodeTests/Correctness/Core/OpOrderingTests.cs
@@ -1,0 +1,79 @@
+using Improbable.Gdk.Core;
+using Improbable.Gdk.Test;
+using Improbable.Gdk.TestUtils;
+using Improbable.Worker.CInterop;
+using NUnit.Framework;
+
+namespace Improbable.Gdk.EditmodeTests
+{
+    [TestFixture]
+    public class Test : MockBase
+    {
+        private const long EntityId = 1;
+
+        [Test]
+        public void Authority_changes_are_kept_in_order_and_applied_in_order()
+        {
+            World
+                .Step(world => world.Connection.CreateEntity(EntityId, GetTemplate()))
+                .Step(world =>
+                {
+                    world.Connection.ChangeAuthority(EntityId, 54, Authority.Authoritative);
+                    world.Connection.ChangeAuthority(EntityId, 54, Authority.NotAuthoritative);
+                })
+                .Step(world =>
+                {
+                    var authChanges = world
+                        .GetSystem<ComponentUpdateSystem>()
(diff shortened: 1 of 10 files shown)


workers/unity/Assets/Tests/EditmodeTests/Correctness/Core/OpOrderingTests.cs
```diff
@@ -0,0 +1,79 @@
...
+using Improbable.Gdk.TestUtils;
+using Improbable.Worker.CInterop;
+using NUnit.Framework;
+
+namespace Improbable.Gdk.EditmodeTests
+{
+    [TestFixture]
+    public class Test : MockBase
+    {
+        private const long EntityId = 1;
```
BryanJY-Wong, line 12: Why do you need this if you just used the literal `1`? 😂

paulbalaji: :this:

jamiebrynes7: Woops, fixed!

workers/unity/Packages/io.improbable.gdk.core/Worker/OpListConverter.cs
```diff
@@ -86,7 +87,9 @@ public bool ParseOpListIntoDiff(OpList opList, CommandMetaData commandMetaData)
                         break;
                     case OpType.AuthorityChange:
                         var authorityOp = opList.GetAuthorityChangeOp(i);
-                        viewDiff.SetAuthority(authorityOp.EntityId, authorityOp.ComponentId, authorityOp.Authority);
+                        viewDiff.SetAuthority(authorityOp.EntityId, authorityOp.ComponentId, authorityOp.Authority,
```
BryanJY-Wong, line 90: Out of curiosity, do we have a standard about using `Function(integer); integer++;` vs `Function(integer++);` since the former is clearer (I usually use the latter)?

paulbalaji (approved):

zeroZshadow (approved):

BryanJY-Wong (approved):

workers/unity/Packages/io.improbable.gdk.core/Worker/OpListConverter.cs
jamiebrynes7: We don't have a standard for it 😃

PR #1467, opened 2020-08-27
Upgrade to Burst 1.3.5 to fix broken compile for performance tests

zeroZshadow:
#### Description
Upgrade to latest burst to fix broken performance test compiles in CI

### DIFF
--- a/ci/perf-test.sh
+++ b/ci/perf-test.sh
@@ -87,7 +87,7 @@ function main {
 
 function anyMatchingTargets {
     if ! [[ -z "${PERF_BRANCH_OVERRIDE}" ]]; then
-        return 1
+        return 0
     fi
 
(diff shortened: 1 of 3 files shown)


jamiebrynes7 (approved):

ci/perf-test.sh
```diff
@@ -86,6 +86,10 @@ function main {
 }
 
 function anyMatchingTargets {
+    if ! [[ -z "${PERF_BRANCH_OVERRIDE}" ]]; then
+        return 0
+    fi
+
```
paulbalaji, line 92: have you tested this works when you don't set `PERF_BRANCH_OVERRIDE`?

zeroZshadow: I have now

PR #1476, opened 2020-09-03
Feature/worker flag protocol logs

BryanJY-Wong:
#### Description
- Enables protocol logs to be toggled during a deployment using worker flags

#### Tests
- Tested on local and cloud deployment
- Wrote some unit tests

#### Documentation
- [x] Update docs
- [x] Changelog

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.core/Worker/ProtocolLogController.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        private bool LogFilter(LogCategory category, LogLevel level)
+        {
+            return enabled && (category == LogCategory.NetworkStatus || category.HasFlag(LogCategory.NetworkTraffic)) && level >= LogLevel.Info;
+        }
+
+        public void Dispose()
+        {
+            callbackSystem.UnregisterWorkerFlagChangeCallback(callbackKey);
```
BryanJY-Wong, line 76: I'm not sure if this is even needed, since this really doesn't need to be deregistered. Thought?

workers/unity/Packages/io.improbable.gdk.core/Worker/ProtocolLogController.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            enabled = !string.IsNullOrWhiteSpace(value) && value.Split(',').Contains(workerId);
+            if (enabled)
+            {
+                workerSystem.EnableLogging();
+            }
+            else
+            {
+                workerSystem.DisableLogging();
+            }
```
jamiebrynes7, line 66: This turns off _all_ logging from the Worker SDK, not just this current logger. Given that this only logs a subset of the information that you can get from the Worker SDK logging, this doesn't seem like the best behaviour?

workers/unity/Packages/io.improbable.gdk.core/Worker/ProtocolLogController.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        private void OnWorkerFlagChanged((string, string) pair)
+        {
+            var (key, value) = pair;
+            if (key != ProtocolLoggingFlag)
+            {
+                return;
+            }
+
+            enabled = !string.IsNullOrWhiteSpace(value) && value.Split(',').Contains(workerId);
```
jamiebrynes7, line 58: Is there any case where the `value` is `null`? If not, I think the first part of this statement is redundant (if its empty or whitespace the second part is always false)

workers/unity/Packages/io.improbable.gdk.core/Worker/ProtocolLogController.cs
jamiebrynes7: Probably not as long we are guaranteed that the lifetime of this object is the same as the connection

workers/unity/Assets/Tests/EditmodeTests/Correctness/Core/WorkerFlagLoggingTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            var template = new EntityTemplate();
+            template.AddComponent(new Position.Snapshot());
+            return template;
+        }
+
+        [Test]
+        public void Logging_disabled_when_ProtocolLoggingFlag_does_not_exist()
+        {
+            World
+                .Step(world => world.Connection.CreateEntity(EntityId, GetTemplate()))
```
jamiebrynes7, line 24: Why do these tests create entities?

workers/unity/Packages/io.improbable.gdk.core/Worker/ProtocolLogController.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+            else
+            {
+                workerSystem.DisableLogging();
+            }
+        }
+
+        private bool LogFilter(LogCategory category, LogLevel level)
+        {
+            return enabled && (category == LogCategory.NetworkStatus || category.HasFlag(LogCategory.NetworkTraffic)) && level >= LogLevel.Info;
```
jamiebrynes7, line 71: I'd consider splitting some of these boolean statements out to make it more legible.

workers/unity/Packages/io.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -145,6 +147,7 @@ protected async Task Connect(IConnectionHandlerBuilder builder, ILogDispatcher l
 
         protected virtual void HandleWorkerConnectionEstablished()
         {
+            protocolLog?.OnWorkerConnected(Worker.World);
```
jamiebrynes7, line 150: Is the intention that the user can optionally call `base.HandleWorkerConnectionEstablished()` to wire this up?

workers/unity/Packages/io.improbable.gdk.core/Worker/WorkerConnector.cs
```diff
@@ -214,6 +217,15 @@ protected ConnectionParameters CreateConnectionParameters(string workerType, ICo
...
+
+        protected ConnectionParameters CreateConnectionParameters(string workerId, string workerType,
+            IConnectionParameterInitializer initializer = null)
+        {
+            protocolLog = new ProtocolLogController(workerId);
+            var @params = CreateConnectionParameters(workerType, initializer);
+            @params.Logsinks.Add(protocolLog.LogsinkParameters);
 
             return @params;
         }
```
jamiebrynes7, line 231: Couldn't this instead be implemented as a `IConnectionParameterInitializer`? Rather than an overload?

jamiebrynes7 (commented):
I wonder why Unity is generating meta files for the Packages in here, they don't do that when I use the project 😕

workers/unity/Packages/io.improbable.gdk.core/Worker/ProtocolLogController.cs
BryanJY-Wong: if the user deletes the flag it returns null.

workers/unity/Assets/Tests/EditmodeTests/Correctness/Core/WorkerFlagLoggingTests.cs
BryanJY-Wong: I somehow thought that was create connection😅

workers/unity/Packages/io.improbable.gdk.core/Worker/WorkerConnector.cs
BryanJY-Wong: I assume users will call `base.HandleWorkerConnectionEstablished()` when overriding the method. Should I use the `OnWorkerCreationFinished` event instead?

jamiebrynes7: You could even call this directly before we call `HandleWorkerConnectionEstablished` in `Connect` rather than the indirection through the callbacks

workers/unity/Packages/io.improbable.gdk.core/Worker/WorkerConnector.cs
BryanJY-Wong: It could, but the `protocolLogController` would need to be moved to the `GameLogicWorkerConnector` because we need the workerId which isn't really known until after the ConnectionFlow gets initialized.

workers/unity/Assets/Playground/Scripts/Worker/GameLogicWorkerConnector.cs
```diff
@@ -19,20 +19,20 @@ private async void Start()
...
+                    new ConnectionParameterInitializer(flow.WorkerId, protocolLogController));
             }
             else
             {
-                flow = new ReceptionistFlow(CreateNewWorkerId(WorkerUtils.UnityGameLogic),
-                    new CommandLineConnectionFlowInitializer());
+                flow = new ReceptionistFlow(CreateNewWorkerId(WorkerUtils.UnityGameLogic), new CommandLineConnectionFlowInitializer());
                 connectionParameters = CreateConnectionParameters(WorkerUtils.UnityGameLogic,
-                    new CommandLineConnectionParameterInitializer());
+                    new ConnectionParameterInitializer(flow.WorkerId, protocolLogController, new CommandLineConnectionParameterInitializer()));
```
BryanJY-Wong, line 35: I think this looks weird, I needed to make `protocolLogController` protected to be able to read it. Should I just put `protocolLogController` in `GameLogicWorkerConnector`?

BryanJY-Wong: Reverted this back to the overload implementation.

PR #1481, opened 2020-09-11
UTY-2606: Improve entity system

BryanJY-Wong:
#### Description
- Refactor `EntitySystem` to use a more Data-Oriented approach.
- Breaking change, EntitiesAdded + EntitiesRemoved are now properties and returns `NativeList<EntityId>` instead of `List<EntityId>`

#### Tests
- Unit Tests in Assets/Tests/EditmodeTests/Core/EntitySystemTests.cs
- Launched local deployment and spawn cubes

#### Documentation
- [x] Changelog

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
+using Unity.Jobs;
 
 namespace Improbable.Gdk.Core
 {
     [DisableAutoCreation]
     [AlwaysUpdateSystem]
     [UpdateInGroup(typeof(SpatialOSReceiveGroup.InternalSpatialOSReceiveGroup))]
-    [UpdateBefore(typeof(SpatialOSReceiveSystem))]
-    public class EntitySystem : ComponentSystem
+    [UpdateAfter(typeof(SpatialOSReceiveSystem))]
```
zeroZshadow, line 11: Do the systems that rely on running AFTER `EntitySystem` have been updated to keep this order change in mind?

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
+                EntityCount = query.CalculateEntityCount();
+                if (EntityIds.Length >= EntityCount)
                 {
-                    entitiesRemoved.Add(entityId);
+                    return;
                 }
 
-                if (entitiesAdded.Count != 0 || entitiesRemoved.Count != 0)
+                EntityIds.Dispose();
+                EntityIds = new NativeArray<EntityId>(EntityCount << 1, Allocator.Persistent);
```
zeroZshadow, line 45: Just do normal math here, its not a bitfield

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
-                    entitiesAdded.Add(entityId);
+                    JobHandle.Complete();
+                    return new NativeSlice<EntityId>(EntityIds, 0, EntityCount);
                 }
+            }
 
-                foreach (var entityId in diff.GetEntitiesRemoved())
+            public void EnsureSize(ref EntityQuery query)
+            {
+                EntityCount = query.CalculateEntityCount();
```
zeroZshadow, line 38: If this expects queries that dont have filters, you can use the `IsEmptyIgnoreFilter` field instead. Much faster

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
+                get
                 {
-                    entitiesAdded.Add(entityId);
+                    JobHandle.Complete();
+                    return new NativeSlice<EntityId>(EntityIds, 0, EntityCount);
                 }
+            }
 
-                foreach (var entityId in diff.GetEntitiesRemoved())
+            public void EnsureSize(ref EntityQuery query)
```
zeroZshadow, line 36: If we're not altering the query, you can change this to `in` instead of `ref`

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
zeroZshadow, line 45: Since we do resizes, would a `NativeList` not work better? It can use the NativeList `GetEnumerator()` instead of the NativeSlice to make iteration nicely without write access.

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
+
+            if (added.EntityCount != 0 || removed.EntityCount != 0)
+            {
+                ViewVersion += 1;
+            }
+        }
+
+        private void UpdateEntitiesAdded()
+        {
+            added.EnsureSize(ref addedQuery);
```
zeroZshadow, line 90: This will only ensure the size is right for the amount of entities we had in the PREVIOUS iteration, that doesn't sound right?

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
+                }).ScheduleParallel(Dependency);
+        }
+
+        private void UpdateEntitiesRemoved()
+        {
+            removed.EnsureSize(ref removedQuery);
+            var buffer = bufferSystem.CreateCommandBuffer().AsParallelWriter();
+            var array = removed.EntityIds;
+            removed.JobHandle = Entities.WithName("EntitiesRemoved")
+                .WithAll<EntitySystemStateComponent>()
```
zeroZshadow, line 110: No need for this, as this constraint is already created with the `in EntitySystemStateComponent entityId`

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
+        }
+
+        private void UpdateEntitiesAdded()
+        {
+            added.EnsureSize(ref addedQuery);
+            var buffer = bufferSystem.CreateCommandBuffer().AsParallelWriter();
+            var array = added.EntityIds;
+            added.JobHandle = Entities.WithName("EntitiesAdded")
+                .WithNone<EntitySystemStateComponent>()
+                .WithAll<SpatialEntityId>()
```
zeroZshadow, line 95: No need for this, as the constraint is already created with the `in SpatialEntityId entityId`

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
+        }
+
         protected override void OnUpdate()
         {
-            entitiesAdded.Clear();
-            entitiesRemoved.Clear();
+            UpdateEntitiesAdded();
+            UpdateEntitiesRemoved();
+
+            Dependency = JobHandle.CombineDependencies(added.JobHandle, removed.JobHandle);
```
zeroZshadow, line 79: This code is a little confusing, as it is not clear where the jobhandles come from.
why not have `UpdateEntitiesAdded` and `UpdateEntitiesRemoved` return their job handles, instead of setting them on `added` and `removed`, then do that step here.

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
+        private EntityQuery addedQuery; // This query needs to be a member of SystemBase
+        private EntityCollection removed;
+        private EntityQuery removedQuery; // This query needs to be a member of SystemBase
+        public int ViewVersion { get; private set; }
+        public NativeSlice<EntityId> EntitiesRemoved => removed.Slice;
+        public NativeSlice<EntityId> EntitiesAdded => added.Slice;
+
+        protected override void OnCreate()
+        {
+            base.OnCreate();
```
zeroZshadow, line 68: base functions for `OnCreate` and `OnDestroy` are empty I think, no need to call.

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,127 @@
...
                 }
             }
         }
 
+        private EndSimulationEntityCommandBufferSystem bufferSystem;
+        private EntityCollection added;
+        private EntityQuery addedQuery; // This query needs to be a member of SystemBase
+        private EntityCollection removed;
+        private EntityQuery removedQuery; // This query needs to be a member of SystemBase
+        public int ViewVersion { get; private set; }
```
zeroZshadow, line 62: newline between privates and public

zeroZshadow:
As there is only 1 system actually using the "added" and "removed" entities, would it not be better to remove the EntitiesSystem and move that functionality into the 1 system user the data?

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
BryanJY-Wong: Apparently it doesn't, I got the example from [Unity's docs](https://docs.unity3d.com/Packages/com.unity.entities@0.2/manual/entities_job_foreach.html) (in the 'Access to EntityQuery object example' section) and the unit test doesn't fail.

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
BryanJY-Wong: After `EntitySystem` previously or now? Using `Ctrl-Shift-F` nothing has a UpdateBefore or After `EntitySystem`

workers/unity/Packages/io.improbable.gdk.core/Systems/EntitySystem.cs
```diff
@@ -1,61 +1,124 @@
...
 
 namespace Improbable.Gdk.Core
 {
     [DisableAutoCreation]
     [AlwaysUpdateSystem]
     [UpdateInGroup(typeof(SpatialOSReceiveGroup.InternalSpatialOSReceiveGroup))]
-    [UpdateBefore(typeof(SpatialOSReceiveSystem))]
-    public class EntitySystem : ComponentSystem
+    [UpdateAfter(typeof(SpatialOSReceiveSystem))]
+    public class EntitySystem : SystemBase
```
BryanJY-Wong, line 12: I don't think we need this system. The only usages are in the WorkerInspector and Subscription. Both of which can use `WorkerSystem.Diff.GetEntitiesRemoved()` or `WorkerSystem.Diff.GetEntitiesAdded()`. 

Thoughts?

PR #1486, opened 2020-11-09
Type cache

zeroZshadow:
#### Description
Removed Temporary components.
Added type cache to avoid spikes when initializing multiple workers.


workers/unity/Packages/io.improbable.gdk.core/Systems/CleanTemporaryComponentsSystem.cs
```diff
@@ -18,17 +18,15 @@ public class CleanTemporaryComponentsSystem : ComponentSystem
 
         protected override void OnCreate()
         {
-            base.OnCreate();
-
             // Find all components with the RemoveAtEndOfTick attribute
```
paulbalaji, line 21: update this comment

paulbalaji (approved):

PR #1488, opened 2020-11-27
Type cache

zeroZshadow:
#### Description
Removed Temporary components.
Added type cache to avoid spikes when initializing multiple workers.


workers/unity/Packages/io.improbable.gdk.core/Utility/TypeCache.cs
```diff
@@ -0,0 +1,27 @@
...
+                    typeof(AutoRegisterSubscriptionManagerAttribute));
+            }
+
+            return subscriptionManagerTypes;
+        }
+
+        public static List<Type> GetComponentMetaClassTypes()
+        {
+            return ReflectionUtility.GetNonAbstractTypes(typeof(IComponentMetaclass));
+        }
```
jamiebrynes7, line 25: Why is one of these cached and the other isn't?

jamiebrynes7 (approved):

PR #1493, opened 2021-01-20
Fix `WorldCommandSender` 

jamiebrynes7:
The code between the generated command senders and the `WorldCommandSender` drifted apart over time and so bugs that had been fixed weren't fixed in the `WorldCommandSender`.

Re-unified the generated & world command senders by using a common base and thus fixed the bug.

Split from #1491


paulbalaji (approved):
neat

zeroZshadow (approved):

